PR #22, opened 2015-01-05
u/fjammes/dm-627

andy-slac:
Reviewing DM-627


admin/python/lsst/qserv/admin/dataloader.py
```diff
@@ -46,6 +46,7 @@
 #----------------------------------
```
andy-slac: What is the reason for renaming this module? I think its original name `dataLoader` already follows DM guidelines for camelCase naming.

andy-slac:
I think this branch contains changes from more than one ticket. I'm confused which changes I have to review for which ticket. Could you split this into one  branch per ticket?

PR #24, opened 2015-01-12
Create (secondary) index only for director table (DM-1720).

fjammes:
Chunk/subchunk index is only needed for director table, data loader
script now ignores index table option for all other tables. Director
table name is determined by dirTable configuration option.

### DIFF
--- a/admin/python/lsst/qserv/admin/partConfig.py
+++ b/admin/python/lsst/qserv/admin/partConfig.py
@@ -53,9 +53,16 @@ class PartConfig(UserDict.UserDict):
     Implemented as a dictionary with some extra methods.
     """
 
-    # keys that must be defined in partitioner config files
-    requiredConfigKeys = ['part.num-stripes', 'part.num-sub-stripes',
-                          'part.default-overlap']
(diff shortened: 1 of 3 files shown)


admin/python/lsst/qserv/admin/partConfig.py
```diff
@@ -109,6 +109,14 @@ def isView(self):
         """Returns True if table is a view"""
         return bool(self.data.get('view', False))
 
+    def isDirector(self, tableName):
+        """
+        Returns True if table is a director table. Director table name
+        is determined by dirTable parameter, if dirTable is not set then
+        Object is assumed to be a director table.
+        """
+        return tableName == self.get('dirTable', 'Object')
```
fjammes: I would like not to have default value for dirTable, this is an important parameter so I think it should always be explicited in cfg files.

andy-slac: OK, I will remove default.

fjammes: Thanks Andy ;-)

PR #27, opened 2015-01-15
Fix parenthesization in rendered IR tree (DM-737)

fritzm:
This fixes parethesization in rendering of BoolTerm trees.  A
BoolTerm::OpPrecedence enum was added, and the various subclasses
of BoolTerm were assigned relative precedence based on their operators
per the SQL92 standard.  The term list render helper
template function was extended to add parentheses when rendering a
contained term iff the operator of the contained term is of lower
precedence than that of the containing term.

A regression test for the reported test case, and some additional
unit test for parenthesis rendering were also added.

### DIFF
--- a/core/modules/global/sqltoken.cc
+++ b/core/modules/global/sqltoken.cc
@@ -53,8 +53,8 @@ struct InsensitiveCompare {
 /// implementations. Possibilities: hashing, lookup tables.
 struct CompareMap {
     CompareMap() {
-        const char* sepWords[] = {"select", "from", "where", "by", "limit"};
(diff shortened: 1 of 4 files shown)


fritzm:
Thanks Daniel -- I've pushed a second commit that factors out a helper for the unit tests and adds a couple three-subterm tests.  Is this along the lines of what you had in mind?

Note that one of the new three-term unit-tests currently fails (because of a spacing issue -- good call to add that test -- there will be another commit coming to address that).

fritzm:
(Fixed bug mentioned above and re-pushed.)

PR #28, opened 2015-01-27
U/fjammes/dm 1677

fritzm:
By Fabrice: Factor multi-error capability out of QueryAction into a standalone utility class, and extend slightly (DM-1677)


core/modules/util/Error.cc
```diff
@@ -0,0 +1,56 @@
...
+}
+
+/** Overload output operator for this class
+ *
+ * @param out
+ * @param multiError
+ * @return an output stream
+ */
+std::ostream& operator<<(std::ostream &out, Error const& error) {
+    out << error.toString();
```
fritzm: It might be a little better to reverse these -- have operator<< be the primitive, and then have toString use operator<<, passing in the local ostringstream.  That way you could avoid constructing extra streams.

core/modules/util/Error.cc
```diff
@@ -0,0 +1,56 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the LSST License Statement and
+ * the GNU General Public License along with this program.  If not,
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+
+#include "Error.h"
```
fritzm: Change to #include "util/Error.h"?

core/modules/util/Error.h
```diff
@@ -0,0 +1,68 @@
...
+
+/** @class
+ * @brief Store a Qserv error
+ *
+ * To be used with util::MultiError
+ *
+ */
+class Error {
+public:
+    Error(int code, std::string msg) :
```
fritzm: Use const std::string &msg here instead, to avoid extra string copy?

core/modules/util/MultiError.cc
```diff
@@ -0,0 +1,61 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the LSST License Statement and
+ * the GNU General Public License along with this program.  If not,
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+
+#include "MultiError.h"
```
fritzm: #include "util/MultiError.h"

core/modules/util/Error.cc
```diff
@@ -0,0 +1,56 @@
+// -*- LSST-C++ -*-
+/*
+ * LSST Data Management System
+ * Copyright 2014 LSST Corporation.
```
jbecla: update copyright line, should say: "Copyright 2014-2015 AURA/LSST."

core/modules/util/Error.cc
jbecla: and a comment "Qserv headers" like we do elsewhere :)

core/modules/util/MultiError.cc
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @param out
+ * @param multiError
+ * @return an output stream
+ */
+std::ostream& operator<<(std::ostream &out,
+        MultiError const& multiError) {
+    out << multiError.toString();
+    return out;
+}
```
fritzm: Reverse these, as noted above, to save extra stream construction?

core/modules/util/Error.cc
```diff
@@ -0,0 +1,56 @@
...
+ * @return an output stream
+ */
+std::ostream& operator<<(std::ostream &out, Error const& error) {
+    out << error.toString();
+    return out;
+}
+
+}
+}
+} // lsst::qserv::util
```
jbecla: I'd put all 3 closing brackets in one line, e.g. "}}} // lsst::qserv::util". That is what we do elsewhere.

core/modules/util/Error.h
```diff
@@ -0,0 +1,68 @@
+// -*- LSST-C++ -*-
+/*
+ * LSST Data Management System
+ * Copyright 2014 LSST Corporation.
```
jbecla: need to update copyright line in every file you touch.

core/modules/util/Error.h
```diff
@@ -0,0 +1,68 @@
...
+    }
+    std::string toString() const;
+
+    int code;
+    std::string msg;
+
+    friend std::ostream& operator<<(std::ostream &out, Error const& error);
+};
+
+}
```
jbecla: }}}

core/modules/util/MultiError.h
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * Store Qserv errors in a throwable vector.
+ * util::Error operator << is used for output.
+ *
+ */
+class MultiError: public std::exception, public std::vector<Error> {
+public:
+
+    std::string toString() const;
+    virtual ~MultiError() throw () {
```
jbecla: alignment (missing 4 spaces)

core/modules/util/MultiError.h
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    friend std::ostream& operator<<(std::ostream &out,
+            MultiError const& multiError);
```
jbecla: alignment looks strange

core/modules/wdb/QueryAction.cc
```diff
@@ -165,6 +159,7 @@ bool QueryAction::Impl::act() {
     bool connOk = _initConnection();
     if(!connOk) { return false; }
 
+
```
jbecla: really needed?

core/modules/util/MultiError.cc
```diff
@@ -0,0 +1,61 @@
...
+ */
+std::string MultiError::toString() const {
+    std::ostringstream oss;
+
+    if (!this->empty()) {
+        std::ostream_iterator<Error> string_it(oss, "\n");
+        if (this->size()>1) oss << "Multi-error:\n";
+        std::copy(this->begin(), this->end(), string_it);
+    }
+    return oss.str();
```
fritzm: This is pretty cool, but it would be nice not to have the trailing newline, since I think logging etc. downstream will typically add it?

core/modules/util/MultiError.h
```diff
@@ -0,0 +1,71 @@
...
+*
+* @ingroup util
+*
+* @brief Implement a generic error container for Qserv
+*
+* @author Fabrice Jammes, IN2P3/SLAC
+*/
+
+#ifndef LSST_QSERV_UTIL_ERROR_H
+#define LSST_QSERV_UTIL_ERRORCONTAINER_H
```
fritzm: include guard symbols mismatched?

core/modules/util/MultiError.h
```diff
@@ -0,0 +1,71 @@
...
+namespace util {
+
+/** @class
+ * @brief Implement a generic error container for Qserv
+ *
+ * Store Qserv errors in a throwable vector.
+ * util::Error operator << is used for output.
+ *
+ */
+class MultiError: public std::exception, public std::vector<Error> {
```
fritzm: Uh oh...  It's is generally considered bad practice to derive from standard containers (for example, historically, std::vector<> didn't have a virtual destructor -- is this still the case?)  You might try googling "derive from standard containers" and read up on reasons why.

I'd suggest composition here instead of inheritance -- try having MultiError hold a vector internally, then implement your own begin(), end() methods that return iterators to the underlying container.

std::exception is explicitly designed to be used as a base class, so that one is fine.

fritzm:
This is a good start, Fabrice -- thanks!

core/modules/util/Error.cc
fjammes: ok, thanks.

core/modules/util/Error.cc
fjammes: ok, thanks.

core/modules/util/Error.h
fjammes: Ok, thanks.
A little detail: I think LSST coding standards recommends: "std::string const& msg", see https://confluence.lsstcorp.org/pages/viewpage.action?pageId=16908706 chapter 5.10

core/modules/util/MultiError.cc
fjammes: Ok thanks.

core/modules/util/Error.cc
fjammes: Ok thanks. Does the Copyright needs to be updated each year on all source files? If yes we should template this, isn't it?

core/modules/util/MultiError.cc
fjammes: ok, this is smart.

core/modules/util/Error.cc
fjammes: Ok thanks,

core/modules/util/Error.h
fjammes: ok thanks.

core/modules/util/Error.h
fjammes: ok thanks

core/modules/wdb/QueryAction.cc
fjammes: ok removed.

core/modules/util/MultiError.h
fjammes: It would be fine to have a automated tool to apply lsst standards?

core/modules/util/Error.cc
ktlim: The copyright needs to be updated whenever a source file is modified.  It can't really be templated, as it needs to be in the original source as distributed via git.
(It's not entirely clear to me whether it should say "2008-2015" based on the entire DM system or "2014-2015" based on the particular file.  The "AURA/LSST" is important, though.)

core/modules/util/MultiError.cc
fjammes: ok thanks.

core/modules/util/Error.cc
jbecla: Would claiming copyright for 2008 in files that were created 6 years later be appropriate?

fjammes:
Hi Fritz, 

Could you have a look, it looks fine now.

PR #30, opened 2015-02-02
Add some missing virtual destructors

fritzm:
(@andy-slac, @wangd: I think we should probably fix these?  If you guys agree, I'll open a ticket, attach this cl and push it in?)

The compiler was flagging with warnings at a few places where we
had class hierarchies without virtual destructors.  Added empty
virtual destructor the the base classes in each of these cases
for safety.

### DIFF
--- a/core/modules/global/ResourceUnit.h
+++ b/core/modules/global/ResourceUnit.h
@@ -102,6 +102,7 @@ private:
 
 class ResourceUnit::Checker {
 public:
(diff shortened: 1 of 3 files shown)


andy-slac:
Totally agree, if there is any virtual method in a class the destructor must be virtual as well. Not sure how did we miss it before, guess we need to pay more attention to compiler output.

PR #33, opened 2015-02-03
U/fjammes/dm 1822

fritzm:


fritzm:
Cool, Fabrice -- could we add a unit test for this in the existing testProtoRowBuffer.cc file as well?

fjammes:
ok done, and currently testing, could you close the pull request please?

fritzm:
I think it gets closed automatically when you commit your merge?  (If I close it manually, it will flag as "denied" I think.)

PR #38, opened 2015-02-18
Tickets/dm 216

smonkewitz:


core/modules/qproc/geomAdapter.h
```diff
@@ -0,0 +1,105 @@
...
+    return boost::make_shared<sg::Circle>(sg::UnitVector3d(center), a);
+}
+
+inline boost::shared_ptr<sg::Ellipse>
+getEllipseFromParams(std::vector<Coordinate> const& params) {
+    // lon, lat, semimajang, semiminang, posangle
+    if(params.size() != 5) {
+        throw QueryProcessingBug("Invalid number or parameters for region");
+    }
+    sg::UnitVector3d uv3(sg::LonLat::fromDegrees(params[0], params[1]));
```
smonkewitz: I think center is a better name for this unit vector.

core/modules/qproc/geomAdapter.h
```diff
@@ -0,0 +1,105 @@
...
+        uv3,
+        sg::Angle::fromDegrees(params[2]),
+        sg::Angle::fromDegrees(params[3]),
+        sg::Angle::fromDegrees(params[4]));
+}
+
+inline boost::shared_ptr<sg::ConvexPolygon>
+getConvexPolyFromParams(std::vector<Coordinate> const& params) {
+    // polygon vertices, min 3 vertices, must get even number of params
+    if((params.size() <= 6) || ((params.size() & 1) != 0)) {
```
smonkewitz: I don't think you should check the size of params here - just check that it has an even number of entries. The constructor will throw if it doesn't like what's being passed in.

core/modules/qproc/geomAdapter.h
```diff
@@ -0,0 +1,105 @@
...
+        throw QueryProcessingBug("Invalid number or parameters for region");
+    }
+    std::vector<sg::UnitVector3d> uv3;
+    for(unsigned i=0; i < params.size(); i += 2) {
+        sg::LonLat vx = sg::LonLat::fromDegrees(params[i], params[i+1]);
+        uv3.push_back(sg::UnitVector3d(vx));
+    }
+    return boost::make_shared<sg::ConvexPolygon>(uv3);
+}
+
```
smonkewitz: In general, have you though about wrapping exceptions coming out of sg constructors in some way? Also, are you using `QueryProcessingBug` instead of some user-input related exception type because (some of?) the parameter length constraints should be enforced in the grammar?

core/modules/qproc/fakeGeometry.h
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    int _stripes;
+    int _subStripes;
+};
+
+}}} // namespace lsst::qserv::ccontrol
+
+#endif // LSST_QSERV_CCONTROL_FAKEGEOMETRY_H
+
```
smonkewitz: Shouldn't this file be deleted?

core/modules/qproc/geomAdapter.h
```diff
@@ -0,0 +1,105 @@
...
+// Qserv headers
+#include "css/StripingParams.h"
+#include "qproc/QueryProcessingBug.h"
+
+
+namespace lsst {
+namespace qserv {
+namespace qproc {
+typedef double Coordinate;
```
smonkewitz: I dislike this typedef, because you are using it for things that I wouldn't classify as a coordinate in the traditional sense of the word (e.g. circle opening angle, ellipse semi-major/semi-minor axis, etc...). Can we just use double?

core/modules/qproc/SecondaryIndex.cc
```diff
@@ -0,0 +1,176 @@
...
+    virtual ChunkSpecVector lookup(query::ConstraintVector const& cv) = 0;
+};
+
+class MySqlBackend : public SecondaryIndex::Backend {
+public:
+    MySqlBackend(mysql::MySqlConfig const& c)
+        : _sqlConnection(c, true) {
+    }
+    virtual ChunkSpecVector lookup(query::ConstraintVector const& cv) {
+    // cv should only contain index constraints
```
smonkewitz: The body of this method needs to be indented.

core/modules/qproc/SecondaryIndex.h
```diff
@@ -0,0 +1,63 @@
...
+#include "mysql/MySqlConfig.h"
+#include "qproc/ChunkSpec.h"
+#include "query/Constraint.h"
+
+namespace lsst {
+namespace qserv {
+namespace qproc {
+
+/// SecondaryIndex handles lookups into a secondary index. Only one instance of
+/// this is necessary: all user queries can share a single instance.
```
smonkewitz: Shouldn't there be one instance per director table? Even if this part of the implementation isn't done yet, I think the comment should reflect that.

smonkewitz: Ah never mind. I see that he details about the table to look things up in is part of the constraint vector.

core/modules/qproc/testIndexMap.cc
```diff
@@ -0,0 +1,181 @@
...
+ *
+ * You should have received a copy of the LSST License Statement and
+ * the GNU General Public License along with this program.  If not,
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+
+ /**
+  * @file
+  *
+  * @brief Test SecondaryIndex and IndexMap operations.
```
smonkewitz: You aren't actually testing IndexMap here at all (or anywhere), as far as I can tell. Consider renaming the file and fixing up the comments?

core/modules/qproc/IndexMap.cc
```diff
@@ -0,0 +1,252 @@
...
+using lsst::sg::ConvexPolygon;
+using lsst::sg::SubChunks;
+using lsst::qserv::qproc::Coordinate;
+
+typedef std::vector<SubChunks> SubChunksVector;
+
+namespace { // File-scope helpers
+template <typename T>
+std::vector<T> convertVec(StringVector const& v) {
+    std::vector<T> out;
```
smonkewitz: out.reserve(v.size())?

core/modules/qproc/IndexMap.cc
```diff
@@ -0,0 +1,252 @@
...
+}
+
+ChunkSpec convertSgSubChunks(SubChunks const& sc) {
+    ChunkSpec cs;
+    cs.chunkId = sc.chunkId;
+    cs.subChunks.resize(sc.subChunkIds.size());
+    std::copy(sc.subChunkIds.begin(), sc.subChunkIds.end(),
+              cs.subChunks.begin());
+    return cs;
+}
```
smonkewitz: Shouldn't the two functions above be in the anonymous namespace?

core/modules/qproc/IndexMap.cc
```diff
@@ -0,0 +1,252 @@
...
+
+bool isIndex(query::Constraint const& c) {
+    return c.name == "sIndex";
+}
```
smonkewitz: Should this be a method of Constraint?

smonkewitz: Oh never mind, I can see that it and the function below is required by <algorithms>. Is it not possible to move this to an anonymous namespace?

core/modules/qproc/IndexMap.cc
```diff
@@ -0,0 +1,252 @@
...
+    return c.name != "sIndex";
+}
+
+////////////////////////////////////////////////////////////////////////
+// IndexMap::PartitioningMap definition and implementation
+////////////////////////////////////////////////////////////////////////
+class IndexMap::PartitioningMap {
+public:
+    explicit PartitioningMap(css::StripingParams const& sp) {
+        sg::Angle overlap(0.000001); // Dummy.
```
smonkewitz: This line is dead.

core/modules/qproc/IndexMap.cc
```diff
@@ -0,0 +1,252 @@
...
+    /// results. Regions are assumed to be joined by implicit "OR" and not "AND"
+    ///
+    SubChunksVector getIntersect(RegionPtrVector const& rv) {
+        SubChunksVector scv;
+        for(RegionPtrVector::const_iterator i=rv.begin(), e=rv.end();
+            i != e;
+            ++i) {
+            if(*i) {
+                SubChunksVector area = getCoverage(**i);
+                scv.insert(scv.end(), area.begin(), area.end());
```
smonkewitz: Different area constraints linked by ORs can overlap, so that the same chunk and/or sub-chunks can appear in scv after this code runs. Do you get rid of duplicates somewhere else to avoid dispatching multiple times on the same chunk/sub-chunk?

core/modules/qproc/IndexMap.cc
```diff
@@ -0,0 +1,252 @@
...
+            }
+        }
+        return scv;
+    }
+
+    inline SubChunksVector getCoverage(Region const& r) {
+        return _chunker->getSubChunksIntersecting(r);
+    }
+    ChunkSpecVector getAllChunks() const {
+        IntVector allChunks = _chunker->getAllChunks();
```
smonkewitz: Here, you assume that int is the same as int32_t.

core/modules/qproc/IndexMap.cc
```diff
@@ -0,0 +1,252 @@
...
+
+        ChunkSpecVector csv;
+        std::transform(scv.begin(), scv.end(),
+                       std::back_inserter(csv), convertSgSubChunks);
+        if(!indexSpecs.empty()) {
+            intersectSorted(csv, indexSpecs); // performs "AND"
+        }
+        return csv;
+    }
+}
```
smonkewitz: I feel like this would be a lot simpler and involve a lot less vector copying/creation etc... if you made SecondaryIndex just ignore non "sIndex" constraints, and ditched the use of `std::transform`, `std::copy`, etc...

core/modules/qproc/IndexMap.cc
```diff
@@ -0,0 +1,252 @@
...
+        fMap["qserv_areaspec_box"] = make<Box>;
+        fMap["qserv_areaspec_circle"] = make<Circle>;
+        fMap["qserv_areaspec_ellipse"] = make<Ellipse>;
+        fMap["qserv_areaspec_poly"] = make<ConvexPolygon>;
+    }
+    typedef std::map<std::string, MakeFunc> Map;
+    Map fMap;
+};
+
+static FuncMap funcMap;
```
smonkewitz: Maybe replace the global map with an inline function and a switch?

core/modules/qproc/ChunkSpec.cc
```diff
@@ -57,12 +60,116 @@ std::ostream& operator<<(std::ostream& os, ChunkSpec const& c) {
...
 }
 
+
+/// ChunkSpecVector intersection.
+/// Computes ChunkSpecVector intersection, overwriting dest with the result.
+///
+/// precondition: Elements in ChunkSpecVector should be minimized in the sense
+/// that there should be only one element for a particular chunkId. i.e., for
+/// all ChunkSpec A element, there is no ChunkSpec B element in the same vector
+/// where A.chunkId == B.chunkId.
```
smonkewitz: Also, don't the vectors need to be sorted on chunk ID?

core/modules/qproc/ChunkSpec.cc
```diff
@@ -57,12 +60,116 @@ std::ostream& operator<<(std::ostream& os, ChunkSpec const& c) {
...
+void intersectSorted(ChunkSpecVector& dest, ChunkSpecVector const& a) {
+    ChunkSpecVector tmp;
+    tmp.swap(dest);
```
smonkewitz: I think this would be clearer if you did the swap with dest right before returning. More importantly, that strategy yields the strong exception safety guarantee.

core/modules/ccontrol/UserQuery.cc
```diff
@@ -274,4 +252,45 @@ void UserQuery::_setupMerger() {
...
+    boost::shared_ptr<qproc::IndexMap> im;
+    std::string dominantDb = _qSession->getDominantDb();
+    if(!_qSession->validateDominantDb()) {
+        // TODO: Revisit this for L3
+        throw UserQueryError("Couldn't determine dominantDb for dispatch");
+    }
+
+    boost::shared_ptr<IntSet const> eSet = _qSession->getEmptyChunks();
+    {
+        //self._emptyChunks = metadata.getEmptyChunks(self.dominantDb)
```
fritzm: Commented-out line -- delete or change to more descriptive TODO if appropriate?

core/modules/qproc/IndexMap.cc
```diff
@@ -0,0 +1,252 @@
...
+        RegionPtrVector rv;
+        std::transform(cv.begin(), cv.end(), std::back_inserter(rv), getRegion);
+
+        SubChunksVector scv = _pm->getIntersect(rv);
+
+        ChunkSpecVector csv;
+        std::transform(scv.begin(), scv.end(),
+                       std::back_inserter(csv), convertSgSubChunks);
+        if(!indexSpecs.empty()) {
+            intersectSorted(csv, indexSpecs); // performs "AND"
```
smonkewitz: This seems to be buggy - `intersectSorted` assumes its input vectors are sorted and don't have multiple entries for the same chunk. `getIntersect` above does not guarantee either property.

core/modules/SConscript
```diff
@@ -49,7 +48,7 @@ if not env.GetOption('clean'):
 # all modules
 allModules = """ccontrol css czar global log mysql obsolete
                 parser proto proxy qana qdisp qproc query
-                rproc sql
+                rproc sql sg
```
fritzm: Should we update to sphgeom for congruence with github version?

core/modules/css/Facade.cc
```diff
@@ -438,17 +445,20 @@ Facade::_tableIsSubChunked(string const& dbName,
...
-    return FacadeFactory::createMemFacade(f);
+    return FacadeFactory::createMemFacade(f, emptyChunkPath);
 }
 
 boost::shared_ptr<Facade>
-FacadeFactory::createMemFacade(std::istream& mapStream) {
+FacadeFactory::createMemFacade(std::istream& mapStream,
+                               std::string const& emptyChunkPath) {
     boost::shared_ptr<css::Facade> cssFPtr(new css::Facade(mapStream));
+    cssFPtr->_emptyChunks.reset(new EmptyChunks(emptyChunkPath));
```
fritzm: Would it be better to do this init internally in the facade ctors?

core/modules/css/SConscript.test
```diff
@@ -31,4 +31,10 @@ programs.append(p)
 utest = env.UnitTest(p)
 env.Append(UNIT_TESTS=utest)
 
+#extDeps = "log log4cxx boost_thread".split()
```
fritzm: remove dead line?

core/modules/czar/bin/startCzar.py
```diff
@@ -39,7 +39,7 @@
 from lsst.qserv.czar import app
 from lsst.qserv.czar import client
 from lsst.qserv.czar import config
-from lsst.qserv.czar import indexing
+#from lsst.qserv.czar import indexing
```
fritzm: remove dead lines?

core/modules/global/stringUtil.h
```diff
@@ -0,0 +1,69 @@
...
+namespace qserv {
+
+/// @return true if a string is safe enough to use as a name in our SQL dialect.
+bool
+inline isNameSafe(std::string::value_type const& c) {
+    std::locale loc;
+    if(std::isalnum(c)) { // Not sure that using the default locale is safe.
+        return true;
+    }
+    switch(c) { // Special cases. '_' is the only one right now.
```
fritzm: Maybe ';' as well (see http://xkcd.com/327/)

core/modules/qproc/ChunkSpec.cc
```diff
@@ -24,7 +24,7 @@
 /**
```
fritzm: non-standard file header (no copyright, etc?)

core/modules/qproc/ChunkSpec.cc
```diff
@@ -57,12 +60,116 @@ std::ostream& operator<<(std::ostream& os, ChunkSpec const& c) {
...
+
+bool ChunkSpec::operator<(ChunkSpec const& rhs) const {
+    typedef std::vector<int32_t>::const_iterator VecIter;
+
+    if(chunkId < rhs.chunkId) return true;
+    else if(chunkId > rhs.chunkId) return false;
+    else {
+        std::pair<VecIter,VecIter> mism
+            = std::mismatch(subChunks.begin(), subChunks.end(),
+                            rhs.subChunks.begin());
```
fritzm: are subChunks and rhs.subChunks guaranteed to be the same length?  (if rhs.subChunks is shorter, behavior of std::mismatch is I believe undefined?)

fritzm: (with the three arg version)

core/modules/qproc/ChunkSpec.cc
```diff
@@ -57,12 +60,116 @@ std::ostream& operator<<(std::ostream& os, ChunkSpec const& c) {
...
+        } else if(mism.second == rhs.subChunks.end()) {
+            return false;
+        } else { // both are valid;
+            return ((*mism.first) < (*mism.second));
+        }
+    }
+}
+
+bool ChunkSpec::operator==(ChunkSpec const& rhs) const {
+    typedef std::vector<int32_t>::const_iterator VecIter;
```
fritzm: dead line?

core/modules/qproc/ChunkSpec.cc
```diff
@@ -57,12 +60,116 @@ std::ostream& operator<<(std::ostream& os, ChunkSpec const& c) {
...
+
+    if(chunkId != rhs.chunkId) return false;
+    return subChunks == rhs.subChunks;
+}
+
+ChunkSpec ChunkSpec::makeFake(int chunkId, bool withSubChunks) {
```
fritzm: const method?

core/modules/qproc/ChunkSpec.h
```diff
@@ -42,20 +43,53 @@ namespace qproc {
...
-/// parsed query statement. Contains A specification of chunkId and subChunkId
+/// parsed query statement. Contains a specification of chunkId and subChunkId
 /// list.
+/// Do not inherit.
 struct ChunkSpec {
 public:
-    ChunkSpec() : chunkId(-1) {}
-    int32_t chunkId;
-    std::vector<int32_t> subChunks;
+    typedef std::vector<int32_t> Int32Vector;
```
fritzm: Maybe move to global/intTypes?

core/modules/qproc/ChunkSpec.h
```diff
@@ -42,20 +43,53 @@ namespace qproc {
...
+    Int32Vector subChunks;
+
     void addSubChunk(int s) { subChunks.push_back(s); }
     bool shouldSplit() const;
+
+    /// @return the intersection with the chunk.
+    /// If both ChunkSpecs have non-empty subChunks, but do not intersect, chunkId is set to be invalid (-1)
+    ChunkSpec intersect(ChunkSpec const& cs) const;
+    void restrict(ChunkSpec const& cs);
+//    void mergeUnion(ChunkSpec const& rhs);
```
fritzm: remove?

core/modules/wdb/QueryAction.cc
```diff
@@ -311,29 +314,33 @@ bool QueryAction::Impl::_dispatchChannel() {
...
+                } // TODO: may want to confirm (cheaply) that
+                // successive queries have the same result schema.
+                // Now get rows...
+                if(!_fillRows(res, numFields)) {
+                    erred = true;
+                }
+                _mysqlConn->freeResult();
+            } // Each query in a fragment
+        } // Each fragment in a msg.
+    } catch(sql::SqlErrorObject const& e) {
```
fritzm: Good enough for now, but could you add a "TODO fritzm" here to revisit/check this error handling strategy?  It's not clear to me that we want to throw all out of this loop on an error (rather than continuing on and reporting on all errors in the disptach via the multi-error?)

core/modules/wdb/QueryAction.cc
```diff
@@ -300,6 +300,9 @@ class ChunkResourceRequest {
     proto::TaskMsg const& _msg;
 };
 
+util::Error makeError(sql::SqlErrorObject const& e) {
```
fritzm: Should we make sql::SqlError a subclass of the new util::Error?  Then this would (may) not be necessary?

core/modules/qproc/geomAdapter.h
wangd: Ah, maybe you're right. Mmmm. okay, I'll put together a basic user-input related exception (or maybe find one I've forgotten about) and return that.

core/modules/qproc/fakeGeometry.h
wangd: I thought it might be useful, should we ever want to mock-out a dependence on sphGeom and do a more restricted test. No?

smonkewitz: I'm not sure how much of that code you could actually use if you wanted to do that. Also, I guess I don't see why you'd mock the geometry related classes. Isn't usage of the region types contained to the guts of IndexMap.cc, making IndexMap a more natural (and simpler) thing to exchange with a mock object?

core/modules/qproc/geomAdapter.h
wangd: ok

core/modules/qproc/geomAdapter.h
wangd: Okay, but I think I still want to catch what the constructor might throw and wrap it in a different exception so I don't have to teach any of the upper layers about what things might be thrown beneath them.

core/modules/qproc/fakeGeometry.h
wangd: Hmm. You're probably right. Okay,

core/modules/qproc/geomAdapter.h
wangd: Sure, as long as I can quote you on saying that we are unlikely (<10%?) to switch to something other than double.

core/modules/qproc/SecondaryIndex.cc
wangd: good catch. sorry about that.

core/modules/qproc/testIndexMap.cc
wangd: I actually would like to have IndexMap testing, but don't want to delay the merge. How about I put a TODO with a sketch of what I think should be tested, and how?

core/modules/qproc/IndexMap.cc
wangd: Looks that way. ok.

core/modules/qproc/IndexMap.cc
wangd: I don't think so. Originally, I hid them inside the body of getIntersect, but the compiler can't resolve the function names that way. I got the message that it had to be accessible in (more or less), a (qualified) global scope. Are anonymous namespaces okay?

core/modules/qproc/IndexMap.cc
wangd: right

core/modules/qproc/IndexMap.cc
wangd: I had logic to do that, but I think there are cases where it's not called, so let me fix that now.

core/modules/qproc/IndexMap.cc
wangd: This is a good point. I flip-flop between making logic "strict" and making it tolerant, and in this case, it's clear that we'll have less to debug if we make it more tolerant. Fixing...

core/modules/qproc/IndexMap.cc
wangd: I'm not sure I follow. How does a switch help? I thought the switch required an integer switch value, and here I'm choosing based on a string value.

core/modules/qproc/geomAdapter.h
smonkewitz: You can quote me :)

core/modules/qproc/IndexMap.cc
smonkewitz: I'd have to try it to be sure, but I'd be surprised if anonymous namespaces didn't work.

core/modules/qproc/IndexMap.cc
wangd: fixed, I think.

core/modules/SConscript
wangd: Not in this ticket. I'd rather make that a separate step.

core/modules/css/Facade.cc
wangd: I'll change it to that way. It doesn't matter much to me, so whatever looks better to the group as a whole.

core/modules/qproc/IndexMap.cc
smonkewitz: Sorry to be unclear - I meant the moral equivalent of a switch (just a bunch of if/else clauses).

core/modules/global/stringUtil.h
wangd: ;) we are really restrictive right now: only alnum + _, so any punctuation or quoting or escaping is sanitized out.

core/modules/qproc/ChunkSpec.cc
wangd: Wait, what? Which file is missing the copyright? (but yes, ChunkSpec.cc needed to be patched for AURA/LSST instead of LSST Corp)

core/modules/SConscript
jbecla: Please open a new ticket

core/modules/qproc/ChunkSpec.cc
wangd: Ah, good. I didn't notice that. Thanks!

core/modules/qproc/ChunkSpec.cc
wangd: It's a class static, so there's no implicit access to an instance for constness to apply to. Right?

core/modules/qproc/ChunkSpec.h
wangd: Added back in. This was a useful function to correct an oversight noticed earlier.

core/modules/qproc/IndexMap.cc
wangd: I reflowed the logic as the later comment suggested, so the code doesn't need these any more, so things should be fine.

core/modules/wdb/QueryAction.cc
wangd: I'll put in a comment in SqlErrorObject that we should consider this. At this point, we pay a couple lines of more code in exchange for a bit of independence and isolation.

core/modules/qproc/ChunkSpec.cc
fritzm: Oops, missed the static-ness -- you are correct.

core/modules/qproc/ChunkSpec.cc
fritzm: (Sorry, I missed the little accordion widget...)

PR #39, opened 2015-02-19
U/salnikov/DM-1618: Implement distributed database creation

fjammes:


admin/bin/qserv-admin.py
```diff
@@ -340,9 +337,79 @@ def _parseShow(self, tokens):
...
+
+            if possibleKeys is not None and key not in possibleKeys:
+                raise QservAdminException(QservAdminException.BAD_CMD,
+                   "Unrecognized option key '%s', possible keys: %s" % \
+                   (key, ' '.join(possibleKeys)))
+
+            kvList.append((key, value))
+
+        return kvList
+
```
fjammes: We may have to use a parser module if parsing will get more complex here? However, this would be done in an other ticket.

admin/python/lsst/qserv/admin/qservAdmin.py
```diff
@@ -486,7 +494,9 @@ def _createTable(self, options, dbName, tableName):
     def getNode(self, nodeName):
         """
         Returns node definition for specified node name. Returned dictionary has
-        keys defined in https://dev.lsstcorp.org/trac/wiki/db/Qserv/CSS#Node-related
+        keys defined in https://dev.lsstcorp.org/trac/wiki/db/Qserv/CSS#Node-related.
+        The state of the node (which does not have separate key in CSS) is returned
+        as 'state' key in the dictionary.
```
fjammes: There is also a css doc here: http://www.slac.stanford.edu/exp/lsst/qserv/2015_02/devel/css.html
All should be in the same place.
I prefer using sphinx so that the doc is up to date with the code, which is not true for Trac.
Is it possible to move merge the trac css doc with the one in sphinx?

admin/python/lsst/qserv/admin/workerMgmt.py
```diff
@@ -0,0 +1,175 @@
...
+                        node type are returned
+        """
+
+        nodes = self.selectDict(state, nodeType)
+
+        # convert to instances
+        return [WorkerAdmin(name=key, qservAdmin=self.css) for key, _ in nodes.items()]
+
+
+    def selectDict(self, state=None, nodeType=None):
```
fjammes: Why not selectNodes(), it seems clearer isn't it?

admin/python/lsst/qserv/admin/workerMgmt.py
```diff
@@ -0,0 +1,175 @@
...
+                # exception in case someone else managed to create database at this instant
+                db.createDb(dbName, mayExist=True)
+                nCreated += 1
+
+                if grantUser:
+                    grant = "GRANT ALL ON {0}.* to '{1}'@'{2}'"
+                    # TODO: grant ALL to user@localhost, we may also want to extend this
+                    # with user@127.0.0.1, or user@%, need to look at his again when I
+                    # have better understanding of deployment model
+                    hostmatch = ['localhost']
```
fjammes: restrict to local connection would be safer, if possible...

admin/tests/test_cssNodes.py
```diff
@@ -207,7 +218,125 @@ def testAddNode(self):
...
+        nodes = mgr.selectDict(state=qservAdmin.NodeState.ACTIVE, nodeType='backup')
+        self.assertEqual(len(nodes), 0)
+        nodes = mgr.selectDict(state=qservAdmin.NodeState.INACTIVE, nodeType='worker')
+        self.assertEqual(len(nodes), 0)
+
 ####################################################################################
 
 if __name__ == "__main__":
-    unittest.main()
+    suite = unittest.TestLoader().loadTestsFromTestCase(TestCssNodes)
```
fjammes: Good to have these use cases examples in test case, thanks!

admin/bin/qserv-admin.py
```diff
@@ -256,19 +262,10 @@ def _parseCreateNode(self, tokens):
...
-            key, value = opt.split('=', 1)
-
-            if key not in requiredKeys:
-                raise QservAdminException(QservAdminException.BAD_CMD,
-                   "Unrecognized option key '%s', possible keys: %s" % \
-                   (key, ' '.join(requiredKeys)))
+        # parse command, 'type' and 'state' are optional, rest is required
+        options = {'nodeName': tokens[0], 'type': 'worker', 'state': NodeState.ACTIVE}
 
+        for key, value in self._parseKVList(tokens[1:], requiredKeys):
```
fjammes: Some logging would be welcome here, at least in DEBUG level.

admin/python/lsst/qserv/admin/qservAdmin.py
```diff
@@ -561,9 +576,36 @@ def addNode(self, nodeName, nodeType, host, runDir, mysqlConn):
...
+            nodes = [nodes]
+
+        for node in nodes:
+            # check that node exists
+            nodeKey = '/NODES/' + node
+            if not self._kvI.exists(nodeKey):
+                raise QservAdminException(QservAdminException.NODE_DOES_NOT_EXIST,
+                                          'Node: ' + node, 'Key: ' + nodeKey)
+
+            # update its state
```
fjammes: Some logging would be welcome here, at least in DEBUG level.

admin/python/lsst/qserv/admin/workerMgmt.py
```diff
@@ -0,0 +1,175 @@
...
+        if state is not None:
+            nodes = [item for item in nodes if item[1].get('state') in state]
+        if nodeType is not None:
+            nodes = [item for item in nodes if item[1].get('type') in nodeType]
+
+        # make dict
+        return dict(nodes)
+
+
+    def createDb(self, dbName, grantUser=None, state=None, nodeType=None, **kwargs):
```
fjammes: Some logging would be welcome in this function, at least in DEBUG level.

admin/bin/qserv-admin.py
andy-slac: I do not expect anything more complex there but if other commands are added that use similar command structure we can think about defining some common parsing method.

admin/python/lsst/qserv/admin/qservAdmin.py
andy-slac: I don't think that sphinx is the right place for that, for me sphinx works better only for documentation that is release-specific. Editing sphinx docs and distributing them is much harder that wiki. Can you easily fix an error in sphinx documentation in the last 5 releases that we built?

admin/python/lsst/qserv/admin/workerMgmt.py
andy-slac: The name is because this method returns dict. All select methods (two of them for now) return "nodes", just in different representation, `select()` returns the set of `WorkerAdmin` instances, `selectDict()` returns dictionary where key is node name and value is another dict. 'selectNodes()` would be too generic name for this method. It is not private.

admin/python/lsst/qserv/admin/workerMgmt.py
andy-slac: Yes, I agree. Will figure it out eventually.

admin/bin/qserv-admin.py
andy-slac: I don't think that qserv-admin does any logging at all. I guess as a standalone app it should be easy to debug it if someone needs that.

admin/python/lsst/qserv/admin/qservAdmin.py
andy-slac: See my other comment. And I think kazoo will print a lot of info here if you enable logging for kazoo.

admin/python/lsst/qserv/admin/workerMgmt.py
andy-slac: I'll add something here.

PR #43, opened 2015-02-26
Improve strictness to reject queries with garbage after a valid query.

fritzm:
Extend the grammar to include a production for supported qserv queries, and
modify the code to use this production instead of a generic sql statement.
Extend ParseException to be constructed with more ANTLR exceptions

### DIFF
--- a/core/modules/parser/SqlSQL2.g
+++ b/core/modules/parser/SqlSQL2.g
@@ -309,6 +309,11 @@ boost::shared_ptr<VoidTwoRefFunc> _qservFctSpecHandler;
 // ---------------
 // Top-level rules
 // ---------------
+// Qserv-supported statement, requires end of file, no statement chaining
+qserv_stmt : 
+      select_stmt (SEMICOLON)* EOF
+    ;
+
 //{ Rule #--- <sql_stmt> is a proxy rule for all types of sql statements
 sql_stmt : 
 	  sql_data_stmt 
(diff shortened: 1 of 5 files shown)


core/modules/parser/SqlSQL2.g
```diff
@@ -309,6 +309,11 @@ boost::shared_ptr<VoidTwoRefFunc> _qservFctSpecHandler;
 // ---------------
 // Top-level rules
 // ---------------
+// Qserv-supported statement, requires end of file, no statement chaining
+qserv_stmt : 
+      select_stmt (SEMICOLON)* EOF
```
fritzm: Does this mean the semicolon is optional?  If so, is that intentional?

core/modules/qproc/testQueryAnalysis.cc
```diff
@@ -814,6 +814,17 @@ BOOST_AUTO_TEST_CASE(dm646) {
...
+    std::string stmt = "SELECT foo FROM Filter f limit 5";
+    std::string expected = "SELECT foo FROM LSST.Filter AS f LIMIT 5";
+    testAndCompare(qsTest, stmt, expected);
+
+    stmt = "SELECT foo from Filter f limit 5 garbage query !#$%!#$";
+    char const expectedErr[] = "ParseException:Parse token mismatch error:expecting EOF, found 'garbage':";
+    boost::shared_ptr<QuerySession> qs = testStmt3(qsTest, stmt, expectedErr);
+
+}
+
```
fritzm: Maybe add some regression checks for supported trailing semicolon scenarios, per comment above?

fritzm:
Looks good to me, modulo just the semicolon scenario questions above.

core/modules/parser/SqlSQL2.g
wangd: Yes, the semicolon is optional, and this is intentional. Jacek has noted to me that the semicolon is indeed optional, and is mainly useful as a statement separator

PR #47, opened 2015-03-13
u/fjammes/dm-1841

andy-slac:
Reviewing DM-1841

### DIFF
--- a/core/modules/util/MultiError.h
+++ b/core/modules/util/MultiError.h
@@ -30,8 +30,8 @@
  * @author Fabrice Jammes, IN2P3/SLAC
  */
 
-#ifndef LSST_QSERV_UTIL_ERROR_H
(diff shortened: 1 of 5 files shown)


core/modules/util/MultiError.h
```diff
@@ -31,7 +31,7 @@
  */
 
 #ifndef LSST_QSERV_UTIL_ERROR_H
-#define LSST_QSERV_UTIL_ERRORCONTAINER_H
+#define LSST_QSERV_UTIL_ERROR_H
```
andy-slac: Correct name for include guard should be `LSST_QSERV_UTIL_MULTIERROR_H`. Also check comment for `#endif`

fritzm:
Merged Mar 16, 2015 -- SHA 1cb0003

PR #49, opened 2015-03-18
Add IN2P3 cluster install procedure

jgates108:
- Add puppet script to install and
  configure Qserv master and worker
  at IN2P3

- Allow use to eups 'latestbuild'
  global tag to setup on cluster nodes
  a Qserv version built on cluster
  build node

- Add documentation

- Fix Qserv configuration procedure
  not to require a writable
  Qserv installation directory

### DIFF
--- a/admin/tools/cluster/cc-in2p3/install-puppet-node.sh
+++ b/admin/tools/cluster/cc-in2p3/install-puppet-node.sh
@@ -1,0 +1,139 @@
+#!/bin/bash
+
+# LSST Data Management System
+# Copyright 2014-2015 LSST Corporation.
+#
+# This product includes software developed by the
+# LSST Project (http://www.lsst.org/).
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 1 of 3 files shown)


admin/tools/cluster/cc-in2p3/install-puppet-node.sh
```diff
@@ -0,0 +1,128 @@
+#!/bin/bash
+
+# LSST Data Management System
+# Copyright 2014 LSST Corporation.
+#
```
jgates108: Copyright should be 2014-2015

admin/tools/cluster/cc-in2p3/install-puppet-node.sh
```diff
@@ -0,0 +1,128 @@
...
+# Edit variables below to specify defaults for a given cluster
+#
+################################################################
+MASTER='ccqserv100.in2p3.fr'
+SHARED_DIR=/sps/lsst/Qserv/stack/
+INSTALL_DIR=/qserv/stack
+QSERV_RUN_DIR=/qserv/qserv-run
+SETUP_OPTS='-t latestbuild -t qserv'
+
```
jgates108: Comment should be added that these are defaults for in2p3 cluster.

fjammes: ok

PR #50, opened 2015-03-18
Added unit tests for qdisp.

jgates108:


core/modules/SConscript
```diff
@@ -115,6 +115,7 @@ def getSrcs(mName):
     return map(lambda s: os.path.join(mName, s), srcs)
 
 def getTgts(mName):
+    print "Calling getTgts(%s)" % mName
```
fjammes: I would recommend to use the logger here. It can be called this way:
state.log.debug(DEBUG_MSG)

core/modules/qdisp/ExecStatus.cc
```diff
@@ -53,6 +53,8 @@ char const* ExecStatus::stateText(ExecStatus::State s) {
     case RESPONSE_ERROR: return "Response error";
     case RESPONSE_DATA: return "Retrieving response data";
     case RESPONSE_DATA_ERROR: return "Error retrieving response";
+    case RESPONSE_DATA_ERROR_OK: return "Error retrieving response session OK";
```
fjammes: Error OK is strange, could it be clarified please?

core/modules/qdisp/Executive.cc
```diff
@@ -168,6 +145,8 @@ void Executive::add(int refNum, Executive::Spec const& s) {
         return;
     }
     ExecStatus& status = _insertNewStatus(refNum, s.resource);
+    _empty.set(false);
+    LOGF_DEBUG("Flag set to false %1%" % _empty.get());
```
fjammes: log4cxx official documentation show a good way to fine tune the logger. An example is here: https://jira.lsstcorp.org/browse/DM-2373.
I'm not sure this should be done in the current ticket.

core/modules/qdisp/Executive.cc
```diff
@@ -194,9 +173,12 @@ bool Executive::join() {
     LOGF_INFO("Query exec finish. %1% dispatched." % _requestCount);
     _reportStatuses();
     if(sCount != _requestCount) {
-        LOGF_INFO("Query exec error:. %1% != %2%" % _requestCount % sCount);
+        LOGF_INFO("Query exec:. %1% != %2%" % _requestCount % sCount);
     }
-    return sCount == _requestCount;
+    bool empty = (sCount == _requestCount);
+    _empty.set(empty);
+    LOGF_DEBUG("Flag set to empty=%1% _empty=%2% sCount=%3% requestCount=%4%" %empty % _empty.get() % sCount % _requestCount);
```
fjammes: Here empty and _empty should be the same. Is it worth logging both?

core/modules/qdisp/Executive.cc
```diff
@@ -367,8 +355,11 @@ bool Executive::_shouldRetry(int refNum) {
     return true;
 }
 
-ExecStatus& Executive::_insertNewStatus(int refNum,
-                                        ResourceUnit const& r) {
+// Returning a reference instead of a shared pointer is an optimization.
+// It requires that the reference does not outlast the pointer in _statuses.
+// At the time of writing, everything in _statuses lasts until this Executive
+// is deleted.
+ExecStatus& Executive::_insertNewStatus(int refNum, ResourceUnit const& r) {
```
fjammes: I don't know if this kind of optimization is worth it, it may break easily and lead to complex bug in maintenance (whereas I've done it sometime). Maybe we should decide a standard way of doing this: for example only optimize shared pointer argument/return object after monitoring has shown it is worth it?

core/modules/qdisp/Executive.cc
```diff
@@ -470,7 +463,6 @@ void Executive::_waitUntilEmpty() {
         count = _requesters.size();
         _reapRequesters(lock);
         if(count != lastCount) {
-            lastCount = count;
```
fjammes: Here lastCount will always be equal to -1, isn't it?

core/modules/qdisp/QueryResource.cc
```diff
@@ -56,8 +56,12 @@ void QueryResource::ProvisionDone(XrdSsiSession* s) { // Step 3
     LOGF_INFO("Provision done");
     if(!s) {
         // Check eInfo in resource for error details
-        int code;
-        char const* msg = eInfo.Get(code);
+        int code = -1;
+        char const* message = eInfo.Get(code);
+        std::string msg = "no message from XrdSsi, code may not be reliable";
```
fjammes: What does this message mean, code is the XrdSsi return code, the xrootd source code, or the czar source code? Could you please clarify that?

core/modules/qdisp/TaskSpec.h
```diff
@@ -1,63 +0,0 @@
-// -*- LSST-C++ -*-
```
fjammes: This code was not used?

core/modules/qdisp/testQDisp.cc
```diff
@@ -0,0 +1,412 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the LSST License Statement and
+ * the GNU General Public License along with this program.  If not,
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+
+#include <string>
```
fjammes: Add

// System headers 

here

core/modules/qdisp/testQDisp.cc
```diff
@@ -0,0 +1,412 @@
...
+// Qserv headers
+#include "ccontrol/MergingRequester.h"
+#include "css/Facade.h"
+#include "global/ResourceUnit.h"
+#include "global/MsgReceiver.h"
+#include "qdisp/MessageStore.h"
+#include "qdisp/XrdSsiMocks.h"
+#include "util/threadSafe.h"
+
+// Quick and easy way to access private members for testing.
```
fjammes: Would it be better to define correct accessors, and document them?

core/modules/SConscript
wangd: I recommend removing the print statement.

core/modules/qdisp/testQDisp.cc
```diff
@@ -0,0 +1,412 @@
...
+	newInstance(int chunkId) {
+		boost::shared_ptr<ChunkMsgReceiverMock> r = boost::make_shared<ChunkMsgReceiverMock>();
+		r->_chunkId = chunkId;
+		return r;
+	}
+	int _chunkId;
+};
+
+// Simple functor for testing that _retryfunc has been called.
+class RetryTest : public util::VoidCallable<void> {
```
fjammes: Doxygen comment?

core/modules/qdisp/testQDisp.cc
```diff
@@ -0,0 +1,412 @@
...
+// This function is run in a separate thread to fail the test if it takes too long
+// for the jobs to complete.
+void timeoutFunc(qdisp::Executive &ex, int jobs, int millisecs) {
+	LOGF_INFO("timeoutFunc");
+	boost::this_thread::sleep(boost::posix_time::milliseconds(millisecs));
+	LOGF_INFO("timeoutFunc sleep over %1% empty=%2%" % millisecs % ex._empty.get());
+	BOOST_REQUIRE(ex._empty.get() == true);
+}
+
+BOOST_AUTO_TEST_CASE(Executive) {
```
fjammes: Overall remark: Good job, these mt tests seems complex to code. I didn't understand the very details of the logic but I hope I'll get in that interesting code in the future ;-)

core/modules/qdisp/ExecStatus.h
```diff
@@ -57,7 +57,8 @@ class ExecStatus {
                  PROVISION_OK, // ???
                  REQUEST, REQUEST_ERROR,
                  RESPONSE_READY, RESPONSE_ERROR,
-                 RESPONSE_DATA, RESPONSE_DATA_ERROR, RESPONSE_DATA_NACK,
+                 RESPONSE_DATA, RESPONSE_DATA_ERROR, RESPONSE_DATA_ERROR_OK,
+				 RESPONSE_DATA_ERROR_CORRUPT, RESPONSE_DATA_NACK,
```
wangd: How about putting RESPONSE_DATA_NACK on the same line where it used to be, and putting RESPONSE_DATA_ERROR_CORRUPT and RESPONSE_DATA_ERROR_OK  on one line together? And hopefully you can line things up as before.

core/modules/qdisp/Executive.cc
jgates108: This is Daniel's optimization and I felt the shift from shared pointer to return a plain old pointer needed a comment. The code, as is, is valid and faster than using shared pointers, so I am inclined to leave it. I think it's also beyond the scope of this ticket to make that change unless the code was incorrect.

core/modules/qdisp/Executive.cc
jgates108: Multiple threads are at work here. The message will always show the first time through, as lastCount = -1, but the _requesters.size can be changed by other threads, and that will trigger the log message to be printed.

core/modules/qdisp/QueryResource.cc
jgates108: eInfo.Get() is supposed to return and error message and set code to the error id from xrootd. If the returned value is NULL, xrootd (XrdSsi) did not supply an error message, and there's reason to believe 'code' is invalid. For unit testing, it is no surprise that xrootd doesn't have an error code available as we haven't actually asked it to do anything.

core/modules/qdisp/QueryResource.cc
```diff
@@ -75,7 +79,11 @@ void QueryResource::ProvisionDone(XrdSsiSession* s) { // Step 3
         _request = 0; // _session now has ownership
     } else {
         int code;
-        char const* msg = eInfo.Get(code);
+        char const* message = eInfo.Get(code);
+        std::string msg = "no message from XrdSsi, code may not be reliable";
+        if (message) {
+        	msg = message;
+        }
```
wangd: Seeing as there are two identical defensive checks against a null ptr returned from eInfo.Get(code), do you think it makes sense to put the logic in an inline function?

core/modules/qdisp/SConscript.test
```diff
@@ -0,0 +1,66 @@
...
+    targetFile.write(headerSfx + "\n")
+    targetFile.write("size_t %s_length = %d;\n" % (arrayName, length))
+    targetFile.close()
+    return None
+
+#bld = Builder(action = buildHeader,
+#              suffix = '.h',
+#              src_suffix = '.kvmap')
+#env.Append(BUILDERS = {'HeaderKvmap' : bld})
+#header = env.HeaderKvmap('testMap')
```
wangd: You can remove the Kvmap-related parts here (buildHeader, charArrayGenerator) because your test code has no need for a test instance of the CSS. I assume this was cribbed from SConscript.test, right? Maybe have a look and try to minimize extDeps so it only includes the libs that are necessary?

core/modules/qdisp/TaskSpec.h
jgates108: It compiles and runs without it, so it really looks that way.

core/modules/qdisp/testQDisp.cc
jgates108: Fritz and I talked about this, and I think we agree that this is an effective way of keeping private members and functions private, but still access them in the test to verify their values. It only applies within this one .cc file and avoids adding functions that will never be used (or should never be used) outside of the unit test.

core/modules/qdisp/ExecStatus.cc
jgates108: I can't say that I fully understand what is happening with this particular error. What it means to say is that there was an error or something unexpected in the response received from xrootd, but the xrootd session is still OK. The RESPONSE_DATA_ERROR_CORRUPT means that there was an error in the xrootd response and there is something wrong with the xrootd session. Really understanding this requires more xrootd knowledge than I currently have.

core/modules/qdisp/TaskSpec.h
wangd: It's used for people who use emacs. There is an LSST-DM emacs c++ mode that is pre-customized to LSST coding standards. I (as an emacs user) would prefer that it not be removed. The lsst emacs modes can be found at: https://dev.lsstcorp.org/cgit/LSST/DMS/devenv/build.git/

core/modules/qdisp/ExecStatus.h
jgates108: I shifted things around. The formatting errors are caused by the editor using tabs instead of spaces, which I've fixed.

core/modules/qdisp/testQDisp.cc
```diff
@@ -0,0 +1,412 @@
...
+	BOOST_CHECK(ex._empty.get() == true);
+	// test multiple threads
+	int threadCount = 10;
+	int jobsPerThread = 2;
+	std::vector<WorkerTest> workers(threadCount);
+	std::vector<boost::shared_ptr<boost::thread> > threads(threadCount);
+	// Run half with long sleep value.
+	int j = 0;
+	for(; j<threadCount/2; ++j) {
+		threads[j] = boost::make_shared<boost::thread>(&WorkerTest::operator(), &(workers[j]), boost::ref(ex), boost::ref(sequence),
```
wangd: Hey, some of these lines are pretty long (140 chars?). Standard 4-6 says 110 columns or less (I usually go for <80, but try not to complain unless they are really long.)

core/modules/qdisp/Executive.cc
fjammes: ok

core/modules/qdisp/Executive.cc
fjammes: So in my understanding lastCount is always equal to -1, and  _requesters.size() can-t be equal to -1, so the test is always true, isn't it?

core/modules/qdisp/ExecStatus.cc
fjammes: Maybe you can add a comment on this in the code?

core/modules/SConscript
fritzm: +1 for removing print statement -- now that it's debugged, how often are we going to want to see this?

core/modules/qdisp/testQDisp.cc
```diff
@@ -0,0 +1,412 @@
...
+	BOOST_CHECK(ex._empty.get() == true);
+	// test adding 4 jobs
+	duration = executiveTest(ex, sequence, chunkId, millis, 4);
+	jobs += 4;
+	LOGF_INFO("duration four %1%" % duration);
+	BOOST_CHECK(ex._empty.get() == true);
+	// test multiple threads
+	int threadCount = 10;
+	int jobsPerThread = 2;
+	std::vector<WorkerTest> workers(threadCount);
```
wangd: Wait, so you are starting a bunch of 'worker' threads that each add 1 or 4 jobs and then join on a single Executive? Executive::add() is not thread-safe or re-entrant. Executive::join() wasn't supposed to be either (though it might be). The intent of an Executive is to have a single manager thread create an instance, add a bunch of things to it, and then wait for things to finish. There should only be one manager thread. Other threads may peek into an Executive object's state. I feel like there is a much more complex usage being tested here. 
We guarantee that Executive never has multiple threads  calling add() or join(), because it is only responsible for dispatch for a single user query, and query analysis/generation is single-threaded. There shouldn't be any benefit(?) to performing add() with parallel threads because the main task of add() is to kick-off xrootd calls (which are async, non-blocking, and using an internal thread pool). The xrootd calls are not supposed to block for any reason.

core/modules/qdisp/testQDisp.cc
```diff
@@ -0,0 +1,412 @@
...
+	qdisp::MessageStore ms;
+	BOOST_CHECK(ms.messageCount() == 0);
+	ms.addMessage(123, 456, "test1");
+	std::string str("test2");
+	ms.addMessage(124, -12, str);
+	ms.addMessage(86, -12, "test3");
+	BOOST_CHECK(ms.messageCount() == 3);
+	BOOST_CHECK(ms.messageCount(-12) == 2);
+	qdisp::QueryMessage qm = ms.getMessage(1);
+	BOOST_CHECK(qm.chunkId == 124 && qm.code == -12 && 	str.compare(qm.description) == 0);
```
wangd: There is some weird spacing here.

core/modules/qdisp/TaskSpec.h
jgates108: The whole file wasn't used, not just the one line.

core/modules/qdisp/testQDisp.cc
```diff
@@ -0,0 +1,412 @@
...
+	LOGF_INFO("MessageStore test end");
+}
+
+BOOST_AUTO_TEST_CASE(QueryResource) {
```
wangd: Wouldn't it be better to mock the requester, so you don't need an executive, a chunkmsgreceiver, a merger, etc.? It's best not to involve an Executive instance if you don't have to.

core/modules/util/threadSafe.h
```diff
@@ -0,0 +1,101 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the LSST License Statement and
+ * the GNU General Public License along with this program.  If not,
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ *
+ *  @author: John Gates,
+ */
+
+#ifndef CORE_MODULES_UTIL_THREADSAFE_H_
```
wangd: Prefer: LSST_QSERV_UTIL_THREADSAFE_H (standard 4-9), like you did with XrdSsiMocks.h .

core/modules/qdisp/testQDisp.cc
fritzm: (If the lines are longer than 80, Daniel can't punch cards)
image

core/modules/qdisp/Executive.cc
jgates108: That line was not supposed to be deleted. Thanks.

core/modules/qdisp/QueryRequest.cc
```diff
@@ -58,7 +58,7 @@ inline void unprovisionSession(XrdSsiSession* session) {
 }
 
 ////////////////////////////////////////////////////////////////////////
-// QueryRequest::Canceller
+// QueryRequest::Canceler
 ////////////////////////////////////////////////////////////////////////
 class QueryRequest::Canceller : public util::VoidCallable<void> {
```
fritzm: I like the sp. fix, but maybe change the name of the actual function as well?

core/modules/qdisp/Executive.cc
```diff
@@ -141,25 +137,6 @@ class Executive::DispatchAction : public util::VoidCallable<void> {
     ExecStatus& _status; ///< Reference back to exec status
 };
 
-class NotifyExecutive : public util::UnaryCallable<void, bool> {
```
wangd: Prefer keeping NotifyExecutive inside .cc, but declaring static
factory function in .h
```
// .h: inside class Executive
static boost::shared_ptr<UnaryCallable<void, bool> >
newNotifier(Executive& e, int refNum);
// .cc
Executive::newNotifier(Executive& e, int refNum) {
 return NotifyExecutive::Ptr(new NotifyExecutive(e, refNum));
}
```
You can remove the static factory function inside NotifyExecutive then.

core/modules/qdisp/Executive.cc
```diff
@@ -181,7 +160,7 @@ void Executive::add(int refNum, Executive::Spec const& s) {
 bool Executive::join() {
     // To join, we make sure that all of the chunks added so far are complete.
     // Check to see if _requesters is empty, if not, then sleep on a condition.
-    _waitUntilEmpty();
+    _reapUntilEmpty();
```
wangd: prefer blockUntilDone instead of reapUntilEmpty. Reason:
waitUntilEmpty made sense to me, because the waiting was the most
important part. The reaping was a means of garbage collecting
previously-in-flight tasks. I understand that "wait" may not sound
appropriate because there is a meaning conflict, so how about "block",
because that's really what's going on, and "until done" because being
"done" is what's important, not that anything is empty or reaped. 

Your suggested "waitAll" is totally fine too.

core/modules/qdisp/Executive.cc
wangd: I think Fabrice is pointing out that you removed the update to lastCount from the loop, so the conditional is always true--it always acts as if the count is new information.

core/modules/qdisp/testQDisp.cc
wangd: Hey, those are some nice cards there. But really, I'm just faster at reading code in narrow windows and I like having the side-by-side comparison (for meld or other diff tools) fit on my small laptop screen with a "reasonable" font size. I'd rather sit on the porch and say 'get off my lawn' than scroll right and lose context, okay? ;)

core/modules/qdisp/testQDisp.cc
```diff
@@ -0,0 +1,412 @@
...
+#undef private
+
+namespace test = boost::test_tools;
+using namespace lsst::qserv;
+
+typedef util::Sequential<int> SequentialInt;
+
+struct TestFixture {
+    TestFixture(void) {
+        // To learn how to dump the map, see qserv/core/css/KvInterfaceImplMem.cc
```
fritzm: Comment here not relevant to test and could be reaped.  Actually, the whole test fixture is not used and could be reaped -- then just change BOOST_FIXTURE_TEST_SUITE to BOOST_TEST_SUITE below?

core/modules/qdisp/testQDisp.cc
```diff
@@ -0,0 +1,412 @@
...
+#include "boost/chrono.hpp"
+#include "boost/chrono/duration.hpp"
+#include "boost/make_shared.hpp"
+
+// LSST headers
+#include "lsst/log/Log.h"
+
+// Qserv headers
+#include "ccontrol/MergingRequester.h"
+#include "css/Facade.h"
```
fritzm: I think this include is not needed?

core/modules/qdisp/QueryRequest.cc
wangd: Please, just make it consistent whatever you pick. Personally, I prefer double ell--maybe it reflects my roots in electrical engineering (never seen "echo cancelation" or "noise cancelation"--it's always "echo cancellation" and "noise cancellation"), but my days are numbered anyway.

core/modules/util/threadSafe.h
```diff
@@ -0,0 +1,101 @@
+// -*- LSST-C++ -*-
```
fritzm: Could/should we use boost::atomic for this stuff instead?

core/modules/qdisp/QueryRequest.cc
fritzm: Yup, just consistent throughout.  I'm good with two ells.  (John sez he tried to put this back but it was more merge lossage.)

core/modules/util/threadSafe.h
jgates108: atomic doesn't have a condition variable. It might be good for Sequential and Flag, but not for FlagNotify.

core/modules/qdisp/testQDisp.cc
ktlim: Ideally the class could be tested solely via its public interface (but not by adding extra, unneeded accessor functions -- instead by seeing the effects of internal state changes).  But if you need access to internals for testing, isn't it better to provide a friend accessor class than this hack?

jgates108: With other changes, the number of private members being accessed dropped, and it was simplest to add a function to access the value of the remaining private element.

PR #51, opened 2015-03-24
Check error from QuerySession before continuing with UserQuery construction

wangd:
Should fix DM-2390


jbecla:
Force closing in github, it was merged on March 30, 2015, SHA 3d87bf5

PR #52, opened 2015-03-26
u/salnikov/dm-2176

andy-slac:
PR for reviewing and merging DM-2176 branch

### DIFF
--- a/admin/templates/configuration/etc/init.d/qserv-wmgr
+++ b/admin/templates/configuration/etc/init.d/qserv-wmgr
@@ -36,7 +36,7 @@ starttimeout=1
 # Disabling buffering in python in order to enable "real-time" logging.
 export PYTHONUNBUFFERED=1
 
-cmd="${PYTHON} ${start} -c ${config} >> ${logfile} 2>&1 &"
+cmd="${PYTHON} ${start} -c ${config} -v >> ${logfile} 2>&1 &"
 
 start_cmd() {
     start -p $pidfile -l $lockfile -t ${starttimeout} $NAME "${cmd}"
--- a/admin/templates/configuration/etc/init.d/qserv-wmgr
+++ b/admin/templates/configuration/etc/init.d/qserv-wmgr
@@ -1,0 +1,79 @@
+#!/bin/sh
+#
+# qserv-wmgr  This shell script takes care of starting and stopping
+#             Qserv worker management service.
+
+### BEGIN INIT INFO
+# Provides: qserv-wmgr
+# Short-Description: start and stop qserv worker management service
+# Description: qserv-wmgr is the Qserv worker management service \
+#              It provides RESTful HTTP interface which is a single \
+#              end-point for all worker communication and control.
+### END INIT INFO
+
+DIR=$(cd "$(dirname "$0")"; pwd -P)
+
+# Source functions library.
+. ${DIR}/qserv-functions
+
+NAME="qserv-wmgr"
+DAEMON=${NAME}
+
+source_sysconfig ${DIR}
+
+# Check variables which are not controlled by application
+check_writable ${NAME} "QSERV_RUN_DIR"
+check_writable ${NAME} "QSERV_PID_DIR"
+check_writable ${NAME} "QSERV_LOG_DIR"
+
+start=qservWmgr.py
+config=${QSERV_RUN_DIR}/etc/${NAME}.cnf
+pidfile=${QSERV_PID_DIR}/${NAME}.pid
+lockfile=${QSERV_RUN_DIR}/var/lock/subsys/${NAME}
+logfile=${QSERV_LOG_DIR}/${NAME}.log
+starttimeout=1
+
+# Disabling buffering in python in order to enable "real-time" logging.
+export PYTHONUNBUFFERED=1
+
+cmd="${PYTHON} ${start} -c ${config} >> ${logfile} 2>&1 &"
+
+start_cmd() {
+    start -p $pidfile -l $lockfile -t ${starttimeout} $NAME "${cmd}"
+    ret=$?
+    [ $ret -eq 0 ] || echo "See startup logfile : ${logfile}"
+    return $ret
+}
+
+stop_cmd() {
+    stop -p $pidfile -l $lockfile $NAME
+}
+
+# See how we were called.
+case "$1" in
+    start)
+        start_cmd
+        ;;
+    stop)
+        stop_cmd
+        ;;
+    restart)
+        stop_cmd
+        start_cmd
+        ;;
+    condrestart|try-restart)
+        if status_of_proc -p ${pidfile} "$DAEMON" "$NAME" >&/dev/null; then
+            stop_cmd
+            start_cmd
+        fi
+        ;;
+    status)
+        status_of_proc -p ${pidfile} "$DAEMON" "$NAME" && exit 0 || exit $?
+        ;;
+    *)
+        echo "Usage: $0 {start|stop|restart|reload|status|condrestart|try-restart}"
+        exit 2
+        ;;
+esac
+
+exit $?
--- a/admin/templates/configuration/etc/qserv-wmgr.cnf
+++ b/admin/templates/configuration/etc/qserv-wmgr.cnf
@@ -21,7 +21,7 @@ JSONIFY_PRETTYPRINT_REGULAR = False
 # ======== application-specific parameters =========
 
 #  database connection parameters, use None for defaults
-DB_HOST = 'localhost'
+DB_HOST = '127.0.0.1'
 DB_PORT = {{MYSQLD_PORT}}
 DB_SOCKET = '{{MYSQLD_SOCK}}'
 
--- a/admin/templates/configuration/etc/qserv-wmgr.cnf
+++ b/admin/templates/configuration/etc/qserv-wmgr.cnf
@@ -1,0 +1,41 @@
+#
+# configuration file for wmgrService, production version
+#
+
+# port number is determined by this
+SERVER_NAME = '127.0.0.1:{{WMGR_PORT}}'
+
+# interface to listen to, defaults to 127.0.0.1
+WMGR_INTERFACE = '0.0.0.0'
+
+# for production set both to 0
+DEBUG = 0
+TESTING = 0
+
+LOGGER_NAME = "wmgr"
+
+JSON_AS_ASCII = False
+JSON_SORT_KEYS = True
+JSONIFY_PRETTYPRINT_REGULAR = False
+
+# ======== application-specific parameters =========
+
+#  database connection parameters, use None for defaults
+DB_HOST = 'localhost'
+DB_PORT = {{MYSQLD_PORT}}
+DB_SOCKET = '{{MYSQLD_SOCK}}'
+
+# regular database account
+DB_USER = '{{QSERV_USER}}'
+DB_PASSWD = ''
+
+# special account used to create new databases
+DB_USER_PRIV = '{{MYSQLD_USER}}'
+DB_PASSWD_PRIV = '{{MYSQLD_PASS}}'
+
+# CSS-related parameters
+USE_CSS = True
+CSS_CONN = '{{QSERV_MASTER}}:{{ZOOKEEPER_PORT}}'
+
+# location of the run directory
+RUN_DIR = "{{QSERV_RUN_DIR}}"
--- a/admin/templates/installation/qserv-meta.conf
+++ b/admin/templates/installation/qserv-meta.conf
@@ -59,6 +59,12 @@ master = 127.0.0.1
 # master = qserv-master-example-name.mydomain.edu
 
 
+[wmgr]
+
+# Port number for worker management service
+port = 5012
+
+
 [zookeeper]
 
 # Port number for zookeeper server
--- a/core/modules/wmgr/SConscript
+++ b/core/modules/wmgr/SConscript
@@ -1,0 +1,10 @@
+import pytarget
+
+Import('env')
+Import('extraTgts')
+
+pyTargets = pytarget.getPyTargets(env, ["lsst","qserv","wmgr"], [])
+extraTgts['dist'].extend(pyTargets)
+
+for f in env.Glob("bin/*py"):
+    extraTgts['dist'].append(('bin', f))
--- a/core/modules/wmgr/bin/qservWmgr.py
+++ b/core/modules/wmgr/bin/qservWmgr.py
@@ -1,0 +1,85 @@
+#!/usr/bin/env python
+
+# LSST Data Management System
+# Copyright 2015 AURA/LSST.
+#
+# This product includes software developed by the
+# LSST Project (http://www.lsst.org/).
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
+
+"""
+Application which provides worker management service (HTTP-based).
+
+@author  Andy Salnikov, SLAC
+"""
+
+#--------------------------------
+#  Imports of standard modules --
+#--------------------------------
+import argparse
+import logging
+import sys
+
+#-----------------------------
+# Imports for other modules --
+#-----------------------------
+from flask import Flask
+from lsst.qserv.wmgr.config import Config
+from lsst.qserv.wmgr import dbMgr
+
+#----------------------------------
+# Local non-exported definitions --
+#----------------------------------
+
+#------------------------
+# Exported definitions --
+#------------------------
+
+def main():
+
+    parser = argparse.ArgumentParser(description='Single-node data loading script for Qserv.')
+
+    parser.add_argument('-v', '--verbose', dest='verbose', default=[], action='append_const',
+                        const=None, help='More verbose output, can use several times.')
+    parser.add_argument('-c', '--config', dest='configFile', default=None, metavar='PATH',
+                        help='Read configuration from provided file. By default configuration'
+                        ' is loaded from a file specified with env. variable WMGRCONFIG.')
+    args = parser.parse_args()
+
+    # configure logging
+    verbosity = len(args.verbose)
+    levels = {0: logging.WARNING, 1: logging.INFO, 2: logging.DEBUG}
+    level = levels.get(verbosity, logging.DEBUG)
+    fmt = "%(asctime)s [%(levelname)s] %(name)s: %(message)s"
+    logging.basicConfig(level=level, format=fmt)
+
+    # instanciate and configure app
+    app = Flask(__name__)
+    if args.configFile is not None:
+        app.config.from_pyfile(args.configFile)
+    else:
+        app.config.from_envvar('WMGRCONFIG')
+    Config.init(app)
+
+    # add few blueprints
+    app.register_blueprint(dbMgr.dbService, url_prefix='/dbs')
+
+    # run
+    app.run()
+
+
+if __name__ == '__main__':
+    sys.exit(main())
--- a/core/modules/wmgr/bin/qservWmgr.py
+++ b/core/modules/wmgr/bin/qservWmgr.py
@@ -78,8 +78,9 @@ def main():
     app.register_blueprint(dbMgr.dbService, url_prefix='/dbs')
     app.register_blueprint(procMgr.procService, url_prefix='/services')
 
-    # run
-    app.run()
+    # run, port number comes from SERVER_HOST configuration
+    host = app.config.get('WMGR_INTERFACE')
+    app.run(host)
 
 
 if __name__ == '__main__':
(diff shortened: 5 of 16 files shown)


admin/templates/configuration/etc/qserv-wmgr.cnf
```diff
@@ -0,0 +1,41 @@
...
+DB_PORT = {{MYSQLD_PORT}}
+DB_SOCKET = '{{MYSQLD_SOCK}}'
+
+# regular database account
+DB_USER = '{{QSERV_USER}}'
+DB_PASSWD = ''
+
+# special account used to create new databases
+DB_USER_PRIV = '{{MYSQLD_USER}}'
+DB_PASSWD_PRIV = '{{MYSQLD_PASS}}'
```
fjammes: I don't like to have password file written in multiple configuration file, this is a huge security issue. This is not the scope of this ticket but we should remember to fix it.

admin/templates/configuration/etc/init.d/qserv-wmgr
```diff
@@ -0,0 +1,79 @@
...
+#
+# qserv-wmgr  This shell script takes care of starting and stopping
+#             Qserv worker management service.
+
+### BEGIN INIT INFO
+# Provides: qserv-wmgr
+# Short-Description: start and stop qserv worker management service
+# Description: qserv-wmgr is the Qserv worker management service \
+#              It provides RESTful HTTP interface which is a single \
+#              end-point for all worker communication and control.
```
fjammes: I'm happy you use the general template, I think it'll ease maintenance.

admin/templates/installation/qserv-meta.conf
```diff
@@ -59,6 +59,12 @@ master = 127.0.0.1
 # master = qserv-master-example-name.mydomain.edu
 
 
+[wmgr]
+
+# Port number for worker management service
+port = 5012
```
fjammes: Good integration in the configuration framework. It shows you've understood it well. So I'm interested if you have ideas to improve it. But not in this ticket of course :-)

core/modules/wmgr/SConscript
```diff
@@ -0,0 +1,10 @@
+import pytarget
```
fjammes: it seems we have two tools to manage python modules: 
- site_scons/pytarget.py
and
- site_scons/site_tools/pymod.py (grep for InstallPythonModule)

core/modules/wmgr/bin/qservWmgr.py
```diff
@@ -0,0 +1,88 @@
...
+    parser = argparse.ArgumentParser(description='Single-node data loading script for Qserv.')
+
+    parser.add_argument('-v', '--verbose', dest='verbose', default=[], action='append_const',
+                        const=None, help='More verbose output, can use several times.')
+    parser.add_argument('-c', '--config', dest='configFile', default=None, metavar='PATH',
+                        help='Read configuration from provided file. By default configuration'
+                        ' is loaded from a file specified with env. variable WMGRCONFIG.')
+    args = parser.parse_args()
+
+    # configure logging
```
fjammes: I will re-use/generalize it in the ticket about logging.

core/modules/wmgr/python/dbMgr.py
```diff
@@ -0,0 +1,692 @@
...
+@dbService.route('', methods=['GET'])
+def dbs():
+    """ Return the list of databases """
+
+    _log.debug('request: %s', request)
+    _log.debug('dbs() GET => get database list')
+
+    # use non-privileged account, may limit the list of databases returned.
+    dbConn = Config.instance().dbConn()
+    dbs = dbConn.execCommandN('SELECT SCHEMA_NAME FROM INFORMATION_SCHEMA.SCHEMATA')
```
fjammes: It doesn't seem to have authentication to read in the database. Do you plan to do this? And how?

core/modules/wmgr/python/dbMgr.py
```diff
@@ -0,0 +1,692 @@
...
+    if not dbName:
+        raise ExceptionResponse(400, "MissingArgument", "Database name argument (db) is missing")
+
+    # validate it
+    _validateDbName(dbName)
+
+    # create database, use privileged account
+    dbConn = Config.instance().privDbConn()
+    try:
+        dbConn.createDb(dbName)
```
fjammes: It doesn't seem to have authentication to WRITE in the database. Do you plan to add this? And how?
I think we should add a minimalist pluggable authentication right now, flask seems to provide that.

https should also be used here. is it possible with Flask?

core/modules/wmgr/python/dbMgr.py
```diff
@@ -0,0 +1,692 @@
...
+    except db.DbException as exc:
+        _log.error('dbCreate(): exception when creating database %s: %s', dbName, exc)
+        if exc.errCode() == db.DbException.DB_EXISTS:
+            raise ExceptionResponse(409, "DatabaseExists", "Database %s already exists" % dbName)
+        raise
+
+    _log.debug('dbCreate(): database %s created', dbName)
+
+    # grant full access to non-privileged account
+    cmd = "GRANT ALL PRIVILEGES ON {0}.* TO '{1}'@'{2}'"
```
fjammes: Same remark: big security issue if the cluster run in a non-isolated/infiltrated environment. We should provide reasonable security as soon as possible.

core/modules/wmgr/python/dbMgr.py
```diff
@@ -0,0 +1,692 @@
...
+    return response
+
+
+@dbService.route('/<dbName>', methods=['DELETE'])
+def dbDelete(dbName):
+    """ Drop database """
+
+    _log.debug('request: %s', request)
+    _log.debug('request.form: %s', request.form)
+    _log.debug('dbDelete() DELETE => drop database')
```
fjammes: This is my prefered one :-) We can be in trouble if this database contains PB of data, and we should secure this as soon as possible.

core/modules/wmgr/python/dbMgr.py
```diff
@@ -0,0 +1,692 @@
...
+
+    dbData = []
+    for dbName in dbs:
+        if dbName not in _specialDbs:
+            dbData += [_dbDict(dbName)]
+
+    return json.jsonify(results=dbData)
+
+
+@dbService.route('', methods=['POST'])
```
fjammes: Flask seems to also allow variable passing from client to server using GET method, isn't it? But it seems you always use the POST method for that. Is there a reason?

core/modules/wmgr/python/dbMgr.py
```diff
@@ -0,0 +1,692 @@
...
+
+    _log.debug('dbCreate(): grants added')
+
+    # return representation for new database, 201 code is for CREATED
+    response = json.jsonify(result=_dbDict(dbName))
+    response.status_code = 201
+    return response
+
+
+@dbService.route('/<dbName>', methods=['DELETE'])
```
fjammes: Dropping the url and the database at the same time is very elegant.

core/modules/wmgr/python/dbMgr.py
```diff
@@ -0,0 +1,692 @@
...
+    dbData = []
+    for dbName in dbs:
+        if dbName not in _specialDbs:
+            dbData += [_dbDict(dbName)]
+
+    return json.jsonify(results=dbData)
+
+
+@dbService.route('', methods=['POST'])
+def dbCreate():
```
fjammes: Will it be possible to use that for distributed data loading at in2p3, it's very interesting.

core/modules/wmgr/python/dbMgr.py
```diff
@@ -0,0 +1,692 @@
...
+    _validateDbName(dbName)
+    _validateTableName(tblName)
+
+    # check the db exists (listTables() does not fail on non-existing databases)
+    dbConn = Config.instance().dbConn()
+    if not dbConn.tableExists(tblName, dbName):
+        raise ExceptionResponse(404, "TableMissing", "Table %s.%s does not exist" % (dbName, tblName))
+
+    # regexp matching chunk table names
+    # TODO: we need some central location for things like this
```
fjammes: I agree. xrootd/cmsd local server should tell us that, isn't it?

core/modules/wmgr/python/dbMgr.py
```diff
@@ -0,0 +1,692 @@
...
+
+@dbService.route('/<dbName>/tables/<tblName>/data', methods=['POST'])
+@dbService.route('/<dbName>/tables/<tblName>/chunks/<int:chunkId>/data', methods=['POST'])
+@dbService.route('/<dbName>/tables/<tblName>/chunks/<int:chunkId>/overlap', methods=['POST'])
+def dataLoad(dbName, tblName, chunkId=None):
+    """
+    Upload data into a table or chunk using the file format supported by mysql
+    command LOAD DATA [LOCAL] INFILE.
+
+    Not implemented yet.
```
fjammes: It seems promising :-)

core/modules/wmgr/python/dbMgr.py
```diff
@@ -0,0 +1,692 @@
...
+    _validateDbName(dbName)
+    _validateTableName(tblName)
+
+    _log.error('dataLoad(): method not implemented yet')
+    raise ExceptionResponse(501, "NotImplemented", "Data loading operations are not implemented yet")
+
+
+@dbService.route('/<dbName>/tables/<tblName>/index', methods=['GET'])
+@dbService.route('/<dbName>/tables/<tblName>/chunks/<int:chunkId>/index', methods=['GET'])
+def getIndex(dbName, tblName, chunkId=None):
```
fjammes: Interesting, this index will be quite big, do you plan to use this feature to create the full index?

core/modules/wmgr/python/procMgr.py
```diff
@@ -0,0 +1,146 @@
...
+
+    _log.debug('request: %s', request)
+
+    if service not in _svcNames:
+        raise ExceptionResponse(404, "InvalidService", "Invalid service name " + service)
+
+    action = request.form.get('action', '').strip()
+    if not action:
+        raise ExceptionResponse(400, "MissingArgument", "Action argument (action) is missing")
+    if action not in ['start', 'stop', 'restart']:
```
fjammes: Two test can be merged easily, isn't it?

core/modules/wmgr/python/xrdMgr.py
```diff
@@ -0,0 +1,271 @@
...
+            raise
+
+def _restartXrootd():
+    """ Restart xrootd process, only if it is running already """
+    initScript = os.path.join(Config.instance().runDir, 'etc/init.d/xrootd')
+
+    cmd = [initScript, 'status']
+    if _runCmd(cmd) == 0:
+        try:
+            cmd = [initScript, 'restart']
```
fjammes: I'm not sure xrootd init.d script returns correct error code.

core/modules/wmgr/test/wmgrDbTest.py
```diff
@@ -0,0 +1,228 @@
...
+#---------------------
+
+#-------------------------------
+#  Unit test class definition --
+#-------------------------------
+
+logging.basicConfig(level=logging.WARNING)
+
+
+class wmgrDbTest(unittest.TestCase):
```
fjammes: Nice example. But is it possible to generate a doc for each of the url this service provide? It would help to use it.

core/modules/wmgr/test/wmgrDbTest.py
```diff
@@ -0,0 +1,228 @@
...
+        tables = [obj['name'] for obj in response['results']]
+        self.assertEqual(tables, [])
+
+        # drop database
+        response = self._getJson(self.app.delete('/dbs/' + dbName))
+
+#
+#  run unit tests when imported as a main module
+#
+if __name__ == "__main__":
```
fjammes: Good job, I think this service will be very useful for Qserv data-loading, and eventually monitoring. You provide lots of major feature for data loading and it seems it can easily be extended. Thanks :-)
It lack a reasonable authentication mechanism, and also a autorisation mechanism (I don't want to multiply account but we might use an account for data loading and a restricted account for monitoring)

core/modules/wmgr/python/dbMgr.py
```diff
@@ -0,0 +1,692 @@
...
+@dbService.route('/<dbName>', methods=['DELETE'])
+def dbDelete(dbName):
+    """ Drop database """
+
+    _log.debug('request: %s', request)
+    _log.debug('request.form: %s', request.form)
+    _log.debug('dbDelete() DELETE => drop database')
+
+    # validate it
+    _validateDbName(dbName)
```
fjammes: This should be done in dbConn.dropDB(), not here. It seems also that MySQLdb could do that for you. Is it already the case in dbConn.dropDB()? If yes then _validateDbName should be removed.

core/modules/wmgr/python/dbMgr.py
fjammes: createDB should also check db name.

core/modules/wmgr/python/dbMgr.py
```diff
@@ -0,0 +1,692 @@
...
+
+    _log.debug('dbCreate(): database %s created', dbName)
+
+    # grant full access to non-privileged account
+    cmd = "GRANT ALL PRIVILEGES ON {0}.* TO '{1}'@'{2}'"
+    user = Config.instance().dbUser
+    hostnames = ['localhost']
+    for host in hostnames:
+        try:
+            dbConn.execCommand0(cmd.format(dbName, user, host))
```
fjammes: Are use and hostname checked for correct syntax? It may be worth doing it? Maybe not if they come from a controlled config file?

core/modules/wmgr/python/dbMgr.py
ktlim: I thought that our plan was to handle end-user authentication and authorization at the czar level, not the worker level.  All communication with the worker MySQL would be through a Unix or localhost socket and so would not generally require any sophisticated authentication.  While the inter-node networks are not isolated, the on-node connection would be.

admin/templates/configuration/etc/qserv-wmgr.cnf
andy-slac: Yes, I agree, we need more consistent and safer approach to authentication, we should discuss it at some point.

admin/templates/installation/qserv-meta.conf
andy-slac: I have too many ideas, most of them are not useful :)

core/modules/wmgr/SConscript
andy-slac: OK, open a ticket.

core/modules/wmgr/python/dbMgr.py
andy-slac: `Config.instance().dbConn()` opens connection and it does authentication using that user name and password from config file

core/modules/wmgr/python/dbMgr.py
andy-slac: GET cannot be used to modify resources. See https://confluence.lsstcorp.org/display/DM/REST+API+General+Guidelines (or any other RESTful introduction).

core/modules/wmgr/python/dbMgr.py
andy-slac: Oh, I guess your question was about authenticating clients with this web service. This is the next step (have no ticket for it yet), we must do it of course.

core/modules/wmgr/python/dbMgr.py
andy-slac: I guess this is again about authenticating clients of the web service. I don't believe something like OpenId is the answer here, we do not want to authenticate against remote server. I believe thing as simple as user name and password or application ID should be enough here.

andy-slac: createDB() is not required to check database name, and it does not say it will do it.

core/modules/wmgr/python/dbMgr.py
andy-slac: I assume that mysql account is already setup and has reasonable security (which we currently do not have).

core/modules/wmgr/python/dbMgr.py
andy-slac: User name comes from config file which I trust, host name is 'localhost'.

core/modules/wmgr/python/dbMgr.py
andy-slac: Not sure what kind of confirmation do you mean, this API is not for end-users. And it will be protected by reasonable auth like the rest of the API.

core/modules/wmgr/python/dbMgr.py
andy-slac: I don't want to trust it.

core/modules/wmgr/python/dbMgr.py
ktlim: Yes, we do need to make sure that the client of the web service is an appropriate one.  A shared secret would be OK for now, and maybe forever.

core/modules/wmgr/python/dbMgr.py
andy-slac: No, this is not xrootd interface, but mysql interface (it answers completely different question).

core/modules/wmgr/python/dbMgr.py
andy-slac: I think preferred mode will be to use per-chunk index, I hope that can fit in memory. Per-table index could probably be large, we may need either to disable it or find other way to return it (e.g. streaming without doing JSON conversion).

core/modules/wmgr/python/procMgr.py
andy-slac: My mistake, the line below should raise different exception (InvalidArgument instead of MissingArgument). I'll fix it.

core/modules/wmgr/python/xrdMgr.py
andy-slac: I think it works OK, at least in my tests.

core/modules/wmgr/test/wmgrDbTest.py
andy-slac: We will at some point, the question is only how to do it, generate from code or as a separate document.

core/modules/wmgr/test/wmgrDbTest.py
andy-slac: I'm not sure yet that we need to provide direct access to this "API". DAta loading for sure will have Python wrappers on top of that.

core/modules/wmgr/python/dbMgr.py
andy-slac: I opened DM-2419 to implement authentication/authorization, should be done next month, I hope

core/modules/wmgr/test/wmgrDbTest.py
andy-slac: Opened DM-2420 to handle documentation for web service.

admin/templates/configuration/etc/qserv-wmgr.cnf
fjammes: ok

admin/templates/installation/qserv-meta.conf
fjammes: It'll be worth discussing it at some point.

core/modules/wmgr/SConscript
fjammes: ok, see DM-2421

core/modules/wmgr/python/dbMgr.py
fjammes: ok fine, it's what I was talking about.

core/modules/wmgr/python/dbMgr.py
fjammes: A shared secret seems ok, but we ought to pass it using httpS, so that it remains secret.

core/modules/wmgr/python/dbMgr.py
fjammes: It was again again about authenticating clients of the web service. OK

core/modules/wmgr/python/dbMgr.py
fjammes: It was again about authenticating clients of the web service. OK.

core/modules/wmgr/python/dbMgr.py
fjammes: Ok thanks for this info, I was doing a mistake with HTTP POST and GET method which are different.

core/modules/wmgr/python/dbMgr.py
fjammes: ok

core/modules/wmgr/python/dbMgr.py
fjammes: ok, it's an interesting problem.

core/modules/wmgr/python/procMgr.py
fjammes: ok

core/modules/wmgr/test/wmgrDbTest.py
fjammes: Thanks

core/modules/wmgr/python/dbMgr.py
fjammes: Ok, but if everybody which use this function has to validate the table name, it leads to code duplicate and can be error-prone. A safe implementation of the validation of the table name in the dropDB function would be better for maintenance. We could rename dropDB to dropDBsafe() if needed.

core/modules/wmgr/python/dbMgr.py
fjammes: ok, hope it won't change :-)

admin/templates/configuration/etc/qserv-wmgr.cnf
jbecla: Why don't we rely on ~/.lsst/<file with credential> and pass it via read_default_file?

core/modules/wmgr/python/dbMgr.py
ktlim: We shouldn't pass the secret at all; we should use something like HTTP Basic Auth.

andy-slac: Do you mean digest auth? For basic we do need encryption, otherwise it's exposed.

ktlim: Oops, yes, got that wrong.  Digest.

PR #54, opened 2015-03-27
U/fjammes/dm 854

jbecla:


core/modules/global/stringUtil.h
```diff
@@ -65,5 +65,19 @@ sanitizeName(std::string const& name) {
     return out;
 }
 
+/**
+ * Convert a string to lowercase
+ *
+ * @param name input string to convert to lowercase
+ * @return the converted string
+ */
+inline std::string toLower(std::string const& name) {
```
andy-slac: I second the comment that we should not write stuff like this ourself and use boost::algorithm::to_lower(_copy).

core/modules/global/vectorUtil.h
```diff
@@ -0,0 +1,71 @@
...
+ * must provide an ouput operator.
+ * Examples:
+ * - works fine for std::vector<std::string>,
+ * - doesn't work for std::multimap.
+ *
+ * @param v an iterable data structure
+ * @return a string containing the serialized data structure
+ */
+template<typename Iterable>
+std::string toString(Iterable const& v) {
```
andy-slac: I'd recommend against defining toString() either as method (in Error classes) or standalone function. This feels like I'm reading Java :) I think C++ way is to prefer `operator<<(ostream, ...)`. And having both toString() and `operator<<(ostream, ...)` defined for the same class is unnecessary in most cases and only adds more lines of code to maintain.

core/modules/global/stringUtil.h
fjammes: ok

core/modules/global/vectorUtil.h
fjammes: I think It's not possible to define `operator<<(ostream, ...)` for std::vector<T>, is it?
Furthermore I don't think that having toString() add more lines of code, indeed if toString() contains only:

    std::ostringstream oss;
    oss << *this;
    return oss.str(); 

This avoid duplicating this code each time you want to put the object in a string, and so it leads to less lines of code if you use toString() more than two times.

fritzm:
Nice work, Fabrice!  Sorry I was late to this one.

core/modules/global/vectorUtil.h
andy-slac: Fabrice, anything is possible in C++, if you can define `toString()` for vectors it should also be possible to define some other construct to format vector into a stream :)

My objection to `toString()` is based on couple of of observations:
- most of the time in our code converting complex objects to string is done to push them to streams or to logging system (logging is also usually based of streams)
- methods like toString() are usually implemented using temporary streams.

So if you write code like `cout << toString(vector)` or `LOGF_DEBUG("vector: %1%" % toString(vector))` it is very inefficient because it creates temporary stream and temporary string(s).

To make it more efficient you have to define `operator<<()` which is implemented without using `toString()`. Then you could implement `toString()` based on `operator<<()` but I'd argue that you should avoid it. In case you really need to convert to string there semi-standard tools which already do the same for types that have `operator<<` defined (like `boost::lexical_cast`), but again most of the time you only need `operator<<` as you don't want to mess with strings.

If you want to know how to implement `operator<<` for vector (or any container) here is the sketch of what I would do (there might be simpler ways):

    namespace detail {
        template <typename Cont> struct _ContInserterHelper {
            const Cont& cont;
        };
        template <typename Cont> std::ostream& operator<<(std::ostream& out, const _ContInserterHelper<Cont>& cins) {
            out << "[";
            const Cont& cont = cins.cont;   // this is container itself
            // print container elements with separators
            return out << "]";
        }
    }
    template <typename Cont> detail::_ContInserterHelper<Cont> ContInserter(const Cont& cont) {
        return detail::_ContInserterHelper<Cont>{cont};
    }

And after that you can do:

    std::vector<int> v;
    std::cout << ContInserter(v);

And this has no overhead or any temporary objects created.

fritzm:
Merged Mar 31, 2015 -- SHA 9b55ebf

PR #57, opened 2015-04-03
Tickets/dm 2481

wangd:


fritzm:
Cool, looks good to me (though I think I need to sit down and sketch out an object lifetime/interaction diagram for Task, QueryAction, FormanImpl::Processor, et. al. to see why the reference wasn't working...)

PR #59, opened 2015-04-13
Implement data loading into mysql tables (DM-2385)

andy-slac:
Uses multipart request (just like web upload forms) to send both data
and options to a server. Present implementation uses named pipe to send
data to mysql server. Flask itself creates temporary files to store
uploaded files for requests >500kB so data still has to go through one
disk write and read on wmgr side.

### DIFF
--- a/core/modules/wmgr/python/dbMgr.py
+++ b/core/modules/wmgr/python/dbMgr.py
@@ -29,8 +29,14 @@ Module defining Flask blueprint for database management.
 #--------------------------------
 #  Imports of standard modules --
 #--------------------------------
+from contextlib import closing, contextmanager
+import gzip
 import logging
+import os
 import re
+import shutil
+import tempfile
+from threading import Thread
 
 #-----------------------------
 # Imports for other modules --
@@ -40,7 +46,9 @@ from .errors import errorResponse, ExceptionResponse
 from flask import Blueprint, json, request, url_for
 from lsst.db import db
 from lsst.qserv.admin.qservAdminException import QservAdminException
+import MySQLdb
 from MySQLdb.constants import FIELD_TYPE
+from werkzeug.urls import url_decode
 
 #----------------------------------
 # Local non-exported definitions --
@@ -123,9 +131,11 @@ dbService = Blueprint('dbService', __name__, template_folder='dbService')
 
 
 @dbService.errorhandler(db.DbException)
+@dbService.errorhandler(MySQLdb.Error)
+@dbService.errorhandler(MySQLdb.Warning)
 def dbExceptionHandler(error):
     """ All leaked DbException exceptions make 500 error """
-    return errorResponse(500, "db.DbException", str(error))
+    return errorResponse(500, error.__class__.__name__, str(error))
 
 @dbService.route('', methods=['GET'])
 def listDbs():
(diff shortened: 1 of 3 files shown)


core/modules/wmgr/python/dbMgr.py
```diff
@@ -604,8 +639,91 @@ def locadData(dbName, tblName, chunkId=None):
...
+
+    # named pipe to send data to mysql, make it in a temporary dir
+    tmpDir = tempfile.mkdtemp()
+    fifoName = os.path.join(tmpDir, 'tabledata.dat')
+    os.mkfifo(fifoName, 0600)
+
+    # do we need to uncompress?
+    compressed = _getArgFlag(options, 'compressed', None)
+    if compressed is None:
+        compressed = data.filename.endswith('.gz')
```
smonkewitz: (Just to satisfy my own curiosity) how is werkzeug deciding what the file name for table-data is?

core/modules/wmgr/python/dbMgr.py
```diff
@@ -589,14 +598,40 @@ def deleteChunk(dbName, tblName, chunkId):
...
 
+    def _copy(tableData, fifoName, compressed):
+        """ Helper method to run data copy in a separate thread """
+        if compressed:
+            src = gzip.GzipFile(fileobj=tableData.stream, mode='rb')
+            dst = open(fifoName, 'wb')
+            shutil.copyfileobj(src, dst)
+            _log.info('uncompressed table data to file %s', fifoName)
+            dst.close()
+            src.close()
```
smonkewitz: Would it be a bit safer to use context managers for src and dst here (in conjunction with contextlib.closing if necessary)?

core/modules/wmgr/python/dbMgr.py
```diff
@@ -604,8 +639,91 @@ def locadData(dbName, tblName, chunkId=None):
     _validateDbName(dbName)
     _validateTableName(tblName)
 
-    _log.error('method not implemented yet')
-    raise ExceptionResponse(501, "NotImplemented", "Data loading operations are not implemented yet")
+    # check that table exists
+    dbConn = Config.instance().dbConn()
+    if not dbConn.tableExists(tblName, dbName):
+        raise ExceptionResponse(404, "TableMissing", "Table %s.%s does not exist" % (dbName, tblName))
```
smonkewitz: Does this test make sense when `chunkId is None`?

core/modules/wmgr/python/dbMgr.py
andy-slac: `data.filename` is the name that client passes in a request, workzeug probably stores data in file with different name (or keeps it in memory).

core/modules/wmgr/python/dbMgr.py
```diff
@@ -604,8 +639,91 @@ def locadData(dbName, tblName, chunkId=None):
...
+
+        # do not need file any more
+        data.close()
+
+        # do not forget to delete temporary stuff
+        try:
+            _log.info('deleting temporary directory %s', tmpDir)
+            shutil.rmtree(tmpDir)
+        except Exception as exc:
+            _log.warning('failed to delete temporary directory %s: %s', tmpDir, exc)
```
smonkewitz: I think this is reasonable, but it may be worth considering the use of context managers for cleanup (so that as much as possible gets cleaned up if you do run into odd exceptions, e.g. inside this finally clause).

core/modules/wmgr/python/dbMgr.py
andy-slac: Yes, makes sense. I'm not very much used to `with` novelties. I'll update it.

core/modules/wmgr/python/dbMgr.py
andy-slac: Do you mean `chunkId is not None`? If `chunkId` is `None` then we load data into `tblName` and it must exist. If `chunkId` is not `None` then we load data into chunk table which has different name constructed below, but assumption here is that "original" table still exist as its schema is used to create chunk tables. The check is not critical, mostly to confirm that we are in a sane environment.

core/modules/wmgr/python/dbMgr.py
andy-slac: OK, I'll look into it, need to learn more about what context managers are there.

PR #62, opened 2015-04-21
Add support for wmgr client authentication (DM-2419).

andy-slac:
Implementation is based on modified Flask extension module
flask_httpauth. We support both Basic and Digest auth selectable via
Flask configuration. Qserv-configure is updated to generate secret file
in etc/wmgr.secret used by wmgr service, same file will be used by wmgr
clients, it has to be protected.


core/modules/wmgr/python/auth.py
```diff
@@ -0,0 +1,121 @@
...
+        _log.debug('wmgr auth type: %s', authType)
+        if authType not in ['none', 'basic', 'digest']:
+            raise RuntimeError("unexpected auth type in configuration: " + authType)
+
+        if authType == "none":
+
+            self._secret = None
+
+        else:
+
```
brianv0: There seems to be a little bit of extra whitespace around this code. Not sure if that's intentional.

core/modules/wmgr/bin/qservWmgr.py
```diff
@@ -73,13 +72,17 @@ def main():
...
+    config.Config.init(app)
 
     # add few blueprints
     app.register_blueprint(dbMgr.dbService, url_prefix='/dbs')
     app.register_blueprint(procMgr.procService, url_prefix='/services')
     app.register_blueprint(xrdMgr.xrdService, url_prefix='/xrootd')
 
+    # check authentication before every request
+    authCheck = auth.Auth(app.config)
+    app.before_request(authCheck.checkAuth)
```
brianv0: app.before_request is probably appropriate since I'm assuming the entire application should require authentication. If that's not the case, and if parts may be exposed without authentication, it may be worthwhile to use an [authentication decorator](http://flask.pocoo.org/docs/0.10/patterns/viewdecorators/).

core/modules/wmgr/python/auth.py
```diff
@@ -0,0 +1,121 @@
...
+                return self._auth.auth_error_callback()
+
+    @staticmethod
+    def readSecret(fileName):
+        """
+        Reads secret file and returns (user, password) tuple
+        """
+        _log.debug('reading wmgr secret from %s', fileName)
+        try:
+            return open(fileName).read().strip().split(':', 1)
```
brianv0: If this is an empty file or the file doesn't contain `:`, this returns a list with only one entry. That would definitely break line 108 but only at runtime. Maybe not a big deal.

brianv0: I was trying to find out the implications of an empty file/empty list, but I think they would be fine. My only concern was that an empty file might have the appearance that auth was enabled but non-authenticated users would always succeed, but I don't believe that to be the case with either digest auth or basic auth.

core/modules/wmgr/python/auth.py
andy-slac: That is just my personal style, spaces are supposed to make it more readable. Or maybe it's font in my editor which reads better with extra blank lines :)

core/modules/wmgr/bin/qservWmgr.py
andy-slac: Yes, auth is supposed to apply to all served resources, if we need resource-specific we can do it with decorators or looking at the request inside `checkAuth()` method. I actually started with adding decorators everywhere, but it's ugly and there is a chance you can miss a resource or two. 

flask_httpauth decorator is hard to use if you want to switch between basic/digest/hmac in configuration.

core/modules/wmgr/python/auth.py
andy-slac: I'll add a check for validity of file contents. You are right, it should stop immediately if file contents is not good.

PR #63, opened 2015-04-22
Changed protocol to break large messages into smaller pieces.

jbecla:


core/modules/rproc/InfileMerger.cc
```diff
@@ -136,7 +136,7 @@ class InfileMerger::Mgr {
 
     /// Enqueue an bundle of result rows to be inserted, as represented by the
     /// Protobufs message bundle
-    void enqueueAction(boost::shared_ptr<WorkerResponse> response);
+    void queMerge(std::shared_ptr<WorkerResponse> response);
```
wangd: You probably want to queueMerge or enqueueMerge

core/modules/util/common.h
```diff
@@ -93,6 +94,81 @@ std::ostream& printList(std::ostream &os, char const* label, C const& c) {
...
+/** Return a string showing the 'edge' of a list as integers.
+ * c is the container, edge is how many elements to include from
+ * the begging and end of the list.
+ */
+template <typename C>
+std::string prettyCharList(C const& c, uint edge) {
+    std::ostringstream os;
+    auto sz = c.size();
+    os << "[";
+    auto j = c.size();
```
wangd: I'll nitpick a little and recommend:
auto j = C::size_type;
or
auto j = sz;

core/modules/util/threadSafe.h
```diff
@@ -27,9 +27,12 @@
 #define LSST_QSERV_UTIL_THREADSAFE_H_
 
 // system headers
+#include <mutex>
+#include <thread>
+#include <utility>
 
 // external headers
-#include "boost/thread.hpp"
+//#include "boost/thread.hpp"
```
wangd: ok to remove this one.

core/modules/util/common.h
```diff
@@ -93,6 +94,81 @@ std::ostream& printList(std::ostream &os, char const* label, C const& c) {
     return os;
 }
 
+template <typename C>
```
wangd: These prettyChar* functions are fine, but they look prone to off-by-one errors. Would you add a test case that calls each of them once? I think a well-intentioned future dev might try to "clean up" things and introduce an error and spend way too long debugging it.

core/modules/wbase/Task.h
```diff
@@ -58,7 +58,7 @@ namespace wbase {
 /// Task encapsulates nearly zero logic, aside from:
 /// * constructors
 /// * poison()
-struct Task : public boost::noncopyable {
+struct Task {
```
wangd: Wait, what happened to the design that needed a change for Task to be copyable? There is/was some code that assumed that a Task wouldn't get copied. Ideally, the assuming code should've been documented (sorry that I forgot). The point may have been to ensure that setPoison/poison calls always get routed to the right Task instance and not another Task instance that looks almost the same.

jgates108: It still not copyable. You can now define copy constructor and equal operator = delete, which makes it non-copyable by prevent default constructors from being used. Lines 79 and 80.

core/modules/util/common.h
jgates108: I'm not sure this is the best place for them. qserv::global::debugUtil.h also looks like another reasonable place to put them. Any thoughts?

PR #65, opened 2015-04-23
Tickets/dm 2547: fix multi-node qservoss

wangd:
A couple of other things came along for the ride, but they are minor things.
* compile-time version embedding
* ResourceUnit test case and robustness
* removal of mysqldump dependency


core/modules/xrdoss/QservOss.h
```diff
@@ -33,10 +33,10 @@
 // System headers
 #include <deque>
 #include <set>
+#include <memory>
```
jbecla: Am i the only one liking nicely sorted includes?

andy-slac: I guess most people do not really care about order or order them by criteria other than alphabetic.

The coding standard item 4-10 (https://confluence.lsstcorp.org/pages/viewpage.action?pageId=16908674#C++Files-IncludeFilesandIncludeStatements) tells that we should order includes (in one group) by dependency first and alphabetically after that. I'm not sure how to guess the dependencies between headers, so we should just probably use alphabetic ordering.

wangd: I care. It was an oversight. I'll fix it.

core/modules/global/ResourceUnit.cc
```diff
@@ -159,26 +169,33 @@ ResourceUnit::_setFromPath(std::string const& path) {
...
+        if(t.token().empty()) { _unitType = GARBAGE; return; }
         _chunk = t.tokenAsInt();
         _ingestLeafAndKeys(t.token());
 
     } else if(rTypeString == prefix(RESULT)) {
         _unitType = RESULT;
+        if(_markGarbageIfDone(t)) { return; }
         t.next();
         _hashName = t.token();
+        if(_hashName.empty()) { _unitType = GARBAGE; return; }
```
fritzm: Hmm, it seems this could be cleaner/simpler...  What if you had a combined next + garbage checker/setter helpers, something like:

    bool NextTokenValidated(string* outToken, bool emptyAllowed=true);
    bool NextTokenValidated(int* outToken);

    const bool ALLOW_EMPTY = true;
    const bool DISALLOW_EMPTY = false;

or some such.  Then a lot of these three or four line clauses could be collapsed to single lines like:

    if (!NextTokenValidated(&tokenDest, DISALLOW_EMPTY)) { return; }

(and yeah, I like using pointers instead of refs for output parameters, since the then required "&" at the call site makes it obvious what's going on.  You do have to check 'em for null then in the helper though.  I think it's worth it for code readability, YMMV...)

fritzm: (Or, maybe this kind of triviality is not the best thing to be spending your time on right now, ha!  This has tests, is not so bad, and I'm fine if you want to roll this in.)

fritzm:
Merged Apr 30, 2015 -- SHA bb663c6

PR #69, opened 2015-05-04
Update default LOAD DATA options for data loader (DM-2679)

andy-slac:


admin/python/lsst/qserv/admin/dataLoader.py
```diff
@@ -811,8 +811,8 @@ def _loadOneFile(self, conn, database, table, file, csvPrefix):
         data = {'file': file}
         # need to know special characters used in csv
         # default delimiter is the same as in partitioner
-        special_chars = {'delimiter': '\t',
-                         'enclose':  '"',
+        special_chars = {'delimiter': '\\t',
```
andy-slac: I believe it's not doing quite what we want, the `conn.literal()` function below transforms this into 3-character string `['\\', '\\', 't']` and we want `['\\', 't']`. Maybe we should leave it for now as it was.

vaikuntht:
I reverted the delimiter change and merged to master. Done.

PR #70, opened 2015-05-04
Implement query cancellation

fritzm:
Details of this commit can be found in u/jbecla/debugKill.
It is a squashed mix of commits from Daniel and Jacek

### DIFF
--- a/core/modules/qdisp/QueryResource.h
+++ b/core/modules/qdisp/QueryResource.h
@@ -77,7 +77,10 @@ public:
 
 private:
     XrdSsiSession* _session; ///< unowned, do not delete.
-    QueryRequest* _request; ///< Owned temporarily, special deletion handling.
+
+    /// Owned temporarily, special deletion handling.
+    QueryRequest* _request;
+
     std::string const _payload; ///< Request payload
     boost::shared_ptr<ResponseRequester> _requester; ///< Response requester
     /// Called upon transaction finish
--- a/core/modules/qdisp/ResponseRequester.h
+++ b/core/modules/qdisp/ResponseRequester.h
@@ -39,6 +39,14 @@ namespace lsst {
 namespace qserv {
 namespace qdisp {
 
+/// ResponseRequester is an interface that handles result bytes. Tasks are
+/// submitted to an Executive instance naming a resource unit (what resource is
+/// required), a request string (task payload), and a requester (handler for
+/// returning bytes). The requester implements logic to process incoming results
+/// and buffers that are sized to the number of bytes expected in the next
+/// segment of results. The requester instance functions as a handle object that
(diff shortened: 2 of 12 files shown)


core/modules/qdisp/ResponseRequester.h
```diff
@@ -39,6 +39,14 @@ namespace lsst {
 namespace qserv {
 namespace qdisp {
 
+/// ResponseRequester is an interface that handles result bytes. Tasks are
+/// submitted to an Executive instance naming a resource unit (what resource is
+/// required), a request string (task payload), and a requester (handler for
+/// returning bytes). The requester implements logic to process incoming results
+/// and buffers that are sized to the number of bytes expected in the next
+/// segment of results. The requester instance functions also functions as a
```
fritzm: typo ("functions also functions")

core/modules/qdisp/ResponseRequester.h
```diff
@@ -101,8 +109,10 @@ class ResponseRequester {
         boost::shared_ptr<CancelFunc> f;
         {
             boost::lock_guard<boost::mutex> lock(_cancelMutex);
-            f.swap(_cancelFunc);
-            _cancelled = true;
+            if(!_cancelled) {
+                f = _cancelFunc;
```
fritzm: The swap() call above nulls out _cancelFunc.  The new code doesn't do this.  Was that an intentional change?

core/modules/qdisp/QueryResource.h
```diff
@@ -58,7 +58,7 @@ class QueryResource : public XrdSsiService::Resource {
                   ExecStatus& status)
         : Resource(rPath.c_str()),
           _session(NULL),
-          _request(NULL),
+          _request(),
```
fritzm: Why this change?  _request should be explicitly init'd since it is a raw pointer

core/modules/qdisp/ResponseRequester.h
jbecla: I believe Daniel removed that because it the _cancelFunc is called only in the _callCancel(), and inside _CallCancel() we now have a protection through the "_cancelled" flag, which ensures that we only call that function once.  But I agree it'll be safer to set the _cancelFunc to null, just as yet another level of protection. I'll make the change.

core/modules/qdisp/QueryResource.h
jbecla: I agree, I missed that, I am not sure what Daniel was thinking, but I'll fix it.

core/modules/qdisp/ResponseRequester.h
jbecla: I tried that, but I backed off, because for some reason it really upsets Qserv badly to the point it segfaults almost every time. Hmm.

PR #71, opened 2015-05-06
Tickets/dm 2698

fritzm:


core/modules/wdb/QueryAction.cc
```diff
@@ -141,9 +141,12 @@ class QueryAction::Impl::Poisoner : public util::VoidCallable<void> {
...
     Poisoner(std::shared_ptr<Impl> i) : _i(i) {}
     void operator()() {
-        _i->poison();
+        std::shared_ptr<Impl> iSharedPtr(_i.lock());
+        if (iSharedPtr) {
+            iSharedPtr->poison();
+        }
     }
-    std::shared_ptr<Impl> _i;
+    std::weak_ptr<Impl> _i;
```
fritzm: Maybe just add a comment saying that poisoners are potentially long lived, so that's why we are using a weak ptr?

jbecla: good idea, will do

PR #72, opened 2015-05-06
Fix mutex use before creation bug.

smonkewitz:
QueryPlugin was protecting its name to factory mapping with a static
mutex. As written, the constructor for that mutex was not guaranteed
to be called before static objects in other translation units started
using it to register plugin factory classes.

### DIFF
--- a/core/modules/qana/QueryPlugin.cc
+++ b/core/modules/qana/QueryPlugin.cc
@@ -34,10 +34,7 @@
 
 // System headers
 #include <map>
-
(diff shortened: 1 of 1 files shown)


fritzm:
Looks good to me!

[aside: I always wonder if this auto-registry-via-statics pattern is really even worth it whenever I see it...  Given that you have to rebuild to change it anyway, I feel a simple static list of the available plugins is simpler, easier to read, and more foolproof...]

PR #73, opened 2015-05-06
Fix DM-2708: QueryResource corrupted after log statement

smonkewitz:
QueryResource constructor was stashing a raw pointer to the string
buffer underlying one of its input arguments into a base class
member variable.  This failed unpredictably in the case where the
input argument was a temporary (depending on what code followed.)

Instead now QueryResource ctor copies the input data onto the
heap, then frees the heap allocation in its dtor.

### DIFF
--- a/core/modules/qdisp/QueryResource.h
+++ b/core/modules/qdisp/QueryResource.h
@@ -24,7 +24,10 @@
 #define LSST_QSERV_QDISP_QUERYRESOURCE_H
 
 // System headers
+#include <cstdlib>
+#include <new>
 #include <string>
+#include <string.h>
 
 // Third-party headers
 #include "boost/shared_ptr.hpp"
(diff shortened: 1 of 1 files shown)


smonkewitz:
Should you:
```
#include <string.h>
#include <cstdlib>
```
in QueryResource.h to avoid depending on other headers bringing these in?

core/modules/qdisp/QueryResource.h
```diff
@@ -56,17 +56,23 @@ class QueryResource : public XrdSsiService::Resource {
...
+          _status(status)
+    {
+        if (rName == NULL) {
+            throw std::bad_alloc();
+        }
     }
 
-    virtual ~QueryResource() {}
+    virtual ~QueryResource() {
+        free(const_cast<char *>(rName)); // clean up heap allocated resource path copy
```
smonkewitz: nitpick: `std::free` or `::free` (depending on whether you included `<cstdlib>` or `<stdlib.h>`)

fritzm: wilco

fritzm:
wilco on includes as well, thx

PR #74, opened 2015-05-07
U/fjammes/dm 2693

jbecla:


fritzm:
Per Fabrice above, decided not to merge.

PR #76, opened 2015-05-07
Include <unistd.h> for gcc 4.9.2

fritzm:
Build under gcc 4.9.2 was complaining about undefined close(),
read(), etc. in a few places.


smonkewitz:
Looks good.

PR #77, opened 2015-05-08
Implement Python client API for wmgr (DM-2593).

andy-slac:
Adding new module wmgr.client which implements Python client API
wrapping HTTP-based wmgr API. This is based on requests Python library,
currently we do not have separate requests package and use requests from
anaconda distribution, so I added anaconda for now as qserv dependency.
This dependency should be removed after we have separate requests
package installed.

There is one new unit test which only tests one small piece of the new
module (multi-part streaming data encoder). I have tested all methods in
interactive session, more testing will be done when I start to
reimplement data loader based on this new module.


fritzm:
Merged May 19, 2015 -- SHA da95be6

PR #83, opened 2015-05-19
Re-implement data loading using new wmgr service (DM-1901).

andy-slac:
Data loader now uses wmgr client library to perform all operation with
data. Changes to the script are mostry transparent, but script options
have changed, instead of scpecifying mysql connection options one has to
provide wmgr connection options. Config templates are extended to
include port number used by wmgr.

CSS structure describing worker nodes has changed as well, instead of
mysql connection parameters it now specifies wmgr port number.
qserv-admin interface has changed accordingly. NodeAdmin class which was
used as a ssh tunnel for mysql connection before now only provides
access to wmgr client instance.

WmgrClient inteface is extended with couple of methods to implement all
features needed by data loader.

This change requires corresponding changes in qserv_testdata packages,
please use same branch until these two packages are merged with master
branch in both repos.


admin/bin/qserv-admin.py
```diff
@@ -273,6 +274,12 @@ def _parseCreateNode(self, tokens):
         # check that all required options are there
         self._checkExist(options, requiredKeys)
 
+        # check obsolete keys
+        for opt in ('runDir', 'mysqlConn'):
+            if opt in options:
+                self._logger.warning(opt + ' option is obsolete')
+                del options[opt]
```
fjammes: I would exit with error here to prevent undefined behaviour of existing calling scripts?

admin/bin/qserv-data-loader.py
```diff
@@ -257,59 +246,17 @@ def run(self):
...
-            kws['unix_socket'] = self.args.mysqlSocket
-
-        try:
-            return mysql.Connection(**kws)
-        except Exception as exc:
-            logging.critical('Failed to connect to mysql database: %s', exc)
-            raise
-
-
-    def _workerDbConnect(self, nodeName, qservAdmin):
```
fjammes: qservAdmin could be renamed qservCssAdmin for better readability?

admin/bin/qserv-data-loader.py
```diff
@@ -146,8 +144,10 @@ def __init__(self):
                            help='Disable CSS updates.')
 
         group = parser.add_argument_group('Database options', 'Options for database connection')
-        group.add_argument('-H', '--host', dest='mysqlHost', default='localhost', metavar='HOST',
-                           help='Host name for czar mysql server, def: %(default)s.')
+        group.add_argument('-H', '--host', dest='czarHost', default='localhost', metavar='HOST',
+                           help='Host name for czar wmgr service, def: %(default)s.')
```
fjammes: It seems weird to have a wmgr service on the czar, could we rename wmgr to something else?

admin/bin/qserv-data-loader.py
```diff
@@ -146,8 +144,10 @@ def __init__(self):
...
+        group.add_argument('-P', '--port', dest='czarPort', default=5012, metavar='PORT_NUMBER', type=int,
+                           help='Port number to use for czar wmgr connection, def: %(default)s.')
```
fjammes: Could we stor czarPort in CSS?

admin/bin/qserv-data-loader.py
```diff
@@ -257,59 +246,17 @@ def run(self):
...
-            db = wAdmin.mysqlConn(**kws)
-            self._dbs.append(db)
-            db.connect()
-            # extract mysql connection object from Db, this is a hack to be used for now
-            # in the future we'll need to use Db API directly in data loader
-            return db._conn
-        except Exception as exc:
-            logging.critical('Failed to connect to mysql database: %s', exc)
-            raise
+            # no CSS, use node name as host name
```
fjammes: Is it usefull to support "no CSS" option?

admin/python/lsst/qserv/admin/dataLoader.py
```diff
@@ -681,102 +619,23 @@ def _createDummyChunk(self, database, table):
...
-                self._makeChunkAndOverlapTable(conn, database, table, 1234567890)
-            else:
-                # TODO: table is a actually a view, need something special. Old loader was
-                # creating new view just by renaming Table to Table_1234567890, I'm not sure
-                # this is a correct procedure. In any case here is the code that does it.
-                cursor = conn.cursor()
-                q = "RENAME TABLE {0}.{1} to {0}.{1}_1234567890".format(database, table)
-
-                self._log.debug('Rename view')
-                self._log.debug('query: %s', q)
```
fjammes: View support removed, isn't it? If yes, good to know :+1:

admin/python/lsst/qserv/admin/dataLoader.py
```diff
@@ -681,102 +619,23 @@ def _createDummyChunk(self, database, table):
...
-                cursor.execute(q)
+            # just make regular chunk with special ID, do not load any data
+            self._makeChunkAndOverlapTable(wmgr, database, table, 1234567890)
 
 
-    def _makeChunkAndOverlapTable(self, conn, database, table, chunkId):
+    def _makeChunkAndOverlapTable(self, wmgr, database, table, chunkId):
```
fjammes: Could we remove this function, which seems to only contains one line?

admin/bin/qserv-data-loader.py
andy-slac: I'll rename it to `css`, qserv and admin are not really relevant here. We'll refactor/rename whole QservAdmin thing in the future, I guess.

admin/python/lsst/qserv/admin/dataLoader.py
```diff
@@ -961,47 +796,45 @@ def _makeIndexMultiNode(self, database, table, metaTable, idxCol):
...
+            columns = (idxCol, 'chunkId', 'subChunkId')
+            indexData = wmgr.getIndex(database, table, chunkId=chunk, columns=columns)
 
-            q = "INSERT INTO {0} VALUES (%s, %s, %s)".format(metaTable)
-            while True:
-                seq = wCursor.fetchmany(1000000)
-                if not seq:
-                    break
-                czarCursor.executemany(q, seq)
+            # dump it into a in-memory file
```
fjammes: May not scale, isn't it? But enough for now.

admin/bin/qserv-admin.py
andy-slac: This is temporary thing to allow migration for the existing clients, after some time it will generate exception for obsolete/unknown options

admin/python/lsst/qserv/admin/nodeAdmin.py
```diff
@@ -21,12 +21,10 @@
...
 
-NodeAdmin is an interface which controls or communicates with qserv
-worker instance. Exact form of control is still to be defined, for now
-we are going to support communication with mysql instance or ability to
-start/stop worker processes. Current implementation is based on SSH
-tunneling (or direct mysql connection if possible), in the future there
-is supposed to be a special service for worker administration.
+NodeAdmin is a primary interface for qserv node administration. Currently
+NodeAdmin is a factory for WmgrClient instances which hides complexity
+of the configuring WmgrClient from various sources (CSS or anything else).
```
fjammes: "of the configuration of WmgrClient" instead?

admin/bin/qserv-data-loader.py
andy-slac: In some sense czar is also a worker as it does some work :) I don't want to do renaming at this point, too many things to change. Maybe when we have nothing else to do. For now people just need to remember that wmgr also manages czar. Or think of wmgr as "world manager" :)

admin/bin/qserv-data-loader.py
andy-slac: Data loader can run without CSS, I do not want CSS to be prerequisite for data loader (yet). For multi-node data loader we actually can use worker definition from CSS, but this is also not a strict requirement.

admin/python/lsst/qserv/admin/nodePool.py
```diff
@@ -129,7 +128,7 @@ def success(p):
                 process_id += 1
                 node = remaining_nodes.pop()
 
-                task = node.getSshCmd(command);
+                task = node.getCommand(command)
```
fjammes: Do I need to test the parallel ssh command in this PR?

admin/templates/configuration/etc/qserv-client.conf
```diff
@@ -58,3 +58,6 @@ sock={{MYSQLD_SOCK}}
 
 port={{ZOOKEEPER_PORT}}
 
+[wmgr]
+port = {{WMGR_PORT}}
+secret = %(qserv_run_dir)s/etc/wmgr.secret
```
fjammes: Did you implement wmgr.secret in configuration script?

admin/bin/qserv-data-loader.py
andy-slac: Mostly for testing. Again, currently data loader does not require CSS to be there, you can load data without CSS just fine. I hope that does not complicate things too much.

admin/tests/test_ssh.py
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env python
```
fjammes: Why renaming this file?

admin/python/lsst/qserv/admin/dataLoader.py
andy-slac: OK

admin/tests/test_ssh.py
```diff
@@ -0,0 +1,110 @@
...
+        self.sock.connect((host, port))
+    def echo(self, data):
+        self.sock.send(data)
+        data = self.sock.recv(1024)
+        return data
+    def close(self):
+        self.sock.close()
+
+
+class TestWorkerAdmin(unittest.TestCase):
```
fjammes: Is it still related to WorkerAdmin, it seems it's ssh only now?

admin/python/lsst/qserv/admin/dataLoader.py
andy-slac: Yes, there is a potential problem. But it is in the memory already, hope is that data size is not too large to make another copy formatted differently.

admin/templates/configuration/etc/qserv-client.conf
andy-slac: Yes, qserv-configure creates that file with some random user/password.

admin/tests/test_ssh.py
andy-slac: Yes, I missed that, will rename it.

core/modules/wmgr/python/dbMgr.py
```diff
@@ -316,36 +325,19 @@ def createTable(dbName):
...
-        if idx < 0:
-            _log.error('schema is missing opening parenthesis: %s', schema)
-            raise ExceptionResponse(400, "InvalidArgument",
-                                    "CREATE TABLE statement has no open parenthesis")
-        schema = schema[idx:]
-        _log.debug('schema from request: %s', schema)
+        # assume that we have correct SQL already
+        try:
+            dbConn.useDb(dbName)
+            dbConn.execCommand0(schema)
```
fjammes: execCommand0() is not a very explicit function name, could you please improve it?

admin/python/lsst/qserv/admin/nodeAdmin.py
andy-slac: OK

core/modules/wmgr/python/dbMgr.py
andy-slac: That method is provided by `db` package, I cannot improve that. But `db` package is in the process of being rewritten, so there will be improvements.

core/modules/wmgr/python/dbMgr.py
```diff
@@ -517,17 +588,42 @@ def createChunk(dbName, tblName):
...
 
-        # make table using DDL from non-chunked one
-        query = "CREATE TABLE {2}.{0} LIKE {2}.{1}".format(chunkTable, tblName, dbName)
-        _log.debug('make chunk table: %s', chunkTable)
-        try:
+        # check if this table is a ctually a view
+        if dbConn.isView(tblName, dbName):
+            
+            # view needs more complicated algorithm to copy its definition, first copy
+            # its current definition, then rename existing view and then re-create it again
```
fjammes: Would i be possible to create chunkTable as a view on tblName view to have simpler algorithm?

core/modules/wmgr/python/dbMgr.py
```diff
@@ -810,7 +944,7 @@ def getIndex(dbName, tblName, chunkId=None):
             if not rows:
                 break
             for row in rows:
-                result.append(list(row))
+                result.append(tuple(row))
```
fjammes: It would be fine to have a documented, high-level picture of data-loader architecture. Indeed it becomes quite complex now. Could you do it soon in a new ticket please?

core/modules/wmgr/python/dbMgr.py
andy-slac: As long as this complicated algorithm produces correct result I do not see any problem with it. And it's not too complicated to understand, it's just a bit more complex than a single SQL statement.

core/modules/wmgr/python/dbMgr.py
andy-slac: Is this documentation enough http://www.slac.stanford.edu/exp/lsst/qserv/2015_05/admin/data-loader.html or do you want internals documented as well?

admin/python/lsst/qserv/admin/dataLoader.py
andy-slac: It's done in wmgr now, data loader does not need to care about that.

admin/python/lsst/qserv/admin/nodePool.py
andy-slac: Would be nice to test it, I do not know how to do that.

core/modules/wmgr/python/dbMgr.py
fjammes: Here, I'd like to have an additional documentation page which summarize data-loading architecture, where does related services run, which protocols are used, who authenticate where, and a hogh-level description of the workflow. I think it could be done shortly with some well-documentend figures?

admin/python/lsst/qserv/admin/nodePool.py
fjammes: ok I'll do it, FYI it"s (briefly) documented here: http://www.slac.stanford.edu/exp/lsst/qserv/2015_05/HOW-TO/cluster-deployment.html#deploy-on-the-cluster

core/modules/wmgr/python/dbMgr.py
fjammes: OK, however it seems the view still point on the (empty) XXX table whereas it should point on the XXX_chunkId table, isn't it? Here we have to guess XXX and rename it XXX_chunkId, isn't it?

admin/bin/qserv-admin.py
fjammes: ok

admin/bin/qserv-data-loader.py
fjammes: OK, but Qserv isn't the whole world, isn't it?

admin/bin/qserv-data-loader.py
fjammes: ok

admin/bin/qserv-data-loader.py
fjammes: ok

admin/bin/qserv-data-loader.py
fjammes: ok

admin/python/lsst/qserv/admin/dataLoader.py
fjammes: ok

core/modules/wmgr/python/dbMgr.py
fjammes: ok

core/modules/wmgr/python/dbMgr.py
andy-slac: That needs a new ticket. Note also that architecture is likely going to change in the future when we have data distribution mechanism in place which is supposed to take over some of the current functions of data loader.

core/modules/wmgr/python/dbMgr.py
andy-slac: In that sense yes, we do not change view definition at this point and we probably should. But it depends entirely on how view is defined which may be complicated. I guess this part is not complete (I only copied whatever old data loader was doing and did not try to improve it) and it needs separate ticket (and an example of the view definition).

admin/bin/qserv-data-loader.py
andy-slac: No, for us there is nothing else in the world but qserv :) Do you have a better name than {{wmgr}}?

core/modules/wmgr/python/dbMgr.py
andy-slac: I opened DM-2817 for this.

core/modules/wmgr/python/dbMgr.py
andy-slac: DM-2818 is opened for this task

admin/bin/qserv-data-loader.py
fjammes: ok, I have no better name yet, sorry...

core/modules/wmgr/python/client.py
```diff
@@ -206,13 +207,6 @@ class ServerResponseError(ClientException):
     def __init__(self, msg, *args):
         ClientException.__init__(self, 'Server response error: ' + msg, *args)
 
-class DataFileError(ClientException):
-    """
-    Exception raised for errors in data file operations
-    """
-    def __init__(self, fileName, *args):
-        ClientException.__init__(self, 'Data file error: ' + fileName, *args)
-
```
vaikuntht: If this is removed, how are file operation errors handled now?

core/modules/wmgr/python/client.py
```diff
@@ -434,7 +471,9 @@ def loadData(self, dbName, tableName, dataFile, chunkId=None, overlap=False,
 
         @param dbName:     Database name
         @param tableName:  Table name
-        @param dataFile:   Path name for file with data
+        @param dataFile:   File object for file with data
+        @param fileName:   File name to use, if data is compressed then pass name which
+                           ends with .gz or set compressed argument to True.
```
vaikuntht: I'm not sure I understand the distinction between dataFile and fileName.

core/modules/wmgr/python/dbMgr.py
```diff
@@ -275,6 +279,10 @@ def createTable(dbName):
     schemaSource:   source for schema name, possible values: 'request', 'css',
                     (default: 'request')
     schema:         complete "CREATE TABLE ..." statement (optional)
+    chunkColumns:   boolean flag, false by default, accepted values:
+                    '0', '1', 'yes', 'no', 'false', 'true'. If set to true then
+                    delete colums "_chunkId", "_subChunkId" from table (if they
```
vaikuntht: Typo: colums -> columns

core/modules/wmgr/python/dbMgr.py
vaikuntht: Adding to this conversation, I have been using execCommandN() for the integration tests and was wondering if it is redundant to have a function that executes N lines vs. just 1 line. execCommandN() should be sufficient for all cases, no?

core/modules/wmgr/python/dbMgr.py
```diff
@@ -359,19 +351,51 @@ def createTable(dbName):
             raise ExceptionResponse(500, "CSSError", "Failed to retrieve table schema from CSS", str(exc))
 
         # schema in CSS is stored without CREATE TABLE, so we are already OK
-
-    # create table
```
vaikuntht: Maybe this comment should be retained .

core/modules/wmgr/python/dbMgr.py
```diff
@@ -359,19 +351,51 @@ def createTable(dbName):
...
+        # Change table schema, drop _chunkId, _subChunkId, add chunkId, subChunkId
+
+        cursor = dbConn._conn.cursor()
+
+        try:
+            table = '`{0}`.`{1}`'.format(dbName, tblName)
+
+            # get current column set
+            q = "SHOW COLUMNS FROM %s" % table
+            cursor.execute(q)
```
vaikuntht: Could dbConn.execCommand0() not have been used here?

core/modules/wmgr/python/dbMgr.py
```diff
@@ -517,17 +588,42 @@ def createChunk(dbName, tblName):
     _log.debug('will create tables: %s', tables)
     for tblType, chunkTable in tables.items():
 
-        # make table using DDL from non-chunked one
-        query = "CREATE TABLE {2}.{0} LIKE {2}.{1}".format(chunkTable, tblName, dbName)
-        _log.debug('make chunk table: %s', chunkTable)
-        try:
+        # check if this table is a ctually a view
```
vaikuntht: Typo: a ctually -> actually

core/modules/wmgr/python/client.py
andy-slac: This exception was used for errors when opening file, now file is opened by client code, so errors in file opening will have to be handled by client.

core/modules/wmgr/python/client.py
andy-slac: dataFile is a file object (the one returned from open() for example) representing actual data to be sent. fileName is a string with the file name, file name is not used to open the file, it is just sent to wmgr service so it can do something smart based on file name (e.g. uncompress it if it sees that name ends with .gz and you did not pass "sompressed" flag).

core/modules/wmgr/python/dbMgr.py
andy-slac: thanks

core/modules/wmgr/python/dbMgr.py
andy-slac: Indeed, `execCommandN()` could be used for anything. You can think of `execCommand0()` is just an optimization on top of `execCommandN()`, the difference is that `execCommand0()` does not care about result returned from query, while `execCommandN()` returns full result as list of rows (and `execCommand1()` returns single row).

core/modules/wmgr/python/dbMgr.py
andy-slac: OK

core/modules/wmgr/python/dbMgr.py
andy-slac: No, `execCommand0()` does not return anything :) But we could use `execCommandN()` here indeed. It's mostly historical, copied from the code which used cursor exclusively. This piece of code will have to be changed anyway when we switch to new `db` implementation based on sqlalchemy.

core/modules/wmgr/python/dbMgr.py
andy-slac: thanks

PR #84, opened 2015-05-21
Call my_thread_end from threads that have called my_init.

smonkewitz:
This is accomplished from the destructor of a thread local
object.

### DIFF
--- a/core/modules/mysql/MySqlConnection.cc
+++ b/core/modules/mysql/MySqlConnection.cc
@@ -30,51 +30,39 @@
 
 // Third-party headers
 #include "boost/lexical_cast.hpp"
+#include "boost/thread/tss.hpp"
 
 // Qserv headers
 #include "mysql/MySqlConfig.h"
 
 
-namespace { // File-scope helpers
-inline void killMySql(MYSQL* mysql, bool useThreadMgmt) {
-    mysql_close(mysql);
-    // Dangerous to use mysql_thread_end(), because caller may belong to a
-    // different thread other than the one that called mysql_init(). Suggest
-    // using thread-local-storage to track users of mysql_init(), and to call
-    // mysql_thread_end() appropriately. Not an easy thing to do right now, and
-    // shouldn't be a big deal because we thread-pool anyway.
+namespace {
+    // A class that calls mysql_thread_end when an instance is destroyed.
+    struct MySqlThreadJanitor {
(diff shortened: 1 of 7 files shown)


core/modules/mysql/MySqlConnection.cc
```diff
@@ -161,24 +149,41 @@ MySqlConnection::selectDb(std::string const& dbName) {
...
+    static std::once_flag initialized;
+    static boost::thread_specific_ptr<MySqlThreadJanitor> janitor;
+
+    struct InitializeMysqlLibrary {
+        void operator()() {
+            int rc = mysql_library_init(0, NULL, NULL);
+            assert(0 == rc);
+            if (!janitor.get()) {
+                janitor.reset(new MySqlThreadJanitor);
+            }
```
fritzm: We wouldn't really expect janitor.get() != NULL, because of the call_once below, right?  Maybe this should be an assert instead of an if?

fritzm:
In general, I like this as cleanup even though its not helping a current leak.  Certainly nice to get rid off the useThreadMgmt stuff.

core/modules/mysql/MySqlConnection.cc
```diff
@@ -161,24 +149,41 @@ MySqlConnection::selectDb(std::string const& dbName) {
...
+    // when the calling thread exists. We do this by allocating a thread
+    // local object that calls mysql_thread_end from its destructor.
+    //
+    // C++11 thread_local storage cannot be used until the baseline compiler is
+    // GCC 4.8, so we must fall back to boost. A significant caveat from the
+    // boost documentation is that "on some platforms, cleanup of thread-
+    // specific data is not performed for threads created with the platform's
+    // native API". Fortunately, even though we are creating threads via
+    // the STL and not boost, cleanup seems to be performed on POSIX systems.
+    // However, this likely does not work on Windows.
```
fritzm: Could you add a "// TODO: rework with std tls once we have it" line at the end of this block comment, so this tech. debt comes up on a TODO grep?

smonkewitz: Good idea.

core/modules/mysql/MySqlConnection.cc
smonkewitz: That's right - I'll make it an assert.

PR #86, opened 2015-05-29
Add PID to czar logs

andy-slac:


core/modules/czar/bin/startCzar.py
```diff
@@ -98,6 +98,7 @@ def main():
     config.printTo(sys.stdout)
 
     # Configure logging
+    newlog.MDCPutPid()
```
andy-slac: I think this line is not needed, this method does not exist.

admin/templates/configuration/etc/log4cxx.properties
```diff
@@ -5,6 +5,6 @@
 log4j.rootLogger=DEBUG, QLogger
 log4j.appender.QLogger=org.apache.log4j.ConsoleAppender
 log4j.appender.QLogger.layout=org.apache.log4j.PatternLayout
-log4j.appender.QLogger.layout.ConversionPattern=%d{MMdd HH:mm:ss.SSS} [%t] %-5p %c{2} (%F:%L) - %m%n
+log4j.appender.QLogger.layout.ConversionPattern=%d{MMdd HH:mm:ss.SSS} [PID:%X{PID}, %t] %-5p %c{2} (%F:%L) - %m%n
```
andy-slac: Why does qserv need PID in the log? I thought it was single-process application.

core/modules/czar/bin/startCzar.py
fjammes: Hi Andy,

My bad, the branch to review is in lsst/log repository, I'll remove this branch in Qserv code.

admin/templates/configuration/etc/log4cxx.properties
fjammes: I think it may be useful if we restart the service? Indeed all the log will be pushed to a centralized database (ElasticSearch) on the cluster, and it can be useful to have the pid here. Furthermore K.T. and Jacek agreed on this in DM-2430 (see ticket comments)

andy-slac: I do not see anybody "agreeing" that qserv have to log PID. For PID is only useful in situations when multiple processes log to the same file simultaneously, PID would help to disentangle interleaved messages then. In all other cases its only taking space on a line.
That log4cxx file is not qserv-specific, I think it can be used by default by other apps. What happens when you use %X{PID} but there is no PID in MDC?

fjammes: OK PID is optional. I think it can be useful to know if Qserv was restarted between two log lines, don't we need it?

If PID isn't in MDC, then nothing is printed here.

PR #87, opened 2015-06-01
DM-2777: Lock BlendScheduler _mapMutex before calling _integrityHelper.

smonkewitz:


jgates108:
Looks fine to me.

abh3:
One admonition, if that calling sequence calls an SSI method that can callback and the mutex is locked again in the callback it better be a recursive mutes or you will deadlock.

PR #88, opened 2015-06-01
DM-2779: lock _runnersMutex before passing _running task queue to scheduler.

smonkewitz:


fritzm:
Looks good to me.

jgates108:
I don't see any problems.

PR #89, opened 2015-06-01
Fix Qserv SsiSession worker race condition.

smonkewitz:
The race was caused by releasing a client request buffer after
enqueuing the corresponding task for processing on another thread.
As written, request processing could complete, freeing the request
object, before release of the request buffer by the enqueuing thread.
The result is a potential memory leak.

To fix this, release the request buffer just after decoding the
protocol buffer message contained in it, and before sending the
decoded message off for processing. In passing, make sure to send
an error message back to the client if protocol buffer decoding
fails, fixing an outstanding TODO.


fritzm:
Nice catch!

jgates108:
Looks fine to me.

PR #90, opened 2015-06-05
U/fjammes/dm 2430

andy-slac:


admin/templates/configuration/etc/init.d/xrootd
```diff
@@ -50,7 +50,12 @@ export QSW_MYSQLDUMP=`which mysqldump`
 QSW_SCRATCHPATH="${QSERV_RUN_DIR}/tmp"
 QSW_SCRATCHDB="qservScratch"
 export QSW_RESULTPATH="${XROOTD_RUN_DIR}/result"
-export LSST_LOG_CONFIG="${QSERV_RUN_DIR}/etc/log4xrootd.properties"
+
+# Log configuration file for worker nodes.
+# xrootd manager doesn't use it.
+# Indeed, log4cxx is only used by Qserv plugin
+# which only runs on xrootd server side (i.e. worker nodes)
```
andy-slac: That comment is somewhat confusing. We run xrootd manager (cmsd) both on worker nodes and master node. And cmsd will also use pour plugins, so it will likely have its output redirected to log4cxx as well.

admin/templates/configuration/etc/log4cxx.czar.properties
```diff
@@ -0,0 +1,27 @@
...
+# used by Qserv-czar
+#
+
+log4j.rootLogger=DEBUG, FILE
+
+# Use this appender to interlace xrootd debug messages with Qserv-czar standard
+# logs. stdout/stderr are redirected to {{QSERV_LOG_DIR}}/qserv-czar-console.log.
+# This workaround is required because Qserv-czar can only outputs xrootd debug
+# message on stderr
+log4j.appender.CONSOLE=org.apache.log4j.ConsoleAppender
```
andy-slac: What is the purpose of this appender, is it used at all?

core/modules/ccontrol/UserQuery.cc
```diff
@@ -92,12 +93,13 @@ namespace qserv {
 
 /// A class that can be used to parameterize a ProtoImporter<TaskMsg> for
 /// debugging purposes
-class ProtoPrinter : public util::UnaryCallable<void, std::shared_ptr<proto::TaskMsg> > {
+class ProtoPrinter final: public util::UnaryCallable<void, std::shared_ptr<proto::TaskMsg> > {
```
andy-slac: Does this work with gcc44?

core/modules/ccontrol/UserQuery.cc
```diff
@@ -92,12 +93,13 @@ namespace qserv {
...
 public:
     ProtoPrinter() {}
-    virtual void operator()(std::shared_ptr<proto::TaskMsg> m) {
+    virtual void operator()(std::shared_ptr<proto::TaskMsg> m) override {
```
andy-slac: Same question for gcc44.

core/modules/czar/python/config.py
```diff
@@ -126,6 +126,15 @@ def printTo(outHandle):
     config.write(outHandle)
     pass
 
+def str():
```
andy-slac: Do no define module-level names (or even local name) that override built-in Python names. What happens if someone says `str(variable)` in the code below?

admin/templates/configuration/etc/log4cxx.czar.properties
fjammes: No it isn't used, if you want to activate it you have to write 

log4j.rootLogger=DEBUG, FILE, CONSOLE

Is it fine to keep it as an example?

core/modules/ccontrol/UserQuery.cc
fjammes: :+1: 
good catch, thanks

core/modules/ccontrol/UserQuery.cc
fjammes: :+1: removed

core/modules/czar/python/config.py
fjammes: :+1: renamed

PR #91, opened 2015-06-06
bug fix: secIndex was incorrectly built for non director tables

smonkewitz:


core/modules/qana/QservRestrictorPlugin.cc
```diff
@@ -593,6 +596,14 @@ QservRestrictorPlugin::_newSecIndexRestrictor(
...
     std::shared_ptr<query::ColumnRef> secIndex =
         resolveAsColumnRef(context, cp->left);
+    if (secIndex)  {
+        std::string theDirTable =
+            context.cssFacade->getDirTable(secIndex->db, secIndex->table);
+        if ( ! theDirTable.empty() ) {
+            LOGF_DEBUG("Setting dirTable for sIndex to %1%" % theDirTable);
+            secIndex->table = theDirTable;
+        }
+    }
```
smonkewitz: I don't think this is the correct fix:
- You are setting the secondary index table without setting the secondary index db. Child tables do not have to be in the same database as their directors.
- You always set the secondary index table to the director table, if there is one. I think you should only do this if the column name matches the director column name for the child table.

This is also not the correct place for the fix - you should embed this in the other `_newSecIndexRestrictor` implementation (the one that handles `IN` clauses), as it is delegated to by this one. If you try to patch things up here, you will fail to fix the problem when clients use `IN` rather than `=`.

Finally, did you verify that `resolveAsColumnRef` returns a copy of the IR node from the query, and not a pointer to the actual node? If it does not, then changing the table / db could potentially hose subsequent analysis passes and we'll have to copy the IR node. If it does then we are fine (but please add a comment to the effect that we are mutating a copy).

jbecla: Tell me more about secondary index db, I think it is the first time I hear we have/support it!
 * I checked https://dev.lsstcorp.org/trac/wiki/db/Qserv/CSS and it is not mentioned
 * I grepped for it in qserv_testdata/case0?/data/*.cfg and we are not setting it there
 * I grepped for dirD it in qserv source code and didn't get any match either (dirC and dirT are found)

Agreed wrt not checking if column name matches. I'll fix that.

Agreed wrt IN

Regarding resolveAsColumnRef - yup, it does call copyAsColumnRef. I did run some tests with valgrind and saw no memory issues whatsoever.

smonkewitz: Sorry, I said that in a confusing way. What I meant was that you should set the db for the `secIndex` variable (a column ref) to the db of the director table. However, we might not be storing the database of a table's director table in the CSS, so you probably cannot do that. I think we really do need to support for this: an L3 child table might very well have an LSST data release Object table as its director, while almost certainly not living in the DR database. So I think this is an oversight in the CSS metadata and needs a ticket.

core/modules/qana/QservRestrictorPlugin.cc
```diff
@@ -571,14 +574,36 @@ QservRestrictorPlugin::_newSecIndexRestrictor(
...
+        p->_params.push_back(dirDb);
+        p->_params.push_back(dirTable);
+        p->_params.push_back(dirCol);
+    } else {
+        LOGF_DEBUG("Using %1%, %2%, %3% as sIndex." %
+                   cr->db % cr->table % cr->column);
+        p->_params.push_back(cr->db);
+        p->_params.push_back(cr->table);
+        p->_params.push_back(cr->column);
+    }
```
smonkewitz: This still doesn't seem right to me - you invoke the alternate code path even when cr->column doesn't match the director column, and you do not resolve the director column name in the director table. Maybe this will be more clear if you just see what I think the logic should look like (untested):
```
    // Build the QsRestrictor
    query::QsRestrictor::Ptr p = std::make_shared<query::QsRestrictor>();
    p->_name = "sIndex";
    // sIndex has paramers as follows:
    // db, table, column, val1, val2, ...

    // Get the director column name
    std::string dirCol = context.cssFacade->getDirColName(cr->db, cr->table);
    if (cr->column == dirCol) {
        // cr may be a column in a child table, in which case we must figure
        // out the corresponding column in the child's director to properly
        // generate a secondary index constraint.
        std::string dirDb = context.cssFacade->getDirDb(cr->db, cr->table);
        std::string dirTable = context.cssFacade->getDirTable(cr->db, cr->table);
        if (dirTable.empty()) {
            dirTable = cr->table;
            if (!dirDb.empty() && dirDb != cr->db) {
                LOGF_ERROR("dirTable missing, but dirDb is set inconsistently for %1%.%2%" % cr->db % cr->table);    
                return query::QsRestrictor::Ptr();
            }
            dirDb = cr->db;
        } else if (dirDb.empty()) {
            dirDb = cr->db;
        }
        if (dirDb != cr->db || dirTable != cr->table) {
            // Lookup the name of the director column in the director table
            dirCol = context.cssFacade->getDirColName(dirDb, dirTable);
            if (dirCol.empty()) {
                LOGF_ERROR("dirCol missing for %1%.%2%" % dirDb % dirTable);    
                return query::QsRestrictor::Ptr();
            }
        }
        LOGF_DEBUG("Using dirDb %1%, dirTable %2%, dirCol %3% as sIndex for %4%.%5%.%6%" %
                   dirDb % dirTable % dirCol % cr->db % cr->table % cr->column);
        p->_params.push_back(dirDb);
        p->_params.push_back(dirTable);
        p->_params.push_back(dirCol);
    } else {
        LOGF_DEBUG("Using %1%, %2%, %3% as sIndex." % cr->db % cr->table % cr->column);
        p->_params.push_back(cr->db);
        p->_params.push_back(cr->table);
        p->_params.push_back(cr->column);
    }
```

Does that make sense to you?

jbecla: Hmm... not really! The block

    if (dirDb != cr->db || dirTable != cr->table) {
        // Lookup the name of the director column in the director table
        dirCol = context.cssFacade->getDirColName(dirDb, dirTable);
        if (dirCol.empty()) {
            LOGF_ERROR("dirCol missing for %1%.%2%" % dirDb % dirTable);    
            return query::QsRestrictor::Ptr();
        }
    }

is not needed because we are setting dirCol earlier, so it is guaranteed to be set.

But most importantly, I always thought that if we set dirColName, we must set dirTable as well (and I assume just-introduced dirDb can be left out, in which case it defaults to current db). Does it really make sense to set dirColName without dirTable? If you agree that dirColName and dirTable always go together, than we need to rewrite it. How about:

    // Build the QsRestrictor
    query::QsRestrictor::Ptr p = std::make_shared<query::QsRestrictor>();
    p->_name = "sIndex";
    // sIndex has paramers as follows:
    // db, table, column, val1, val2, ...

    // Get the director db/tb/col info
    std::string dirDb = context.cssFacade->getDirDb(cr->db, cr->table);
    std::string dirTable = context.cssFacade->getDirTable(cr->db, cr->table);
    std::string dirCol = context.cssFacade->getDirColName(cr->db, cr->table);
    if (dirDb.empty() && dirTable.empty() && dirCol.empty()) {
        LOGF_DEBUG("Using %1%, %2%, %3% as sIndex." % cr->db % cr->table % cr->column);
        p->_params.push_back(cr->db);
        p->_params.push_back(cr->table);
        p->_params.push_back(cr->column);
    } else {
        if (dirTable.empty() && !dirCol.empty() ) {
            LOGF_ERROR("dirCol missing, dirTable set to %1%" % dirTable);
            return query::QsRestrictor::Ptr();
        }
        if (dirCol.empty() && !dirTable.empty() ) {
            LOGF_ERROR("dirTable missing, dirCol set to %1%" % dirCol);
            return query::QsRestrictor::Ptr();
        }
        if (!dirDb.empty() && (dirTable.empty() or dirCol.empty()) {
            LOGF_ERROR("dirTable/dirCol missing, dirDb set to %1%" % dirDb);
            return query::QsRestrictor::Ptr();
        }
        if (dirDb.empty()) {
            dirDb = cr->db;
        }
        if (dirCol != cr->col) {
            LOGF_ERROR("dirCol=%1 set inconsistently for %2%.%3%.%4%" % \
		       dirColcr->db % cr->table % cr->col);
            return query::QsRestrictor::Ptr();
        }
        p->_params.push_back(dirDb);
        p->_params.push_back(dirTable);
        p->_params.push_back(dirCol);
    }

smonkewitz: No.

The first time you set dirCol, you set it to the name of the column **for the current table**. You have to set it again to the name of the column **in the director table**, once you know what the director table is (and that it is not equal to the current table). Consider for example a Source table with a director column named "directorId" that corresponds to the "objectId" column in a director table named Object.

As far as whether dirColName should be allowed without a dirTable - no idea. I thought it might be reasonable to default to the current table if no director table is explicitly specified. If we should be checking that dirTable and dirColName are always set together, it's a very small mod to the initial proposed snippet. FWIW I find your sequence of if statements very confusing, you are using `or` (non-idiomatic C++) instead of `||`, and the final error when cr->col != dirCol is not correct.

jbecla: Forget the || vs or, it is a typo, (I was just tweaking python code).

The way I imagined it was:
    dirDb = getDirDb("myDb", "Source");           // say it is "", so we default to current
    dirTable = getDirTable("myDb", "Source"); // dirTable gets set to "Object" 
    dirCol = getDirCol("myDb", "Source");        // dirCol gets set to "objectId"
and there we have everything we need to know about dirDb/dirTable/dirCol

I think I hear you are saying that after
    dirTable = getDirTable("myDb", "Source"); // dirTable gets set to "Object" 
if the dirTable has been set to something different than current table, the
    dirCol = getDirCol("myDb", "Source");        // dirCol gets set to "objectId"
should really become
    dirCol = getDirCol("myDb", "Object");        // dirCol gets set to "objectId"
Hmm... I guess we could chain it like that, but somehow I always imagined that we should set dirTable AND dirCol together, and we don't need to follow the chain: ("what is my dirTable?", "ok, it is X", "let me go to X and check its dirCol"... ). If we do that, then I'd argue that if we set dirTable, we should DISALLOW dirCol, and it should be required to go to dirTable to get dirCol.

jbecla: Serge we are both wrong. We have both secIndexColName and dirColName per table. We really need to clean that up. How about the following:
 * get rid of secIndexColName
 * if table is a director table, dirColName must be set, and dirColTable must not be set and dirDb must not be set
 * if the table is not a director table, dirTable must be set, optionally dirDb maybe set, dirColName must not be set.

PR #93, opened 2015-06-08
Added unit test.

fritzm:


core/modules/ccontrol/userQueryProxy.cc
```diff
@@ -79,44 +79,77 @@ class UserQueryManager : public SessionManager<UserQuery::Ptr> {
 };
 static UserQueryManager uqManager;
 
-std::string const& UserQuery_getError(int session) {
-    return uqManager.get(session)->getError();
+std::string UserQuery_getError(int session) {
+    std::string s;
+    try {
+        s = uqManager.get(session)->getError();
+    } catch (std::exception& e) {
```
fritzm: Nit: should these catches be const&, since you don't intend to modify in your handlers?

PR #99, opened 2015-06-26
U/fjammes/dm 2885

jbecla:


core/modules/ccontrol/queryMsg.cc
```diff
@@ -43,19 +43,23 @@ int queryMsgGetCount(int session) {
 }
 
 // Python call: msg, chunkId, code, timestamp = queryMsgGetMsg(session, idx)
-std::string queryMsgGetMsg(int session, int idx, int* chunkId, int* code, time_t* timestamp) {
+std::string queryMsgGetMsg(int session, int idx, int* chunkId, int* code, std::string* severity, time_t* timestamp) {
```
jbecla: I'd suggest to use references (&) instead of pointers (*) for both severity and timestamp.

core/modules/qdisp/MessageStore.cc
```diff
@@ -52,10 +85,14 @@ void MessageStore::addMessage(int chunkId, int code, std::string const& descript
...
-            QueryMessage(chunkId, code, description, std::time(0)));
+            QueryMessage(chunkId, code, description, std::time(0), severity));
     }
 }
 
+void MessageStore::addErrorMessage(std::string const& description) {
+	addMessage(NOTSET, NOTSET, description, Severity::ERROR);
+}
+
 const QueryMessage MessageStore::getMessage(int idx) {
```
jbecla: how about returning by reference instead of making a copy of QueryMessage?

core/modules/ccontrol/queryMsg.cc
jbecla: Also, maybe return const std::string&?

core/modules/ccontrol/queryMsg.cc
```diff
@@ -43,19 +43,23 @@ int queryMsgGetCount(int session) {
...
 
     qdisp::QueryMessage msg = UserQuery_get(session).getMessageStore()->getMessage(idx);
     *chunkId = msg.chunkId;
     *code = msg.code;
     *timestamp = msg.timestamp;
+    *severity = qdisp::to_string(msg.severity);
```
jbecla: if you use references when passing arguments, you won't have odd pointers here :)

core/modules/ccontrol/queryMsg.cc
```diff
@@ -43,19 +43,23 @@ int queryMsgGetCount(int session) {
...
     return msg.description;
 }
 
-void queryMsgAddMsg(int session, int chunkId, int code,
-                            std::string const& message) {
+void queryMsgAddMsg(int session, int chunkId, int code, std::string const& message,
+					std::string const& severity /* = "INFO" */) {
```
jbecla: alignment

core/modules/ccontrol/queryMsg.h
```diff
@@ -37,9 +37,10 @@ int queryMsgGetCount(int session);
 
 // Python call: msg, chunkId, code, timestamp = queryMsgGetMsg(session, idx)
 // int* chunkId, int* code, time_t* timestamp matches with %apply directive to help SWIG
-std::string queryMsgGetMsg(int session, int idx, int* chunkId, int* code, time_t* timestamp);
+std::string queryMsgGetMsg(int session, int idx, int* chunkId, int* code, std::string* severity, time_t* timestamp);
 
-void queryMsgAddMsg(int session, int chunkId, int code, std::string const& message);
+void queryMsgAddMsg(int session, int chunkId, int code, std::string const& message,
+		            std::string const& severity = "INFO");
```
jbecla: alignment

core/modules/czar/python/app.py
```diff
@@ -340,7 +339,8 @@ def __init__(self, query, context, setSessionId, resultName=""):
         pass
 
     def _reportError(self, chunkId, code, message):
-        logger.dbg("reporting", chunkId, code, message)
+        # TODO remove lsst/log wrapper
```
jbecla: should we deal with the todo now?

core/modules/ccontrol/queryMsg.h
```diff
@@ -37,9 +37,10 @@ int queryMsgGetCount(int session);
 
 // Python call: msg, chunkId, code, timestamp = queryMsgGetMsg(session, idx)
```
jbecla: Several files, including this one need copyright header fixed. 

core/modules/ccontrol/queryMsg.h: * Copyright 2012-2014 LSST Corporation.
core/modules/czar/python/app.py:# Copyright 2008-2014 AURA/LSST.
core/modules/czar/python/proxy.py:# Copyright 2008-2014 LSST Corporation.
core/modules/qdisp/JobStatus.cc: * Copyright 2014 LSST Corporation.
core/modules/util/testMultiError.cc: * Copyright 2014 LSST Corporation.

core/modules/ccontrol/queryMsg.h
```diff
@@ -37,9 +37,10 @@ int queryMsgGetCount(int session);
...
 // int* chunkId, int* code, time_t* timestamp matches with %apply directive to help SWIG
-std::string queryMsgGetMsg(int session, int idx, int* chunkId, int* code, time_t* timestamp);
+std::string queryMsgGetMsg(int session, int idx, int* chunkId, int* code, std::string* severity, time_t* timestamp);
```
jbecla: * --> &

core/modules/ccontrol/queryMsg.cc
ktlim: Actually, for output parameters, pointers are generally preferred as they make clear that the parameter is different from others.  Another option is to package all the outputs into a struct or class or `pair<>` that can be returned.

core/modules/ccontrol/queryMsg.cc
ktlim: See below for the arguments.  For the return value, a reference to string is somewhat dangerous and only appropriate if you know that the underlying string has a lifetime as long or longer than any user of the result (e.g. because it is a constant).

core/modules/czar/python/proxy.py
```diff
@@ -50,9 +63,11 @@ class Lock:
     requests when it is waiting on mysql queries that it knowingly
     initiates in its C-layer, whereas it treats its Lua plugin code as a
     single-threaded black-box."""
-    createTmpl = "CREATE TABLE IF NOT EXISTS %s (chunkId INT, code SMALLINT, message CHAR(255), timeStamp FLOAT) ENGINE=MEMORY;"
+    createTmpl = "CREATE TABLE IF NOT EXISTS %s " \
+                 "(chunkId INT, code SMALLINT, message CHAR(255), " \
+                 "severity ENUM ('DEBUG', 'INFO', 'ERROR'), timeStamp FLOAT) ENGINE=MEMORY;"
```
jbecla: what about "warning"?

core/modules/czar/python/proxy.py
```diff
@@ -50,9 +63,11 @@ class Lock:
     requests when it is waiting on mysql queries that it knowingly
     initiates in its C-layer, whereas it treats its Lua plugin code as a
     single-threaded black-box."""
-    createTmpl = "CREATE TABLE IF NOT EXISTS %s (chunkId INT, code SMALLINT, message CHAR(255), timeStamp FLOAT) ENGINE=MEMORY;"
+    createTmpl = "CREATE TABLE IF NOT EXISTS %s " \
+                 "(chunkId INT, code SMALLINT, message CHAR(255), " \
```
jbecla: 255 might not be enough to store message.

core/modules/proxy/mysqlProxy.lua
```diff
@@ -624,11 +624,11 @@ function queryProcessing()
         q2 = "SELECT * FROM " .. resultTableName
         proxy.queries:append(2, string.char(proxy.COM_QUERY) .. q2,
                              {resultset_is_needed = true})
-	q3 = "DROP TABLE " .. msgTableName
+        q3 = "DROP TABLE " .. msgTableName
```
jbecla: alignment!

core/modules/proxy/mysqlProxy.lua
```diff
@@ -708,23 +708,31 @@ function read_query_result(inj)
     -- we injected query with the id=1 (for messaging and locking purposes)
     if (inj.type == 1) then
         print("q1 - ignoring")
+        local error_msg = ""
         for row in inj.resultset.rows do
-           print("   " .. row[1] .. " " .. row[2] .. " " .. tostring(row[3]) .. " " .. row[4])
-           if (tonumber(row[2]) < 0) then -- errors have code < 0
+           if (row[4] == "ERROR") then
               queryErrorCount  = queryErrorCount + 1
```
jbecla: please put a comment near the place when you create schema for that table saying that we are sensitive to column positions.

core/modules/proxy/mysqlProxy.lua
```diff
@@ -708,23 +708,31 @@ function read_query_result(inj)
...
-              return err.setAndSend(ERR_QSERV_RUNTIME,
-                                    "Error during execution:\n"..
-                                        row[2] .. " " .. tostring(row[3]) .. " (" .. row[1] .. ")")
+              error_msg = error_msg .. "\n" .. tostring(row[3])
+              -- WARN czar never returns multiple errors for now
+              if (queryErrorCount > 1) then
+                  error_msg = "\n-- WARN multiple errors"
+              end
+           else
+              print("   chunkId: " .. row[1] .. ", code: " .. row[2] .. ", msg: " .. tostring(row[3]) .. ", timestamp: " .. row[5])
```
jbecla: print? why not logging?

jbecla: oh i see, we are in lua, so indeed that might be too much

core/modules/qdisp/Executive.cc
```diff
@@ -63,6 +63,12 @@
 #include "qdisp/XrdSsiMocks.h"
 
 namespace {
+
+LOG_LOGGER getLogger() {
+        static const LOG_LOGGER _logger(LOG_GET("lsst.qserv.qdisp.Executive"));
```
jbecla: alignment looks wrong

core/modules/qdisp/Executive.cc
```diff
@@ -156,50 +162,55 @@ class Executive::DispatchAction : public util::VoidCallable<void> {
...
-    ExecStatus::Ptr _execStatus; ///< Points at status in Executive::_statusMap
+    int _jobId;
+    JobDescription _jobDescription;
+    JobStatus::Ptr _jobStatus; ///< Points at status in Executive::_statusMap
 };
 
-/// Add a spec to be executed. Not thread-safe.
-void Executive::add(int refNum, Executive::Spec const& s) {
+/* Add a new job to executive queue, if not already in. Not thread-safe.
+ *
```
jbecla: remove extraneous line

core/modules/qdisp/Executive.cc
```diff
@@ -220,61 +231,68 @@ bool Executive::join() {
...
     }
-
-    LOGF_INFO("Query exec finish. %1% dispatched." % _requestCount);
-    _reportStatuses();
-    if(sCount != _requestCount) {
-        LOGF_INFO("Query exec:. %1% != %2%" % _requestCount % sCount);
+    if(sCount == _requestCount) {
+        LOGF_INFO("Query execution succeed: %1% jobs dispatched and completed." % _requestCount);
+    }
+    else {
```
jbecla: move else to previous line

core/modules/qdisp/Executive.cc
```diff
@@ -220,61 +231,68 @@ bool Executive::join() {
         std::lock_guard<std::mutex> lock(_statusesMutex);
         sCount = std::count_if(_statuses.begin(), _statuses.end(), successF::f);
     }
-
-    LOGF_INFO("Query exec finish. %1% dispatched." % _requestCount);
-    _reportStatuses();
-    if(sCount != _requestCount) {
-        LOGF_INFO("Query exec:. %1% != %2%" % _requestCount % sCount);
+    if(sCount == _requestCount) {
+        LOGF_INFO("Query execution succeed: %1% jobs dispatched and completed." % _requestCount);
```
jbecla: succeed --> succeeded

core/modules/qdisp/Executive.cc
```diff
@@ -504,14 +524,16 @@ void Executive::_reapRequesters(std::unique_lock<std::mutex> const&) {
...
-    StatusMap::const_iterator i,e;
-    for(i=_statuses.begin(), e=_statuses.end(); i != e; ++i) {
-        ExecStatus::Info info = i->second->getInfo();
+    for(auto i=_statuses.begin(), e=_statuses.end(); i != e; ++i) {
+        JobStatus::Info info = i->second->getInfo();
         std::ostringstream os;
-        os << ExecStatus::stateText(info.state)
-           << " " << info.stateCode;
+        if (LOG_CHECK_LVL(getLogger(), LOG_LVL_TRACE)) {
+                LOGF(getLogger(), LOG_LVL_TRACE, "%1%" % info.state);
```
jbecla: alignment

core/modules/qdisp/Executive.cc
```diff
@@ -519,6 +541,13 @@ void Executive::_reportStatuses() {
         _messageStore->addMessage(info.resourceUnit.chunk(),
                                   info.state, os.str());
     }
+	}
```
jbecla: alignment (whole section)

core/modules/qdisp/Executive.cc
```diff
@@ -556,8 +585,8 @@ void Executive::_waitAllUntilEmpty() {
 }
 
 std::ostream& operator<<(std::ostream& os,
-                         Executive::StatusMap::value_type const& v) {
-    ExecStatus const& es = *(v.second);
+                         Executive::JobStatusPtrMap::value_type const& v) {
+    JobStatus const& es = *(v.second);
     os << v.first << ": " << es;
```
jbecla: no need to create "es", you can write it as os << v.first << ": " << *(v.second);

core/modules/qdisp/Executive.h
```diff
@@ -65,8 +65,13 @@ class Executive {
         static std::string getMockStr() {return "Mock";};
     };
 
-    /// Specification for something executable by the Executive
-    struct Spec {
+    /** Description of a job managed by the executive
+     *
+     * Launch a chunk query against a xrootd ressource and
```
jbecla: ressource typo.

core/modules/qdisp/Executive.h
```diff
@@ -65,8 +65,13 @@ class Executive {
...
+     * retrieve the result
+     *
```
jbecla: remove extra line

core/modules/qdisp/Executive.h
```diff
@@ -136,12 +149,20 @@ class Executive {
     std::shared_ptr<MessageStore> _messageStore; ///< MessageStore for logging
     XrdSsiService* _service; ///< RPC interface
     RequesterMap _requesters; ///< RequesterMap for results from submitted tasks
-    StatusMap _statuses; ///< Statuses of submitted tasks
+    JobStatusPtrMap _statuses; ///< Statuses of submitted tasks
+
+    /** Execution errors */
```
jbecla: let's do comments consistently. Looks like a bit higher we are using "/// < text"

core/modules/qdisp/JobStatus.cc
```diff
@@ -0,0 +1,96 @@
+// -*- LSST-C++ -*-
+/*
+ * LSST Data Management System
+ * Copyright 2014 LSST Corporation.
```
jbecla: 2015

core/modules/qdisp/JobStatus.cc
```diff
@@ -0,0 +1,96 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the LSST License Statement and
+ * the GNU General Public License along with this program.  If not,
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+
+#include "qdisp/JobStatus.h"
```
jbecla: add "// Class header"

core/modules/qdisp/JobStatus.cc
```diff
@@ -0,0 +1,96 @@
...
+	state_message[JobStatus::PROVISION_ERROR] = "Error accessing resource";
+	state_message[JobStatus::PROVISION_NACK] =  "Error accessing resource (delayed)";
+	state_message[JobStatus::REQUEST] = "Sending request to resource";
+	state_message[JobStatus::REQUEST_ERROR] = "Error sending request";
+	state_message[JobStatus::RESPONSE_READY] = "Response ready";
+	state_message[JobStatus::RESPONSE_ERROR] = "Response error";
+	state_message[JobStatus::RESPONSE_DATA] = "Retrieving response data";
+	state_message[JobStatus::RESPONSE_DATA_ERROR] = "Error retrieving response data";
+	state_message[JobStatus::RESPONSE_DATA_ERROR_OK] = "Error retrieving response, session is OK";
+	state_message[JobStatus::RESPONSE_DATA_ERROR_CORRUPT] =  "Error retrieving response session is corrupt";
```
jbecla: extra space

core/modules/qdisp/JobStatus.cc
```diff
@@ -0,0 +1,96 @@
...
+    : resourceUnit(resourceUnit_),
+      state(UNKNOWN),
+      stateCode(0) {
+    stateTime = ::time(NULL);
+}
+
+
+std::ostream& operator<<(std::ostream& os, JobStatus::State const& state) {
+
+	std::map<JobStatus::State, std::string> state_message;
```
jbecla: maybe let's make this map static instead of redoing it for every operator<< call

core/modules/qdisp/JobStatus.cc
```diff
@@ -0,0 +1,96 @@
...
+	state_message[JobStatus::RESPONSE_DATA_NACK] = "Error in response data";
+	state_message[JobStatus::RESPONSE_DONE] = "Finished retrieving result";
+	state_message[JobStatus::RESULT_ERROR] = "Error in worker result data.";
+	state_message[JobStatus::MERGE_OK] = "Merge complete";
+	state_message[JobStatus::MERGE_ERROR] = "Error merging result";
+	state_message[JobStatus::COMPLETE] = "Complete (success)";
+
+	auto it = state_message.find(state);
+	if (it != state_message.end()) {
+		os << it->second;
```
jbecla: alignment

core/modules/qdisp/JobStatus.cc
```diff
@@ -0,0 +1,96 @@
...
+	} else
+		os << "State error (unrecognized)";
```
jbecla: alignment, and pls add {} brackets for else

core/modules/qdisp/JobStatus.cc
```diff
@@ -0,0 +1,96 @@
...
+	return os;
+}
+
+std::ostream& operator<<(std::ostream& os, JobStatus const& es) {
+    JobStatus::Info info = es.getInfo();
+    return os << info;
+}
+
+std::ostream& operator<<(std::ostream& os, JobStatus::Info const& info) {
+    // At least 26 byes, according to "man ctime", but might be too small.
```
jbecla: byes typo

core/modules/qdisp/JobStatus.cc
```diff
@@ -0,0 +1,96 @@
...
+    const int BLEN=64;
+    char buffer[BLEN];
+    struct tm mytm;
+    char const timefmt[] = "%Y%m%d-%H:%M:%S";
+    int tsLen = ::strftime(buffer, BLEN, timefmt,
+                           ::localtime_r(&info.stateTime, &mytm));
```
jbecla: i think I have seen someone mentioned we should avoid local time. @ktlim, can you comment?
Also, it might be worth looking at C++ time functions and simplify this section.

core/modules/qdisp/JobStatus.h
```diff
@@ -63,13 +69,20 @@ class ExecStatus {
...
+     *
+     *	Useful for logging and error reporting
+     *
+     * TODO: Save past state history:
+     *  - resourceUnit should be extracted from Info (beware of mutex)
+     *  - Info should be put in a vector
+     */
+    void updateInfo(State s, int code=0, std::string const& desc=_empty) {
         std::lock_guard<std::mutex> lock(_mutex);
 #if 0
```
jbecla: do we still want this disabled block of code? Remove maybe?

core/modules/qdisp/MessageStore.cc
```diff
@@ -39,11 +39,44 @@ namespace lsst {
 namespace qserv {
 namespace qdisp {
 
+namespace {
+	const int NOTSET = -1;
```
jbecla: alignment

core/modules/qdisp/MessageStore.cc
```diff
@@ -39,11 +39,44 @@ namespace lsst {
...
+}
+
 ////////////////////////////////////////////////////////////////////////
 // public
 ////////////////////////////////////////////////////////////////////////
 
-void MessageStore::addMessage(int chunkId, int code, std::string const& description) {
+/** Output operator for Severity */
+std::ostream& operator<<(std::ostream& os, Severity const& severity) {
+	switch (severity) {
```
jbecla: what is going on with misalignment? Is your editor playing tricks on you?

core/modules/qdisp/MessageStore.cc
```diff
@@ -39,11 +39,44 @@ namespace lsst {
...
+	case INFO:
+		os << "INFO";
+		break;
+	case ERROR:
+		os << "ERROR";
+		break;
```
jbecla: do we want warning here?

core/modules/qdisp/MessageStore.cc
```diff
@@ -52,10 +85,14 @@ void MessageStore::addMessage(int chunkId, int code, std::string const& descript
     {
         std::lock_guard<std::mutex> lock(_storeMutex);
         _queryMessages.insert(_queryMessages.end(),
-            QueryMessage(chunkId, code, description, std::time(0)));
+            QueryMessage(chunkId, code, description, std::time(0), severity));
     }
 }
 
+void MessageStore::addErrorMessage(std::string const& description) {
+	addMessage(NOTSET, NOTSET, description, Severity::ERROR);
```
jbecla: alignment

core/modules/qdisp/MessageStore.h
```diff
@@ -43,26 +44,70 @@ namespace lsst {
...
 public:
-    void addMessage(int chunkId, int code, std::string const& description);
+    void addMessage(int chunkId, int code, std::string const& description, Severity severity_ = INFO);
+
+    /** Add an error message
+     *
+     * This message will be sent to proxy via message table, in order to be displayed to mysql client
+     * console. chunkId and code are equal to NOTSET because this message can aggregate multiple
+     * error messages in multiple files.
+     *
```
jbecla: remove empty line

core/modules/rproc/InfileMerger.cc
```diff
@@ -78,13 +78,18 @@ using lsst::qserv::proto::WorkerResponse;
 using lsst::qserv::rproc::InfileMergerConfig;
 using lsst::qserv::rproc::InfileMergerError;
 
+LOG_LOGGER getLogger() {
+        static const LOG_LOGGER _logger(LOG_GET("lsst.qserv.rproc.InfileMerger"));
```
jbecla: alignment

core/modules/qdisp/JobStatus.cc
ktlim: If you produce a timestamp, it must always say what timezone it is in.  This can be accomplished by using UTC and appending "Z" or "+0000", or it can be accomplished by including the timezone in the (ISO-standardized) output.

core/modules/ccontrol/UserQuery.cc
```diff
@@ -192,12 +192,10 @@ void UserQuery::submit() {
         std::shared_ptr<MergingRequester> mr
             = std::make_shared<MergingRequester>(cmr, _infileMerger,
                                                  chunkResultName);
-        qdisp::Executive::Spec s = { ru,
-                                     ss.str(),
-                                     mr};
+        qdisp::Executive::JobDescription jobDesc = { ru, ss.str(), mr};
```
andy-slac: Extra space after curly brace

core/modules/czar/python/app.py
```diff
@@ -340,7 +339,8 @@ def __init__(self, query, context, setSessionId, resultName=""):
...
+        logger.dbg("Reporting:\n\tchunkId=[%s]\n\tcode=[%s]\n\tmessage=[%s]\n" % (chunkId, code, message))
```
andy-slac: Does this need to be multi-line message?
Use lazy formatting for logging (I think logger wrapper should be able to support it).

core/modules/qdisp/Executive.cc
```diff
@@ -156,50 +162,55 @@ class Executive::DispatchAction : public util::VoidCallable<void> {
...
+    if (_empty.get()) {
+        _empty.set(false);
+        LOGF(getLogger(), LOG_LVL_TRACE, "Flag \"Empty\" set to false by jobId %1%" % jobId);
+    }
 
     ++_requestCount;
-    std::string msg = "Exec add pth=" + s.resource.path();
-    LOGF_INFO("%1%" % msg);
-    _messageStore->addMessage(s.resource.chunk(), log::MSG_MGR_ADD, msg);
+    std::string msg = "Executive: Add job with path=" + jobDescription.resource.path() + "";
```
andy-slac: Why adding empty string at the end?

core/modules/qdisp/Executive.cc
```diff
@@ -504,14 +524,16 @@ void Executive::_reapRequesters(std::unique_lock<std::mutex> const&) {
     }
 }
 
-void Executive::_reportStatuses() {
+void Executive::_updateProxyMessages() {
+	{
```
andy-slac: Check indentations in the whole function.

core/modules/qdisp/Executive.cc
```diff
@@ -504,14 +524,16 @@ void Executive::_reapRequesters(std::unique_lock<std::mutex> const&) {
...
     std::lock_guard<std::mutex> lock(_statusesMutex);
-    StatusMap::const_iterator i,e;
-    for(i=_statuses.begin(), e=_statuses.end(); i != e; ++i) {
-        ExecStatus::Info info = i->second->getInfo();
+    for(auto i=_statuses.begin(), e=_statuses.end(); i != e; ++i) {
+        JobStatus::Info info = i->second->getInfo();
```
andy-slac: `JobStatus::Info const& info`? Or `auto const& info` if you started playing with auto already

core/modules/qdisp/Executive.cc
```diff
@@ -504,14 +524,16 @@ void Executive::_reapRequesters(std::unique_lock<std::mutex> const&) {
...
         std::ostringstream os;
-        os << ExecStatus::stateText(info.state)
-           << " " << info.stateCode;
+        if (LOG_CHECK_LVL(getLogger(), LOG_LVL_TRACE)) {
```
andy-slac: Why is this if needed?

core/modules/qdisp/JobStatus.cc
andy-slac: Indeed this is horribly inefficient code. Please return old `switch()` back, it's much better than a map.

core/modules/qdisp/JobStatus.cc
```diff
@@ -0,0 +1,96 @@
...
+		os << it->second;
+	} else
+		os << "State error (unrecognized)";
+	return os;
+}
+
+std::ostream& operator<<(std::ostream& os, JobStatus const& es) {
+    JobStatus::Info info = es.getInfo();
+    return os << info;
+}
```
andy-slac: No need to make a copy of `JobStatus::Info`, `os << es.getInfo()` works perfectly well.

core/modules/qdisp/JobStatus.cc
```diff
@@ -0,0 +1,96 @@
...
+
+std::ostream& operator<<(std::ostream& os, JobStatus::Info const& info) {
+    // At least 26 byes, according to "man ctime", but might be too small.
+    const int BLEN=64;
+    char buffer[BLEN];
+    struct tm mytm;
+    char const timefmt[] = "%Y%m%d-%H:%M:%S";
```
andy-slac: This is a non-standard time format (even ignoring timezone absence).

core/modules/qdisp/JobStatus.cc
```diff
@@ -0,0 +1,96 @@
...
+    int tsLen = ::strftime(buffer, BLEN, timefmt,
+                           ::localtime_r(&info.stateTime, &mytm));
+    std::string ts(buffer, tsLen);
```
andy-slac: There is no need to create temporary string just to send buffer to a stream.

core/modules/qdisp/JobStatus.h
```diff
@@ -37,15 +37,21 @@ namespace lsst {
 namespace qserv {
 namespace qdisp {
 
-/// ExecStatus instances receive timestamped reports of execution State. This
-/// allows a manager object to receive updates on status without exposing its
-/// existence to a delegate class. The ExecStatus class could be extended to
-/// save all received reports to provide a timeline of state changes, but this
-/// is not currently implemented.
-class ExecStatus {
```
andy-slac: Why do you rename this class, what is wrong with `ExecStatus`?

core/modules/qdisp/MessageStore.cc
andy-slac: I think it's inserting TAB, we should use 4-space indentation.

core/modules/qdisp/MessageStore.cc
```diff
@@ -39,11 +39,44 @@ namespace lsst {
...
+		os << "ERROR";
+		break;
+	default:
+		os << "UNKNOWN";
+	}
+	return os;
+}
+
+/** Convert Severity to std::string */
+std::string to_string(Severity const& severity) {
```
andy-slac: If this method is used frequently it should be implemented more efficiently

core/modules/rproc/InfileMerger.cc
```diff
@@ -78,13 +78,18 @@ using lsst::qserv::proto::WorkerResponse;
...
+
 /// @return a timestamp id for use in generating temporary result table names.
 std::string getTimeStampId() {
     struct timeval now;
     int rc = gettimeofday(&now, NULL);
     if (rc != 0) {
-        throw InfileMergerError(InfileMergerError::INTERNAL,
-                                "Failed to get timestamp.");
+        std::string msg = "Failed to get timestamp.";
+        throw InfileMergerError(InfileMergerError::INTERNAL, msg);
```
andy-slac: Why is this better?

core/modules/util/Error.h
```diff
@@ -53,10 +53,23 @@ class Error {
...
-            code(code), msg(msg) {
+    Error() :
+        status(NONE) {
+    }
+
+    Error(int code, std::string const& msg = "") :
+        status(NONE), code(code), msg(msg) {
     }
 
     virtual ~Error() {
```
andy-slac: We should remove virtual (and whole destructor)

core/modules/wdb/QueryAction.cc
```diff
@@ -284,7 +284,9 @@ void QueryAction::Impl::_transmit(bool last) {
     std::string resultString;
     _result->set_continues(!last);
     if (!_multiError.empty()) {
-        std::string msg = _multiError.toString();
+
+        std::string chunkId = std::to_string((*_msg).chunkid());
```
andy-slac: This does not compile with gcc44. Please try to build on NCSA cluster before sending code for review (always).
You can use `boost::lexical_cast` before we all switch to more decent gcc version

core/modules/ccontrol/queryMsg.h
andy-slac: Do you have to pass string here? Can you pass Severity enum instead to help with type/range checking?

core/modules/qdisp/Executive.cc
```diff
@@ -349,33 +367,35 @@ int Executive::getNumInflight() {
...
-    std::lock_guard<std::mutex> lock(_statusesMutex);
     std::ostringstream os;
-    std::for_each(_statuses.begin(), _statuses.end(), printMapEntry(os, "\n"));
-    LOGF_ERROR("%1%" % os.str());
+    {
+        std::lock_guard<std::mutex> lock(_statusesMutex);
+        std::for_each(_statuses.begin(), _statuses.end(), printMapEntry(os, "\n"));
+    }
+    LOGF(getLogger(), LOG_LVL_ERROR, "%1%" % os.str());
     return os.str();
```
andy-slac: `os.str()` is called twice here, which is not efficient.

core/modules/util/Error.h
```diff
@@ -53,10 +53,23 @@ class Error {
...
+    } status;
 
-    Error(int code, std::string const& msg) :
-            code(code), msg(msg) {
+    Error() :
+        status(NONE) {
+    }
+
+    Error(int code, std::string const& msg = "") :
+        status(NONE), code(code), msg(msg) {
```
andy-slac: Comment for the whole class - in general I do not like modifiable classes especially if class is simple like this one. It's easier and safer to work with objects that do not need "extended construction". This class constructors do not provide any way to initialize status, to set status to something else than NONE one has to construct it first and then modify later. You should add status parameter to non-default constructor. And I'd recommend to disallow modifications of data members entirely.

andy-slac: I'd also recommend to name constructor arguments differently from data members `code(code)` looks a bit weird.

core/modules/ccontrol/queryMsg.cc
fjammes: ok with K.T.

core/modules/qdisp/MessageStore.cc
fjammes: ok with K.T. here, furthermore I didn't change this line in this ticket.

core/modules/ccontrol/queryMsg.cc
fjammes: ok for pointer output parameters, like in libc :-)

core/modules/ccontrol/queryMsg.cc
fjammes: ok

core/modules/czar/python/app.py
fjammes: we should do it in a new ticket, why was this wrapper implemented? FYI, it seems it only concatenate its parameters and send it to lsst/log. If lsst/log API is stable now, removing the wrapper would simplify the code.

core/modules/ccontrol/queryMsg.h
fjammes: :+1:

core/modules/czar/python/proxy.py
fjammes: Storing messages inside TEXT might be resource consuming, isn't it? If yes, we may improve it later, depending on our user experience on log/error message length on the mysql proxy/client?

core/modules/qdisp/Executive.cc
fjammes: :+1:

core/modules/qdisp/Executive.cc
fjammes: :+1:

core/modules/qdisp/Executive.cc
fjammes: :+1:

core/modules/qdisp/Executive.cc
fjammes: :+1:

core/modules/qdisp/Executive.cc
fjammes: :+1:

core/modules/qdisp/Executive.cc
fjammes: :+1:

core/modules/qdisp/Executive.cc
fjammes: :+1:

core/modules/qdisp/Executive.h
fjammes: :+1:

core/modules/qdisp/Executive.h
fjammes: :+1:

core/modules/qdisp/Executive.h
fjammes: O, but I try to follow LSST documentation standards: https://confluence.lsstcorp.org/display/LDMDG/Documentation+Standards#DocumentationStandards-RequiredDocumentationStyle
and they looks different from the previous standards used by Qserv. Shall we switch for new comment and create a low priority ticket to convert everything?

core/modules/qdisp/JobStatus.cc
fjammes: :+1:

core/modules/qdisp/JobStatus.cc
fjammes: :+1:

core/modules/qdisp/Executive.h
jbecla: I guess

core/modules/czar/python/proxy.py
jbecla: actually MEMORY tables can't contain TEXT. I think we just need to make sure we have a reasonable fixed size CHAR (and maybe check whether we are hitting the limit, and log it somewhere)

fjammes: ok, DM-3035 created.

core/modules/qdisp/Executive.h
fjammes: DM-3036 created

core/modules/czar/python/app.py
fjammes: DM-3037 created

core/modules/ccontrol/queryMsg.h
fjammes: :+1:  done

core/modules/proxy/mysqlProxy.lua
fjammes: :+1:

core/modules/qdisp/JobStatus.cc
fjammes: ok, sorry for that

core/modules/qdisp/JobStatus.cc
fjammes: :+1:

core/modules/czar/python/proxy.py
fjammes: Only "INFO" and "ERROR" are useful for message. lua log "INFO" and send "ERROR" to client. So I removed "DEBUG" for now.

core/modules/qdisp/JobStatus.cc
fjammes: :+1: done

core/modules/qdisp/JobStatus.h
fjammes: :+1:

core/modules/qdisp/MessageStore.cc
fjammes: :+1:

core/modules/qdisp/MessageStore.cc
fjammes: code remove operator<< is no more required.

fjammes: I've tuned eclipse with 4-space indentation, thanks.

core/modules/qdisp/MessageStore.cc
fjammes: Not for now

core/modules/qdisp/MessageStore.cc
fjammes: :+1:

core/modules/qdisp/MessageStore.h
fjammes: :+1:

core/modules/rproc/InfileMerger.cc
fjammes: :+1: sorry for all these alignment issues, I'll be more careful

core/modules/ccontrol/UserQuery.cc
fjammes: :+1:

core/modules/czar/python/app.py
fjammes: Ok I set a oneliner message.

No, logger wrapper doesn't support lazy formatting, it only join parameter using ' ' as a separator. FYI I opened DM-3037 to remove this wrapper.

core/modules/qdisp/Executive.cc
fjammes: :+1:

core/modules/qdisp/Executive.cc
fjammes: :+1:

core/modules/qdisp/Executive.cc
fjammes: :+1:

core/modules/qdisp/Executive.cc
fjammes: :+1: I removed this trace, whole message is logged in a addMessage() below.

core/modules/qdisp/JobStatus.cc
fjammes: :+1:

core/modules/qdisp/JobStatus.cc
fjammes: :+1: , replaced with "%FT%T%z" which is ISO8601 compliant. Wasn't able to easily print milliseconds here.

core/modules/qdisp/JobStatus.cc
fjammes: :+1:

core/modules/qdisp/JobStatus.h
fjammes: I renamed refnum to jobId, Spec to JobDescription and ExecStatus to JobStatus, I think this make more sense because all of this entities are related to a Job (which is what run on the worker for 1 chunk query). I'll do it in separate ticket in the future. Please note that this refactoring allowed me to better understand the code ans was very useful to complete the ticket.

core/modules/qdisp/MessageStore.cc
fjammes: severity enum is now swigged, as a consequence this method was unused and then removed.

core/modules/rproc/InfileMerger.cc
fjammes: :+1: reverted

core/modules/util/Error.h
fjammes: done

core/modules/qdisp/Executive.cc
fjammes: :+1: I created a std::string

core/modules/util/Error.h
fjammes: :+1: added getter/setter, change constructor, add DM-2996. Please note that I integrated existing Error structures in util::Error, so I had to deal with existant, that's why I didn't go to far in refactoring in the first place.

core/modules/ccontrol/queryMsg.cc
```diff
@@ -43,19 +44,23 @@ int queryMsgGetCount(int session) {
 }
 
 // Python call: msg, chunkId, code, timestamp = queryMsgGetMsg(session, idx)
-std::string queryMsgGetMsg(int session, int idx, int* chunkId, int* code, time_t* timestamp) {
+std::string queryMsgGetMsg(int session, int idx, int* chunkId, int* code, lsst::qserv::MessageSeverity* severity, time_t* timestamp) {
```
jbecla: I'd shorten this line - can you put last arg in separate line? Also, you don't need "last::qserv::" in front (revisit this part in other places too)

core/modules/ccontrol/queryMsg.h
```diff
@@ -37,9 +40,10 @@ int queryMsgGetCount(int session);
 
 // Python call: msg, chunkId, code, timestamp = queryMsgGetMsg(session, idx)
 // int* chunkId, int* code, time_t* timestamp matches with %apply directive to help SWIG
-std::string queryMsgGetMsg(int session, int idx, int* chunkId, int* code, time_t* timestamp);
+std::string queryMsgGetMsg(int session, int idx, int* chunkId, int* code, lsst::qserv::MessageSeverity* severity, time_t* timestamp);
 
-void queryMsgAddMsg(int session, int chunkId, int code, std::string const& message);
+void queryMsgAddMsg(int session, int chunkId, int code, std::string const& message,
+                    lsst::qserv::MessageSeverity const& severity = MessageSeverity::MSG_INFO);
```
jbecla: see? You have MessageSeverity here (once) without "last::qserv::"

core/modules/czar/python/app.py
```diff
@@ -292,7 +289,7 @@ def __init__(self, query, context, setSessionId, resultName=""):
         @param context a user context object containing conditions and a user query factory
         @param setSessionId - unary function. a callback so this object can provide
                           a handle (sessionId) for the caller to access query
-                          messages.
+                           messages.
         @param resultName name of result table for query results."""
```
jbecla: why this extra space that is breaking alignment? I'd put "messages." in previous line.

core/modules/czar/python/app.py
```diff
@@ -329,7 +326,7 @@ def __init__(self, query, context, setSessionId, resultName=""):
         except:
             self._error = "Unexpected error: " + str(sys.exc_info())
             logger.err(self._error, traceback.format_exc())
-            self._reportError(-1,
+            self._addProxyMessage(-1,
                               msgCode.MSG_QUERY_INIT,
                               "Initialize Query: " + self.queryStr);
```
jbecla: realign

core/modules/czar/python/proxy.py
```diff
@@ -49,10 +62,19 @@ class Lock:
     acquiring a lock on the table. mysql-proxy can service other
     requests when it is waiting on mysql queries that it knowingly
     initiates in its C-layer, whereas it treats its Lua plugin code as a
-    single-threaded black-box."""
-    createTmpl = "CREATE TABLE IF NOT EXISTS %s (chunkId INT, code SMALLINT, message CHAR(255), timeStamp FLOAT) ENGINE=MEMORY;"
+    single-threaded black-box.
+    """
+
+
```
jbecla: remove one extra empty line?

core/modules/czar/python/proxy.py
```diff
@@ -49,10 +62,19 @@ class Lock:
...
+    createTmpl = "CREATE TABLE IF NOT EXISTS %s " \
+                 "(chunkId INT, code SMALLINT, message CHAR(255), " \
+                 "severity ENUM ('INFO', 'ERROR'), timeStamp FLOAT) ENGINE=MEMORY;"
```
jbecla: fyi, ";" at the end of queries sent through c api is not needed, it is only needed in mysql client. I'd remove it for cleanness.

core/modules/global/constants.h
```diff
@@ -41,5 +41,15 @@ const char ENV_XRDID[] = "XRDNAME";
 
 const char SUBCHUNKDB_PREFIX[] = "Subchunks_";
 const char SCISQLDB_PREFIX[] = "scisql_";
+
```
jbecla: copyright year not changed in this file

core/modules/qdisp/Executive.cc
```diff
@@ -407,25 +427,25 @@ void Executive::_setup() {
...
-        _retryMap[refNum] = 1;
+        _retryMap[jobId] = 1;
     } else if(i->second < MAX_RETRY) {
-        _retryMap[refNum] = 1 + i->second;
+        _retryMap[jobId] = 1 + i->second;
     } else {
         return false;
     }
     return true;
 }
```
jbecla: hmm, this code:
   if (...) {
    } else {
        return false;
    }
    return true;
}
makes no sense, I am not blaming you but we should fix it!

core/modules/qdisp/Executive.cc
```diff
@@ -407,25 +427,25 @@ void Executive::_setup() {
...
 
-ExecStatus::Ptr Executive::_insertNewStatus(int refNum,
+JobStatus::Ptr Executive::_insertNewStatus(int jobId,
                                             ResourceUnit const& r) {
```
jbecla: looks misaligned

core/modules/qdisp/Executive.cc
```diff
@@ -460,11 +480,11 @@ bool Executive::_track(int refNum, RequesterPtr r) {
     return true;
```
jbecla: same here, the combination of if/else is wrong.

core/modules/qdisp/MessageStore.h
```diff
@@ -34,35 +34,52 @@
...
     int code;
     std::string description;
     std::time_t timestamp;
+    MessageSeverity severity;
 };
 
 class MessageStore {
 public:
-    void addMessage(int chunkId, int code, std::string const& description);
+    void addMessage(int chunkId, int code, std::string const& description, MessageSeverity severity_ = MessageSeverity::MSG_INFO);
```
jbecla: break the line into two

core/modules/util/Error.h
```diff
@@ -74,6 +96,47 @@ class Error {
...
+        this->msg = msg;
+    }
+
+    int getStatus() const {
+        return status;
+    }
+
+    void setStatus(int status) {
+        this->status = status;
+    }
```
jbecla: I'd make all of the above one-liners: "void setStatus(int status) { this->status = status; }" etc.

core/modules/util/Error.h
```diff
@@ -40,6 +40,33 @@ namespace lsst {
 namespace qserv {
 namespace util {
```
jbecla: #ifndef in this file is non-conventional: UTIL_ERROR_H_, should be LSST_QSERV_UTIL_ERROR_H. While you are at it, could you also fix:

qana/CheckAggregation.h:#ifndef QANA_CHECKAGGREGATION_H_
qmeta/QMeta.h:#ifndef QMETA_QMETA_H
query/typedefs.h:#ifndef QUERY_TYPES_H_
util/Error.h:#ifndef UTIL_ERROR_H_

core/modules/util/Error.h
```diff
@@ -74,6 +96,47 @@ class Error {
...
+     */
+    bool isNone() {
+        return (code==util::ErrorCode::NONE);
+    }
+
+private:
+
+    int code;
+    std::string msg;
+    int status;
```
jbecla: maybe prepend with underscores (all 3)

andy-slac: No may be, but strong have to :) 
3-10. Private class variables and methods MUST have underscore prefix.

core/modules/ccontrol/MergingRequester.cc
```diff
@@ -227,11 +227,10 @@ bool MergingRequester::_merge() {
...
 
 void MergingRequester::_setError(int code, std::string const& msg) {
-    LOGF_INFO("setError");
-    LOGF_INFO("setError %1% %2%" % code % msg);
+    LOGF_INFO("setError: code: %1%, message: %2%" % code % msg);
     std::lock_guard<std::mutex> lock(_errorMutex);
-    _error.code = code;
-    _error.msg = msg;
+    _error.setCode(code);
+    _error.setMsg(msg);
```
andy-slac: `_error = Error(code, msg);` instead? Again, I do not like modifiable value-types, I'd prefer that setCode() method and other did not even exist in this class. And maybe `MergingRequester::_setError()` does not have to take two parametes but should take `Error` instead? But this is not for this ticket.

core/modules/ccontrol/queryMsg.cc
```diff
@@ -43,19 +44,23 @@ int queryMsgGetCount(int session) {
...
     *code = msg.code;
     *timestamp = msg.timestamp;
+    *severity = msg.severity;
     return msg.description;
 }
 
-void queryMsgAddMsg(int session, int chunkId, int code,
-                            std::string const& message) {
+void queryMsgAddMsg(int session, int chunkId, int code, std::string const& message,
+                    lsst::qserv::MessageSeverity const& severity /* = "INFO" */) {
```
andy-slac: `/* = "INFO" */` should not be here.

core/modules/qdisp/Executive.cc
```diff
@@ -223,74 +233,81 @@ bool Executive::join() {
...
         }
-        LOGF_ERROR("Executive: error executing refnum=%1%. Code=%2% %3%" %
-                   refNum % e.code % e.msg);
     }
-    _unTrack(refNum);
+    _unTrack(jobId);
     if(!success) {
-        LOGF_ERROR("Executive: requesting squash (cause refnum=%1% with code=%2% %3%)" % refNum % e.code % e.msg);
+        LOGF(getLogger(), LOG_LVL_ERROR,
+                 "Executive: requesting squash, cause: jobId=%1% failed (code=%2% %3%)" % jobId % err.getCode() % err.getMsg());
```
andy-slac: Indentation, and make line shorter, it's hard to review code which is wider than github panel size.

core/modules/qdisp/Executive.cc
```diff
@@ -352,33 +369,36 @@ int Executive::getNumInflight() {
...
-    std::for_each(_statuses.begin(), _statuses.end(), printMapEntry(os, "\n"));
-    LOGF_ERROR("%1%" % os.str());
-    return os.str();
+    {
+        std::lock_guard<std::mutex> lock(_statusesMutex);
+        std::for_each(_statuses.begin(), _statuses.end(), printMapEntry(os, "\n"));
+    }
+    auto const & msg_progress = os.str();
+    LOGF(getLogger(), LOG_LVL_ERROR, "%1%" % msg_progress);
+    return std::move(msg_progress);
```
andy-slac: My experience with std::move is very limited and I'm not sure that it does anything useful here. I'd probably just rely on good old NRVO:

    std::string msg_progress = os.str();
    ....
    return msg_progress;

core/modules/qdisp/Executive.cc
```diff
@@ -499,20 +519,26 @@ void Executive::_reapRequesters(std::unique_lock<std::mutex> const&) {
...
-        std::ostringstream os;
-        os << ExecStatus::stateText(info.state)
-           << " " << info.stateCode;
-        if(!info.stateDesc.empty()) {
-            os << " (" << info.stateDesc << ")";
+void Executive::_updateProxyMessages() {
+    {
+        std::lock_guard<std::mutex> lock(_statusesMutex);
+        for(auto i=_statuses.begin(), e=_statuses.end(); i != e; ++i) {
+            auto const & info = i->second->getInfo();
```
andy-slac: For consistency use `const&` and not `const &`

core/modules/qdisp/Executive.cc
```diff
@@ -499,20 +519,26 @@ void Executive::_reapRequesters(std::unique_lock<std::mutex> const&) {
...
+                os << " (" << info.stateDesc << ")";
+            }
+            os << " " << info.stateTime;
+            _messageStore->addMessage(info.resourceUnit.chunk(),
+                                      info.state, os.str());
+        }
+    }
+    {
+        std::lock_guard<std::mutex> lock(_errorsMutex);
+        if (_multiError.size()>0) {
```
andy-slac: `if (!_multiError.empty()) {` should be ~better (or my-preferred most-readable variant `if (not _multiError.empty()) {`)

core/modules/qdisp/JobStatus.cc
```diff
@@ -0,0 +1,162 @@
...
+std::string const JobStatus::_empty;
+
+JobStatus::Info::Info(ResourceUnit const& resourceUnit_)
+    : resourceUnit(resourceUnit_),
+      state(UNKNOWN),
+      stateCode(0) {
+    stateTime = ::time(NULL);
+}
+
+void JobStatus::updateInfo(JobStatus::State s, int code /* =0 */, std::string const& desc /* =_empty */) {
```
andy-slac: Commented-out defaults should not be here.

core/modules/qdisp/JobStatus.cc
```diff
@@ -0,0 +1,162 @@
...
+    _info.state = s;
+    _info.stateCode = code;
+    _info.stateDesc = desc;
+}
+
+std::ostream& operator<<(std::ostream& os, JobStatus::State const& state) {
+    switch(state)
+    {
+    case JobStatus::UNKNOWN:
+        os << "Unknown";
```
andy-slac: I'm not sure how well g++ can optimize this code (reducing ~20 calls to operator<<() into one), I'd probably try to help it a bit by using temporary `char const*` variable:

    char const* msg = 0;
    switch (state) {
    case ...:
        msg = "...";
        break;
    default: 
        msg = "...";
    }
    return os << msg;

core/modules/qdisp/MessageStore.cc
```diff
@@ -34,28 +34,39 @@
...
+namespace {
+    const int NOTSET = -1;
+}
+
 ////////////////////////////////////////////////////////////////////////
 // public
 ////////////////////////////////////////////////////////////////////////
 
-void MessageStore::addMessage(int chunkId, int code, std::string const& description) {
+void MessageStore::addMessage(int chunkId, int code, std::string const& description, MessageSeverity severity /* = INFO */) {
```
andy-slac: `/* = INFO */` does not belong here.

core/modules/qdisp/MessageStore.h
```diff
@@ -34,35 +34,52 @@
 
 // System headers
 #include <ctime>
+#include <iostream>
```
andy-slac: I don't think this is needed here

core/modules/qdisp/testQDisp.cc
```diff
@@ -398,8 +397,8 @@ class CancellableRequester : public qdisp::ResponseRequester {
         return true;
     }
     virtual void errorFlush(std::string const& msg, int code) {
-        _lastError.msg = msg;
-        _lastError.code = code;
+        _lastError.setMsg(msg);
+        _lastError.setCode(code);
```
andy-slac: `_lastError = Error(code, msg);`

core/modules/rproc/InfileMerger.cc
```diff
@@ -77,14 +77,19 @@ using lsst::qserv::proto::ProtoImporter;
 using lsst::qserv::proto::WorkerResponse;
 using lsst::qserv::rproc::InfileMergerConfig;
 using lsst::qserv::rproc::InfileMergerError;
+using lsst::qserv::util::ErrorCode;
```
andy-slac: There is one use of `ErrorCode` and eight `util::ErrorCode` in this file, may be better stick to `util::ErrorCode` everywhere?

core/modules/util/Error.h
andy-slac: I'd make all of set* method disappear, in my opinion that are not need and interface would be _much_ better without them.

core/modules/qdisp/JobStatus.cc
ktlim: This one is probably OK; only one call will be executed, and the space costs are trivial.

core/modules/util/Error.h
ktlim: I agree with AndyS.

core/modules/ccontrol/queryMsg.cc
fjammes: :+1:

fjammes: :+1:

core/modules/czar/python/app.py
fjammes: :+1:

core/modules/czar/python/app.py
fjammes: :+1:

core/modules/czar/python/proxy.py
fjammes: :+1:

core/modules/czar/python/proxy.py
fjammes: :+1:

core/modules/global/constants.h
fjammes: :+1:

core/modules/qdisp/Executive.cc
fjammes: :+1:, I've added a boolean should_retry, hope it's fine now. I didn't write this code, so I can't fine tune-it in a deeper level quickly.

core/modules/qdisp/Executive.cc
fjammes: :+1:

core/modules/qdisp/Executive.cc
fjammes: :+1: , this function is unused, I removed it.

core/modules/ccontrol/MergingRequester.cc
fjammes: I agree this clearer, but using `_error = Error(code, msg);` will create one additional error object (_error is created with default constructor while constructing parent object) and call copy constructor. Is it as performant than calling setters?

core/modules/qdisp/MessageStore.h
fjammes: :+1: , and I also documented this method

core/modules/util/Error.h
fjammes: :+1: done

core/modules/util/Error.h
fjammes: :+1:

core/modules/util/Error.h
fjammes: :+1:

core/modules/ccontrol/MergingRequester.cc
fjammes: :+1: but I'm still interested to know about question below.

core/modules/qdisp/Executive.cc
fjammes: :+1: this might interest you: https://userstyles.org/styles/108591/github-wide

core/modules/qdisp/Executive.cc
fjammes: :+1:

core/modules/qdisp/Executive.cc
fjammes: :+1:

core/modules/qdisp/Executive.cc
fjammes: :+1:

core/modules/qdisp/JobStatus.cc
fjammes: :+1: default added to doc

core/modules/qdisp/JobStatus.cc
fjammes: :+1:

core/modules/qdisp/MessageStore.h
fjammes: :+1:

core/modules/qdisp/testQDisp.cc
fjammes: :+1:

core/modules/rproc/InfileMerger.cc
fjammes: :+1: , I deleted all using directive for consistence

core/modules/ccontrol/MergingRequester.cc
andy-slac: Reasonable compiler should be able to optimize that temporary away (if all methods are inline). Even if not the performance penalty should be negligible unless you need to create billions of Error instances.

fjammes: Ok it seems reasonable as Qserv requirement is to management billion of 
stars, and not billion of errors :-)

PR #100, opened 2015-06-29
DM-2643: Migrate Qserv to ssi v2

fritzm:
Adapt Qserv to the latest XrdSsi implementation, which has some backward-compato-breaking API changes (all error returns are now via callback, new "Provider" API to locate service objects).

This is some prep work to streamline taking patches/fixes from AndyH. while we are pursuing the distributed query bug -- we cannot take this patch into master until we update the xrootd package in eups.  I'd propose that we do this on the upcoming release in a few days, though?

Note that this cl does not include changes to take advantage of the new feature of the ssi API that would allow us to implement QueryResource directly within the xrdsvc binary and dispense with the xrdoss binary entirely; I will create a new JIRA ticket for that.

### DIFF
--- a/core/modules/qdisp/XrdSsiMocks.h
+++ b/core/modules/qdisp/XrdSsiMocks.h
@@ -48,7 +48,7 @@ class Executive;
 class XrdSsiServiceMock : public XrdSsiService
 {
 public:
-    virtual bool Provision(Resource *resP, unsigned short timeOut=0);
+    virtual void Provision(Resource *resP, unsigned short timeOut=0);
     XrdSsiServiceMock(Executive *executive) : _executive(executive) {};
     void setGo(bool go) {
         _go.set(go);
@@ -70,11 +70,11 @@ class XrdSsiSessionMock : public XrdSsiSession
 public:
     XrdSsiSessionMock(char *sname, char *sloc=0) : XrdSsiSession(sname, sloc) {}
     virtual ~XrdSsiSessionMock() {}
-    static const char* getMockString(bool tf) {
-        return (tf) ? "MockTrue" : "MockFalse";
+    static const char* getMockString() {
+        return "MockTrue";
     }
-    virtual bool ProcessRequest(XrdSsiRequest *reqP, unsigned short tOut){
-        std::string s = getMockString(true);
+    virtual void ProcessRequest(XrdSsiRequest *reqP, unsigned short tOut){
+        std::string s = getMockString();
         bool res = s.compare(sessName) == 0;
         LOGF_INFO("sessName=%1% res=%2%" % sessName % res);
         // Normally, reqP->ProcessResponse() would be called, which invokes
@@ -84,7 +84,6 @@ public:
         if (r) {
             r->cleanup();
         }
-        return res;
     };
     virtual bool Unprovision(bool forced=false){return true;};
 
--- a/core/modules/qdisp/testQDisp.cc
+++ b/core/modules/qdisp/testQDisp.cc
@@ -266,24 +266,12 @@ BOOST_AUTO_TEST_CASE(QueryResource) {
     r->ProvisionDone(NULL);
     BOOST_CHECK(status.getInfo().state  == qdisp::ExecStatus::PROVISION_NACK);
     char buf[20];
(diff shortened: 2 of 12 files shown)


abh3:
Hi Fritz,

Good work! I will also be checking in a set of patches to fix some other 
issues:

1) Use old-style locking to provide a strong memory model (fixes helgrind 
complaints).

2) Always do callbacks on an independent thread instead of the response 
handler thread (avoids deadlock when soket interrupts have been masked).

Andy

core/modules/xrdsvc/plugin.cc
```diff
@@ -20,30 +20,57 @@
  * the GNU General Public License along with this program.  If not,
```
jgates108: Copyright

core/modules/qdisp/XrdSsiMocks.h
```diff
@@ -84,7 +84,7 @@ class XrdSsiSessionMock : public XrdSsiSession
         if (r) {
             r->cleanup();
         }
-        return res;
+        return;
```
jbecla: I'd drop this return.

core/modules/xrdsvc/plugin.cc
```diff
@@ -20,30 +20,57 @@
...
+    }
+
+    virtual bool Init(
+        XrdSsiLogger* logP,
+        XrdSsiCluster* clsP,
+        char const* cfgFn,
+        char const* parms,
+        int argc,
+        char **argv)
+    {
```
jgates108: Perhaps, collapse the arguments down to one or two lines as we don't seem to be using most of them.

core/modules/xrdsvc/plugin.cc
```diff
@@ -20,30 +20,57 @@
...
+namespace lsst {
+namespace qserv {
+namespace xrdsvc {
 
-extern "C" {
-XrdSsiService *XrdSsiGetServerService(XrdSsiLogger  *logP,
-                                      XrdSsiCluster *clsP,
-                                      const char    *cfgFn,
-                                      const char    *parms)
+class SsiProviderServer : public XrdSsiProvider
```
jbecla: I'd stay consistent with the rest of qserv: put the bracket on previous line (in many places in this file)

core/modules/qdisp/testQDisp.cc
```diff
@@ -266,23 +266,11 @@ BOOST_AUTO_TEST_CASE(QueryResource) {
...
-            qdisp::Executive::newNotifier(ex, refNum),
-            retryTest, status);
-    r->ProvisionDone(&xsMockFalse);
-    BOOST_CHECK(status.getInfo().state  == qdisp::ExecStatus::REQUEST_ERROR);
-    BOOST_CHECK(retryTest->_retryCalled == true);
-    // At this point, r has been deleted by QueryResource::ProvisionDone
-
-    LOGF_INFO("QueryResource test 2");
-    retryTest->_retryCalled = false;
+    strcpy(buf, qdisp::XrdSsiSessionMock::getMockString(true));
```
jgates108: Should getMockString lose it's boolean argument? I'm not sure it is ever false now.

core/modules/qdisp/XrdSsiMocks.h
fritzm: Yup, will do, thx.

core/modules/qdisp/testQDisp.cc
fritzm: I wasn't super clear on how the test plumbing worked downstream of here -- that is, I didn't want to remove functionality in the mocks that you think might be used in other tests?  On your recommend, happy to remove?

core/modules/xrdsvc/plugin.cc
fritzm: Yup, thx!

core/modules/xrdsvc/plugin.cc
fritzm: Ok.

core/modules/xrdsvc/plugin.cc
fritzm: Ok.

PR #102, opened 2015-07-01
Tickets/dm 2953

fritzm:


fritzm:
Looks good to me -- thanks for the cleanup work!

PR #103, opened 2015-07-08
Adding test suite for SqlTransaction class (DM-3090)

andy-slac:
This is not really a unit test as it needs database instance up and
running.

Small change to SqlTransaction class itself, abort() and commit() now
return bool value which is true if operation succeeds, this matches
better other interfaces in this package.

### DIFF
--- a/core/modules/sql/testSqlTransaction.cc
+++ b/core/modules/sql/testSqlTransaction.cc
@@ -1,0 +1,259 @@
+/*
+ * LSST Data Management System
+ * Copyright 2015 AURA/LSST.
+ *
+ * This product includes software developed by the
+ * LSST Project (http://www.lsst.org/).
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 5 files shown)


core/modules/sql/testSqlTransaction.cc
```diff
@@ -0,0 +1,260 @@
...
+    }
+    ~PerTestFixture() {
+        // drop table
+        std::string query = "DROP TABLE " FULL_TABLE_NAME;
+        SqlErrorObject errObj;
+        sqlConn->runQuery(query, errObj);
+    }
+
+    static TestDBGuard testDB;
+    std::shared_ptr<SqlConnection> sqlConn;
```
fritzm: Nit: Should this be unique_ptr instead?  It looks like the pointer-smartness is only being used for lifetime control (pointer is deref'd everywhere below before being passed, unless I missed somthing?)

For that matter, could this just be an embedded member var instead of a pointer?

andy-slac: You are right, we don't need anything smart here. I'll update it.

PR #104, opened 2015-07-09
Tickets/dm 3110

andy-slac:


admin/python/lsst/qserv/admin/qservAdmin.py
```diff
@@ -314,7 +314,7 @@ def _createDbLikeLocked(self, dbName, templateDbName):
             self._createDbLockSection(dbP)
             self._kvI.set(dbP, "READY")
         except KvException as e:
-            self._logger.error("Failed to create database '%s' like '%s', error was: %s",
+            self._logger.error("Failed to create db '%s' like '%s', error was: %s",
```
andy-slac: One small trick - I prefer to use `%r` instead of `'%s'` in cases when I want to have quoted (and escaped) strings on output, frees me from the need to think whether do I need to escape quotes inside string.

core/modules/ccontrol/UserQueryFactory.cc
```diff
@@ -101,19 +103,19 @@ UserQueryFactory::newUserQuery(std::string const& query,
...
     if(!qs->getError().empty()) {
-        LOGF_INFO("Invalid query: %s" % qs->getError());
+        LOGF(_log, LOG_LVL_INFO, "Invalid query: %s" % qs->getError());
         sessionValid = false;
     }
     if(!qs->getError().empty()) {
-        LOGF_INFO("Invalid query: %s" % qs->getError());
+        LOGF(_log, LOG_LVL_INFO, "Invalid query: %s" % qs->getError());
         sessionValid = false;
     }
```
andy-slac: Does the code above repeat the same `if () {...}` three times (once with ERROR and twice with INFO)?

andy-slac: And it must be `%1%` instead of `%s`.

core/modules/css/python/kvInterface.py
```diff
@@ -605,8 +608,6 @@ def getChildren(self, k):
         try:
             self._logger.info("GETCHILDREN '%s'" % (k))
             c = self._kvi.getChildren(k)
```
andy-slac: This line should be removed as well

core/modules/czar/python/css.py
```diff
@@ -86,7 +86,6 @@ def __init__(self, **kwargs):
                           "timeout" : 10000 }
         elif "config" in kwargs:
             self._cfg = kwargs["config"]
-            #raise KvException(KvException.MISSING_PARAM, str(self._cfg))
             self._logger.info("Using css config: %s" %  str(self._cfg))
```
andy-slac: `str()` is superfluous here.

core/modules/czar/python/css.py
```diff
@@ -134,19 +133,16 @@ def refresh(self):
         return self.refreshZk()
 
     def refreshFile(self):
-        """Refresh using self._file . Rereads full state from
-        self._filename . There isn't a clear way to avoid reading the
-        whole file, so we won't, seeing as this mode is primarily
-        (exclusively?) for debugging.
-        entire tree.
+        """Refresh using self._file . Rereads full state from self._filename .
```
andy-slac: Spaces before dots

core/modules/czar/python/css.py
```diff
@@ -134,19 +133,16 @@ def refresh(self):
...
+        as this mode is primarily (exclusively?) for debugging.
         """
         self.snapshot = KvInterfaceImplMem(self._file)
 
     def refreshZk(self):
-        """Refresh using self._zk .Currently, pulls completely new
-        state. In the future, it should check modification/creation
-        times against previous state in order to avoid visiting the
-        entire tree.
+        """Refresh using self._zk .Currently, pulls completely new state. In the
```
andy-slac: Dot-space instead of space-dot

core/modules/proto/FakeProtocolFixture.h
```diff
@@ -74,8 +73,8 @@ class FakeProtocolFixture {
         }
```
andy-slac: `s = f.mutable_subchunks();` repeats for both if and else, can be moved outside if, and merged with declaration.

admin/python/lsst/qserv/admin/qservAdmin.py
jbecla: All right, I am replacing this consistently in all logging messages. If you think that is too much, let me know :)

core/modules/ccontrol/UserQueryFactory.cc
jbecla: Indeed. I suspect it happened during resolving merge conflicts. It wasn't me though, it is in cc9fcf9fc2.

PR #110, opened 2015-07-28
add qserv-restart.sh, for convenience

andy-slac:


fritzm:
Yes, please, thanks!

admin/templates/configuration/bin/qserv-restart.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+{{QSERV_RUN_DIR}}/bin/qserv-stop.sh
+{{QSERV_RUN_DIR}}/bin/qserv-start.sh
+{{QSERV_RUN_DIR}}/bin/qserv-status.sh
```
andy-slac: "restart" usually means "stop + start", not "stop + start + status", if you need to run qserv-status because qserv-start does not shown some info then maybe qserv-start needs to be modified to display that info.

I would probably make qserv-start conditional and only run it if qserv-stop returns OK code like:

    {{QSERV_RUN_DIR}}/bin/qserv-stop.sh && {{QSERV_RUN_DIR}}/bin/qserv-start.sh

fritzm: It's very convenient to have "status" there also, to see the PID for attaching a debugger, tho...

andy-slac: yes, I know, but you probably want it in qserv-start as well?

fritzm: makes sense

jbecla: Ok, I switched to conditional restart, I kept the qserv-status.sh. I don't want to redo all the init.d scripts, it was supposed to be a trivial 0.1 SP, low-hanging fruit type thing. I favor practicality over perfection :)

ktlim: Practicality is fine as long as we record somewhere the need for cleanup later.

PR #111, opened 2015-07-29
Tickets/DM-3161

andy-slac:


core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,319 @@
...
+ *
+ * You should have received a copy of the LSST License Statement and
+ * the GNU General Public License along with this program.  If not,
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+
+/**
+  * @file
+  *
+  * @brief Interface to the Common State System - MySql-based implementation.
```
vaikuntht: Just poking in - I thought CSS was Central State System?

andy-slac:
One general comment about commit messages, they have to be formatted like:

    One-line brief summary of the commit.

    Multi-line detailed description separated from first line by one empty line...

core/modules/css/KvInterface.h
```diff
@@ -42,18 +42,28 @@ namespace css {
 
 class KvInterface {
 public:
+
+    /**
+     * Sets a max length for kvKey.
+     * If it is changed it must also be changed in the
+     * CssData SQL Schema as well.
+     */
+    const unsigned int kMaxKeyLength = 255;
```
andy-slac: This is declared as data member of a class, which we do not want or need. Either declare is as static or make it an enum.

Our coding guidelines say:
> Named constants (including enumeration values) MUST be all uppercase using underscore to separate words

so the name should look like `MAX_KEY_LENGTH`

core/modules/css/KvInterface.h
```diff
@@ -42,18 +42,28 @@ namespace css {
...
+
     virtual ~KvInterface() {};
 
     /**
      * Create a key/value pair.
-     * Throws KeyExistsError if the key already exists (or if any other problem,
-     * e.g., a connection error is detected).
+     * key must be shorter than kMaxKeyLength
```
andy-slac: Start sentence with capital letter and end with dot. What happens if key length exceeds the limit, should we add special exception for that?

core/modules/css/KvInterfaceImplMySql.cc
andy-slac: I believe it is indeed Central (but it is also common :)

admin/templates/configuration/tmp/configure/sql/CssData.sql
```diff
@@ -0,0 +1,33 @@
...
+USE `qservCssData` ;
+
+-- -----------------------------------------------------
+-- Table `qservCssData`.`kvData`
+-- -----------------------------------------------------
+CREATE TABLE IF NOT EXISTS `qservCssData`.`kvData` (
+  `kvKey` CHAR(255) NOT NULL COMMENT 'slash delimited path',
+  `kvVal` TEXT(1024) NULL COMMENT 'value for key',
+  `parentKey` CHAR(255) NULL COMMENT 'slash delimited path of parent key',
+  PRIMARY KEY (`kvKey`)  COMMENT '')
```
andy-slac: We should probably add foreign key constraint for parent key here to enforce referential integrity. I'd also think of defining `kvId` integer auto-increment column to be used as primary key, so that `parentKey` refers to that instead of `kvKey` column. `kvKey` then has to be made unique.

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,319 @@
...
+
+private:
+    sql::SqlErrorObject _errObj; // this must be declared before _trans
+    sql::SqlTransaction _trans;
+};
+
+
+KvInterfaceImplMySql::KvInterfaceImplMySql(mysql::MySqlConfig const& mysqlConf)
+: _conn(mysqlConf) {
+
```
andy-slac: This empty line does not have to exist :)

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,319 @@
...
+KvInterfaceImplMySql::create(std::string const& key, std::string const& value) {
+    if (key.size() > kMaxKeyLength) {
+        LOGF(_logger, LOG_LVL_DEBUG, "create - rejecting; key too long: %1%" % key);
+        throw CssError("key too long");
+    }
+
+    KvTransaction transaction(_conn);
+
+    boost::filesystem::path keyPathObj(key);
+    boost::filesystem::path parentKeyPathObj = keyPathObj.parent_path();
```
andy-slac: These two variables are not used below

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,319 @@
...
+
+    std::string parentKey = boost::filesystem::path(key).parent_path().string();
```
andy-slac: That will not work correctly on non-Linux platforms.

core/modules/css/KvInterface.h
andy-slac: And this does not compile without `static` on gcc44. Jenkins can be used to test building at >1 platform at a time: https://ci.lsst.codes/job/stack-os-matrix/build?delay=0sec

core/modules/css/CssError.h
```diff
@@ -48,6 +52,12 @@ class CssError : public std::runtime_error {
 public:
     explicit CssError(std::string const& msg)
         : std::runtime_error(msg) {}
+
+    CssError(sql::SqlErrorObject const& sqlErr)
+        : CssError("Error from mysql: ("
```
andy-slac: Nice new feature, but it does not compile with gcc44.

core/modules/css/CssError.h
```diff
@@ -48,6 +52,12 @@ class CssError : public std::runtime_error {
 public:
     explicit CssError(std::string const& msg)
         : std::runtime_error(msg) {}
+
+    CssError(sql::SqlErrorObject const& sqlErr)
```
andy-slac: Add `explicit` here too, and to all single argument constructors in this file.

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,319 @@
...
+
+    boost::format fmQuery("INSERT INTO cssData (kvKey, kvVal, parentKey) VALUES ('%1%', '%2%', '%3%');");
+    fmQuery % key;
+    fmQuery % value;
+    fmQuery % parentKey;
```
andy-slac: What about Bobby Tables key? :)

You need to properly escape all string values with `_conn.escapeString(string)`. And I would not use boost format here, it's kind of expensive, just concatenating strings is more light-weight.

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,319 @@
...
+    fmQuery % value;
+    std::string query = fmQuery.str();
+
+    // run query
+    sql::SqlErrorObject errObj;
+    LOGF(_logger, LOG_LVL_DEBUG, "set - executing query: %1%" % query);
+    if (not _conn.runQuery(query, errObj)) {
+        LOGF(_logger, LOG_LVL_ERROR, "%1% failed with err:%2%" % query % errObj.errMsg());
+        throw CssError(errObj);
+    } else if (errObj.isSet()) {
```
andy-slac: When `errObj.isSet()` is true `_conn.runQuery()` always returns false, so the second branch will never be executed

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,319 @@
...
+    errObj.reset();
+    std::vector<std::string> ret;
+    if (not results.extractFirstColumn(ret, errObj)) {
+        LOGF(_logger, LOG_LVL_ERROR, "getChildren - failed to extract children from %1% failed with err:%2%" % query % errObj.errMsg());
+        throw CssError((boost::format("getChildren - failed to extract children error:%1% from query:%2%") % errObj.errMsg() % query).str());
+    }
+
+    transaction.commit();
+
+    return ret;
```
andy-slac: I think this returns full path names of all children, other kvInterface implementations return relative names, we should make them all consistent.

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,319 @@
...
+}
+
+
+bool
+KvInterfaceImplMySql::exists(std::string const& key) {
+    KvTransaction transaction(_conn);
+
+    std::string val;
+    sql::SqlErrorObject errObj;
+    return _getValFromServer(key, val, errObj);
```
andy-slac: This may be suboptimal for large values, I'd implemented this separately with `SELECT COUNT(*) FROM cssData WHERE kvKey = ...`

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,319 @@
...
+
+    return ret;
+}
+
+
+void
+KvInterfaceImplMySql::deleteKey(std::string const& key) {
+    KvTransaction transaction(_conn);
+
+    boost::format fmQuery("DELETE FROM cssData WHERE kvKey='%1%';");
```
andy-slac: What happens if key has children?

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,319 @@
...
+        throw CssError(errObj);
+    }
+
+    // limit the row count to 1 - there should not be multiple rows for any key.
+    auto affectedRows = resultsObj.getAffectedRows();
+    if (affectedRows < 1) {
+        LOGF(_logger, LOG_LVL_ERROR, "SQL DELETE FROM failed (no such key): %1%" % query);
+        throw NoSuchKey(errObj);
+    }
+    else if (affectedRows > 1) {
```
andy-slac: `else` should be on the same line with '}'

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,319 @@
...
+
+std::string
+KvInterfaceImplMySql::_get(std::string const& key, std::string const& defaultValue, bool throwIfKeyNotFound) {
+    std::string val;
+    sql::SqlErrorObject errObj;
+    if (not _getValFromServer(key, val, errObj)) {
+        if (throwIfKeyNotFound) {
+            throw NoSuchKey(errObj);
+        }
+        return defaultValue;
```
andy-slac: This can potentially break NRVO, it may be better to replace it with `val = defaultValue;`.

core/modules/css/KvInterfaceImplMySql.cc
andy-slac: `_getValFromServer` can return false when either the key is missing or something is wrong on mysql side. You probably want to look at `errObj` and throw an exception if it has an error condition set.

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,319 @@
...
+
+    transaction.commit();
+}
+
+
+std::string
+KvInterfaceImplMySql::_get(std::string const& key, std::string const& defaultValue, bool throwIfKeyNotFound) {
+    std::string val;
+    sql::SqlErrorObject errObj;
+    if (not _getValFromServer(key, val, errObj)) {
```
andy-slac: Same here, false returned from _getValFromServer means two things, you want to check `errObj` and always throw (something that is not NoSuchKey) if error is set in it.

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    KvTransaction transaction(_conn);
+
+    boost::filesystem::path keyPathObj(key);
+    boost::filesystem::path parentKeyPathObj = keyPathObj.parent_path();
+
+    std::string parentKey = boost::filesystem::path(key).parent_path().string();
+
+    boost::format fmQuery("INSERT INTO cssData (kvKey, kvVal, parentKey) VALUES ('%1%', '%2%', '%3%');");
```
andy-slac: Database name that you defined in SQL file is `qservCssData`, not `cssData`.

andy-slac: Table name in SQL file is 'kvData`, not `cssData`

andy-slac: What happens if parent key does not exist in database?

core/modules/css/testKvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,200 @@
...
+
+
+BOOST_AUTO_TEST_CASE(CreateAndGetKV) {
+    kvInterface->create("/foo/testKey", "testValue");
+    BOOST_CHECK_EQUAL(kvInterface->get("/foo/testKey"), "testValue");
+}
+
+
+BOOST_AUTO_TEST_CASE(CreateWithBackslash) {
+    kvInterface->create("\\foo\\testKey", "testValue");
```
andy-slac: Hmm, are we required to support backslashes?

core/modules/css/testKvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,200 @@
...
+    kvInterface->create("setAndGetChildren", "");
+    // then create children to use in test:
+    kvInterface->create("setAndGetChildren/child0", "abc");
+    kvInterface->create("setAndGetChildren/child1", "abc");
+    kvInterface->create("setAndGetChildren/child2", "abc");
+    std::vector<std::string> children;
+    BOOST_REQUIRE_NO_THROW(children = kvInterface->getChildren("setAndGetChildren"));
+    BOOST_REQUIRE(children.size() == 3U);
+    std::sort(children.begin(), children.end());
+    BOOST_CHECK_EQUAL(children[0], "setAndGetChildren/child0");
```
andy-slac: This should be `BOOST_CHECK_EQUAL(children[0], "child0");`

core/modules/css/KvInterfaceImplMySql.cc
n8pease: If there’s no data associated with a parent key, does it have to exist?

core/modules/css/testKvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,200 @@
...
+    // verify the change:
+    BOOST_CHECK_EQUAL(kvInterface->get("SetNonExistentKey"), "toANewValue");
+}
+
+
+BOOST_AUTO_TEST_CASE(KeyTooLong) {
+    std::string tooLongKey("abcdefghijklmnopqrstuvwxyz1234abcdefghijklmnopqrstuvwxyz1234abcdefghijklmnopqrstuvwxyz1234abcdefghijklmnopqrstuvwxyz1234abcdefghijklmnopqrstuvwxyz1234abcdefghijklmnopqrstuvwxyz1234abcdefghijklmnopqrstuvwxyz1234abcdefghijklmnopqrstuvwxyz1234abcdefghijklmnopabcdefghijklmnopqrstuvwxyz1234abcdefghijklmnopqrstuvwxyz1234abcdefghijklmnopqrstuvwxyz1234abcdefghijklmnopqrstuvwxyz1234abcdefghijklmnopqrstuvwxyz1234abcdefghijklmnopqrstuvwxyz1234abcdefghijklmnopqrstuvwxyz1234abcdefghijklmnopqrstuvwxyz1234abcdefghijklmnop");
+    BOOST_CHECK_THROW(kvInterface->set(tooLongKey, "to value"), lsst::qserv::css::CssError);
+}
+
```
andy-slac: Maybe one more test for NoSuchKey exception for `get('/missing key')`?

core/modules/css/testKvInterfaceImplMySql.cc
n8pease: I don’t know, but I thought I would check that it works at least as well as that. I can nuke it entirely though. Any preference? Should I sanitize against backslash characters in key strings?

core/modules/css/KvInterfaceImplMySql.cc
andy-slac: I believe yes. Otherwise it would be hard to discover the whole tree.

core/modules/css/testKvInterfaceImplMySql.cc
andy-slac: I do not think we have any requirements on the content of the key string except that forward slash is a separator. I'd just ignore any fancy stuff in the string, and hope people know what they are doing.

core/modules/css/KvInterface.h
jbecla: Regarding names, see https://jira.lsstcorp.org/browse/DM-3115

core/modules/css/KvInterfaceImplMySql.cc
jbecla: We need better unit tests to expose these problems.

n8pease: I'll add tests for that (I hadn't identified it as a requirement previously).

n8pease: In the case where the parent key does not yet exist: is that an error, or should parent key(s) be created as needed?

andy-slac: I'd try to make it more convenient to client and created all parents as needed. There may be some implications w.r.t. locking and but we'll deal with that later.

jbecla: I was just saying that if (a) table name is wrong then unit test should pick it up (unless it is consistently wrong everywhere), and (b) we need to test for various conditions: parent key exist / does not.

admin/templates/configuration/tmp/configure/sql/CssData.sql
```diff
@@ -0,0 +1,36 @@
...
+-- Table `qservCssData`.`kvData`
+-- -----------------------------------------------------
+CREATE TABLE IF NOT EXISTS `qservCssData`.`kvData` (
+  `kvKey` CHAR(255) NOT NULL COMMENT 'slash delimited path',
+  `kvVal` TEXT(1024) NOT NULL COMMENT 'value for key',
+  `parentKvId` INT UNSIGNED NULL COMMENT 'id of parent key',
+  `kvId` INT UNSIGNED NOT NULL AUTO_INCREMENT,
+  PRIMARY KEY (`kvId`),
+  UNIQUE INDEX `kvKey_UNIQUE` (`kvKey` ASC))
+ENGINE = InnoDB;
```
andy-slac: Some improvements for this schema I think may be useful:
- I believe it's custom to define primary key column(s) at the begging, I'd move `kvId` to the top
- add foreign key constraint for `parentKvId` to make sure that data is consistent
- `TEXT(1024)` is the same storage type as `TEXT` and its size is limited to 64 kBytes, we probably want larger limit so I suggest to replace it with `MEDIUMTEXT`
- `UNSIGNED` is mysql-ism, we want to avoid it

core/modules/css/CssError.h
```diff
@@ -36,6 +36,10 @@
 #include <map>
 #include <stdexcept>
 #include <string>
+#include <boost/lexical_cast.hpp>
+
+#include "util/Issue.h"
```
andy-slac: I don't think anything from this header is used below?

core/modules/css/KvInterface.h
```diff
@@ -42,18 +42,29 @@ namespace css {
...
+     * If the parent key does not exist it will be created with an empty value.
+     * Throws KeyExistsError if the key already exists
+     * Throws CssError for other problems (e.g., a connection error is detected).
      */
     virtual void create(std::string const& key, std::string const& value) = 0;
 
     /**
      * Set a key/value pair. If the key already exists, its value is
      * overwritten.
+     * key must be shorter than kMaxKeyLength
```
andy-slac: MAX_KEY_LENGTH

core/modules/css/KvInterface.h
```diff
@@ -42,18 +42,29 @@ namespace css {
...
     virtual ~KvInterface() {};
 
     /**
-     * Create a key/value pair.
-     * Throws KeyExistsError if the key already exists (or if any other problem,
-     * e.g., a connection error is detected).
+     * Create a slash-delimited key-value pair.
+     * Key must be shorter than MAX_KEY_LENGTH.
+     * If the parent key does not exist it will be created with an empty value.
+     * Throws KeyExistsError if the key already exists
```
andy-slac: For Doxygen comments I'd prefer to have special doxygen markup like:

    @throws KeyExistsError if the key already exists

core/modules/css/KvInterface.h
```diff
@@ -86,14 +98,15 @@ class KvInterface {
...
-     * e.g., a connection error is detected)
+     * Throws NoSuchKey if the key does not exist
+     * Throws CssError for other problems (e.g., a connection error is detected).
      */
     virtual std::vector<std::string> getChildren(std::string const& key) = 0;
 
     /**
-     * Delete a key.
-     * Throws CssRunTimeError on failure.
+     * Delete a key, and all of it's children (if they exist)
```
andy-slac: its

admin/templates/configuration/tmp/configure/sql/CssData.sql
n8pease: all the documentation about foreign keys talks about separate tables. will they work on the same table?

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,503 @@
...
+        return true;
+    }
+}
+
+// todo this is copied more or less directly from QMetaTransaction. Refactor to shared location?
+// proper place is to port it into db module
+// https://github.com/LSST/db
+// right now it's python only.
+// create a ticket to make generalized code to be used between c++ and python
+class KvTransaction
```
andy-slac: nitpick - I'd put this class in anonymous namespace. But it has to disappear anyways so I don't worry too much.

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,503 @@
...
+KvInterfaceImplMySql::KvInterfaceImplMySql(mysql::MySqlConfig const& mysqlConf)
+: _conn(mysqlConf) {
+
+}
+
+
+void
+KvInterfaceImplMySql::_findParentId(std::string const& childKvKey, bool* hasParent, unsigned int* parentKvId) {
+
+    // This should only be called when a transaction is in process.
```
andy-slac: another nitpick - one way to ensure that transaction has started is to require it (transaction object) as a parameter to this method (even if you do not do anything with that object).

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,503 @@
...
+
+
+KvInterfaceImplMySql::KvInterfaceImplMySql(mysql::MySqlConfig const& mysqlConf)
+: _conn(mysqlConf) {
+
+}
+
+
+void
+KvInterfaceImplMySql::_findParentId(std::string const& childKvKey, bool* hasParent, unsigned int* parentKvId) {
```
andy-slac: I'd probably just return an integer value from this method, >=0 is the parent key, and -1 if parent key does not exist

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,503 @@
...
+    }
+
+    size_t loc = childKvKey.find_last_of(KEY_PATH_DELIMITER);
+    if (loc == 0) {
+        *hasParent = false;
+        return;
+    }
+    std::string parentKey = childKvKey.substr(0, loc);
+    boost::format fmQuery("SELECT kvId FROM kvData WHERE kvKey='%1%'");
+    fmQuery % parentKey;
```
andy-slac: This needs escaping, like any string that you interpolate into query.

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,503 @@
...
+KvInterfaceImplMySql::_findParentId(std::string const& childKvKey, bool* hasParent, unsigned int* parentKvId) {
+
+    // This should only be called when a transaction is in process.
+
+    if (hasParent == NULL || parentKvId == NULL) {
+        throw CssError("null out-var ptr");
+    }
+
+    size_t loc = childKvKey.find_last_of(KEY_PATH_DELIMITER);
+    if (loc == 0) {
```
andy-slac: Does this method return expected result if slash is not even in the key string?

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,503 @@
...
+    LOG_LOGGER _logger = LOG_GET("lsst.qserv.css.KvInterfaceImplMySql");
+    const char KEY_PATH_DELIMITER('/');
+
+    /**
+     * @brief Helper for getting the value from an INT row
+     * @param results the object containing results information
+     * @param value the output var to return the result
+     * @return true if an int result could be obtained else false
+     */
+    bool extractIntValueFromSqlResults(sql::SqlResults* results, unsigned int* value) {
```
andy-slac: I think you can pass results by reference.

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,503 @@
...
+    // A transaction should be in process here.
+
+    unsigned int parentKvId(0);
+    bool hasParent(false);
+    _findParentId(key, &hasParent, &parentKvId);
+
+    boost::format fmQuery;
+    if (hasParent) {
+        fmQuery = boost::format("INSERT INTO kvData (kvKey, kvVal, parentKvId) VALUES ('%1%', '%2%', '%3%')");
+        fmQuery % key % value % parentKvId;
```
andy-slac: String escaping here and below.

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,503 @@
...
+            break;
+        }
+    }
+
+    std::string getIdQuery = "SELECT LAST_INSERT_ID()";
+    sql::SqlResults results;
+    errObj.reset();
+    if (not _conn.runQuery(getIdQuery, results, errObj)) {
+        throw CssError(errObj);
+    }
```
andy-slac: Do you need these 6 lines?

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,503 @@
...
+
+    unsigned int kvId = _conn.getInsertId();
+    LOGF(_logger, LOG_LVL_DEBUG, "_create - executed query: %1%, kvId is:%2%" % query % kvId);
+    return kvId;
+}
+
+
+// todo verify this will create parents as needed
+void
+KvInterfaceImplMySql::set(std::string const& key, std::string const& value) {
```
andy-slac: _validateKey/_validateValue here?

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,503 @@
...
+        LOGF(_logger, LOG_LVL_ERROR, "exists - %1% failed with err:%2%" % query % errObj.errMsg());
+        throw CssError((boost::format("exists - error:%1% from query:%2%") % errObj.errMsg() % query).str());
+    }
+
+    unsigned int count(0);
+    if (not extractIntValueFromSqlResults(&results, &count)) {
+        throw CssError("failed to extract int value from query");
+    }
+
+    bool foundIt(false);
```
andy-slac: Why `foundIt`? Just return `count == 1`?

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,503 @@
...
+        throw CssError(errObj);
+    }
+
+    transaction.commit();
+}
+
+
+bool
+KvInterfaceImplMySql::exists(std::string const& key) {
+    KvTransaction transaction(_conn);
```
andy-slac: _validateKey()?

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,503 @@
...
+        throw CssError("multiple keys for key");
+    }
+    transaction.commit();
+    return foundIt;
+}
+
+
+std::vector<std::string>
+KvInterfaceImplMySql::getChildren(std::string const& parentKey) {
+
```
andy-slac: _validateKey here or in _getChildrenFullPath

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,503 @@
...
+    transaction.commit();
+}
+
+
+void
+KvInterfaceImplMySql::_delete(std::string const& key) {
+    // A transaction should be in process here.
+
+    // recursively delete child keys first
+    std::vector<string> childKeys = _getChildrenFullPath(key);
```
andy-slac: THere should be more efficient way to delete them than just doing one-by-one recursively, but I'm OK with this for now, we can do optimizations later.

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,503 @@
...
+
+    // trim off the parent key, leaving only the last item in the path.
```
andy-slac: I think this comment comes from some other place :)

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,503 @@
...
+    for (std::vector<std::string>::iterator strItr = childKeys.begin(); strItr != childKeys.end(); ++strItr) {
+        _delete(*strItr);
+    }
+
+    boost::format fmQuery("DELETE FROM kvData WHERE kvKey='%1%'");
+    fmQuery % key;
+    std::string query = fmQuery.str();
```
andy-slac: make sure key is escaped

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,503 @@
...
+    // Note regarding starting a transaction here:
+    // Most private methods (in fact all of them) require that a transaction already be in progress. On the other hand:
+    // this method is protected, and is called by KvInterface with no transaction in progress so here one must be started.
+    KvTransaction transaction(_conn);
+
+    std::string val;
+    sql::SqlErrorObject errObj;
+
+    boost::format fmQuery("SELECT kvVal FROM kvData WHERE kvKey='%1%'");
+    fmQuery % key;
```
andy-slac: escape key

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,503 @@
...
+    std::string query = fmQuery.str();
+
+    sql::SqlResults results;
+    if (not _conn.runQuery(query, results, errObj)) {
+        LOGF(_logger, LOG_LVL_ERROR, "_get - query failed: %1%" % query);
+        if (throwIfKeyNotFound) {
+            throw CssError(errObj);
```
andy-slac: Error here means something more serious than "key not found", I believe you have to always throw here.

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,503 @@
...
+        } else {
+            val = defaultValue;
+        }
+    } else {
+        errObj.reset();
+        if (not results.extractFirstValue(val, errObj)) {
+            if (throwIfKeyNotFound) {
```
andy-slac: Also here error may mean something else besides "result set is empty".

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,503 @@
...
+            }
+        }
+    }
+
+    transaction.commit();
+    return val;
+}
+
+
+bool KvInterfaceImplMySql::_getIdFromServer(std::string const& key, unsigned int* id) {
```
andy-slac: could also return an int here, negative for not found.

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,503 @@
...
+    sql::SqlErrorObject errObj;
+    std::string retStr;
+    if (not _conn.escapeString(key, retStr, errObj)) {
+        throw CssError(errObj);
+    }
+    // verify that:
+    // - key is less than max length
+    // - first character should be a delimiter
+    // - last character should not be a delimiter
+    if (key.size() > KvInterface::MAX_KEY_LENGTH ||
```
andy-slac: I think you want to do checking before escaping it, escaping can make string longer (e.g. replace ü with 4-character sequence \xFC)

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -0,0 +1,503 @@
...
+            key.find_first_of(KEY_PATH_DELIMITER) != 0 ||
+            key.find_last_of(KEY_PATH_DELIMITER) == key.size()-1) {
+        LOGF(_logger, LOG_LVL_DEBUG, "create - rejecting key: %1%" % retStr);
+        throw CssError("invalid key");
+    }
+    return retStr;
+}
+
+
+std::string KvInterfaceImplMySql::_validateValue(std::string const& value) {
```
andy-slac: This does not validate anything, only escapes, I'd rename it to avoid confusion.

admin/templates/configuration/tmp/configure/sql/CssData.sql
andy-slac: It should unless mysql is broken.

core/modules/css/KvInterfaceImplMySql.cc
jbecla: can we make "results" const?

n8pease: SqlResults::extractFirstValue is not const so the short answer is no.
I could look into making the extract... methods in SqlResults const (or the ones that can be), but I had decided not to get too much into that. Should I do it?

core/modules/css/KvInterfaceImplMySql.cc
n8pease: oh, I replaced those with _conn.getInsertId() when I found that method, but forgot to delete these. thanks.

core/modules/css/KvInterfaceImplMySql.cc
andy-slac: I think logically those operations should not be const. Const operations usually do not change state and those operations certainly do. You cannot call `SqlResults::extractFirstValue()` twice, after first call corresponding mysql result disappears.  Better/easier leave it as it is now.

jbecla: I am fine with that

core/modules/css/KvInterfaceImplMySql.cc
n8pease: it's true. however very little information is provided by the sql module about what error occurred, so this does what it can with the info it has. I'll at least add a log line that prints the error message. To handle it programmatically we could set an error number to the errObj, inside SqlResults::extractFirstValue. I don't see error codes defined anywhere inside the sql module, should we add a namespace for that?

core/modules/css/KvInterfaceImplMySql.cc
n8pease: my experience is that overloading the meaning of a return value this way is less elegant than a slightly more complex but well defined api...

core/modules/css/KvInterfaceImplMySql.cc
n8pease: longer, right. but the string has to fit in the database table so checking has to be done after escaping.

core/modules/css/KvInterfaceImplMySql.cc
n8pease: OTOH we might want to add validation to the value, which is why I put it in and named it thus. Otherwise it'd be as easy to just do the escaping inline I think.

core/modules/css/KvInterfaceImplMySql.cc
andy-slac: It may be better to use different method here, e.g. extractFirstColumn(), so you can handle all conditions properly.

core/modules/css/KvInterfaceImplMySql.cc
andy-slac: In database it's still stored as a single byte (or character), it's only textual SQL query representation that needs this escaping.

core/modules/css/KvInterfaceImplMySql.cc
andy-slac: I do not think we'll ever need to validate values, and having `validate()` method which does not validate anything but does escaping is confusing. I'd actually suggest separating validation and escaping, validation should happen on method arguments (unescaped), and escaping should happen at a point when you build SQL query.

core/modules/css/KvInterfaceImplMySql.cc
n8pease: actually - do I even need to check the size, or will the insert just fail if it's too long?

andy-slac: mysql is very ugly in that respect - by default it silently truncates long strings. We should probably use non-default option to force it generate errors in this case. I'd still keep that check.

core/modules/css/CssError.h
```diff
@@ -33,9 +33,12 @@
 #define LSST_QSERV_CSSERROR_H
 
 // System headers
-#include <map>
 #include <stdexcept>
 #include <string>
+#include <boost/lexical_cast.hpp>
```
jbecla: we consider boost headers "third party headers", so add appropriate comment above, and switch to " "

core/modules/css/CssError.h
```diff
@@ -134,6 +149,7 @@ class VersionMismatchError : public CssError {
         : CssError("CSS version number mismatch: expected=" + expected +", actual=" + actual) {}
 };
 
+
```
jbecla: not necessary?

core/modules/sql/SqlConnection.h
```diff
@@ -125,9 +125,19 @@ class SqlConnection {
...
      */
     std::string escapeString(std::string const& rawString) const;
 
+    /**
+     * Escape string for use inside SQL statements.
+     * @return true if the escaped string could be created.
+     * @note this method will attempt to connect if the connection is not already estabilshed.
+     */
+    bool escapeString(std::string const& rawString, std::string& escapedString,
+            SqlErrorObject& errObj);
```
jbecla: alignment

PR #112, opened 2015-07-30
Improve czar-worker communication debugging (DM-3223)

andy-slac:


core/modules/ccontrol/MergingRequester.cc
```diff
@@ -93,12 +94,16 @@ bool MergingRequester::flush(int bLen, bool last) {
     case MsgState::HEADER_SIZE_WAIT:
         _response->headerSize = static_cast<unsigned char>(_buffer[0]);
         if (!proto::ProtoHeaderWrap::unwrap(_response, _buffer)) {
-            _setError(log::MSG_RESULT_DECODE,
-                std::string("Error decoding proto header for ") + getStateStr(_state));
+             std::ostringstream os;
```
andy-slac: Indentation.

core/modules/ccontrol/MergingRequester.cc
```diff
@@ -93,12 +94,16 @@ bool MergingRequester::flush(int bLen, bool last) {
...
+            os << "From:" << _wName << "Error decoding proto header for " << getStateStr(_state);
```
andy-slac: nitpick: this could still be done with just string concatenation which should be somewhat more efficient that creating ostringstream.

core/modules/ccontrol/MergingRequester.cc
```diff
@@ -93,12 +94,16 @@ bool MergingRequester::flush(int bLen, bool last) {
...
+            _setError(log::MSG_RESULT_DECODE, os.str());
             _state = MsgState::HEADER_ERR;
             return false;
         }
-        LOGF_DEBUG("HEADER_SIZE_WAIT: Resizing buffer to %1%" % _response->protoHeader.size());
+        if (_wName.compare("~") == 0) {
```
andy-slac: What is wrong with simple `_wName == "~"`?

core/modules/ccontrol/MergingRequester.cc
```diff
@@ -115,6 +120,9 @@ bool MergingRequester::flush(int bLen, bool last) {
                 LOGF_INFO("Messages continues, waiting for next header.");
                 _state = MsgState::RESULT_EXTRA;
                 _buffer.resize(proto::ProtoHeaderWrap::PROTO_HEADER_SIZE);
+            } else {
+              LOGF_INFO("Messages ends, setting last=true");
```
andy-slac: indentation should be 4 spaces

core/modules/ccontrol/MergingRequester.cc
```diff
@@ -137,22 +145,16 @@ bool MergingRequester::flush(int bLen, bool last) {
...
+        // We shouldn't wind up here. _buffer.size(0) and last=true should end communication.
+        // fall-through
     case MsgState::HEADER_ERR:
     case MsgState::RESULT_ERR:
-        _setError(log::MSG_RESULT_ERROR, "Unexpected message");
+         {
+            std::ostringstream eos;
+            eos << "Unexpected message From:" << _wName << " flush state=" << getStateStr(_state) << " last=" << last;
+            LOGF_ERROR(eos.str().c_str());
+            _setError(log::MSG_RESULT_ERROR, eos.str().c_str());
```
andy-slac: do you need `.c_str()` here?

core/modules/ccontrol/MergingRequester.cc
```diff
@@ -137,22 +145,16 @@ bool MergingRequester::flush(int bLen, bool last) {
...
-        return true;
+        // We shouldn't wind up here. _buffer.size(0) and last=true should end communication.
+        // fall-through
     case MsgState::HEADER_ERR:
     case MsgState::RESULT_ERR:
-        _setError(log::MSG_RESULT_ERROR, "Unexpected message");
+         {
+            std::ostringstream eos;
+            eos << "Unexpected message From:" << _wName << " flush state=" << getStateStr(_state) << " last=" << last;
+            LOGF_ERROR(eos.str().c_str());
```
andy-slac: I'd replace it with `LOGF_ERROR("%1%" % eos.str());` just in case `_wName` can contain percent character in it.

core/modules/ccontrol/MergingRequester.h
```diff
@@ -123,6 +123,7 @@ class MergingRequester : public qdisp::ResponseRequester {
     bool _flushed; ///< flushed to InfileMerger?
     std::mutex _cancelledMutex; ///< Protect check/write of cancel flag.
     bool _cancelled; ///< Cancelled?
+    std::string _wName; // worker name
```
andy-slac: `///<` like all other declarations?

core/modules/qdisp/QueryRequest.cc
```diff
@@ -153,6 +154,13 @@ bool QueryRequest::_importStream() {
     bool retrieveInitiated = false;
     // Pass ResponseRequester's buffer directly.
     std::vector<char>& buffer = _requester->nextBuffer();
+    LOGF_DEBUG("QueryRequest::_importStream buffer.size=%1%" % buffer.size());
+    for(auto iter=buffer.begin(); iter != buffer.end(); iter++) *iter=8;
```
andy-slac: Our guidelines recommend multi-line for:

    for(auto iter=buffer.begin(); iter != buffer.end(); iter++) {
        *iter=8;
    }

Also prefer `++ iter` over `iter ++`.

Is 8 some magical number everyone is supposed to know about?

core/modules/qdisp/QueryRequest.cc
```diff
@@ -153,6 +154,13 @@ bool QueryRequest::_importStream() {
...
+    std::ostringstream dbg;
+    const void* pbuf = (void*)(&buffer[0]);
+    dbg << "_importStream->GetResponseData size=" << buffer.size();
+    dbg << ' ' << pbuf << ' ' <<  util::prettyCharList(buffer, 25);
+    LOGF_INFO(dbg.str().c_str());
```
andy-slac: YOu could probably do all of that without temporary ostringstream:

    LOGF_INFO("_importStream->GetResponseData size=%1% %2% %3%" % 
              buffer.size() % pbuf % util::prettyCharList(buffer, 25));

Otherwise do `LOGF_INFO("%1%" % dbg.str());` as it may have % in the resulting string

core/modules/qdisp/QueryRequest.cc
```diff
@@ -202,8 +210,13 @@ void QueryRequest::ProcessResponseData(char *buff, int blen, bool last) { // Ste
             _status.updateInfo(JobStatus::COMPLETE);
             _finish();
         } else {
-            LOGF_INFO("ProcessResponse waiting for more data");
             std::vector<char>& buffer = _requester->nextBuffer();
+            for(auto iter=buffer.begin(); iter != buffer.end(); iter++) *iter=9;
```
andy-slac: Same comments here and lines below

core/modules/wdb/QueryAction.cc
```diff
@@ -164,6 +167,9 @@ QueryAction::Impl::Impl(QueryActionArg const& a)
       _poisoned(false),
       _sendChannel(a.task->sendChannel),
       _user(a.task->user) {
+    char buf[MAX_NAME_LENGTH];
+    gethostname(buf, MAX_NAME_LENGTH);
+    _hostName = buf;
```
andy-slac: If there is more than one instance of this class I'd avoid calling `gethostname()` repeatedly  with simple caching like:

    namespace {
        std::string const& hostname() {
            static std::string cachedHostname;
            if (cachedHostname.empty()) {
                char buf[64];
                gethostname(buf, sizeof buf);
                cachedHostname = buf;
            }
            return cachedHostname;
        }
    }

core/modules/wdb/QueryAction.cc
```diff
@@ -123,6 +125,7 @@ class QueryAction::Impl {
     wbase::Task::Ptr _task;
     std::shared_ptr<ChunkResourceMgr> _chunkResourceMgr;
     std::string _dbName;
+    std::string _hostName;
```
andy-slac: If you do something like what I suggested below you probably don't need this data member, just use `::hostname()` everywhere.

core/modules/wdb/QueryAction.cc
gpdf: Shouldn't that constant be POSIX's HOST_NAME_MAX?

And note that if the name is truncated there is no guarantee that a trailing null byte will be written to the array.

http://pubs.opengroup.org/onlinepubs/9699919799/functions/gethostname.html

core/modules/ccontrol/MergingRequester.cc
jbecla: there is extra space before std::

core/modules/qdisp/QueryRequest.cc
jgates108: Deleted the line. Just wanted a non-zero value so we could see if buffer was being altered.

core/modules/wdb/QueryAction.cc
jgates108: I added a mutex and wound up with this

std::string const& getHostname() {
    static std::mutex mx;
    static std::string cachedHostname;
    std::lock_guard<std::mutex> lg(mx);
    if (cachedHostname.empty()) {
        char buf[HOST_NAME_MAX+1];
        buf[HOST_NAME_MAX] = '\0';
        gethostname(buf, sizeof buf);
        cachedHostname = buf;
    }
    return cachedHostname;
}

andy-slac: If you need this to be thread-safe then it would be easier to rely on static initialization for that:

    std::string initHostName() {
            char buf[HOST_NAME_MAX+1];
            buf[HOST_NAME_MAX] = '\0';
            gethostname(buf, sizeof buf - 1);
            return std::string(buf);
    }
    std::string const& getHostname() {
        static std::string const cachedHostname = initHostName();
        return cachedHostname;
    }

And to make it super-safe I added ` - 1` to `sizeof buf`.

gpdf: Also one should always check system service return codes, such as from gethostname().  What to do with the knowledge of an error in this case depends on the bigger picture of this class.  With the design as it stands there seems to be little alternative but to throw an exception.

core/modules/wdb/QueryAction.cc
```diff
@@ -63,6 +63,8 @@
 #include "wdb/ChunkResource.h"
 #include "wdb/QueryAction.h"
 
+#define MAX_NAME_LENGTH 50
```
gpdf: I've already commented on using the POSIX constant HOST_NAME_MAX but just for information-transfer note that the actual value of that constant on Linux is typically 64, so this was not only an unnecessary constant definition but a bad value - and a good illustration of the need to follow POSIX more closely.  Please make sure that this idea is scrubbed from the rest of the code base.

PR #114, opened 2015-08-05
Include sciSQL in master node configuration

vaikuntht:
Running mysql tests in multi-node integration tests require adding scisql to the master node configuration.


vaikuntht:
Only `qserv-check-integration.py` does, see comment in other repo regarding this. It can be changed to run just like `qserv-test-integration.py` if we want.

PR #115, opened 2015-08-18
CSS/QMeta interaction in czar (DM-3163)

andy-slac:
Czar is now updating query metadata database when it processes query.
    Some pieces are still missing though - we do not have user information
    for example, and all queries are SYNC currently, so this will need to be
    extended later when we work on those parts. Also missing is worker
    assignments to query chunks, I have not managed to figure out yet which
    part of code has worker info, to be done later as well (and that is only
    relevant to ASYNC queries and only when we need job migration between
    czars, so it can wait).
    
    Czar script now has one additional option which can set czar name, by
    default czar name is built from host name and czar port number. Czar
    configuration file was also extended to include parameters for qmeta
    database connection options (currently the same as czar database).

    Facade::containsTable() method was updated to not only check key
    existence in CSS but also to check key value, table is assumed to exist
    only if key value is "READY", false is returned for all other values.
    
    UserQuery::submit() now checks that tables exist after registering
    registering a query in qmeta database, it table does not exist (e.g.
    table was marked as "to be deleted") query is marked as failed. This
    needed some updates to UserQuery::join() method as well because of
    complicated logic of error/exception handling.
    
    Currently this code does not do much because we are still using CSS
    snapshot which is read-only and table existence is checked even before
    query is submitted.


core/modules/ccontrol/UserQueryFactory.cc
```diff
@@ -43,6 +43,7 @@
 #include "mysql/MySqlConfig.h"
 #include "qdisp/Executive.h"
 #include "qproc/QuerySession.h"
+#include "qmeta/QMetaMysql.h"
```
jbecla: sort alphabetically (move one line up)

core/modules/css/testFacade.cc
```diff
@@ -99,28 +99,28 @@ struct FacadeFixture {
...
 
         p = "/DBS/dbB/TABLES";
         kv.push_back(make_pair(p, ""));
-        kv.push_back(make_pair(p + "/Exposure", ""));
-        kv.push_back(make_pair(p + "/MyObject", ""));
+        kv.push_back(make_pair(p + "/Exposure", "READY"));
+        kv.push_back(make_pair(p + "/MyObject", "READY"));
         kv.push_back(make_pair(p + "/MyObject/partitioning", ""));
         kv.push_back(make_pair(p + "/MyObject/partitioning/lonColName", "ra_PS"));
         kv.push_back(make_pair(p + "/MyObject/partitioning/latColName", "decl_PS"));
```
jbecla: would it make sense to test states other than "READY" here? e.g. maybe some conditions that result in failure?

core/modules/ccontrol/UserQueryFactory.cc
andy-slac: OK

core/modules/css/testFacade.cc
andy-slac: I just wanted for this test to continue working after my change to the implementation of `Facade::_containsTable()` method. We could add separate test for different value of table key, I'd prefer to do it when we work on re-designing `Facade` stuff.

PR #117, opened 2015-08-24
Improve Docker container management

fritzm:
* Create image for unconfigured Qserv stack
* Add script to produce Qserv image for
  worker and master instances

* Switch to debian:jessie as Docker distrib
  - Update Debian:jessie system dependencies

* Build and test host-based source code in a container
  - Add qserv and dev account in Docker image.
  - Add script to change qserv user and
    group id in image


fjammes:
Thanks Fritz, I've taken your remarks in account, refactored Dockerfiles and re-tested everything. Is the review complete?

admin/tools/docker/build-configured-images.sh
```diff
@@ -0,0 +1,94 @@
...
+DEST_IMAGE="$IMAGE-worker"
+set -x
+docker run --cidfile="$CID_FILE" -u qserv "$IMAGE" $SCRIPT
+set +x
+
+CONTAINER_ID=$(cat $CID_FILE)
+rm $CID_FILE
+
+MSG="Create Qserv worker image"
+DEST_IMAGE="$IMAGE-worker"
```
fritzm: Redundant with line 72?

admin/tools/docker/build-configured-images.sh
```diff
@@ -0,0 +1,94 @@
...
+# Create Qserv master image
+DEST_IMAGE="$IMAGE-master"
+docker run --cidfile="$CID_FILE" -u qserv "$IMAGE" $SCRIPT -m
+
+CONTAINER_ID=$(cat $CID_FILE)
+rm $CID_FILE
+
+MSG="Create Qserv master image"
+docker commit --message="$MSG" --author="Fabrice Jammes" "$CONTAINER_ID" "$DEST_IMAGE"
+echo "Image $DEST_IMAGE created successfully"
```
fritzm: Would it make sense to have two additional Dockerfiles for worker and master that "FROM" to build on the base image, then considerably simplify this script?   I think that might be more in line with standard practice.  Docker commit is not frequently used.

doc/source/install/docker.rst
```diff
@@ -1,28 +1,85 @@
...
+Develop and test 
+================
+
+Set uid in qserv:dev w.r.t your host machine user account. This creates a qserv:dev-uid image which can mount host source code in a container with correct permissions. 
+Host might be a development machine or a continuous integration server.
+
+.. code-block:: bash
+
+   ./change-uid.sh
+
```
fritzm: Hmm, okay for now I guess...   Could one just -v in /etc/group, /etc/passwd, and use --user as an alternative for dev instead of spinning a whole add'l container?  User namespacing is supposed to be coming to docker "real soon now" (1.9?)

admin/tools/docker/build-configured-images.sh
```diff
@@ -0,0 +1,94 @@
...
+# Create Qserv worker image
+TMP_DIR=$(mktemp -d -t docker_qserv_uid.XXXXXX)
+CID_FILE="$TMP_DIR"/cid
+
+DEST_IMAGE="$IMAGE-worker"
+set -x
+docker run --cidfile="$CID_FILE" -u qserv "$IMAGE" $SCRIPT
+set +x
+
+CONTAINER_ID=$(cat $CID_FILE)
```
jhoblitt: This is extremely minor but the preferred bash way of expressing this would be as `CONTAINER_ID=$(<$CID_FILE)` to avoid shelling out to run `cat`.

admin/tools/docker/build-configured-images.sh
jhoblitt: I share @fritzm opinion that this would be more 'typical' as two separate `Dockerfiles`.

admin/tools/docker/build-configured-images.sh
```diff
@@ -0,0 +1,94 @@
...
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <http://www.lsstcorp.org/LegalNotices/>.
```
jhoblitt: I believe we aren't required to put the license boilerplate into every file.  @frossie - can you confirm that this can be replaced with a LICENSE file?

admin/tools/docker/change-uid.sh
```diff
@@ -0,0 +1,97 @@
...
+SCRIPT="/home/dev/scripts/change-uid.sh"
+
+_UID=$(id -u)
+GID=$(id -g)
+
+TMP_DIR=$(mktemp -d -t docker_qserv_uid.XXXXXX)
+CID_FILE="$TMP_DIR"/cid
+
+set -x
+docker run --cidfile="$CID_FILE" -u root "$IMAGE" $SCRIPT $_UID $GID
```
jhoblitt: Why isn't this part of the initial container build?  If it needs to be a separate step, I think doing it as `Dockerfile` would cut down on the boilerplate.

admin/tools/docker/dev/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+    chown -R dev:dev /home/dev/scripts && \
+    chmod g+w stack/ && \
+    chmod -R g+w stack/Linux64/mysql
+
+USER dev
+
+WORKDIR /home/dev
+RUN mkdir src
+
+# Source code should be mounted using 'docker run -v'
```
jhoblitt: If the source code needs to be mounted as a volume, why is it then being cloned directly?

admin/tools/docker/dev/scripts/startup.py
```diff
@@ -0,0 +1,15 @@
+hooks.config.Eups.colorize = True
+hooks.config.Eups.userTags += ["git"]
+
+def cmdHook(Eups, cmd, opts, args):
```
jhoblitt: A comment explaining why this callback is needed would be helpful.

admin/tools/docker/latest/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+COPY scripts/install-stack.sh scripts/install-stack.sh
+
+# Hack to enable diagnose of stack install failure
+RUN /bin/bash /qserv/scripts/install-stack.sh || echo 'INSTALL FAILED' > /qserv/STACK_INSTALL_ERROR
+
+COPY scripts/*.sh scripts/
+
+USER root
+
+RUN chown -R qserv:qserv /qserv/scripts/
```
jhoblitt: I'm guessing it is intentional that no ports are `EXPOSE`'d and there isn't a default `CMD`?

admin/tools/docker/useful-cmd.sh
```diff
@@ -1,11 +1,11 @@
 # These commands require Docker on your machine and that you belong to 'docker' group
```
jhoblitt: I believe this file is really documentation -- is there a better place for it to live?

doc/source/install/docker.rst
jhoblitt: I'm fairly sympathetic to not wanting to -v in everything under `/etc/`.  I think the current approach is reasonable until a uid map can be setup.

fritzm: Okay, sounds good!

jhoblitt:
I haven't sat down and run the shell scripts in this PR, so I don't' feel that I have a good mental map of how they all hang together; I can add very little upon @fritzm 's excellent comments.  As a general observation, I suspect some code repetition could be reduced by using series of chained `Dockerfiles in place of some of the shell scripts.  I'm also very much in favor of removing the LICENSE boilerplate from every file.

admin/tools/docker/build-configured-images.sh
fjammes: :+1:

admin/tools/docker/build-configured-images.sh
fjammes: Ok I agree, but in the future I'll need to pass some parameters (like czar fqdn) before creating these images, and Dockerfile doesn't support that. So I propose to let it like that for now.

admin/tools/docker/change-uid.sh
fjammes: Here I can't pass uid/gid as parameters to a `Dockerfile`, that's why I need to use `docker run`

admin/tools/docker/dev/Dockerfile
fjammes: :+1: , removed the cloning thing

admin/tools/docker/dev/scripts/startup.py
fjammes: done

admin/tools/docker/latest/Dockerfile
fjammes: Yes, this will be done in a future ticket, I need to be more experienced about docker/qserv integration to perform fin-tuning.

admin/tools/docker/useful-cmd.sh
fjammes: It's integrated in the sphynx doc, see bottom of `doc/source/install/docker.rst`

fjammes:
Is there a policy about the LICENSE boilerplate? For now it's in each Qserv source file. It would be convenient to have it only once.

admin/tools/docker/dev/Dockerfile
fjammes: I add extensive comment here, indeed I had to workaround a tricky eups(and maybe docker?) problem

timj:
Re: Licensing. See [RFC-45](https://jira.lsstcorp.org/browse/RFC-45).

admin/tools/docker/useful-cmd.sh
jhoblitt: +1 This is a nice solution.

admin/tools/docker/change-uid.sh
jhoblitt: OK - sounds like you want to pickup the uid/gid of the user doing the build dynamically instead of setting them to known value.  I can't think of a solution for that's much cleaner than this, sadly.

admin/tools/docker/useful-cmd.sh
fjammes: Thanks :-)

PR #118, opened 2015-08-26
Tickets/dm 3450

fjammes:


admin/templates/configuration/etc/init.d/qserv-czar
```diff
@@ -46,6 +49,17 @@ export QSW_RESULTDIR=${QSERV_RUN_DIR}/qserv-run/tmp
...
+export XRDSSIDEBUG=1
+
+# Increase timeouts, without that, long-running queries fail.
+# The proper fix is coming, see DM-3433
+export XRD_REQUESTTIMEOUT=64000
+export XRD_STREAMTIMEOUT=64000
+export XRD_DATASERVERTTL=64000
+export XRD_TIMEOUTRESOLUTION=64000
+
+
```
fjammes: We should put these 'configuration' parameters in QSERV_RUN_DIR/etc/sysconfig/xrootd and source it, it would be easier to maintain than in a startup script, which shouldn't contains configuration /param/values. Same think for PYTHONUNBUFFERED.

jbecla: Fabrice. I have nothing against it, but... we are doing plenty of similar things already in init.d/qserv-czar and init.d/zookeeper, and I'd rather not do this sort of surgery right now (partly because I am not entirely clear how to transport things to QSERV_RUN_DIR/etc/). If you can quickly send me some directions I will try to implement it tomorrow (Thu), otherwise we will deal with that in the future. So, should I put it in ./templates/configuration/etc/sysconfig/ and call a "cp" in some other script?

fjammes: Jacek,

I think that environment variables defined in init.d/qserv-czar should be related to generic and low-level configuration (for example log and pid file location), so advanced configuration parameters related to service core shouldn't be here (it shouldn't even be in env variables but in configuration files but we have no choice here).

What you could do is simply move these export directives to the "Environment variables" section of:
admin/templates/configuration/etc/sysconfig/qserv
and comment that it is related to qserv-czar

PR #119, opened 2015-08-26
Added a lock to detect mysql going away or multiple workers (DM-2930)

andy-slac:


core/modules/wdb/ChunkResource.cc
```diff
@@ -196,9 +206,38 @@ class Backend {
         }
         return true;
     }
+
     void discard(ScTableVector const& v) {
         _discard(v.begin(), v.end());
     }
+
+    enum LockStatus {UNLOCKED, LOCKED_OTHER, LOCKED_OURS};
+    std::string strLockStatus(LockStatus ls) {
```
andy-slac: I'd like to advocate again to prefer stream insert operators (`std::ostream& operator<<(std::ostream&, LockStatus)`) over string conversion methods with random names. In that way you don't need to remember the name of the method and can do `out << ls` instead of `out << strLockStatus(ls)`. Boost format is supposed to work correctly with the types which define stream inserters.

core/modules/wdb/ChunkResource.cc
```diff
@@ -235,8 +279,112 @@ class Backend {
...
+    LockStatus _memLockStatus() {
+        std::string sql = "SELECT uid FROM " + _lockDbTbl + " WHERE keyId = 1;";
+        sql::SqlResults results;
+        sql::SqlErrorObject err;
+        if (!_sqlConn.runQuery(sql, results, err)) {
+            // Assuming UNLOCKED should be safe as either it must be LOCKED_OURS to continue
+            // or we are about to try to lock. Failure to lock will cause the program to exit.
+            std::ostringstream emsg;
+            emsg << "memLockStatus query failed, assuming UNLOCKED. " << sql << " err=" << err.printErrMsg();
+            LOGF_WARN("%1%" % emsg.str());
```
andy-slac: Replace three lines above with:

    LOGF_WARN("memLockStatus query failed, assuming UNLOCKED. %1% err=%2%" % sql % err.printErrMsg());

core/modules/wdb/ChunkResource.cc
```diff
@@ -235,8 +279,112 @@ class Backend {
...
+        sql::SqlErrorObject err;
+        if(!_sqlConn.runQuery(query, err)) {
+            LOGF_ERROR("Lock failed, exiting. query=%1%  err=%2%" % query % err.printErrMsg());
+            exit(EXIT_FAILURE);
+        }
+    }
+
+    /// Return the status of the lock on the in memory tables.
+    LockStatus _memLockStatus() {
+        std::string sql = "SELECT uid FROM " + _lockDbTbl + " WHERE keyId = 1;";
```
andy-slac: Semicolon is not needed at the end of query if there is only one query.

core/modules/wdb/ChunkResource.cc
```diff
@@ -235,8 +279,112 @@ class Backend {
...
+            std::ostringstream emsg;
+            emsg << "memLockStatus query failed, assuming UNLOCKED. " << sql << " err=" << err.printErrMsg();
+            LOGF_WARN("%1%" % emsg.str());
+            return UNLOCKED;
+        }
+        std::string uidStr;
+        if (!results.extractFirstValue(uidStr, err)) {
+            std::ostringstream emsg;
+            emsg << "memLockStatus unexpected results, assuming LOCKED_OTHER. err=" << err.printErrMsg();
+            LOGF_WARN("%1%" % emsg.str());
```
andy-slac: Same here, no need for temporary stream.

core/modules/wdb/ChunkResource.cc
```diff
@@ -235,8 +279,112 @@ class Backend {
...
+            return LOCKED_OTHER;
+        }
+        int uid = atoi(uidStr.c_str());
+        if (uid != _uid) {
+            std::ostringstream emsg;
+            emsg << "memLockStatus LOCKED_OTHER wrong uid. Expected " << _uid << " got " << uid
+                << " err=" << err.printErrMsg();
+            LOGF_WARN("%1%" % emsg.str());
```
andy-slac: Same thing here.

core/modules/wdb/ChunkResource.cc
```diff
@@ -235,8 +279,112 @@ class Backend {
...
+        _lockDb = MEMLOCKDB;
+        _lockTbl = MEMLOCKTBL;
+        _lockDbTbl = _lockDb + "." + _lockTbl;
+        LockStatus mls = _memLockStatus();
+        if (mls != UNLOCKED) {
+            LOGF_WARN("Memory tables were not released cleanly! LockStatus=%1%" % strLockStatus(mls));
+        }
+
+        // Lock the memory tables.
+        std::string sql = "CREATE DATABASE IF NOT EXISTS " + _lockDb + ";";
```
andy-slac: Extra semicolon, not needed

core/modules/wdb/ChunkResource.cc
jgates108: I like streams in error messages as I have more faith in the compiler to catch errors during compilation than during run-time, and exercising this code before it is needed is not simple.

core/modules/wdb/ChunkResource.cc
```diff
@@ -235,8 +279,112 @@ class Backend {
...
+        sql = "TRUNCATE TABLE " + _lockDbTbl;
+        _execLockSql(sql);
+        std::ostringstream insert;
+        insert << "INSERT INTO " << _lockDbTbl << " (keyId, uid) VALUES(1, " << _uid << " );";
+        _execLockSql(insert.str());
+
+        // Delete any old in memory databases. They could be empty or wrong.
+        // Empty tables would prevent new tables from being created.
+        std::ostringstream os;
+        os << "SHOW DATABASES LIKE '" << SUBCHUNKDB_PREFIX << "%';";
```
andy-slac: You can do the same with simple string concatenation, it's much more efficient than creating temporary stream.

core/modules/wdb/ChunkResource.cc
```diff
@@ -235,8 +279,112 @@ class Backend {
...
+        sql::SqlResults results;
+        sql::SqlErrorObject err;
+        if (!_sqlConn.runQuery(os.str(), results, err)) {
+            std::ostringstream emsg;
+            emsg << "Memory database query failed, exiting. " << os.str() << " err=" << err.printErrMsg();
+            LOGF_ERROR("%1%" % emsg.str());
```
andy-slac: One more unnecessary tmp stream

core/modules/wdb/ChunkResource.cc
andy-slac: `SUBCHUNKDB_PREFIX` contains underscore which is a wildcard character in mysql LIKE. IT can potentially match more than what you need (e.g. in addition to "Subchunks_123" it will also match "SubchunksDoNotDelete"). I'd add additional check in the for loop below to verify that name actually starts with `SUBCHUNKDB_PREFIX`.

core/modules/wdb/ChunkResource.cc
```diff
@@ -167,6 +171,11 @@ typedef std::vector<ScTable> ScTableVector;
 
 class Backend {
 public:
+    virtual ~Backend() {
+        if (!_isFake) {
+            _memLockRelease();
```
andy-slac: My guess is that release should not be called when it is detected that some other process has this lock. I think `memLockCheck()` calls `exit()` when it detects this situation. Do we have guarantee that Backend's destructor has no chance of being called when exit() happens? I'd probably add some protection in case destructor is called but we detected that lock is not ours.

core/modules/wdb/ChunkResource.cc
smonkewitz: LOGF_WARN is using boost::format under the hood, which provides the same level of compile time safety as streams (as far as I know, boost::format is in fact implemented on top of streams), so I don't understand how doing what Andy suggests would move any validity checking from compile time to run time. Can you elaborate?

For what it's worth, I also prefer the concision of Andy's snippet above, and favor the removal of temporary streams elsewhere.

andy-slac: We just talked with John about it, I think his argument has some reason. Basically boost format does not protect you at compile time from things like `"%1% %3%" % x % y` (providing inconsistent format string) and instead it causes run-time crash which is hard to unit-test.

smonkewitz: Ah I see - that makes sense. This makes me wish that log provided macros that worked with streams instead of boost::format. (I know, I know - too many macros already).

core/modules/wdb/ChunkResource.cc
smonkewitz: Sometimes it's nice to have both (so that you don't have to explicitly create temporary streams if you just want a string or figure out how to convert things to a string if you just have a stream), with one implemented in terms of the other. I'd be happier if this function used a standard name (for us) like `toString`, and I don't understand why it is a non-static member of Backend.

core/modules/wdb/ChunkResource.cc
```diff
@@ -235,8 +279,112 @@ class Backend {
         }
     }
 
+    /// Run the 'query'. If it fails, terminate the program.
+    void _execLockSql(std::string const& query) {
+        LOGF_DEBUG("execLockSql %1%" % query);
+        sql::SqlErrorObject err;
+        if(!_sqlConn.runQuery(query, err)) {
+            LOGF_ERROR("Lock failed, exiting. query=%1%  err=%2%" % query % err.printErrMsg());
```
smonkewitz: If you are going with temp streams, why not use one here?

core/modules/wdb/ChunkResource.cc
```diff
@@ -196,9 +206,38 @@ class Backend {
...
+            return true;
+        }
+        LockStatus mls = _memLockStatus();
+        if (mls != LOCKED_OURS) {
+            LOGF_ERROR("memLockCheck could not verify this program owned the memory table lock, Exiting.");
+            exit(EXIT_FAILURE);
+            return false;
+        }
+        return true;
+    }
```
smonkewitz: I don't think this function's return value is useful, and the name could be clearer. Would something like:
```
void requireMemLockOwnership() {
    if (!_isFake && _memLockStatus() != LOCKED_OURS) {
        LOGF_ERROR(...)
        exit(...);
    }
}
```
be better?

core/modules/wdb/ChunkResource.cc
smonkewitz: `keyId` doesn't seem to be needed or used. I'd suggest dropping it.

core/modules/wdb/ChunkResource.cc
```diff
@@ -235,8 +279,112 @@ class Backend {
...
+        _execLockSql(sql);
+    }
+
     bool _isFake;
     sql::SqlConnection _sqlConn;
+
+    // Memory lock table members.
+    std::string _lockDb;
+    std::string _lockTbl;
+    std::string _lockDbTbl;
```
smonkewitz: Can't you get rid of all the `_lockXXX` member fields and just use the constants you've defined elsewhere?

core/modules/wdb/ChunkResource.cc
```diff
@@ -235,8 +279,112 @@ class Backend {
...
+            LOGF_WARN("%1%" % emsg.str());
+            return LOCKED_OTHER;
+        }
+        return LOCKED_OURS;
+    }
+
+    /// Take the lock on the in memory tables, even if held by another.
+    // Terminate the program if we can't take the lock.
+    // This must be run before any other operations on in memory tables.
+    void _memLockLock() {
```
smonkewitz: I think `_memLockAcquire` would be a better name for this.

More importantly, I think these semantics are problematic, essentially because what you've implemented isn't a lock. You take ownership of a mysql instance immediately rather than waiting for the owner to release it, and while you do make sure that a previous worker commits suicide as soon as it notices a loss of ownership, it will not realize it immediately. Therefore it seems to me that there is some window where two workers could conceivably operate on each-others state and thus potentially return corrupted results.

To me, it makes more sense for a worker to kill itself when it notices that the mysql instance it is trying to acquire has already been claimed. This is slightly more inconvenient from an administrative standpoint - you have to take some action if a worker dies without having released its mysql instance, but I think it's a safer design.

It also seems to give slightly simpler code - acquiring a mysql instance could just be:
```
CREATE DATABASE IF NOT EXISTS q_memoryLockDb;
CREATE TABLE IF NOT EXISTS q_memoryLockDb.memoryLockTbl (whatever INT UNIQUE) ENGINE=MEMORY;
INSERT INTO q_memoryLockDb.memoryLockTbl VALUE (0);
```
(where you exit if the insert fails because someone else has already inserted 0), while explicitly releasing it would have the same effect as restarting mysql:
```
TRUNCATE TABLE q_memoryLockDb.memoryLockTbl
```
What do you think?

smonkewitz: Actually, maybe it would be better to just get rid of the explicit release code entirely (sidestepping Andy's concern about the Backend destructor) and note that if the worker exits, you must also restart mysqld.

core/modules/wdb/ChunkResource.cc
jgates108: I generally believe that macros are evil, but I'd support adding one for logging streams.

core/modules/wdb/ChunkResource.cc
jgates108: I think that requiring a new copy of mysql to be started if the worker dies is reasonable, probably even a good idea. I do, however, like being explicit about what worker owns the entry in the table. 

Getting rid of "TRUNCATE TABLE ..." will cause the worker to fail to acquire the lock, and everything else should work as normal. If restarting mysql causes issues, we can just add those 2 line back and the new worker will always win again.

smonkewitz: I'm OK with storing a pid in the table for informational purposes.

PR #124, opened 2015-09-01
DM-3584: Port Qserv to OS/x

timj:
These patches allow Qserv to be built on OS X with clang.

### DIFF
--- a/core/modules/mysql/MySqlConnection.cc
+++ b/core/modules/mysql/MySqlConnection.cc
@@ -29,6 +29,7 @@
 #include "mysql/MySqlConnection.h"
 
 // Third-party headers
+#include "boost/format.hpp"
 #include "boost/lexical_cast.hpp"
 #include "boost/thread/tss.hpp"
 
@@ -144,7 +145,7 @@ MySqlConnection::cancel() {
     }
     // KILL QUERY only, not KILL CONNECTION.
     int threadId = mysql_thread_id(_mysql);
-    std::string killSql = "KILL QUERY " + boost::lexical_cast<int>(threadId);
+    std::string killSql = boost::str(boost::format("KILL QUERY ") % threadId);
     rc = mysql_real_query(killMysql, killSql.c_str(), killSql.size());
     mysql_close(killMysql);
     if(rc) {
--- a/core/modules/proto/ProtoHeaderWrap.cc
+++ b/core/modules/proto/ProtoHeaderWrap.cc
@@ -43,7 +43,7 @@ const size_t ProtoHeaderWrap::PROTOBUFFER_DESIRED_LIMIT = 2000000;
 const size_t ProtoHeaderWrap::PROTOBUFFER_HARD_LIMIT = 64000000;
 
 std::string ProtoHeaderWrap::wrap(std::string& protoHeaderString) {
-    unsigned char phSize = static_cast<unsigned char>(protoHeaderString.size());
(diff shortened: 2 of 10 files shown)


core/modules/qana/QueryMapping.h
```diff
@@ -29,6 +29,7 @@
   */
 
 // System headers
+#include <string>
```
fritzm: qserv convention is to keep these alphabetized per category

core/modules/query/Constraint.cc
```diff
@@ -33,6 +33,7 @@
 
 // System headers
 #include <iterator>
+#include <iostream>
```
fritzm: alphabetize

core/modules/query/Constraint.h
```diff
@@ -33,6 +33,7 @@
 // System headers
 #include <string>
 #include <vector>
+#include <iosfwd>
```
fritzm: alphabetize

core/modules/tests/testIter.cc
```diff
@@ -25,6 +25,7 @@
 #include <cstring>
 #include <map>
 #include <string>
+#include <fcntl.h>
```
fritzm: alphabetize

core/modules/xrdsvc/ChannelStream.h
```diff
@@ -24,6 +24,7 @@
 #define LSST_QSERV_XRDSVC_CHANNELSTREAM_H
 
 // System headers
+#include <string>
```
fritzm: alphabetize

core/modules/proto/ProtoHeaderWrap.cc
```diff
@@ -43,7 +43,7 @@ const size_t ProtoHeaderWrap::PROTOBUFFER_DESIRED_LIMIT = 2000000;
 const size_t ProtoHeaderWrap::PROTOBUFFER_HARD_LIMIT = 64000000;
 
 std::string ProtoHeaderWrap::wrap(std::string& protoHeaderString) {
-    unsigned char phSize = static_cast<unsigned char>(protoHeaderString.size());
+    char phSize = static_cast<char>(protoHeaderString.size());
```
fritzm: hmm, that's odd -- what's the actual text of the error message?  I would think an explicit static cast to unsigned char should be enough here...

core/modules/mysql/MySqlConnection.cc
```diff
@@ -31,6 +31,7 @@
 // Third-party headers
 #include "boost/lexical_cast.hpp"
 #include "boost/thread/tss.hpp"
+#include "boost/format.hpp"
```
fritzm: alphabetize

core/modules/mysql/MySqlConnection.cc
```diff
@@ -144,7 +145,7 @@ MySqlConnection::cancel() {
     }
     // KILL QUERY only, not KILL CONNECTION.
     int threadId = mysql_thread_id(_mysql);
-    std::string killSql = "KILL QUERY " + boost::lexical_cast<int>(threadId);
+    std::string killSql = (boost::format("KILL QUERY %d") % threadId).str();
```
fritzm: I think the free str operator with boost::format is slight more idiomatic:

```c++
std::string killSql = boost::str(boost::format("KILL QUERY ") % threadId);
```

PR #126, opened 2015-09-03
DM-3619 Move env variables related to xrootd/czar and unlimit in etc/sysconfig/qserv

jbecla:


fritzm:
Merged Sep 8, 2015 -- SHA 45a5c00

PR #128, opened 2015-09-05
Disable duplication of sources in variant_dir (DM-3707).

andy-slac:
With all other commits on this branch this change will support building
with or without duplication of sources. Duplication is turned off for
now.


fritzm:
Wow, that was non-trivial, thanks!  Looks good to me.

PR #129, opened 2015-09-09
Fixes for DM-3779: clean up gcc and eclipse code analyzer warns

fritzm:


core/modules/qana/CheckAggregation.h
```diff
@@ -63,7 +63,7 @@ class CheckAggregation {
      * @return void
      */
     inline void operator()(query::ValueExpr::FactorOp const& fo) {
-        if (!fo.factor.get());
+        if (!fo.factor.get()) {}
```
andy-slac: Does this do anything at all, my impression is that it can be removed entirely?

core/modules/qproc/testQueryAna.h
```diff
@@ -108,6 +104,8 @@ std::string buildFirstParallelQuery(QuerySession& qs, bool withSubChunks=true) {
     return first.queries[0];
 }
 
+} // anonymous namespace
```
andy-slac: I'm not sure about utility of the anonymous namespaces in header files, if it needs to be "hidden" then I think our standard practice is to put it inside `detail` namespace.

core/modules/wbase/Task.h
```diff
@@ -75,7 +75,7 @@ struct Task {
         bool operator()(Ptr const& x, Ptr const& y);
     };
 
-    explicit Task() : _poisoned{false} {}
+    explicit Task() : entryTime(0), _poisoned{false} {}
```
andy-slac: We should use consistent initialization syntax (braces vs parens) at least in the scope of single constructor.

core/modules/qana/CheckAggregation.h
fritzm: Good point -- will remove.

core/modules/qproc/testQueryAna.h
fritzm: I agree.  It was already there so I changed as little as possible.  I'll take a look at doing something better here.

core/modules/wbase/Task.h
fritzm: Good catch -- will fix.

core/modules/qana/CheckAggregation.h
timj: It was added by @wangd in commit c6dcfc0bba09c109bf4af09fb305f3e9158fd309 but there wasn't an explanation as to what it was doing. I hope it wasn't relying on a side-effect of `get()`.

fritzm: I checked -- it is just a member accessor.

core/modules/qproc/testQueryAna.h
fritzm: So, this .h file is simply included into a bunch of unit tests which are linked into separate executables.  Probably just losing the namespace entirely here is reasonable.

The other alternative would be to separate the implementations of these helpers from their decls and put the into a separate .cc which is #included into each of the tests (not sure it is worth it).

PR #130, opened 2015-09-14
Tickets/dm 3562

jgates108:


core/modules/ccontrol/MergingRequester.cc
```diff
@@ -62,12 +63,15 @@ MergingRequester::MergingRequester(
       _tableName(tableName),
```
andy-slac: Can you change this line to use curly braces for consistency

core/modules/ccontrol/UserQuery.cc
```diff
@@ -195,18 +193,12 @@ void UserQuery::submit() {
...
-        cmr = ChunkMsgReceiver::newInstance(cs.chunkId, _messageStore);
-        std::shared_ptr<MergingRequester> mr
-            = std::make_shared<MergingRequester>(cmr, _infileMerger,
-                                                 chunkResultName);
-        qdisp::Executive::JobDescription jobDesc = {ru, ss.str(), mr};
-
         int refNum = ++_sequence;
-        _executive->add(refNum, jobDesc);
-        ss.str(""); // reset stream
+        qdisp::JobDescription jobDesc(refNum, ru, ss.str(), std::make_shared<MergingHandler>(cmr, _infileMerger, chunkResultName));
```
andy-slac: Line is too long.

core/modules/qdisp/Executive.cc
```diff
@@ -150,61 +119,52 @@ Executive::Executive(Config::Ptr c, std::shared_ptr<MessageStore> ms)
...
+    LOGF_DEBUG("Executive::add(%1%)" % jobDesc.toString());
+    if(_cancelled.get()) {
+        LOGF(getLogger(), LOG_LVL_INFO, "Executive already cancelled, ignoring add(%1%)" % jobDesc.id());
         return;
     }
-    bool trackOk = _track(jobId, jobDescription.requester); // Remember so we can join
-    if(!trackOk) {
-        LOGF(getLogger(), LOG_LVL_WARN, "Ignoring duplicate add(%1%)" % jobId);
+    // Create the JobQuery and put it in the map.
+    JobStatus::Ptr jobStatus = std::make_shared<JobStatus>();
```
andy-slac: `auto` here should be quite readable, and line below

core/modules/qdisp/Executive.cc
```diff
@@ -150,61 +119,52 @@ Executive::Executive(Config::Ptr c, std::shared_ptr<MessageStore> ms)
...
+    MarkCompleteFunc::Ptr mcf = std::make_shared<MarkCompleteFunc>(this, jobDesc.id());
+    JobQuery::Ptr jobQuery(new JobQuery(this, jobDesc, jobStatus, mcf));
```
andy-slac: `make_shared` instead of explicit `new`?

core/modules/qdisp/Executive.cc
```diff
@@ -150,61 +119,52 @@ Executive::Executive(Config::Ptr c, std::shared_ptr<MessageStore> ms)
...
+ * Return true if it was successfully added to the map.
+ */
+bool Executive::_addJobToMap(JobQuery::Ptr job) {
+    int jobId = job->getId();
+    if (_jobMap.find(jobId) != _jobMap.end()) {
+        return false;
+     }
+     _jobMap[jobId] = job;
+     return true;
 }
```
andy-slac: This whole function is equivalent to:

    return _jobMap.insert(job->getId(), job).second;

(except that it is less efficient as it needs 2 lookups).

core/modules/qdisp/Executive.cc
```diff
@@ -213,17 +173,18 @@ bool Executive::join() {
     _waitAllUntilEmpty();
     // Okay to merge. probably not the Executive's responsibility
     struct successF {
-        static bool f(Executive::JobStatusPtrMap::value_type const& entry) {
-            JobStatus::Info const& esI = entry.second->getInfo();
+        static bool f(Executive::JobMap::value_type const& entry) {
+            JobQuery::Ptr job = entry.second;
+            JobStatus::Info const& esI = job->getStatus()->getInfo();
             LOGF_INFO("entry state:%1% %2%)" % (void*)entry.second.get() % esI);
```
andy-slac: `job.get()` instead of `entry.second.get()`?

core/modules/qdisp/Executive.cc
```diff
@@ -238,27 +199,27 @@ bool Executive::join() {
...
+    ResponseHandler::Error err;
     LOGF(getLogger(), LOG_LVL_INFO, "Executive::markCompleted(%1%,%2%)" % jobId % success);
     if(!success) {
         {
-            std::lock_guard<std::mutex> lock(_requestersMutex);
-            RequesterMap::iterator i = _requesters.find(jobId);
-            if(i != _requesters.end()) {
-                err = i->second->getError();
+            std::lock_guard<std::mutex> lock(_incompleteJobsMutex);
+            auto i = _incompleteJobs.find(jobId);
```
andy-slac: Compiler can sure tell what is the type of `i` here but I find it harder to understand after this change. Renaming variable to `iter` can help me to guess what to expect.

core/modules/qdisp/Executive.cc
```diff
@@ -275,80 +236,56 @@ void Executive::markCompleted(int jobId, bool success) {
...
-                toSquash = i->second; // Remember which one to squash
-            } else {
-                needToWarn = true;
-            }
-        } else {
-            throw Bug("Executive::requestSquash() with invalid jobId");
+        std::lock_guard<std::mutex> lock(_jobsMutex);
+        auto iter = _jobMap.find(jobId);
+        if(iter == _jobMap.end()) {
+            LOGF_WARN("requesSquash invalid jobID %1%" % jobId);
```
andy-slac: requestSquash

core/modules/qdisp/Executive.cc
```diff
@@ -150,61 +119,52 @@ Executive::Executive(Config::Ptr c, std::shared_ptr<MessageStore> ms)
...
-    bool trackOk = _track(jobId, jobDescription.requester); // Remember so we can join
-    if(!trackOk) {
-        LOGF(getLogger(), LOG_LVL_WARN, "Ignoring duplicate add(%1%)" % jobId);
+    // Create the JobQuery and put it in the map.
+    JobStatus::Ptr jobStatus = std::make_shared<JobStatus>();
+    MarkCompleteFunc::Ptr mcf = std::make_shared<MarkCompleteFunc>(this, jobDesc.id());
+    JobQuery::Ptr jobQuery(new JobQuery(this, jobDesc, jobStatus, mcf));
+    jobQuery->setupWeakThis(jobQuery);
+    if(!_addJobToMap(jobQuery)) {
+        LOGF_ERROR("Executive ignoring duplicate job add(%1%)" % jobQuery->getId());
```
andy-slac: Why do you prefer to log to global/root logger instead of the logger returned by `getLogger()` defined here? We should try to keep it consistent in one file.

core/modules/qdisp/Executive.cc
```diff
@@ -275,80 +236,56 @@ void Executive::markCompleted(int jobId, bool success) {
...
+    LOGF_INFO("Enqueued jobs for cancelling...done");
     {
-        std::vector<RequesterPtr>::iterator i, e;
-        for(i=pendingRequesters.begin(), e=pendingRequesters.end(); i != e; ++i) {
+        for(auto i=jobsToSquash.begin(), e=jobsToSquash.end(); i != e; ++i) {
             // Could get stuck because it waits on xrootd,
             // which may be waiting on a thread blocked in _unTrack().
-            // Don't do this while holding _requestersMutex
-            (**i).cancel();
+            // Don't do this while holding _statusesMutex
```
andy-slac: Is it _jobsMutex, I can't find _statusesMutex?

core/modules/qdisp/Executive.cc
```diff
@@ -275,80 +236,56 @@ void Executive::markCompleted(int jobId, bool success) {
...
+            jobsToSquash.push_back(i->second);
         }
     }
-    LOGF_INFO("%1%\nLoop cancel all queries..." % os.str());
-    LOGF_INFO("Enqueued requesters for cancelling...done");
+    LOGF_INFO("Enqueued jobs for cancelling...done");
     {
-        std::vector<RequesterPtr>::iterator i, e;
-        for(i=pendingRequesters.begin(), e=pendingRequesters.end(); i != e; ++i) {
+        for(auto i=jobsToSquash.begin(), e=jobsToSquash.end(); i != e; ++i) {
```
andy-slac: we have c++11 now, can do:

    for (const auto& jobPtr: jobsToSquash) {

core/modules/qdisp/Executive.cc
```diff
@@ -462,12 +353,12 @@ bool Executive::_track(int jobId, RequesterPtr r) {
 void Executive::_unTrack(int jobId) {
     bool untracked = false;
     {
-        std::lock_guard<std::mutex> lock(_requestersMutex);
-        RequesterMap::iterator i = _requesters.find(jobId);
-        if(i != _requesters.end()) {
-            _requesters.erase(i);
+        std::lock_guard<std::mutex> lock(_incompleteJobsMutex);
+        auto i = _incompleteJobs.find(jobId);
```
andy-slac: i -> iter

core/modules/qdisp/Executive.cc
```diff
@@ -475,41 +366,43 @@ void Executive::_unTrack(int jobId) {
...
-        }
-        if(!reaped) {
-            break;
+    for(auto i=_incompleteJobs.begin(), e=_incompleteJobs.end(); i != e;) {
+        if(!i->second->getDescription().respHandler()->getError().isNone()) {
+            // Requester should have logged the error to the messageStore
+            LOGF_INFO("Executive (%1%) reaped requester for jobId=%2%" % (void*)this % i->first);
+            auto eraseMe = i;
+            ++i;
+            _incompleteJobs.erase(eraseMe);
```
andy-slac: These three lines can be replaced with:

    i = _incompleteJobs.erase(i);

core/modules/qdisp/Executive.cc
```diff
@@ -475,41 +366,43 @@ void Executive::_unTrack(int jobId) {
...
+ *
+ * @see python module lsst.qserv.czar.proxy.unlock()
+ */
 void Executive::_updateProxyMessages() {
     {
-        std::lock_guard<std::mutex> lock(_statusesMutex);
-        for(auto i=_statuses.begin(), e=_statuses.end(); i != e; ++i) {
-            auto const& info = i->second->getInfo();
+        std::lock_guard<std::mutex> lock(_jobsMutex);
+        for(auto i=_jobMap.begin(), e=_jobMap.end(); i != e; ++i) {
```
andy-slac: Use range-for?

core/modules/qdisp/Executive.cc
```diff
@@ -551,21 +444,41 @@ void Executive::_waitAllUntilEmpty() {
...
+        _allJobsComplete.wait_for(lock, statePrintDelay);
     }
 }
 
-std::ostream& operator<<(std::ostream& os,
-                         Executive::JobStatusPtrMap::value_type const& v) {
-    os << v.first << ": " << *(v.second);
+std::ostream& operator<<(std::ostream& os, Executive::JobMap::value_type const& v) {
+    JobStatus::Ptr status = v.second->getStatus();
+    os << v.first << ": " << status;
```
andy-slac: `*status`?

core/modules/qdisp/Executive.cc
```diff
@@ -551,21 +444,41 @@ void Executive::_waitAllUntilEmpty() {
...
+std::string JobDescription::toString() const {
+    std::ostringstream os;
+    os << "job(id=" << _id << " payload=" << _payload << ")";
+    return os.str();
+}
+
+std::ostream& operator<<(std::ostream& os, JobDescription const& jd) {
+    os << jd.toString();
+    return os;
+}
```
andy-slac: I'd do it other way round to avoid temporary stream creation in toString().

core/modules/qdisp/Executive.h
```diff
@@ -44,16 +46,39 @@ class XrdSsiService;
...
 class MessageStore;
 class QueryResource;
-class ResponseRequester;
+
+/** Description of a job managed by the executive
+ */
+class JobDescription {
+public:
+    JobDescription(int id, ResourceUnit resource, std::string payload,
+        std::shared_ptr<ResponseHandler> respHandler)
```
andy-slac: `resource`, `payload` and `respHandler` can be passed as const ref for efficiency

core/modules/qdisp/Executive.h
```diff
@@ -107,44 +121,22 @@ class Executive {
...
-     *
-     *  @param jobId id of the job related to current chunk query
-     *  @param r pointer to requester which will store chunk query result
-     *
-     *  @return true if (jobId,r) was added to _requesters
-     *          false if this entry was previously in the map
-     */
-    bool _track(int refNum, RequesterPtr r);
+
+    bool _track(int refNum, std::shared_ptr<JobQuery> r);
```
andy-slac: pass shared pointer by const ref?

core/modules/qdisp/Executive.h
```diff
@@ -107,44 +121,22 @@ class Executive {
...
     void _unTrack(int refNum);
+    bool _addJobToMap(std::shared_ptr<JobQuery> job);
```
andy-slac: const ref?

core/modules/qdisp/Executive.h
```diff
@@ -155,31 +147,46 @@ class Executive {
...
-}; // class Executive
+};
+
+class MarkCompleteFunc {
+public:
+    typedef std::shared_ptr<MarkCompleteFunc> Ptr;
+
+    MarkCompleteFunc(Executive* e, int jobId) : _executive(e), _jobId(jobId) {}
+
+    virtual void operator()(bool success) {
```
andy-slac: Is it polymorphic, do you need `virtual` here? If yes then destructor should be virtual too.

core/modules/qdisp/JobQuery.cc
```diff
@@ -0,0 +1,124 @@
...
+// Qserv headers
+#include "qdisp/Executive.h"
+#include "qdisp/QueryRequest.h"
+#include "qdisp/QueryResource.h"
+
+namespace lsst {
+namespace qserv {
+namespace qdisp {
+
+void logErr(std::string const& msg, JobQuery* jq) {
```
andy-slac: Put it inside anonymous namespace

core/modules/qdisp/JobQuery.cc
```diff
@@ -0,0 +1,124 @@
...
+    bool cancelled = _executive->getCancelled();
+    bool handlerReset = _jobDescription.respHandler()->reset();
+    if (!cancelled && handlerReset) {
+        Ptr thisPtr = _weakThis.lock();
+        if (!thisPtr) {
+            logErr("runJob() _weakThis was null!", this);
+            return false;
+        }
+        {
+            QueryResource::Ptr qr = std::make_shared<QueryResource>(thisPtr);
```
andy-slac: `auto` should be fine here, type is clear from RHS

core/modules/qdisp/JobQuery.cc
```diff
@@ -0,0 +1,124 @@
...
+    LOGF_DEBUG("runJob %1%" % toString());
+    if (!_executive) {
+        logErr("runJob failed _executive=NULL", this);
+        return false;
+    }
+    bool cancelled = _executive->getCancelled();
+    bool handlerReset = _jobDescription.respHandler()->reset();
+    if (!cancelled && handlerReset) {
+        Ptr thisPtr = _weakThis.lock();
+        if (!thisPtr) {
```
andy-slac: I do not understand how this could happen. Isn't it equivalent to `this == nullptr`?

If you need to make shared pointer from `this` then standard library has support for it already, `std::enable_shared_from_this<T>`

core/modules/qdisp/JobQuery.cc
```diff
@@ -0,0 +1,124 @@
...
+            logErr("runJob() _weakThis was null!", this);
+            return false;
+        }
+        {
+            QueryResource::Ptr qr = std::make_shared<QueryResource>(thisPtr);
+            std::lock_guard<std::recursive_mutex> lock(_rmutex);
+            if (shouldRetry()) {
+                ++_attemptsToRun;
+                if (_executive == NULL) {
```
andy-slac: `nullptr`

core/modules/qdisp/JobQuery.h
```diff
@@ -0,0 +1,122 @@
...
+
+    int getId() const { return _jobDescription.id(); }
+    JobDescription& getDescription() { return _jobDescription; }
+    JobStatus::Ptr getStatus() { return _jobStatus; }
+
+    void provisioningFailed(std::string const& msg, int code);
+    std::shared_ptr<QueryResource> getQueryResource() { return _queryResourcePtr; }
+    void freeQueryResource() { _queryResourcePtr.reset(); }
+
+    void setQueryRequest(std::shared_ptr<QueryRequest> qr) { _queryRequestPtr = qr; }
```
andy-slac: by const ref

core/modules/qdisp/JobQuery.h
```diff
@@ -0,0 +1,122 @@
...
+    // Values that need mutex protection
+    mutable std::recursive_mutex _rmutex;
+    int _attemptsToRun;
+
+    // xrootd items
+    std::shared_ptr<QueryResource> _queryResourcePtr;
+    std::shared_ptr<QueryRequest> _queryRequestPtr;
+
+    // Cancellation
+    util::Flag<bool> _cancelled;
```
andy-slac: Need header for this type

core/modules/qdisp/JobQuery.h
```diff
@@ -0,0 +1,122 @@
+// -*- LSST-C++ -*-
```
andy-slac: I do not think we are still required to bow to emacs-super-gods :)

core/modules/qdisp/QueryRequest.cc
```diff
@@ -59,71 +59,46 @@ inline void unprovisionSession(XrdSsiSession* session) {
...
-    : _session(session),
-      _payload(payload),
-      _requester(requester),
-      _finishFunc(finishFunc),
-      _retryFunc(retryFunc),
-      _status(status),
-      _finishStatus(QueryRequest::ACTIVE) {
-    LOGF_INFO("New QueryRequest with payload(%1%)" % payload.size());
-    _registerSelfDestruct();
+QueryRequest::QueryRequest( XrdSsiSession* session, std::shared_ptr<JobQuery> const jobQuery) :
```
andy-slac: ref after const?

core/modules/qdisp/QueryRequest.h
```diff
@@ -50,7 +51,7 @@ class BadResponseError : public std::exception {
     BadResponseError(std::string const& s_)
         : std::exception(),
           s("BadResponseError:" + s_) {}
-    ~BadResponseError() throw() {}
+    virtual ~BadResponseError() throw() {}
     virtual char const* what() throw() {
```
andy-slac: exact signature for this is:

    virtual const char* what() const noexcept;

(const is certainly missing here), same for RequestError class

timj: This is one of the warnings that I picked up from clang in [DM-3803](https://jira.lsstcorp.org/browse/DM-3803).

core/modules/qdisp/QueryRequest.h
```diff
@@ -92,13 +93,10 @@ class RequestError : public std::exception {
...
 public:
+    typedef std::shared_ptr<QueryRequest> Ptr;
     QueryRequest(
         XrdSsiSession* session,
-        std::string const& payload,
-        std::shared_ptr<ResponseRequester> const requester,
-        std::shared_ptr<util::UnaryCallable<void, bool> > const finishFunc,
-        std::shared_ptr<util::VoidCallable<void> > const retryFunc,
-        JobStatus& status);
+        std::shared_ptr<JobQuery> const jobQuery);
```
andy-slac: const ref

core/modules/qdisp/QueryRequest.h
```diff
@@ -117,36 +115,32 @@ class QueryRequest : public XrdSsiRequest {
...
-    std::shared_ptr<util::VoidCallable<void> > _retryFunc;
-    /// Reference to an updatable Status
-    JobStatus& _status;
+    util::Flag<bool> _retried; ///< Protect against multiple retries of _jobQuery from a single QueryRequest.
+    util::Flag<bool> _calledMarkComplete; ///< Protect against multiple calls to MarkCompleteFunc
+                                          /// from a single QueryRequest.
 
     std::mutex _finishStatusMutex;
-    enum FinishStatus { ACTIVE, FINISHED, CANCELLED, ERROR } _finishStatus;
+    enum FinishStatus { ACTIVE, FINISHED, CANCELLED, ERROR } _finishStatus; // _finishStatusMutex
```
andy-slac: `CANCELLED` is not used anymore?

core/modules/qdisp/QueryResource.cc
```diff
@@ -44,46 +44,53 @@
 #include "qdisp/JobStatus.h"
 #include "qdisp/QueryRequest.h"
 #include "qdisp/QueryResource.h"
-#include "qdisp/ResponseRequester.h"
 
 namespace lsst {
 namespace qserv {
 namespace qdisp {
 
+QueryResource::QueryResource(std::shared_ptr<JobQuery> jobQuery)
```
andy-slac: const ref

core/modules/qdisp/QueryResource.cc
```diff
@@ -44,46 +44,53 @@
...
+      // this char* must live as long as this object, so copy it on heap
+    : Resource(::strdup(jobQuery->getDescription().resource().path().c_str())),
+      _xrdSsiSession(NULL), _jobQuery(jobQuery) {
+    if (rName == NULL) {
```
andy-slac: `nullptr`

core/modules/qdisp/QueryRequest.h
andy-slac: Yes, true, I get it from there. Sorry for not mentioning your name here :)

core/modules/qdisp/QueryResource.h
```diff
@@ -40,58 +41,38 @@ namespace lsst {
...
-          _requester(requester),
-          _finishFunc(finishFunc),
-          _retryFunc(retryFunc),
-          _status(status) {
-        if (rName == NULL) {
-            throw std::bad_alloc();
-        }
-    }
+    typedef std::shared_ptr<QueryResource> Ptr;
+    QueryResource(std::shared_ptr<JobQuery> jobQuery);
```
andy-slac: const ref

core/modules/qdisp/ResponseRequester.h
```diff
@@ -20,8 +20,8 @@
  * the GNU General Public License along with this program.  If not,
  * see <http://www.lsstcorp.org/LegalNotices/>.
  */
-#ifndef LSST_QSERV_QDISP_RESPONSEREQUESTER_H
-#define LSST_QSERV_QDISP_RESPONSEREQUESTER_H
+#ifndef LSST_QSERV_QDISP_RESPONSEHANDLER_H
+#define LSST_QSERV_QDISP_RESPONSEHANDLER_H
```
andy-slac: This should be done after renaming the file, but probably OK if you are going to rename it soon.

core/modules/qdisp/ResponseRequester.h
```diff
@@ -45,16 +48,14 @@ namespace qdisp {
...
     typedef util::Error Error;
 
-    typedef util::VoidCallable<void> CancelFunc;
-
-    typedef std::shared_ptr<ResponseRequester> Ptr;
-    ResponseRequester() : _cancelled(false) {}
-    virtual ~ResponseRequester() {}
+    typedef std::shared_ptr<ResponseHandler> Ptr;
+    ResponseHandler() {}
+    void setJobQuery(std::shared_ptr<JobQuery> jobQuery) { _jobQuery = jobQuery; }
```
andy-slac: const ref

core/modules/util/threadSafe.h
```diff
@@ -74,17 +74,20 @@ class Flag {
     /** Sets flag value to 'val' and returns the old value of flag.
      */
     virtual T set(T val) {
```
andy-slac: General comment - we should probably replace this with `std::atomic<T>`. Definitely not for this ticket.

core/modules/qdisp/Executive.h
```diff
@@ -44,16 +46,39 @@ class XrdSsiService;
...
 namespace qdisp {
+
+class JobQuery;
 class MessageStore;
 class QueryResource;
-class ResponseRequester;
+
+/** Description of a job managed by the executive
+ */
+class JobDescription {
```
andy-slac: Maybe worth moving it now to a separate file?

core/modules/qdisp/Executive.h
```diff
@@ -155,31 +147,46 @@ class Executive {
...
+    XrdSsiService* _xrdSsiService; ///< RPC interface
+    JobMap _jobMap; ///< Contains information about all jobs.
+    JobMap _incompleteJobs; ///< Map of incomplete jobs.
 
     /** Execution errors */
     util::MultiError _multiError;
 
     int _requestCount; ///< Count of submitted tasks
-    bool _cancelled; ///< Has execution been cancelled?
+    util::Flag<bool> _cancelled; ///< Has execution been cancelled?
```
andy-slac: `std::atomic<bool>` instead?

core/modules/qdisp/JobQuery.cc
```diff
@@ -0,0 +1,124 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the LSST License Statement and
+ * the GNU General Public License along with this program.  If not,
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+
+// System headers
```
andy-slac: add <sstream>

core/modules/qdisp/QueryRequest.h
timj: :smile:

core/modules/qdisp/QueryRequest.h
jgates108: I delete that. I replaced CANCELLED with the  _cancelled flag. Changing the state from ACTIVE to CANCELLED resulted in  logic in _errorFinish being skipped and it complicated other things as well.

core/modules/ccontrol/MergingRequester.cc
jgates108: The compiler doesn't like the change to curly braces.

timj: which compiler?

core/modules/qdisp/QueryResource.cc
jgates108: I haven't upgraded my compiler, I was going to after this ticket and either clean this up as part of DM-2722 or put in another ticket to do the more minor things. Also, that's why I'm using NULL and not 0.

core/modules/ccontrol/MergingRequester.cc
jgates108: Just upgraded to g++ (GCC) 4.9.2 20150212 (Red Hat 4.9.2-6), and it still doesn't like it.
build/ccontrol/MergingRequester.cc:66: error: no matching function for call to 'std::basic_string<char, std::char_traits<char>, std::allocator<char> >::basic_string(<brace-enclosed initializer list>)'

core/modules/qdisp/ResponseRequester.h
```diff
@@ -45,16 +48,14 @@ namespace qdisp {
 /// segment of results. The requester instance functions as a handle object that
```
fritzm: s/requester/handler throughout this comment

core/modules/qdisp/QueryResource.h
```diff
@@ -40,58 +41,38 @@ namespace lsst {
...
 
+/** This class is used to request a session from xrootd via Provision
+ * ProvisionDone is called by xrootd in some thread with a XrdSsiSession.
+ * ProvisionDone creates a QueryRequest using the session and JobQuery,
+ * then it calls xrootd's ProcessRequest with that object.
+ *
+ * This objects existence is controlled by _jobQuery. It must call
+ * _jobQuery->freeQueryResource() before leaving ProvisionDone.
+ * If provisionDone is
+ */
```
fritzm: Finish/fixup this comment pls

core/modules/qdisp/QueryResource.h
```diff
@@ -40,58 +41,38 @@ namespace lsst {
...
-    }
+    typedef std::shared_ptr<QueryResource> Ptr;
+    QueryResource(std::shared_ptr<JobQuery> jobQuery);
 
-    virtual ~QueryResource() {
-        std::free(const_cast<char *>(rName)); // clean up heap allocated resource path copy
-    }
+    virtual ~QueryResource();
 
     void ProvisionDone(XrdSsiSession* s);
```
fritzm: Add "virtual" (or the new "override") here for clarity?

core/modules/qdisp/QueryResource.cc
```diff
@@ -44,46 +44,53 @@
...
-#include "qdisp/ResponseRequester.h"
 
 namespace lsst {
 namespace qserv {
 namespace qdisp {
 
+QueryResource::QueryResource(std::shared_ptr<JobQuery> jobQuery)
+      // this char* must live as long as this object, so copy it on heap
+    : Resource(::strdup(jobQuery->getDescription().resource().path().c_str())),
+      _xrdSsiSession(NULL), _jobQuery(jobQuery) {
```
fritzm: Given that the QueryResource lifetime is now mediated _jobQuery, do we still need the heap hack?

core/modules/qdisp/QueryResource.cc
```diff
@@ -44,46 +44,53 @@
 #include "qdisp/JobStatus.h"
 #include "qdisp/QueryRequest.h"
 #include "qdisp/QueryResource.h"
```
fritzm: redundant #include

core/modules/qdisp/QueryResource.h
```diff
@@ -40,58 +41,38 @@ namespace lsst {
...
     const char* eInfoGet(int &code);
 
-    std::string const& getPayload() const { return _payload; }
+    std::shared_ptr<JobQuery> getJobQuery() { return _jobQuery; }
+    bool getCancelled();
```
fritzm: Is this used?

core/modules/qdisp/QueryResource.cc
jgates108: QueryResource needs the information in _jobQuery to make the QueryRequest (such as JobStatus). It's easier to keep that all in _jobQuery than passing the one pointer instead of passing and/or copying things. I think it is also clearer.

core/modules/qdisp/QueryResource.cc
```diff
@@ -44,46 +44,53 @@
...
+QueryResource::~QueryResource() {
+    LOGF_DEBUG("~QueryResource()");
+    std::free(const_cast<char *>(rName)); // clean up heap allocated resource path copy
+}
+
 /// May not throw exceptions because the calling code comes from
 /// xrootd land and will not catch any exceptions.
 void QueryResource::ProvisionDone(XrdSsiSession* s) { // Step 3
-    LOGF_INFO("Provision done");
+    _provisionDoneHelper(s);
```
fritzm: why the extra layer of the helper excluding the single line below?

core/modules/qdisp/QueryResource.h
jgates108: Yes, in unit tests and to avoid creating QueryRequest if the job was cancelled.

core/modules/util/threadSafe.h
fritzm: Agree -- let's please drop a TODO: comment and put an issue on the backlog so we don't forget?

core/modules/qdisp/QueryResource.cc
```diff
@@ -44,46 +44,53 @@
 #include "qdisp/JobStatus.h"
```
fritzm: Are <cassert>, <iostream>, <sstream> still needed here?

core/modules/qdisp/QueryResource.cc
jgates108: It's too make sure that freeQueryResource gets called regardless of which logic path is followed.

core/modules/qdisp/QueryResource.cc
fritzm: Yes, got that.  I was asking if this strdup/free is still necessary (given that the JobQuery that gives it to us is going to outlive us?)

core/modules/qdisp/QueryRequest.h
```diff
@@ -36,13 +36,14 @@
 #include "XrdSsi/XrdSsiRequest.hh"
 
 // Local headers
+#include "qdisp/Executive.h"
+#include "qdisp/JobQuery.h"
 #include "util/Callable.h"
```
fritzm: No longer needed?

core/modules/qdisp/QueryRequest.h
fritzm: But shouldn't the enum entry go away then too?

core/modules/qdisp/JobStatus.h
```diff
@@ -85,8 +85,7 @@ class JobStatus {
     void updateInfo(State s, int code=0, std::string const& desc="");
```
fritzm: update include guards above to new filename

core/modules/qdisp/JobStatus.h
```diff
@@ -51,7 +51,7 @@ namespace qdisp {
 class JobStatus {
```
fritzm: #include global/ResourceUnit.h aboce no longer needed here?

core/modules/qdisp/ResponseRequester.h
jgates108: I expect to rename before merging with master.

core/modules/qdisp/Executive.cc
```diff
@@ -475,41 +366,43 @@ void Executive::_unTrack(int jobId) {
...
-            break;
+    for(auto i=_incompleteJobs.begin(), e=_incompleteJobs.end(); i != e;) {
+        if(!i->second->getDescription().respHandler()->getError().isNone()) {
+            // Requester should have logged the error to the messageStore
+            LOGF_INFO("Executive (%1%) reaped requester for jobId=%2%" % (void*)this % i->first);
+            auto eraseMe = i;
+            ++i;
+            _incompleteJobs.erase(eraseMe);
+        } else {
+            ++i;
```
jbecla: Move ++i outside of the if/else

core/modules/ccontrol/MergingRequester.cc
```diff
@@ -62,12 +63,15 @@ MergingRequester::MergingRequester(
...
       _response{new WorkerResponse()},
       _flushed{false},
-      _cancelled{false},
       _wName{"~"}  {
```
jbecla: extra space

core/modules/qdisp/Executive.cc
jgates108: Yes, finally we can do that :)

core/modules/qdisp/JobQuery.cc
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the LSST License Statement and
+ * the GNU General Public License along with this program.  If not,
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
```
jbecla: A short description (1-2 sentences) explaining what the JobQuery is would be very useful.

jbecla: you have in in header so it is not too bad :)

core/modules/qdisp/JobQuery.h
```diff
@@ -0,0 +1,122 @@
...
+    virtual ~JobQuery() {
+        LOGF_DEBUG("~JobQuery _jobId=%1%" % getId());
+    }
+    int getAttemptsToRun() const {
+        std::lock_guard<std::recursive_mutex> lock(_rmutex);
+        return _attemptsToRun;
+    }
+    int getMaxRetries() const { return 5; }
+    bool shouldRetry() const { return getAttemptsToRun() < getMaxRetries(); }
+    virtual bool runJob();
```
jbecla: Why is it virtual?

core/modules/qdisp/JobQuery.h
```diff
@@ -0,0 +1,122 @@
...
+
+    int getId() const { return _jobDescription.id(); }
+    JobDescription& getDescription() { return _jobDescription; }
+    JobStatus::Ptr getStatus() { return _jobStatus; }
+
+    void provisioningFailed(std::string const& msg, int code);
+    std::shared_ptr<QueryResource> getQueryResource() { return _queryResourcePtr; }
```
jbecla: Would be useful to cluster together all get functions, and separately modifiers.

core/modules/qdisp/Executive.cc
jgates108: Good point. These are now operation on 2 different maps and 2 different mutex, so this extra bit shouldn't be needed.

core/modules/qdisp/QueryRequest.h
```diff
@@ -117,36 +115,32 @@ class QueryRequest : public XrdSsiRequest {
     virtual void ProcessResponseData(char *buff, int blen, bool last);
 
     void cancel();
+    bool getCancelled();
```
jbecla: consider calling it isCancelled()

core/modules/qdisp/JobQuery.h
jbecla: This also does not change once set. Does it mean it should be in the section above?

core/modules/qdisp/JobQuery.h
```diff
@@ -0,0 +1,122 @@
...
+    JobDescription _jobDescription;
+    std::shared_ptr<MarkCompleteFunc> _markCompleteFunc;
+    std::weak_ptr<JobQuery> _weakThis; ///< Connection to the shared_ptr for this object.
+
+    // JobStatus has its own mutex.
+    JobStatus::Ptr _jobStatus; ///< Points at status in Executive::_statusMap
+
+    // Values that need mutex protection
+    mutable std::recursive_mutex _rmutex;
+    int _attemptsToRun;
```
jbecla: call it _nAttemptsToRun, or _runAttemptsCounter or something like that to indicate it is a counter.

core/modules/qdisp/JobQuery.h
```diff
@@ -0,0 +1,122 @@
...
+    Executive* _executive;
+    JobDescription _jobDescription;
+    std::shared_ptr<MarkCompleteFunc> _markCompleteFunc;
+    std::weak_ptr<JobQuery> _weakThis; ///< Connection to the shared_ptr for this object.
+
+    // JobStatus has its own mutex.
+    JobStatus::Ptr _jobStatus; ///< Points at status in Executive::_statusMap
+
+    // Values that need mutex protection
+    mutable std::recursive_mutex _rmutex;
```
jbecla: it guards _attemptsToRun, right? Let's call it _attemptsToRunMutex (well, rename accordingly, _nAttemptsToRunMutex, or _runAttemptsCounterMutex, or... ?)

core/modules/qdisp/JobQuery.cc
```diff
@@ -0,0 +1,124 @@
...
+    if (!cancelled && handlerReset) {
+        Ptr thisPtr = _weakThis.lock();
+        if (!thisPtr) {
+            logErr("runJob() _weakThis was null!", this);
+            return false;
+        }
+        {
+            QueryResource::Ptr qr = std::make_shared<QueryResource>(thisPtr);
+            std::lock_guard<std::recursive_mutex> lock(_rmutex);
+            if (shouldRetry()) {
```
jbecla: I'd get rid of shouldRetry() and getAttemptsToRun(). Both are used only once. You are locking a mutex here, then again inside getAttemptsToRun(). Maybe it would not have to be recursive_mutex if we clean that?

core/modules/qdisp/JobQuery.h
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    virtual ~JobQuery() {
+        LOGF_DEBUG("~JobQuery _jobId=%1%" % getId());
+    }
+    int getAttemptsToRun() const {
+        std::lock_guard<std::recursive_mutex> lock(_rmutex);
+        return _attemptsToRun;
+    }
+    int getMaxRetries() const { return 5; }
```
jbecla: hmm, magic hidden number. Define as enum as private class variable instead, and write a comment why 5 is a reasonable number and what should one worry about if anything when changing that to say 2 or 100

jbecla: Also, is there a reason why it is public? I'd go through all functions you have public in this class and make sure they need to be public.

core/modules/qdisp/JobQuery.h
jbecla: I found why! testQDisp.cc

core/modules/qdisp/testQDisp.cc
```diff
@@ -63,48 +63,67 @@ class ChunkMsgReceiverMock : public MsgReceiver {
...
-    virtual ~RetryTest() {}
-    virtual void operator()() {
-        _retryCalled = true;
-        LOGF_INFO("_retryCalled=%1%" % _retryCalled);
+    typedef std::shared_ptr<JobQueryTest> Ptr;
+    JobQueryTest(qdisp::Executive* executive,
+              qdisp::JobDescription const& jobDescription,
+              qdisp::JobStatus::Ptr jobStatus,
+              qdisp::MarkCompleteFunc::Ptr markCompleteFunc)
+       : qdisp::JobQuery(executive, jobDescription, jobStatus, markCompleteFunc),
```
jbecla: this section looks misaligned

core/modules/qdisp/testQDisp.cc
```diff
@@ -63,48 +63,67 @@ class ChunkMsgReceiverMock : public MsgReceiver {
...
+        return true;
+    }
+    bool retryCalled;
+
+    // Create a fresh JobQueryTest instance. If you're making this to get a QueryRequestObject,
+    // set createQueryRequest=true and pass an xsSession pointer.
+    // set createQueryResource=true to get a QueryResource.
+    // Special ResponseHandlers need to be defined in JobDescription.
+    static JobQueryTest::Ptr getJobQueryTest(qdisp::Executive* executive,
+            qdisp::JobDescription jobDesc,
```
jbecla: alignment

core/modules/qdisp/Executive.cc
jgates108: Nice that C++11 supports that for maps.

core/modules/qdisp/JobQuery.h
jgates108: The wrath of the emacs super gods is on your head :)

core/modules/qdisp/Executive.cc
jgates108: I think it has to stay in the else case. i = _incompleteJobs.erase(i) should increment the iterator to the next object. another increment will skip things.

core/modules/qdisp/QueryRequest.h
jgates108: What I meant was I read Andy's comment and then deleted CANCELLED from the enum. I had meant to do it earlier.

core/modules/qdisp/testQDisp.cc
fritzm: FWIW, I kind of like this alignment -- I find it really hard to read when params and inits are at the same indent level with no visual distinction?

core/modules/qdisp/QueryResource.cc
fritzm: Fair enough.  RAII pattern would also work.  Personal taste -- I'm okay with it either way -- maybe add a quicky comment about why helper -- when I saw it I was expecting it was being called from more than one place?

core/modules/qdisp/JobQuery.h
jgates108: Putting constants in functions is recommended in the coding standards (doesn't make much difference to me). As for where 5 came from, it's from the previous code and I don't know of a good reason for that value, but I don't want to change it without a good reason.

core/modules/qdisp/JobStatus.h
jgates108: Resource unit was duplicated between JobDescription and JobStatus. No need for 2 copies.

core/modules/qdisp/testQDisp.cc
jgates108: The colon here makes a decent visual break. The one below is pretty bad without some difference in indentation.

core/modules/qdisp/QueryResource.cc
jgates108: Have any RAII tricks better than a small friend class that makes the call when destroyed?

jgates108: I think I came up with something better in this case.

core/modules/qdisp/testQDisp.cc
jbecla: i was talking about pushing the last 3 arguments 4 spaces to the right to align them with the 1st one. I guess I should have inserted my comment one line higher. I have nothing against the line starting with colon.

core/modules/qdisp/QueryResource.cc
fritzm: Well, it turns out you could do this, but its sort of obscure so I dunno...

```c++#include <iostream>

using namespace std;

// This is generic, only needed once:
template<typename T>
using Janitor = std::unique_ptr<T,std::function<void(T*)>>;

// Here's a sample thing with a cleanup method:
class Hazard
{
public:
    void Cleanup() { cout << "cleaning up!" << endl; }
};

// Here's a usage of the generic janitor:
int main(int argc, const char *argv[])
{
    Hazard aHazard;
    Janitor<Hazard> janitor(&aHazard, [](Hazard *h){ h->Cleanup(); });
    return 0;
}
```

core/modules/qdisp/ResponseRequester.h
```diff
@@ -32,29 +32,28 @@
...
 
-/// ResponseRequester is an interface that handles result bytes. Tasks are
+class JobQuery;
+
+/// ResponseHandler is an interface that handles result bytes. Tasks are
 /// submitted to an Executive instance naming a resource unit (what resource is
-/// required), a request string (task payload), and a requester (handler for
-/// returning bytes). The requester implements logic to process incoming results
+/// required), a request string (task payload), and a handler for returning bytes.
+/// The requester implements logic to process incoming results
```
fritzm: "requester" in line above should be "ResponseHandler"?

core/modules/qdisp/QueryRequest.h
```diff
@@ -36,13 +36,14 @@
 #include "XrdSsi/XrdSsiRequest.hh"
 
 // Local headers
+#include "qdisp/Executive.h"
```
fritzm: I don't see executive mentioned below -- is this needed?

core/modules/qdisp/QueryRequest.cc
```diff
@@ -59,71 +59,45 @@ inline void unprovisionSession(XrdSsiSession* session) {
...
-    LOGF_INFO("New QueryRequest with payload(%1%)" % payload.size());
-    _registerSelfDestruct();
+QueryRequest::QueryRequest( XrdSsiSession* session, std::shared_ptr<JobQuery> const& jobQuery) :
+    _session(session), _jobQuery(jobQuery), _jobDesc(_jobQuery->getDescription()) {
+    LOGF_INFO("New QueryRequest with payload(%1%)" % _jobDesc.payload().size());
 }
 
 QueryRequest::~QueryRequest() {
+    LOGF_DEBUG("~QueryRequest");
     unprovisionSession(_session);
```
fritzm: unprovisionSession() seems to be called from this place only.  Just put the def inline here?

core/modules/qdisp/QueryRequest.cc
```diff
@@ -59,71 +59,45 @@ inline void unprovisionSession(XrdSsiSession* session) {
...
 char* QueryRequest::GetRequest(int& requestLength) {
-    requestLength = _payload.size();
+    requestLength = _jobDesc.payload().size();
     LOGF_DEBUG("Requesting, payload size: [%1%]" % requestLength);
     // Andy promises that his code won't corrupt it.
-    return const_cast<char*>(_payload.data());
+    return const_cast<char*>(_jobDesc.payload().data());
 }
 
+// Deleting the buffer (payload) would cause us problems.
```
fritzm: maybe elaborate a little here?

core/modules/qdisp/QueryRequest.cc
```diff
@@ -151,24 +124,21 @@ bool QueryRequest::ProcessResponse(XrdSsiRespInfo const& rInfo, bool isOk) {
...
 bool QueryRequest::_importStream() {
-    bool retrieveInitiated = false;
-    // Pass ResponseRequester's buffer directly.
-    std::vector<char>& buffer = _requester->nextBuffer();
+    bool success = false;
+    // Pass ResponseHandler's buffer directly.
+    std::vector<char>& buffer = _jobDesc.respHandler()->nextBuffer();
     LOGF_DEBUG("QueryRequest::_importStream buffer.size=%1%" % buffer.size());
     const void* pbuf = (void*)(&buffer[0]);
     LOGF_INFO("_importStream->GetResponseData size=%1% %2% %3%" %
```
fritzm: Should these sorts of things log at level DEBUG and not INFO?

core/modules/qdisp/JobQuery.h
```diff
@@ -0,0 +1,132 @@
...
+        _queryResourcePtr.reset();
+    }
+
+    void setQueryRequest(std::shared_ptr<QueryRequest> const& qr) {
+        std::lock_guard<std::recursive_mutex> lock(_rmutex);
+        _queryRequestPtr = qr;
+    }
+    std::shared_ptr<QueryRequest> getQueryRequest() {
+        std::lock_guard<std::recursive_mutex> lock(_rmutex);
+        return _queryRequestPtr; }
```
fritzm: formatting?

core/modules/qdisp/JobQuery.h
```diff
@@ -0,0 +1,132 @@
...
+// System headers
+#include <atomic>
+#include <memory>
+#include <mutex>
+
+// LSST headers
+#include "lsst/log/Log.h"
+
+// Qserv headers
+#include "qdisp/Executive.h"
```
fritzm: It looks like you could just forward-declare Executive here instead of \#include

jgates108: The definition for JobDescription is in that file as well, although I intend to move that to its own file before merging.

core/modules/qdisp/QueryRequest.cc
jbecla: Yes! We are making this mistake in so many places I stopped pointing it out. I am hoping one day we will do a complete sweep through the code and move info to debug.

core/modules/qdisp/Executive.h
```diff
@@ -99,7 +114,7 @@ class Executive {
     /// Squash everything. should we block?
     void squash();
```
fritzm: Should we maybe rename this to "Cancel"?

core/modules/qdisp/Executive.h
```diff
@@ -153,33 +146,46 @@ class Executive {
...
+    virtual void operator()(bool success) {
+        if (_executive) {
+            _executive->markCompleted(_jobId, success);
+        }
+    }
+
+private:
+    Executive* _executive;
+    int _jobId;
+};
```
fritzm: For your consideration: this whole class could be replaced with just `std::function<void(bool)>`.  Then the executive, id, and the markCompleted logic could be rolled up at point of assignment in a one-liner lambda, e.g. instead of calling the ctor.

I like that approach in general because it keeps the logic of what's going to happen on the callback local at the place where it's installed (instead of having to go look up another helper class) and it cuts down on the number of small special purpose glue/helper classes which I find distracting when reading the code.  YMMV!

fritzm: (Additionally, if you use std::function type I think you could dispense with the shared ptr you have tracking this thing?)

core/modules/qdisp/Executive.cc
```diff
@@ -357,103 +275,57 @@ struct printMapEntry {
...
+    return _incompleteJobs.size();
 }
 
 std::string Executive::getProgressDesc() const {
     std::ostringstream os;
     {
-        std::lock_guard<std::mutex> lock(_statusesMutex);
-        std::for_each(_statuses.begin(), _statuses.end(), printMapEntry(os, "\n"));
+        std::lock_guard<std::mutex> lock(_jobsMutex);
+        std::for_each(_jobMap.begin(), _jobMap.end(), printMapEntry(os, "\n"));
```
fritzm: If this is the only use of printMapEntry, consider dispensing with helper and rolling it directly in here as a lambda?

```c++
std::ostringstream os;
bool first = true;
{
    std::lock_guard<std::mutex> lock(_jobsMutex);
    std::for_each(_jobMap.begin(), _jobMap.end(), [&](Executive::JobMap::value_type const& entry){
        JobQuery::Ptr job = entry.second;
        if(!first) { os << "\n"; }
        os << "Ref=" << entry.first << " " << job;
        first = false;
    }));
}
```

fritzm: Oh -- or just a range based loop?

jgates108: Already replaced with range based loop.

fritzm:
Awesome, thanks for digging in on this one, we really needed this!

core/modules/qdisp/Executive.h
jgates108: There are some definite trade offs here. The lambdas require less code and they make the unit tests a little clearer, in that it is clear the function capability is not being used when it is a no-op. On the other hand, the function is called a very long way from the lambdas are defined, and hunting down the source of a std::function<void(bool)> function call is significantly harder than locating MarkCompleteFunc. The full blown functor also has a bit more flexibility to do other things, which might be needed to get cancellation to work. I'm still on the fence.

PR #131, opened 2015-09-15
Tickets/dm 3253

n8pease:


core/modules/css/python/testKvInterface.py
```diff
@@ -55,56 +113,51 @@ def testCreateGetSetDelete(self):
...
+
+    def testPerformance(self):
+        n = 10 # set it to something larger for real test...
+        start = time.clock()
+        for i in range(1,n+1):
+            k = "/unittest/node_%02i" % i
+            #print ("creating %s --> %i" % (k, i))
+            self._kvI.create(k, str(i))
+        elapsed = time.clock()-start
+        print "It took", elapsed, "to create", n, "entries"
```
timj: Please us print function

core/modules/css/python/testKvInterface.py
```diff
@@ -27,27 +27,85 @@
...
+    schema = schemaFile.read()
+    errObj = cssLib.SqlErrorObject()
+    sqlCfg = getConfig()
+    # replace production schema name with test schema:
+    schema = schema.replace("qservCssData", sqlCfg.dbName)
+    sqlCfg.dbName = ''
+    sqlConn = cssLib.SqlConnection(sqlCfg)
+    sqlConn.runQuery(schema, errObj)
+    if errObj.isSet():
+        print "setupDatabase error:%s" %(errObj.printErrMsg())
```
timj: print function

core/modules/css/python/testKvInterface.py
```diff
@@ -55,56 +110,51 @@ def testCreateGetSetDelete(self):
...
+
+    def testPerformance(self):
+        n = 10 # set it to something larger for real test...
+        start = time.clock()
+        for i in range(1,n+1):
+            k = "/unittest/node_%02i" % i
+            #print ("creating %s --> %i" % (k, i))
+            self._kvI.create(k, str(i))
+        elapsed = time.clock()-start
+        print("It took" + str(elapsed) + "to create" + str(n) + "entries")
```
timj:
```python
from __future__ import print_function
```
is allowed in our code.

n8pease: can you please be specific about what you're asking for?

timj: I'm asking for that line to be added to the top of the file so it's clear we are using the print function rather than relying on python2.7 to ignore the parentheses. It will be less confusing later on if someone edits the print by adding a "," to the end and then a tuple prints out.

Adding the line also prevents `futurize` from grumbling.

core/modules/css/cssLib.i
```diff
@@ -47,7 +53,118 @@ namespace std {
     %template(StringVector) vector<string>;
 };
 
+%include "sql/SqlErrorObject.h"
+
+%{
+    void setPythonException(const std::string& exceptionStr, const lsst::qserv::css::CssError& ex) {
+        PyObject* module = PyImport_ImportModule("lsst.qserv.css");
+        if (!module)
+            throw;
```
andy-slac: My SWIG is very rusty, what happens here if you throw C++ exception here, is it caught by SWIG layer or does it crash whole thing?

core/modules/css/cssLib.i
```diff
@@ -47,7 +53,118 @@ namespace std {
     %template(StringVector) vector<string>;
 };
 
+%include "sql/SqlErrorObject.h"
+
+%{
+    void setPythonException(const std::string& exceptionStr, const lsst::qserv::css::CssError& ex) {
+        PyObject* module = PyImport_ImportModule("lsst.qserv.css");
```
andy-slac: This object will need Py_DECREF() too when you are done

core/modules/css/cssLib.i
```diff
@@ -47,7 +53,118 @@ namespace std {
...
+        if (!module)
+            throw;
+        PyObject* exception = PyObject_GetAttrString(module, exceptionStr.c_str());
```
andy-slac: And Py_DECREF(exception) will be needed too.

core/modules/css/cssLib.i
```diff
@@ -47,7 +53,118 @@ namespace std {
...
+        if (not exception)
+            throw;
+        PyObject* what = PyString_FromString(ex.what());
```
andy-slac: I'd replace this line and the rest of this function with one line:

    PyErr_SetString(exception, ex.what());

core/modules/css/cssLib.i
```diff
@@ -47,7 +53,118 @@ namespace std {
...
+        Py_DECREF(type);
+        Py_DECREF(instance);
+    }
+%}
+
+%exception {
+    try {
+        $action
+    }
+    catch (lsst::qserv::css::NoSuchDb e) {
```
andy-slac: `}` and `catch` should be on the same line according to out coding guidelines

core/modules/css/python/testKvInterface.py
n8pease: thanks. ok, will do.

andy-slac: Or use `logging` then you don't need to care about `print` complications :wink:

andy-slac: The replaced version needs some spaces in strings.

core/modules/css/python/testKvInterface.py
```diff
@@ -27,21 +27,76 @@
...
+    errObj = cssLib.SqlErrorObject()
+    sqlCfg = getConfig()
+    # replace production schema name with test schema:
+    schema = schema.replace("qservCssData", sqlCfg.dbName)
+    sqlCfg.dbName = ''
+    sqlConn = cssLib.SqlConnection(sqlCfg)
+    sqlConn.runQuery(schema, errObj)
+    if errObj.isSet():
+        print("setupDatabase error: "  + errObj.printErrMsg())
+        return False
```
andy-slac: Maybe better to throw here, the return code is not checked and running test after this failure is probably pointless.

core/modules/css/cssLib.i
```diff
@@ -47,7 +53,118 @@ namespace std {
...
+%include "mysql/MySqlConfig.h" 
+%include "sql/SqlConnection.h"
+
+%include "css/KvInterfaceImplMySql.h"
 %include "global/constants.h"
+
+%pythoncode %{ 
+class CssError(Exception):
+    def __init__(self, msg):
+        self.msg = msg
```
andy-slac: Replace with:

    Exception.__init__(self, msg)

core/modules/css/python/testKvInterface.py
```diff
@@ -27,21 +27,76 @@
...
+    
+def initConnection():
+    """
+    @brief connects to the database and sets up a table with the schema to run the tests
+    @return True if conected & schema loaded, else false
+    """
+    # need config without database name
+    sqlConfigLocal = sqlConfig
+    sqlConfigLocal.dbName = ""
+    sqlConn = cssLib.SqlConnection(sqlConfigLocal)
```
andy-slac: is this supposed to be `global sqlConn`? OTOH `initConnection()` is not used anywhere, just drop it.

core/modules/css/python/testKvInterface.py
```diff
@@ -27,21 +27,76 @@
...
+        return None
+    cfg = cssLib.MySqlConfig()
+    cfg.username = cfgParser.get('mysql', 'user')
+    cfg.password = cfgParser.get('mysql', 'passwd')
+    cfg.hostname = cfgParser.get('mysql', 'host')
+    cfg.port = cfgParser.getint('mysql', 'port')
+    cfg.dbName = testDatabaseName
+    return cfg
+
+
```
andy-slac: There are invisible spaces on this line

core/modules/css/python/testKvInterface.py
```diff
@@ -27,21 +27,76 @@
...
+    """
+    @brief connects to the database and sets up a table with the schema to run the tests
+    @return True if conected & schema loaded, else false
+    """
+    # need config without database name
+    sqlConfigLocal = sqlConfig
+    sqlConfigLocal.dbName = ""
+    sqlConn = cssLib.SqlConnection(sqlConfigLocal)
+
+
```
andy-slac: Four evil hiding spaces here too :scream:

core/modules/css/python/testKvInterface.py
```diff
@@ -27,21 +27,76 @@
...
+    sqlCfg = getConfig()
+    # replace production schema name with test schema:
+    schema = schema.replace("qservCssData", sqlCfg.dbName)
+    sqlCfg.dbName = ''
+    sqlConn = cssLib.SqlConnection(sqlCfg)
+    sqlConn.runQuery(schema, errObj)
+    if errObj.isSet():
+        print("setupDatabase error: "  + errObj.printErrMsg())
+        return False
+    return True
```
andy-slac: One trailing space on this line. `git diff --check` finds these before you commit.

core/modules/css/cssLib.i
```diff
@@ -31,9 +31,15 @@ Access to the classes from the qserv_css library
...
 #include "css/constants.h"
+#include "css/CssError.h"
 #include "css/KvInterface.h"
 #include "css/KvInterfaceImplMem.h"
+#include "css/KvInterfaceImplMySql.h"
 #include "global/constants.h"
+#include "global/stringTypes.h"
+#include "mysql/MySqlConfig.h"
+#include "sql/SqlErrorObject.h"
+#include "sql/SqlConnection.h"
```
jbecla: Swap order to keep alphabetical

core/modules/css/python/testKvInterface.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 
 # LSST Data Management System
-# Copyright 2013-2014 AURA/LSST.
+# Copyright 2013-2015 AURA/LSST.
 #
```
jbecla: You are creating a database and tables here but do not destroy them. It is working now because you have "create if not exists". We will run into troubles with this approach if we ever want to change schema (because old schema is left behind), or multiple users run the same test at the same time (because db name is hardcoded). I'd recommend to create unique db name each time the test starts, and destroy it at the end of the test.

core/modules/css/python/testKvInterface.py
```diff
@@ -55,56 +110,51 @@ def testCreateGetSetDelete(self):
...
+        elapsed = time.clock()-start
+        print("It took" + str(elapsed) + "to create" + str(n) + "entries")
+        self._kvI.deleteKey("/unittest")
 
 def main():
-    setLogging()
     unittest.main()
 
 if __name__ == "__main__":
+    requiredRunDir = "lsst/qserv/core/modules"
```
jbecla: Is this passing in jenkins?

core/modules/css/cssLib.i
n8pease: AFAIK it rethrows the current C++ exception (which does crash python). If the css module can't be imported I'm not really sure what other option there is... I guess it could raise a python exception, that would probably be better.

andy-slac: I think crashing due to uncaught C++ exception is not very nice, it's better to raise Python exception. You can just do `return` instead of `throw` in this case because import had already set Python error/exception state, so people will see "Import error: ... not found" or something like this.

PR #133, opened 2015-09-21
Migrate Qserv to C++11 nullptr (DM-2956)

jbecla:


core/modules/qdisp/QueryRequest.cc
```diff
@@ -33,6 +33,7 @@
 #include "qdisp/QueryRequest.h"
 
 // System headers
+#include <cstddef>
```
fritzm: I don't see any nullptr changes elsewhere in this file -- is this #include still needed, or was the NULL usage factored away before you rebased?

timj: Yes, in the original branch it was:
```diff
diff --git a/core/modules/qdisp/QueryRequest.cc b/core/modules/qdisp/QueryRequest.cc
index ffe5d74..1c88ceb 100644
--- a/core/modules/qdisp/QueryRequest.cc
+++ b/core/modules/qdisp/QueryRequest.cc
@@ -33,6 +33,7 @@
 #include "qdisp/QueryRequest.h"
 
 // System headers
+#include <cstddef>
 #include <iostream>
 
 // LSST headers
@@ -66,7 +67,7 @@ public:
     Canceller(QueryRequest* qr) : _queryRequest(qr) {}
     virtual ~Canceller() { delete _queryRequest; }
     virtual void operator()() {
-        if (_queryRequest != NULL) {
+        if (_queryRequest) {
             _queryRequest->cancel();
         }
     }
@@ -189,7 +190,7 @@ void QueryRequest::ProcessResponseData(char *buff, int blen, bool last) { // Ste
     if(blen < 0) { // error, check errinfo object.
         int eCode;
         const char* chs = eInfo.Get(eCode);
-        std::string reason = (chs == NULL) ? "Null" : chs;
+        std::string reason = (chs == nullptr) ? "Null" : chs;
         _status.updateInfo(JobStatus::RESPONSE_DATA_NACK, eCode, reason);
         LOGF_ERROR("ProcessResponse[data] error(%1%,\"%2%\")" % eCode % reason);
         _requester->errorFlush("Couldn't retrieve response data:" + reason, eCode);
```

jbecla: I wondered if anyone would ask :). I added it when doing original migration and then John touch these lines so it got removed during rebasing. I figured since we are using nullptr in that file, so it'd be good to keep the include, although clearly some other header is bringing it. I am ok with removing that include if you think it is better.

timj: `nullptr` is still in that file so there doesn't seem to be any harm in keeping the include in there.

fritzm: Oh, if its used elsewhere in the file then let's keep the direct 
include, thanks -- I missed that!

PR #134, opened 2015-09-24
Synchronize access to database (DM-3943)

andy-slac:
Added synchromization to all QMetaMysql methods that access database.
Normally protection is only needed around send/receive methods on the
same physical connection but we also manage transaction state so
single-query synchronization is not enough.

### DIFF
--- a/core/modules/qmeta/QMetaMysql.h
+++ b/core/modules/qmeta/QMetaMysql.h
@@ -23,6 +23,7 @@
 #define LSST_QSERV_QMETA_QMETAMYSQL_H
 
 // System headers
+#include <mutex>
 
 // Third-party headers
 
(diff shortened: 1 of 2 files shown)


core/modules/qmeta/QMetaMysql.h
```diff
@@ -235,6 +236,7 @@ class QMetaMysql : public QMeta {
 private:
 
     sql::SqlConnection _conn;
+    std::mutex _dbMutex;    ///< Synchronizes access to connection
```
jgates108: Might be better to call this _connMutex with comment Synchronize access to _conn.

andy-slac: The comment should be better indeed. Not all access to _conn needs to be synchronized, only some operation (basically sending query and receiving data back) are not thread safe. But we end up synchronizing practically everything because we also want to synchronize transaction handling (no overlapping transactions on the same connection from different threads).

PR #135, opened 2015-09-26
Remove dependency on mysqldb in wmgr (DM-3947)

jbecla:


core/modules/wmgr/python/config.py
```diff
@@ -104,8 +104,8 @@ def dbEngine(self):
         _log.debug('creating new connection %s', kwargs)
         if self.dbPasswd: kwargs['password'] = self.dbPasswd
         inst = getEngineFromArgs(**kwargs)
-        # ignore mysql warnings
-        warnings.filterwarnings("ignore", category=MySQLdb.Warning)
+        # ignore SQLAlchemy warnings
+        warnings.filterwarnings("ignore", category=SAWarning)
```
andy-slac: I am not sure that it does what you expect it to do or that it does anything at all. Can you try to remove `filterwarnings()` and see if it integration tests run OK.

core/modules/wmgr/python/config.py
```diff
@@ -118,8 +118,8 @@ def privDbEngine(self):
         _log.debug('creating new connection %s', kwargs)
         if self.dbPasswdPriv: kwargs['password'] = self.dbPasswdPriv
         inst = getEngineFromArgs(**kwargs)
-        # ignore mysql warnings
-        warnings.filterwarnings("ignore", category=MySQLdb.Warning)
+        # ignore SQLAlchemy warnings
+        warnings.filterwarnings("ignore", category=SAWarning)
```
andy-slac: same here

core/modules/wmgr/python/dbMgr.py
```diff
@@ -916,7 +902,7 @@ def getIndex(dbName, tblName, chunkId=None):
         allRows = dbConn.execute(query)
 
         if allRows.keys():
-            descr = [dict(name=d[0], type=_typeCode2Name(d[1])) for d in allRows.keys()]
+            descr = [dict(name=d[0], type=utils.typeCode2Name(dbConn, d[1])) for d in allRows.keys()]
```
andy-slac: This is totally broken, it was actully broken by previous commit which I failed to notice at that time. Old pre-sqlalchemy code used `cursor.description` to get type codes here, `keys()` only return the list of column names. `d[1]` above returns second character of the column name which is not a type code. We need to fix this, and this should be related to what Brian was doing with dax_dbserv when it needs to return column types. Maybe we should factor out Brian's code and move it to db.utils, for now I think it should be fie to replace `allRows.keys()` with `allRows.cursor.description`.

jbecla: Clearly integration tests are not exercising this code... can you think of a way to address that?

core/modules/wmgr/python/config.py
jbecla: Works fine without filterwarnings, so I'll drop these lines

core/modules/wmgr/python/dbMgr.py
andy-slac: This code runs, it's jut not crashing and it returns nonsense. `typeCode2Name()` returns `None` when it cannot figure out the type, and the result is not actually used by anything yet.

PR #136, opened 2015-09-28
Add CssAccess class providing new interface for CSS (DM-2733)

andy-z:
This new class replaces Facade and has additional methods for update operations on CSS. Many methods of Facade that returned individual bits were replaced with methods returning combined info (like TableParams, PartTableParams, etc.) to keep data consistent. Unit tests for practically all methods of CssAccess are added as well.

Czar is now using new CssAccess class, Facade has been removed. Passess all unit tests and integration tests.

Extend KvInterface API to support operations needed by CssAccess.


core/modules/css/CssAccess.cc
```diff
@@ -0,0 +1,855 @@
...
+
+// Construct from KvInterface instance and empty chunk list instance
+CssAccess::CssAccess(std::shared_ptr<KvInterface> const& kvInterface,
+                     std::shared_ptr<EmptyChunks> const& emptyChunks,
+                     std::string const& prefix)
+    : _kvI(kvInterface), _emptyChunks(emptyChunks), _prefix(prefix) {
+}
+
+int
+CssAccess::cssVersion() {
```
jbecla: This function can be const, can it not?

core/modules/css/CssAccess.cc
```diff
@@ -0,0 +1,855 @@
...
+    auto dbMap = _getSubkeys(_prefix + "/DBS/" + dbName, {"partitioningId"});
+    auto const& partId = dbMap["partitioningId"];
+    if (partId.empty()) {
+        // if database is not defined throw an exception, otherwise return default values
+        _assertDbExists(dbName);
+        return striping;
+    }
+
+    // get all keys
+    std::string pkey = _prefix + "/PARTITIONING/_" + partId;
```
jbecla: pKey

core/modules/css/CssAccess.cc
```diff
@@ -0,0 +1,855 @@
...
+CssAccess::createDb(std::string const& dbName,
+                    StripingParams const& striping,
+                    std::string const& storageClass,
+                    std::string const& releaseStatus) {
+    LOGF(_log, LOG_LVL_DEBUG, "createDb(%1%)" % dbName);
+
+    std::string partId;
+    if (striping.stripes > 0) {
+        // need to define partitioning structure
+
```
jbecla: Let's remove comment or define the structure

core/modules/css/CssAccess.cc
```diff
@@ -0,0 +1,855 @@
...
+    LOGF(_log, LOG_LVL_DEBUG, "dropDb(%1%)" % dbName);
+
+    std::string key = _prefix + "/DBS/" + dbName;
+
+    // key is supposed to exist and key.json is optional
+    try {
+        LOGF(_log, LOG_LVL_DEBUG, "dropDb: try to delete packed key: %s.json" % key);
+        _kvI->deleteKey(key + ".json");
+    } catch (NoSuchKey const& exc) {
+        LOGF(_log, LOG_LVL_DEBUG, "dropDb: packed key is not there");
```
jbecla: "is not there" --> "not found" here and below

core/modules/css/CssAccess.cc
```diff
@@ -0,0 +1,855 @@
...
+std::vector<std::string>
+CssAccess::getTableNames(std::string const& dbName, bool readyOnly) const {
+    LOGF(_log, LOG_LVL_DEBUG, "getTableNames(%1%)" % dbName);
+
+    std::string key = _prefix + "/DBS/" + dbName + "/TABLES";
+    std::vector<std::string> names;
+    try {
+        names = _kvI->getChildren(key);
+    } catch (NoSuchKey const& exc) {
+        LOGF(_log, LOG_LVL_DEBUG, "getTableNames: key is not found: %s" % key);
```
jbecla: "key not found", or add "is" above

core/modules/css/CssAccess.cc
```diff
@@ -0,0 +1,855 @@
...
+    if (val == "DOES_NOT_EXIST") {
+        // table key is not there at all, throw if database name is not good
+        _assertDbExists(dbName);
+        LOGF(_log, LOG_LVL_DEBUG, "containsTable: key is not there: %1%" % key);
+        return false;
+    }
+    LOGF(_log, LOG_LVL_DEBUG, "containsTable: key value: %1%" % val);
+    // if key value is not "READY" it likely means table is in the process
+    // of being deleted, which is the same as if it does not exist
+    if (readyOnly) return val == "READY";
```
jbecla: I think we should use brackets, even for one liners, so
if (readyOnly) {
    return val == "READY";
}

core/modules/css/CssAccess.cc
```diff
@@ -0,0 +1,855 @@
...
+    return schema;
+}
+
+MatchTableParams
+CssAccess::getMatchTableParams(std::string const& dbName,
+                               std::string const& tableName) const {
+    LOGF(_log, LOG_LVL_DEBUG, "getMatchTableParams(%1%, %2%)" % dbName % tableName);
+
+    std::string const tableKey = _prefix + "/DBS/" + dbName + "/TABLES/" + tableName;
+
```
jbecla: you had less empty lines for similar code in functions above

core/modules/css/CssAccess.cc
```diff
@@ -0,0 +1,855 @@
...
+        throw KeyValueError(tableKey + "/partitioning", "one of the sub-keys is not numeric: " +
+                            std::string(exc.what()));
+    }
+
+    return params;
+}
+
+TableParams
+CssAccess::getTableParams(std::string const& dbName, std::string const& tableName) const {
+
```
jbecla: remove empty line

core/modules/css/CssAccess.cc
```diff
@@ -0,0 +1,855 @@
...
+    return params;
+}
+
+void
+CssAccess::createTable(std::string const& dbName,
+                       std::string const& tableName,
+                       std::string const& schema,
+                       PartTableParams const& partParams) {
+
```
jbecla: remove empty line

core/modules/css/CssAccess.cc
```diff
@@ -0,0 +1,855 @@
...
+            std::make_pair("subChunks", std::to_string(int(partParams.subChunks))),
+        };
+        // only store overlap if non-zero
+        if (partParams.overlap != 0.0) {
+            partMap.insert(std::make_pair("overlap", std::to_string(partParams.overlap)));
+        }
+        _storePacked(tableKey + "/partitioning", partMap);
+    }
+
+    // done
```
jbecla: "// done" not needed?

core/modules/css/CssAccess.cc
andy-slac: It's static.

core/modules/css/CssAccess.cc
andy-slac: ok

core/modules/css/CssAccess.cc
andy-slac: Well, the comment says that we need to define that structure in case when `stripes > 0` and code in the if block actually does it, so we do define structure there.

core/modules/css/CssAccess.cc
andy-slac: ok

core/modules/css/CssAccess.cc
andy-slac: ok

core/modules/css/CssAccess.cc
andy-slac: Our coding guidelines has example of one-line `if`without braces, so it should be ok.

core/modules/css/CssAccess.cc
andy-slac: Is it too hard to read?

core/modules/css/CssAccess.cc
andy-slac: I changed it to `// done, can mark table as ready`

core/modules/css/CssAccess.cc
andy-slac: ok

core/modules/css/CssAccess.cc
andy-slac: ok

core/modules/css/CssAccess.cc
jbecla: Ah i see. It is at the same level as the comments below "// create unique key for it and get its id" etc, that is what mislead me. I am not sure how to fix it though.

core/modules/css/CssAccess.cc
jbecla: it just takes more real estate on the screen that it could have :)

core/modules/css/CssAccess.h
```diff
@@ -0,0 +1,422 @@
...
+ *  likely migrate to some different interface (e.g. become a part of
+ *  secondary index) so it should be considered as temporary here.
+ *
+ *  This is a concrete class, instances can be copied around and all copies
+ *  share the same KvInterface instance (and empty chunk list).
+ */
+
+class CssAccess {
+public:
+
```
jgates108: I kind of like Daniel's typdef std::shared_ptr<CssAccess> Ptr;

core/modules/css/CssAccess.cc
```diff
@@ -0,0 +1,848 @@
...
+std::vector<std::string>
+CssAccess::getDbNames() const {
+    std::string p = _prefix + "/DBS";
+    return _kvI->getChildren(p);
+}
+
+bool
+CssAccess::containsDb(std::string const& dbName) const {
+    if (dbName.empty()) {
+        LOGF(_log, LOG_LVL_DEBUG, "Empty database name passed.");
```
jgates108: What are our current logging criteria? This looks like it shouldn't happen often and might be good to know when it does, which seems like info level to me.

core/modules/css/CssAccess.cc
jgates108: Maybe drop the "need to". That and the blank line makes it seem like something still needs to be done.

core/modules/css/CssAccess.cc
```diff
@@ -0,0 +1,848 @@
...
+        LOGF(_log, LOG_LVL_DEBUG, "dropDb: try to delete packed key: %s.json" % key);
+        _kvI->deleteKey(key + ".json");
+    } catch (NoSuchKey const& exc) {
+        LOGF(_log, LOG_LVL_DEBUG, "dropDb: packed key is not found");
+    }
+    try {
+        LOGF(_log, LOG_LVL_DEBUG, "dropDb: try to delete regular key: %s" % key);
+        _kvI->deleteKey(key);
+    } catch (NoSuchKey const& exc) {
+        LOGF(_log, LOG_LVL_DEBUG, "dropDb: regular key is not found");
```
jgates108: I'm wondering if this should be info level. If so, it should probably include the key name.

core/modules/css/CssAccess.cc
```diff
@@ -0,0 +1,848 @@
...
+    return names;
+}
+
+bool
+CssAccess::containsTable(std::string const& dbName, std::string const& tableName, bool readyOnly) const {
+    LOGF(_log, LOG_LVL_DEBUG, "containsTable(%1%, %2%)" % dbName % tableName);
+
+    std::string const key = _prefix + "/DBS/" + dbName + "/TABLES/" + tableName;
+    // If key is not there pretend that its value is not "READY"
+    std::string const val = _kvI->get(key, "DOES_NOT_EXIST");
```
jgates108: Is it possible for a key to have the value "DOES_NOT_EXIST"? I'm not that familiar with the schema here.

core/modules/css/CssAccess.cc
```diff
@@ -0,0 +1,848 @@
...
+    }
+
+    // fill the structure
+    params.dirDb = paramMap["partitioning/dirDb"];
+    params.dirTable = paramMap["partitioning/dirTable"];
+    params.dirColName = paramMap["partitioning/dirColName"];
+    params.latColName = paramMap["partitioning/latColName"];
+    params.lonColName = paramMap["partitioning/lonColName"];
+    params.partitioned =
+            paramMap.count("partitioning") + paramMap.count("partitioning.json") > 0;
```
jgates108: I don't think this line needs to be split.

core/modules/css/CssAccess.cc
```diff
@@ -0,0 +1,848 @@
...
+        if (iter != paramMap.end()) {
+            params.subChunks = std::stoi(iter->second);
+        }
+        iter = paramMap.find("partitioning/overlap");
+        if (iter != paramMap.end()) {
+            params.overlap = std::stod(iter->second);
+        }
+    } catch (std::exception const& exc) {
+        throw KeyValueError(tableKey + "/partitioning", "one of the sub-keys is not numeric: " +
+                            std::string(exc.what()));
```
jgates108: Worthy of logging?

core/modules/css/CssAccess.cc
```diff
@@ -0,0 +1,848 @@
...
+    params.match.dirTable2 = paramMap["match/dirTable2"];
+    params.match.dirColName2 = paramMap["match/dirColName2"];
+    params.match.flagColName = paramMap["match/flagColName"];
+    params.partitioning.dirDb = paramMap["partitioning/dirDb"];
+    params.partitioning.dirTable = paramMap["partitioning/dirTable"];
+    params.partitioning.dirColName = paramMap["partitioning/dirColName"];
+    params.partitioning.latColName = paramMap["partitioning/latColName"];
+    params.partitioning.lonColName = paramMap["partitioning/lonColName"];
+    params.partitioning.partitioned =
+            paramMap.count("partitioning") + paramMap.count("partitioning.json") > 0;
```
jgates108: It seems like maybe the subkeys and all the params.x = paramMap["x"] should be part of TableParams. With a getSubkeys() function and a loadFromMap(&map) function. It looks like there's a lot in common between MatchTableParams, TableParams, and the other Params structures, which could be leveraged to get rid of some nearly duplicate code. Might be better to do it in another ticket though.

core/modules/css/CssAccess.h
```diff
@@ -0,0 +1,422 @@
...
+    /**
+     *  Factory method for creating CssAccess instance based on in-memory KV.
+     *  Data will be loaded from a file specified as an argument,
```
jgates108: Ends with comma, and argument to what?

core/modules/css/CssAccess.h
```diff
@@ -0,0 +1,422 @@
...
+     */
+    static std::shared_ptr<CssAccess> makeMemCss(std::string const& mapPath,
+                                                 std::string const& emptyChunkPath);
+
+    /**
+     *  Factory method for creating CssAccess instance based on in-memory KV.
+     *  Data will be loaded from a stream specified as an argument,
+     */
```
jgates108: ditto

core/modules/css/CssAccess.cc
jgates108: Looks like everything but getTableParams() has been deprecated according to the header file, so a lot of the similar code should be deleted at some point?

core/modules/css/CssAccess.h
```diff
@@ -0,0 +1,422 @@
...
+    // subclass CssAccess and expose these methods for testing
+
+    /**
+     * Throws NoSuchDb exception if the given database does not exist.
+     */
+    void _assertDbExists(std::string const& dbName) const;
+
+    /**
+     * Get values of specified sub-keys of a given key. This method knows how to unpack
+     * packed keys. rEturned map has sub-key names as keys, if sub-key is missing then
```
jgates108: rEturned

core/modules/css/CssError.h
```diff
@@ -57,6 +57,9 @@ class CssError : public std::runtime_error {
                 + boost::lexical_cast<std::string>(sqlErr.errNo())
```
jgates108: Header copyright should be updated to 2015

core/modules/css/KvInterfaceImplMem.cc
```diff
@@ -92,14 +94,26 @@ KvInterfaceImplMem::KvInterfaceImplMem(string const& filename) {
...
-        throw KeyExistsError(key);
+std::string
+KvInterfaceImplMem::create(string const& key, string const& value, bool unique) {
+    LOGF_DEBUG("create(%1%, %2%, unique=%3%)" % key % value % int(unique));
+
+    string path = key;
+    if (unique) {
+        // append unique suffix, in-memory KVI is not meant for large-scale
+        // storage, so we can do dumb brute-force loop
+        int seq = 0;
```
jgates108: Would it make sense to use a random number here?

core/modules/css/CssError.h
```diff
@@ -57,6 +57,9 @@ class CssError : public std::runtime_error {
...
                 + ") "
                 + sqlErr.errMsg()) {}
+
+    // this name will be used to find corresponding Python exception type
+    virtual std::string typeName() const { return "CssError"; }
```
jgates108: Add override to to all the child class typeName functions.

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -202,12 +203,65 @@ KvInterfaceImplMySql::_findParentId(std::string const& childKvKey, bool* hasPare
...
+            ++ uniqueId;
+            std::ostringstream str;
+            str << std::setfill('0') << std::setw(10) << uniqueId;
+            path = key + str.str();
+            try {
+                _create(path, value, false, transaction);
+                break;
+            } catch (KeyExistsError const& exc) {
+                // exists already, try next
+            }
```
jgates108: This will work, but picking a random 10 digit integer and seeing if the full key name is in use might be faster/simpler. I guess it really depends on how many unique identifiers are needed for one name, of which I don't really have any idea.

core/modules/css/testCssAccess.cc
```diff
@@ -0,0 +1,644 @@
+// -*- LSST-C++ -*-
+
+/*
+ * See COPYRIGHT file at the top of the source tree
+ */
+
```
jgates108: I thought copyright year still had to be put at the top of the file?

core/modules/css/CssAccess.cc
jbecla: Not INFO. Debug is better. Info should print what you'd normally want to show to users by default without turning it on. That is my understanding anyway. We are abusing INFO in our code.

core/modules/css/CssAccess.h
andy-slac: I can why it can be attractive but it's not very common convention. I hope with `auto` the need for  these typedefs will be reduced.

core/modules/css/CssAccess.cc
andy-slac: I agree, by default (and default usually means INFO) at this level there should not be anything printed at all. Higher level code may selectively decide to print something with info but low-level packages should be silent.

core/modules/css/CssAccess.cc
andy-slac: OK.

core/modules/css/CssAccess.cc
andy-slac: No, it's purely debugging stuff, "info" message comes in the exception.

core/modules/css/CssAccess.cc
andy-slac: What mattes here is that key is "READY" or not, "DOES_NOT_EXIST" is the same as not "READY", I could use any other default value there.

core/modules/css/CssAccess.cc
andy-slac: ok

core/modules/css/CssAccess.cc
andy-slac: I added ERROR message in addition to exception (indeed exception does not have full info, logging should help)

core/modules/css/CssAccess.cc
andy-slac: I prefer keep that sort of code out of TableParams, this is just one way of constructing TableParms which is CssAccess-specific, I do not want to make TableParams depending on the logic inside CssAccess.

YEs, I hope we can reduce duplication, but it needs some refactoring and small redesign on czar side.

core/modules/css/CssAccess.h
andy-slac: Argument to a method :smile_cat: Fixed that, thanks.

core/modules/css/CssAccess.h
andy-slac: fixEd :smile:

core/modules/css/CssError.h
andy-slac: ok

core/modules/css/KvInterfaceImplMem.cc
andy-slac: Couple of reasons not to:
- I do not like random numbers in general (except in simulation)
- we need unique numbers, random is not the same as unique
I'd like to keep that unique number sequential (how it was done in zookeeper) or at least monotonic. It costs one more database query but we should not need to do it very frequently.

core/modules/css/CssError.h
andy-slac: ok

core/modules/css/KvInterfaceImplMySql.cc
andy-slac: Checking that key is in use may not be the best way - it needs additional query and without proper locking it does not guarantee that you can actually create that key (there may be other concurrent clients trying to do the same).

core/modules/css/testCssAccess.cc
andy-slac: I can't find that in RFC-45.

core/modules/css/KvInterfaceImplMem.cc
ktlim: I think the point here (and below?) is that random numbers are less likely to collide than sequential numbers, so they're faster to find a unique value.  (If the collision window is very short, then it doesn't matter, of course.)  There's no impact on testing or anything else since you can't know what unique value you would get anyway.

core/modules/css/CssAccess.cc
jgates108: I guess I've done differently in the past where we stopped recording debug level info in production but still wanted enough information coming through in the logs so we'd have some idea where to look when things break.

core/modules/css/KvInterfaceImplMem.cc
jgates108: That's what I was getting at. If seq is a random number between 1 and a million, there will pretty much never be a collision, and if there is, increment should get a safe value quickly. It's more complicated below.

PR #137, opened 2015-09-28
Tickets/dm 3936

jgates108:


core/modules/qdisp/QueryRequest.h
```diff
@@ -136,6 +136,7 @@ class QueryRequest : public XrdSsiRequest {
     bool _cancelled {false}; ///< true if cancelled, protected by _finishStatusMutex.
 
     std::shared_ptr<QueryRequest> _keepAlive; ///< Used to keep this object alive during race condition.
+    int _jobId {-1}; ///< for debugging only.
```
jbecla: remove?

jgates108: I'd actually like to keep these around until DM-3946 (although that seems to be worker cleanup) just to make sure everything gets cleaned up properly when queries are cancelled. I'll put comments in the code to that effect.

PR #141, opened 2015-10-12
Tickets/dm 3946

jgates108:


core/modules/wbase/MsgProcessor.h
```diff
@@ -38,22 +37,23 @@ namespace proto {
...
-    /// This allows the caller to request work stoppage.
-    virtual R operator()(A1 taskMsg, A2 replyChannel) = 0;
+/** MsgProcessor implementations handle incoming TaskMsg objects by creating a Task to write their
+ * results over a SendChannel
+ */
+struct MsgProcessor {
+	virtual ~MsgProcessor() {}
+    /// @return a pointer to the Task so it can be cancelled or tracked.
+	virtual std::shared_ptr<Task> processMsg(std::shared_ptr<proto::TaskMsg> const& taskMsg,
+	                                         std::shared_ptr<SendChannel> const& replyChannel) = 0;
```
andy-slac: I still see TABs here.

core/modules/wbase/SendChannel.h
```diff
@@ -42,16 +42,17 @@ namespace wbase {
...
-    typedef util::VoidCallable<void> ReleaseFunc;
-    typedef std::shared_ptr<ReleaseFunc> ReleaseFuncPtr;
-    typedef long long Size;
+    using Ptr = std::shared_ptr<SendChannel>;
+    struct ReleaseFunc {
+        using Ptr = std::shared_ptr<ReleaseFunc>;
+        virtual ~ReleaseFunc() {}
+        virtual void operator()() = 0;
+    };
+    using Size = long long;
```
andy-slac: I think `size_t` is more portable, or `ssize_t` if you want signed type. Or maybe `uint64_t` from <cstdint> if you want 64-bit type.

core/modules/wbase/MsgProcessor.h
jgates108: A few slipped through. I'll hunt them down.

core/modules/wbase/SendChannel.h
```diff
@@ -42,16 +42,17 @@ namespace wbase {
 /// accepts only one call to send bytes, unless the sendStream call is used.
 class SendChannel {
 public:
-    typedef util::VoidCallable<void> ReleaseFunc;
-    typedef std::shared_ptr<ReleaseFunc> ReleaseFuncPtr;
-    typedef long long Size;
+    using Ptr = std::shared_ptr<SendChannel>;
+    struct ReleaseFunc {
```
andy-slac: Is there reason to prefer explicit struct to `util::VoidCallable`? It looks like it is doing the same thing. Should we think about using `std::function` for all of these?

core/modules/wbase/Task.cc
```diff
@@ -87,13 +88,10 @@ Task::ChunkIdGreater::operator()(Task::Ptr const& x, Task::Ptr const& y) {
 ////////////////////////////////////////////////////////////////////////
 // Task
 ////////////////////////////////////////////////////////////////////////
-std::string const
-Task::defaultUser = "qsmaster";
+std::string const Task::defaultUser = "qsmaster";
```
andy-slac: Nice we don't have passwords hard-coded, only user names :scream_cat:

core/modules/wcontrol/Foreman.h
```diff
@@ -50,68 +50,93 @@
...
-/// Factory function for Foreman that hooks in a scheduler
-Foreman::Ptr
-newForeman(Foreman::Scheduler::Ptr s);
+/// Foreman is a pooling thread manager that is pluggable with different scheduling objects.
+///
+class Foreman : public wbase::MsgProcessor{
+public:
+	using Ptr = std::shared_ptr<Foreman>;
+	static Foreman::Ptr newForeman(Scheduler::Ptr const& s);
+    Foreman(Scheduler::Ptr s);
```
andy-slac: `const&`?

core/modules/wbase/SendChannel.h
jgates108: I'll need to look into it more, but it can be tough to track back std::function or util::VoidCallablke<void> back to their source. It's pretty easy to find all the children of ReleaseFunc, see what they do, fix all of them, etc. I think it makes the code easier to read and we don't need generic capabilities here.

core/modules/wcontrol/Foreman.h
```diff
@@ -50,68 +50,93 @@
...
+    friend class RunnerMgr;
+
+protected:
+    void _startRunner(wbase::Task::Ptr const& t);
+
+    std::shared_ptr<wdb::ChunkResourceMgr> _chunkResourceMgr;
+    std::mutex _mutex;
+    Scheduler::Ptr _scheduler;
+    std::unique_ptr<RunnerMgr> _rManager;
+    LOG_LOGGER _log {LOG_GET("Foreman")};
```
andy-slac: Should it be `static`, or even better remove it from header and hide in anonymous namespace in .cc file?

core/modules/wcontrol/RunnerMgr.cc
```diff
@@ -0,0 +1,175 @@
...
+        std::for_each(tq->begin()+1, tq->end(), StartTaskF(_f));
+    }
+    return tq->front();
+}
+
+LOG_LOGGER RunnerMgr::getLog() {
+    return _f._log;
+}
+/// matchHash: helper functor that matches queries by hash.
+class matchHash {
```
andy-slac: Is this class used anywhere?

core/modules/wcontrol/RunnerMgr.h
```diff
@@ -0,0 +1,72 @@
...
+
+namespace lsst {
+namespace qserv {
+namespace wcontrol {
+
+/// Run an event driven thread that is fed by the Foreman's Scheduler.
+class Runner : public std::enable_shared_from_this<Runner> {
+public:
+	using Ptr = std::shared_ptr<Runner>;
+    Runner(RunnerMgr& rm, wbase::Task::Ptr firstTask);
```
andy-slac: `Ptr const&`?

core/modules/wcontrol/RunnerMgr.h
```diff
@@ -0,0 +1,72 @@
...
+};
+
+/// Keep track of the existing Runner objects for the Foreman.
+///
+class RunnerMgr {
+public:
+    RunnerMgr(Foreman& f);
+    RunnerMgr(RunnerMgr&) = delete;
+    RunnerMgr operator=(RunnerMgr&) = delete;
+    void registerRunner(Runner::Ptr const& r, wbase::Task::Ptr t);
```
andy-slac: Why is one arg const ref and other is not? Same for all Task::Ptr below.

core/modules/wdb/ChunkResource.cc
```diff
@@ -608,14 +608,11 @@ class ChunkResourceMgr::Impl : public ChunkResourceMgr {
 ////////////////////////////////////////////////////////////////////////
 // ChunkResourceMgr
 ////////////////////////////////////////////////////////////////////////
-std::shared_ptr<ChunkResourceMgr>
-ChunkResourceMgr::newMgr(mysql::MySqlConfig const& c) {
+ChunkResourceMgr::Ptr ChunkResourceMgr::newMgr(mysql::MySqlConfig const& c) {
     return std::shared_ptr<ChunkResourceMgr>(new Impl(c));
```
andy-slac: make_shared?

core/modules/wdb/ChunkResource.cc
```diff
@@ -608,14 +608,11 @@ class ChunkResourceMgr::Impl : public ChunkResourceMgr {
...
-
 }
 
-std::shared_ptr<ChunkResourceMgr>
-ChunkResourceMgr::newFakeMgr() {
+ChunkResourceMgr::Ptr ChunkResourceMgr::newFakeMgr() {
     return std::shared_ptr<ChunkResourceMgr>(new Impl());
```
andy-slac: here too.

core/modules/wdb/QueryRunner.cc
```diff
@@ -63,136 +64,87 @@
...
+        Release(wbase::Task::Ptr t, wbase::TaskQueryRunner *tqr) : _t{t}, _tqr{tqr} {}
+        ~Release() { _t->freeTaskQueryRunner(_tqr); }
+    private:
+        wbase::Task::Ptr _t;
+        wbase::TaskQueryRunner *_tqr;
+    };
+    Release release(_task, this);
+
+    if (_task->getCancelled()) {
+        LOGF(_log, LOG_LVL_DEBUG, "runQuery, task was cancelled before it started." % _task->dbName);
```
andy-slac: `%1%` is missing in format string

core/modules/wdb/QueryRunner.cc
```diff
@@ -278,17 +224,15 @@ bool QueryAction::Impl::_fillRows(MYSQL_RES* result, int numFields) {
...
-/** Transmit result data with its header.
- * If 'last' is true, this is the last message in the result set
- * and flags are set accordingly.
- */
-void QueryAction::Impl::_transmit(bool last) {
+/// Transmit result data with its header.
+/// If 'last' is true, this is the last message in the result set
+/// and flags are set accordingly.
+void QueryRunner::_transmit(bool last) {
     LOGF_DEBUG("_transmit last=%1%" % last);
```
andy-slac: make all logging to go consistently to the same logger?

core/modules/wdb/QueryRunner.cc
```diff
@@ -278,17 +224,15 @@ bool QueryAction::Impl::_fillRows(MYSQL_RES* result, int numFields) {
...
     std::string resultString;
     _result->set_continues(!last);
     if (!_multiError.empty()) {
-
-        std::string chunkId = boost::lexical_cast<std::string>((*_msg).chunkid());
+        std::string chunkId = boost::lexical_cast<std::string>(_task->msg->chunkid());
```
andy-slac: I think we can start switching to `std::to_string` from boost stuff.

core/modules/wdb/QueryRunner.h
```diff
@@ -0,0 +1,115 @@
...
+};
+
+/// On the worker, run a query related to a Task, writing the results to a table or supplied SendChannel.
+///
+class QueryRunner : public wbase::TaskQueryRunner, public std::enable_shared_from_this<QueryRunner> {
+public:
+    QueryRunner(QueryRunnerArg const& a);
+    // Having more than one copy of this would making tracking its progress difficult.
+    QueryRunner(QueryRunner&) = delete;
+    QueryRunner operator=(QueryRunner&) = delete;
```
andy-slac: I think const is missing in the argument in the above two lines

core/modules/wdb/QueryRunner.h
```diff
@@ -0,0 +1,115 @@
...
+    MYSQL_RES* _primeResult(std::string const& query); ///< Obtain a result handle for a query.
+
+    bool _fillRows(MYSQL_RES* result, int numFields);
+    void _fillSchema(MYSQL_RES* result);
+    void _initMsgs();
+    void _initMsg();
+    void _transmit(bool last);
+    void _transmitHeader(std::string& msg);
+
+    LOG_LOGGER _log;
```
andy-slac: static or move to .cc?

core/modules/wcontrol/RunnerMgr.h
jgates108: There were actually a few issues with Runner and RunnerMgr that all stemmed from 

    std::thread thrd{Runner{*_rManager, t}};

Runner had to be copyable for the above to work, which caused problems. The following allows Runner to be non-copyable and allows its costructor to take `Task::Ptr const& t` as an argument, but if you use `wbase::Task::Ptr const& t` as an argument in the lambda function, you can't use const& in the Runner constructor.

        auto f = [this](wbase::Task::Ptr t) {
            std::unique_ptr<Runner> rp{new Runner{*(this->_rManager), t}};
            (*rp)();
        };
        std::thread thrd{f, t};
        thrd.detach();

core/modules/wdb/ChunkResource.cc
jgates108: Runs into the private constructor problem.

core/modules/wdb/QueryRunner.cc
jgates108: Thanks

core/modules/wbase/Base.cc
```diff
@@ -31,17 +31,15 @@
 #include "wbase/Base.h"
 
 // System headers
-#ifdef __SUNPRO_CC
-#include <sys/md5.h>
-#else // Linux?
 #include <openssl/md5.h>
```
fritzm: alphabetize?

core/modules/wbase/Base.cc
```diff
@@ -373,4 +319,4 @@ void StringBuffer2::_setSize(unsigned size) {
     _bufferSize = size;
 }
 
-}}} // namespace lsst::qserv::wbase
+}}} // namespace
```
fritzm: I like listing the namespaces closed -- do you not?

core/modules/wbase/MsgProcessor.h
```diff
@@ -28,7 +28,6 @@
 #include <memory>
 
 // Qserv headers
```
fritzm: remove?

core/modules/wbase/MsgProcessor.h
```diff
@@ -38,22 +37,23 @@ namespace proto {
...
-                                  std::shared_ptr<proto::TaskMsg>,
-                                  std::shared_ptr<SendChannel> > {
-public:
-    /// @return a cancellation function.
-    /// This allows the caller to request work stoppage.
-    virtual R operator()(A1 taskMsg, A2 replyChannel) = 0;
+/** MsgProcessor implementations handle incoming TaskMsg objects by creating a Task to write their
+ * results over a SendChannel
+ */
+struct MsgProcessor {
```
fritzm: Couldn't this whole class just be a std::function typedef now?

core/modules/wbase/SendChannel.cc
```diff
@@ -62,7 +62,7 @@ class NopChannel : public SendChannel {
     }
 };
 
-std::shared_ptr<SendChannel> SendChannel::newNopChannel() {
+SendChannel::Ptr SendChannel::newNopChannel() {
     std::shared_ptr<NopChannel> n = std::make_shared<NopChannel>();
```
fritzm: maybe just a direct return, consistent with newStringChannel below?

core/modules/wbase/SendChannel.h
fritzm: I'm +1 for std::function instead of derive-and-specialize functors, FWIW.

core/modules/wbase/Task.h
```diff
@@ -50,6 +50,28 @@ namespace lsst {
...
+    virtual bool runQuery()=0;
+    virtual void cancel()=0;
+};
+
+class Task;
+
+/** Base class for scheduling Tasks.
+ * Allows the scheduler to take appropriate action when a task is cancelled.
+ */
+class TaskScheduler {
```
fritzm: again could be just a std::function I think if you buy into that.  Otherwise, you've used struct for some of these elsewhere, so if we're keeping them we should use struct or class consistently?

core/modules/wbase/Task.h
```diff
@@ -75,8 +97,8 @@ struct Task {
         bool operator()(Ptr const& x, Ptr const& y);
     };
 
-    explicit Task() : entryTime(0), _poisoned(false) {}
-    explicit Task(TaskMsgPtr t, std::shared_ptr<SendChannel> sc);
+    explicit Task() {}
```
fritzm: "explicit" doesn't mean anything except for single-argument constructors?

core/modules/wbase/Task.h
```diff
@@ -63,10 +85,10 @@ struct Task {
...
-    typedef std::shared_ptr<Task> Ptr;
-    typedef proto::TaskMsg_Fragment Fragment;
-    typedef std::shared_ptr<Fragment> FragmentPtr;
-    typedef std::shared_ptr<proto::TaskMsg> TaskMsgPtr;
+    using Ptr =  std::shared_ptr<Task>;
+    using Fragment = proto::TaskMsg_Fragment;
+    using FragmentPtr = std::shared_ptr<Fragment>;
+    using TaskMsgPtr = std::shared_ptr<proto::TaskMsg>;
 
     struct ChunkEqual {
```
fritzm: Hmm, how are these functors used?  Should they really be part of the public Task interface?   Are they good candidates to move to e.g. anon. namespace of some implementation file?

core/modules/wbase/Task.h
```diff
@@ -85,21 +107,24 @@ struct Task {
...
-    std::mutex _mutex; // Used for handling poison
-    std::shared_ptr<util::VoidCallable<void> > _poisonFunc;
-    bool _poisoned; ///< To prevent multiple-poisonings
+    std::atomic<bool> _cancelled{false};
+    TaskQueryRunner::Ptr _taskQueryRunner;
+    std::weak_ptr<TaskScheduler> _taskScheduler;
 };
-typedef std::deque<Task::Ptr> TaskQueue;
-typedef std::shared_ptr<TaskQueue> TaskQueuePtr;
+using TaskQueue =  std::deque<Task::Ptr>;
```
fritzm: Maybe move these inside class def as Task::Queue and Task::QueuePtr?

core/modules/wbase/Task.h
```diff
@@ -85,21 +107,24 @@ struct Task {
...
     std::string dbName; ///< dominant db
     std::string user; ///< Incoming username
-    time_t entryTime; ///< Timestamp for task admission
+    time_t entryTime {0}; ///< Timestamp for task admission
     char timestr[100]; ///< ::ctime_r(&t.entryTime, timestr)
     // Note that manpage spec of "26 bytes"  is insufficient
 
-    void poison(); ///< Call the previously-set poisonFunc
-    void setPoison(std::shared_ptr<util::VoidCallable<void> > poisonFunc);
+    void cancel(); ///< Call the previously-set poisonFunc
```
fritzm: Update comment above to refer to the new "cancel" instead of the old "poison"

core/modules/wbase/Task.h
```diff
@@ -26,13 +26,13 @@
 #define LSST_QSERV_WBASE_TASK_H
 
 // System headers
+#include <atomic>
 #include <deque>
 #include <memory>
 #include <mutex>
 #include <string>
 
 // Qserv headers
```
fritzm: remove?

core/modules/wbase/Task.cc
```diff
@@ -102,36 +100,36 @@ Task::Task(Task::TaskMsgPtr t, std::shared_ptr<SendChannel> sc) : entryTime(0) {
...
-            func.swap(_poisonFunc);
-        }
-        _poisoned = true;
+/// Flag the Task as cancelled, try to stop the SQL query, and try to remove it from the schedule.
+void Task::cancel() {
+    if (_cancelled.exchange(true)) {
+        // Was already cancelled.
+        return;
+    }
+    auto qr = _taskQueryRunner;
```
fritzm: auto& maybe?

core/modules/wbase/Task.cc
```diff
@@ -102,36 +100,36 @@ Task::Task(Task::TaskMsgPtr t, std::shared_ptr<SendChannel> sc) : entryTime(0) {
...
+    if (qr != nullptr) {
+        qr->cancel();
     }
-    if(func) {
-        (*func)();
+    TaskScheduler::Ptr sched(_taskScheduler);
```
fritzm: use same syntax as similar block immediately above?

core/modules/wcontrol/Foreman.cc
```diff
@@ -39,353 +38,56 @@
...
-    wbase::TaskQueuePtr _running;
-};
-////////////////////////////////////////////////////////////////////////
-// Foreman factory function
-////////////////////////////////////////////////////////////////////////
-Foreman::Ptr
-newForeman(Foreman::Scheduler::Ptr sched) {
-    ForemanImpl::Ptr fmi = std::make_shared<ForemanImpl>(sched);
+Foreman::Ptr Foreman::newForeman(Scheduler::Ptr const& sched) {
+    Foreman::Ptr fmi = std::make_shared<Foreman>(sched);
```
fritzm: maybe just direct return?

core/modules/wcontrol/Foreman.cc
```diff
@@ -39,353 +38,56 @@
...
     _chunkResourceMgr = wdb::ChunkResourceMgr::newMgr(c);
-
     _rManager.reset(new RunnerMgr(*this));
     assert(s); // Cannot operate without scheduler.
-
 }
-ForemanImpl::~ForemanImpl() {
+Foreman::~Foreman() {
+	LOGF(_log, LOG_LVL_DEBUG, "Foreman::~Foreman()");
     // FIXME: Poison and drain runners.
```
fritzm: s/poison/cancel

core/modules/wcontrol/Foreman.cc
```diff
@@ -39,353 +38,56 @@
...
 }
 
-void
-ForemanImpl::_startRunner(wbase::Task::Ptr t) {
-    std::thread thrd{Runner{*_rManager, t}};
-    thrd.detach();
+/// Create and queue a Task from a TaskMsg and a replyChannel.
+wbase::Task::Ptr Foreman::processMsg(std::shared_ptr<proto::TaskMsg> const& taskMsg,
+                                     std::shared_ptr<wbase::SendChannel> const& replyChannel) {
+    wbase::Task::Ptr t = std::make_shared<wbase::Task>(taskMsg, replyChannel);
```
fritzm: auto?

core/modules/wcontrol/Foreman.cc
```diff
@@ -39,353 +38,56 @@
...
     assert(_scheduler);
-    wbase::TaskQueuePtr newReady;
-    {
-        std::lock_guard<std::mutex> lock(_runnersMutex);
-        newReady = _scheduler->newTaskAct(task, _running);
-    }
+    auto newReady = _rManager->queueTask(task, _scheduler);
     // Perform only what the scheduler requests.
     if(newReady.get() && (newReady->size() > 0)) {
         wbase::TaskQueue::iterator i = newReady->begin();
```
fritzm: maybe use range-based for here?

core/modules/wdb/QueryRunner.cc
```diff
@@ -40,6 +40,7 @@
 // Third-party headers
 #include <boost/lexical_cast.hpp>
 #include <mysql/mysql.h>
+#include <wdb/QueryRunner.h>
```
fritzm: class header should come first in its own section

core/modules/wcontrol/Foreman.cc
jgates108: I expect it will change significantly with the next ticket which deals with queueing on the worker.

core/modules/wbase/Base.cc
jgates108: Main issue for me is that I like to know it's the end of a namespace instead of a class or function. It's easy for the actual names listed in the comment to fall out of sync with the actual names, and the comment duplicates what's in the code instead of explaining the code.

core/modules/wbase/MsgProcessor.h
jgates108: It's now the parent class to Foreman, and it's a direct call to Forman::processMsg instead of creating a functor to do the same thing.

core/modules/wbase/Task.h
jgates108: I'm holding off on these until I get to the queuing ticket.

core/modules/wbase/Task.cc
jgates108: I wanted a copy as _taskQueryRunner can get reset elsewhere. I added a comment.

core/modules/wbase/Task.h
jgates108: Also, these are transplants not new code.

core/modules/wbase/Base.cc
fritzm: The counterpoint here for me would be that the redundant info in the comments gives context to the closing braces far away from the opening braces.  So I don't have to go diving back to the top of whatever file and pick through the pre-namespace content and try to match up any nested braces in between to know their intent.

That said, I don't really care that much.  But let's at least be consistent -- if we decide to change this we should do it everywhere in the source, and I'd recommend a separate big-bang commit for that?

core/modules/wbase/Task.h
fritzm: Cool on future.  Re the transplant argument: ya, but doesn't mean you can't take the opportunity to fix/clean e.g. struct vs. class for consistency as you make a pass through the file (like various other minor consistency and style fixes you've done throughout?)

core/modules/wbase/Base.cc
jbecla: let's keep it how it was please :). We don't create / rename these names often, there is value in keeping the names

PR #143, opened 2015-10-13
fDM-2334: Remove unneeded xrootd oss plugin

abh3:
This requires the latest XRDSsi version. Fritz is working on integrating it.


core/modules/xrdsvc/SsiProvider.cc
```diff
@@ -0,0 +1,148 @@
...
+#include "xrdsvc/XrdName.h"
+
+/******************************************************************************/
+/*                               G l o b a l s                                */
+/******************************************************************************/
+
+XrdSsiProvider *XrdSsiProviderServer = 
+                new  lsst::qserv::xrdsvc::SsiProviderServer;
+
+XrdSsiProvider *XrdSsiProviderLookup = XrdSsiProviderServer;
```
andy-slac: This confuses me a lot. Are these xrootd globals? Is it possible/reasonable to define global symbol in a dynamically-loaded object? Is there a leak here, who is going to destroy XrdSsiProviderServer?

core/modules/xrdsvc/SsiProvider.cc
```diff
@@ -0,0 +1,148 @@
...
+/******************************************************************************/
+/*                                  I n i t                                   */
+/******************************************************************************/
+  
+namespace lsst {
+namespace qserv {
+namespace xrdsvc {
+
+bool SsiProviderServer::Init(XrdSsiLogger* logP,  XrdSsiCluster* clsP, char const* cfgFn,
+          char const*   parms, int argc, char **argv) {
```
andy-slac: After today's debate on too many comments I'm totally not saying that there are too many spaces in the above two lines :)

core/modules/xrdsvc/SsiProvider.cc
```diff
@@ -0,0 +1,148 @@
...
+
+    lsst::qserv::xrdsvc::XrdName x;
+
+// Establish our instance name (many different qservs may be running here).
+//
```
andy-slac: I think the style throughout qserv code is to keep comments indented with code, and it's recommended (as SHOULD) by coding standard.

core/modules/xrdsvc/SsiProvider.cc
```diff
@@ -0,0 +1,148 @@
...
+    _name = x.getName();
+
+// Save the ssi logger as it places messages in anoher file than our log.
```
andy-slac: another

core/modules/xrdsvc/SsiProvider.cc
```diff
@@ -0,0 +1,148 @@
...
+//
+    _chunkInventory.reset(new wpublish::ChunkInventory(_name));
+
+// If we are a data provider (i.e. xrootd) then we need to get the service
+// object. It will print the exported paths. Otherwise, we need to print them
+// here. This is a bit kludgy and should be corrected when we transition to a
+// single shared memory inventory object which should do this by itself.
+//
+    if(clsP && clsP->DataContext()) {
+        _service = std::unique_ptr<SsiService>(new SsiService(logP));
```
andy-slac: `_service.reset()` as above?

core/modules/xrdsvc/SsiProvider.cc
```diff
@@ -0,0 +1,148 @@
...
+
+XrdSsiProvider::rStat SsiProviderServer::QueryResource(char const* rName,
+        char const* contact) {
+
+// Extract db and chunk from path and validate result
+//
+    ResourceUnit ru(rName);
+    if(ru.unitType() != ResourceUnit::DBCHUNK) {
+        // FIXME: Do we need to support /result here?
+        LOGF_INFO("SsiProvider Query %1% invalid" % rName);
```
andy-slac: Should this be logged at DEBUG level instead? I think INFO for us means "normal" output, and I'm not sure that we want a message about every query for normal logging level.

core/modules/xrdsvc/SsiProvider.h
```diff
@@ -41,28 +53,29 @@ class SsiProviderServer : public XrdSsiProvider
...
+                      char const* parms, int argc, char** argv);
 
-    virtual rStat QueryResource(char const*rName, char const* contact=0) {
-        // Not called on this object but part of pure virtual interface,
-        // so we need to provide at least this dummy implementation.
-        return isPresent;
-    }
+    virtual rStat QueryResource(char const* rName, char const* contact=0);
+
+                   SsiProviderServer() : _cmsSsi(0), _logSsi(0) {}
```
andy-slac: I totally agree that we should not care about white space in C++ but this line is totally mis-aligned :)

core/modules/xrdsvc/SsiProvider.h
```diff
@@ -41,28 +53,29 @@ class SsiProviderServer : public XrdSsiProvider
...
-        return true;
-    }
+                      char const* parms, int argc, char** argv);
 
-    virtual rStat QueryResource(char const*rName, char const* contact=0) {
-        // Not called on this object but part of pure virtual interface,
-        // so we need to provide at least this dummy implementation.
-        return isPresent;
-    }
+    virtual rStat QueryResource(char const* rName, char const* contact=0);
```
andy-slac: maybe add `override` to virtual methods?

core/modules/xrdsvc/SsiProvider.h
```diff
@@ -41,28 +53,29 @@ class SsiProviderServer : public XrdSsiProvider
...
+
+                   SsiProviderServer() : _cmsSsi(0), _logSsi(0) {}
+    virtual      ~SsiProviderServer() {}
```
andy-slac: If you want destructor inline then I believe that for all `unique_ptr` members the template types must be complete, otherwise it should fail to compile (if you include this header into a file which has no definition for ChunkInventory). It's better to move destructoor definition to .cc file.

core/modules/xrdsvc/SsiProvider.h
```diff
@@ -24,14 +24,26 @@
 /// implementation. Link this file when building a plugin to be used as
 /// ssi.svclib.
 
+#ifndef LSST_QSERV_XRDSVC_SSIPROVIDER_H
+#define LSST_QSERV_XRDSVC_SSIPROVIDER_H
+
 // System headers
-#include <memory>
```
andy-slac: I think `<memory>` is still useful here, you are using `unique_ptr` below.

core/modules/xrdsvc/SsiProvider.cc
abh3: Sure, I'll change that and follow it in the future. Seems like the rule encourages a lot of leading white space (well, a mandatory 4 leading white spaces everywhere :-)

core/modules/xrdsvc/SsiProvider.cc
abh3: Well, right now we have two log files until we figure out how to combine them into one. This one goes to the ssi log so we know that in that context initialization was started.

core/modules/xrdsvc/SsiProvider.cc
abh3: A bit more context please.

core/modules/xrdsvc/SsiProvider.cc
abh3: I just copied the previous code here. You are right, the present/absent lines should be debug. The invalid line should be logged as it should never happen.

core/modules/xrdsvc/SsiProvider.h
abh3: Good point. Do we actually use this consistently throughout? I don't think so. Perhaps it should be a rule.

core/modules/xrdsvc/SsiProvider.cc
andy-slac: Yes, spaces make code more readable. IT's not really a mandatory 4-space indent, I think namespace-level code has no indent, so if you comment code at namespace level there will be no indent. Inside normal methods/classes it is indeed multiples of 4 spaces.

core/modules/xrdsvc/SsiProvider.cc
andy-slac: For worker side (xrootd side) we have special log4cxx appender configured which redirects LOG_SMTHG() output to xrootd log (I'd prefer it other way round but xrootd logging is not message-oriented so it's hard to do). SO in principle for workers we should have just one log file. Or do I miss something else?

core/modules/xrdsvc/SsiProvider.cc
andy-slac: few lines above there is `_chunkInventory.reset(...)` method used to change the value of unique_ptr, and here there is an assignment operator from temporary which does the same thing . For consistency I'd use one or another in the same method, `reset()` is shorter and more readable, so I like this better:

     _service.reset(new SsiService(logP));

core/modules/xrdsvc/SsiProvider.h
andy-slac: Yes, :+1: for a rule, it's just that we started with complete C++11 support only few weeks ago so it's not very widespread yet. `override` is extremely useful tool and should be used everywhere.

core/modules/xrdsvc/SsiProvider.cc
abh3: Ah, but that still violates the rules in my view because the whole point 
here is that code is not indented relative to the comment. In my view, the 
comment is an "if" and the code that follows is the subject of the 
contextual "if" which makes you naturally indent code relative to the 
comment. That is not allowed; making comments take on the same importance 
as the code. That's rarely the case for no other reason is that the 
comment may be outdated. Comments should be treated as subjective 
statements on what is actually happening. Hence, the code should take on 
more importance. The rules don't recognize that. I'm OK with it but I 
think it's not very realistic.

Andy

core/modules/xrdsvc/SsiProvider.cc
abh3: Well, we can change that. All we need to do is come up with exactly how 
you want log messages to be routed and that can be built into the XRootD 
infrastructure. You are not the only person that has made that point. What 
no one has wanted to do is spend time to come up with the "rules".

Andy

core/modules/xrdsvc/SsiProvider.h
abh3: Agreed!

core/modules/xrdsvc/SsiProvider.cc
andy-slac: I can give you my own view on how it should be done. This is not good place to discuss it, can you open a separate ticket for it?

core/modules/xrdsvc/SsiProvider.cc
```diff
@@ -0,0 +1,160 @@
...
+                             char const*   cfgFn, char const*    parms,
+                             int           argc,  char**         argv) {
+
+    lsst::qserv::xrdsvc::XrdName x;
+
+    // Establish our instance name (many different qservs may be running here).
+    //
+    _name = x.getName();
+
+    // Save the ssi logger as it places messages in anoher file than our log.
```
andy-slac: another instead of anoher?

core/modules/xrdsvc/SsiProvider.cc
```diff
@@ -0,0 +1,160 @@
...
+    //
+    _chunkInventory.reset(new wpublish::ChunkInventory(_name));
+
+    // If we are a data provider (i.e. xrootd) then we need to get the service
+    // object. It will print the exported paths. Otherwise, we need to print
+    // them here. This is kludgy and should be corrected when we transition to a
+    // single shared memory inventory object which should do this by itself.
+    //
+    if(clsP && clsP->DataContext()) _service.reset(new SsiService(logP));
+      else {
```
andy-slac: The indentation is confusing here, I'd make it just:

    if(clsP && clsP->DataContext()) {
        _service.reset(new SsiService(logP));
    } else {

PR #144, opened 2015-10-14
Replace qservAdmin.py use with CssAccess (DM-3574)

andy-slac:
Removed qservAdmin and Python kvInterface and replaced it with SWIG-wrapped CssAccess.

Refactored czar's UserQueryFactory to not make snapshots and instantiate CSS in C++ code, thus removing need for SWIG wrapping on czar side.

Removed all dependencies on zookeeper and kazoo.


admin/bin/qserv-admin.py
```diff
@@ -44,9 +44,39 @@
...
+        ToolError.__init__(self, "Missing parameter: " + str(parm))
+
+class ExcMissingConfig(ToolError):
+    def __init__(self, config):
+        ToolError.__init__(self, "Config file not found: " + str(config))
+
+class ExcUnreadableConfig(ToolError):
+    def __init__(self, config):
+        ToolError.__init__(self, "Can't access the config file: " + str(config))
+
```
smonkewitz: The LSST coding standard says that exception class names should end in Error. I don't see a good reason to introduce a new naming convention here, but maybe there is one. Also, if these class are local, consider having the names reflect that (i.e. leading underscore)?

admin/bin/qserv-admin.py
```diff
@@ -154,7 +186,7 @@ def receiveCommands(self):
         cmd = ''
         prompt = self._prompt
         while True:
-            line = raw_input(prompt).decode("utf-8").strip()
+            line = raw_input(prompt).strip()
```
smonkewitz: I don't think this line or its previous version is correct in Python 2 (consider what happens when grabbing input in a non-ASCII compatible encoding). I'm not that experienced with string encoding, but I think it might be more correct to say something like:
```
line = raw_input().decode(sys.stdin.encoding or locale.getpreferredencoding()).encode('utf-8')
```

admin/bin/qserv-admin.py
```diff
@@ -207,65 +238,62 @@ def _parseCreateDatabase(self, tokens):
...
-                   "Invalid argument '%s', should be <dbName>.<tbName>" % dbTbName2)
-            (dbName2, tbName2) = dbTbName2.split('.')
-            self._impl.createTableLike(dbName, tbName, dbName2, tbName2,
-                                       options)
+            if options.get('match', False):
+                matchParams = css.MatchTableParams(options['dirTable1'], options['dirColName1'],
+                                                   options['dirTable2'], options['dirColName2'],
+                                                   options['flagColName'])
+                chunkLevel = 2 if int(options['subChunks']) else 1
+                self._css.createMatchTable(dbName, tbName, options['schema'], matchParams, chunkLevel)
```
smonkewitz: Match tables are not sub-chunked, so I'm not sure it makes sense to talk about a chunk level here (although I guess it also doesn't hurt).

admin/bin/qserv-admin.py
andy-slac: OK, will fix.

admin/bin/qserv-admin.py
```diff
@@ -414,28 +446,17 @@ def _parseKVList(self, tokens, possibleKeys=None):
...
-                   "Invalid argument '%s', should be <key>=<value>" % opt)
+                raise ExcIllegalCommand("Invalid argument '%s', should be <key>=<value>" % opt)
             key, value = opt.split('=', 1)
 
             if possibleKeys is not None and key not in possibleKeys:
-                raise QservAdminException(QservAdminException.BAD_CMD,
-                   "Unrecognized option key '%s', possible keys: %s" % \
-                   (key, ' '.join(possibleKeys)))
+                raise ExcIllegalCommand("Unrecognized option key '%s', possible keys: %s" % \
+                                        (key, ' '.join(possibleKeys)))
```
smonkewitz: here and on line 449 - use `str.format()` instead of `%`?

admin/bin/qserv-admin.py
```diff
@@ -485,19 +505,17 @@ def _setDefault(self, opts, key, defaultValue, force=False):
                 "param '" + key + "' not found, will use default: " + str(defaultValue))
             opts[key] = defaultValue
         elif force and opts[key] != defaultValue:
-            raise QservAdminException(
-                QservAdminException.WRONG_PARAM,
-                "Got '%s' expected '%s'" % (opts[key], defaultValue))
+            raise ExcWrongParam("Got '%s' expected '%s'" % (opts[key], defaultValue))
```
smonkewitz: `{}` instead of `%`?

admin/python/lsst/qserv/admin/chunkMapping.py
```diff
@@ -85,9 +86,17 @@ def _getChunks(self):
         if self.css is None:
             return
 
-        for table in self.css.tables(self.database):
+        try:
+            tables = self.css.getTableNames(self.database, False)
+        except css.NoSuchDb:
+            return
+
+        for table in self.css.getTableNames(self.database, False):
```
smonkewitz: Can't you reuse `tables` instead of asking the CSS for table names twice?

admin/bin/qserv-admin.py
andy-slac: The unicode problem is somewhat bigger that I'm prepared to handle here. I removed `decode()` because new CSS interface is SWIGed C++ code and it does not handle unicode at all. Previous implementation was using kazoo which required unicode. For now we'll be passing bytes directly to C++ and hope that no one uses anything outside ASCII and no fancy encodings like utf16. I'm not sure how it's going to work for Python3 (I mean SWIG).

admin/bin/qserv-admin.py
andy-slac: Are they always chunked? I should make a cleaner API if sub-chunking does not make sense.

admin/bin/qserv-admin.py
andy-slac: OK, will change everywhere in a file

admin/bin/qserv-admin.py
smonkewitz: As far as I can tell, the initial decode is bogus regardless of what goes on in the C++ layer - it can cause lookup failures for command names in the python layer. My suggestion aims to take care of arbitrary user encodings by first decoding to a unicode instance, then encoding back to a UTF-8 `str`, which is ASCII compatible. 

If you don't want to deal with any of this now, maybe file an issue to fix this so we don't forget?

admin/python/lsst/qserv/admin/chunkMapping.py
andy-slac: Yes, I guess this was the intent, lost to sloppy editing :see_no_evil: Thanks

admin/python/lsst/qserv/admin/dataLoader.py
```diff
@@ -684,17 +671,33 @@ def _updateCss(self, database, table):
...
-
-        self._log.info('Creating table CSS info')
-        self.css.createTable(database, table, options)
+        schema = self._schemaForCSS(database, table)
+
+        if options.get('match', False):
+            matchParams = css.MatchTableParams(options['dirTable1'], options['dirColName1'],
+                                               options['dirTable2'], options['dirColName2'],
+                                               options['flagColName'])
+            chunkLevel = 2 if options['subChunks'] else 1
```
smonkewitz: Same comment regarding match tables and sub-chunking as in qserv-admin.py. This whole code block also looks very similar to the code I commented on there - not sure it's worth it, but maybe factor into a helper function?

admin/python/lsst/qserv/admin/dataLoader.py
```diff
@@ -709,7 +712,7 @@ def _schemaForCSS(self, database, table):
         schema = self.czarWmgr.tableSchema(database, table)
         # strip CREATE TABLE
         i = schema.find('(')
-        return schema[i:]
+        return schema[i:].encode('utf_8')
```
smonkewitz: OK, now I'm confused :smile:. Why are you encoding to a `str` here?

admin/bin/qserv-admin.py
andy-slac: I'll open a ticket to remind us to do it later. I think "right" approach to unicode is a to always handle strings in unicode internally and only encode them to pass to APIs that accept bytes. Question is should CssAccess accept bytes only or should it be better accept unicode.

admin/python/lsst/qserv/admin/dataLoader.py
andy-slac: Because it's a big unicode mess :cry: CssAccess cannot accept unicode, but wmgr (being a web service client) returns unicode strings, so this string needs to be encoded at some point.  I could do it at a point where I call CssAccess or here. No other strings in the module are unicode, so I just say I better drop unicode completely and do it here.

core/modules/css/CssAccess.cc
```diff
@@ -53,31 +68,106 @@ namespace qserv {
...
+        mysqlConfig.hostname = _mapGet(config, "hostname");
+        mysqlConfig.username = _mapGet(config, "username");
+        mysqlConfig.password = _mapGet(config, "password");
+        mysqlConfig.dbName = _mapGet(config, "database");
+        mysqlConfig.socket = _mapGet(config, "socket");
+        auto portStr = _mapGet(config, "port", "");
+        if (portStr.empty()) portStr = "0";
+        try {
+            // tried to use std::stoi() here but it returns OK for strings like "0xFSCK"
+            mysqlConfig.port = boost::lexical_cast<unsigned>(portStr);
```
smonkewitz: I'm totally OK with deferring this for now, but `std::stoi` can be made to tell you when it fails to consume the entire input string.

core/modules/css/KvInterfaceImplMem.cc
```diff
@@ -194,12 +206,35 @@ KvInterfaceImplMem::getChildren(string const& key) {
...
     if ( ! exists(key) ) {
         throw NoSuchKey(key);
     }
     _kvMap.erase(key);
 }
 
+std::string KvInterfaceImplMem::dumpKV() {
+    std::string result;
+    for (auto& pair: _kvMap) {
+        if (not result.empty()) result += '\n';
```
smonkewitz: Aren't we supposed to use `!` in C++?

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -356,11 +364,11 @@ KvInterfaceImplMySql::getMany(std::vector<std::string> const& keys) {
     std::string query = "SELECT kvKey, kvVal FROM kvData WHERE kvKey IN (";
     bool first = true;
     for (auto& key: keys) {
+        if (not first) query += ", ";
```
smonkewitz: `!`?

core/modules/css/CssAccess.cc
andy-slac: I think the way to detect it in call to `stoi()` is something like this:

    size_t pos;
    auto value = std::stoi(str, &pos);
    if (pos != str.size()) {
        // error
    }

Which I do not particularly like, it's not much easier to use than good old `strtol()` (but maybe I'm too used to boost stuff).

core/modules/css/KvInterfaceImplMem.cc
andy-slac: I thought I was writing Python :) 
We have discussed that already some time ago, I think nobody had violently objected using "Python-like" operators. And I'm not alone doing it actually. Does it make it harder to read?

admin/bin/qserv-admin.py
smonkewitz: They are always chunked.

To expand on this a bit: you can of course have an unpartitioned table that records some join result, which is a "match" table in some sense. But there is no need to distinguish amongst different types of unpartitioned tables since those can be ignored by the query rewriting logic.

core/modules/css/CssAccess.cc
smonkewitz: Yup. It's easy to provide a wrapper that hides this annoyance, but I don't think it's worth it unless/until we get rid of boost completely.

core/modules/css/KvInterfaceImplMem.cc
smonkewitz: It's jarring (this is the only C++ code base where I've seen not used), but I can't claim it's harder to read.

smonkewitz:
OK, I think that's it from me. Nice work - I especially appreciate the attention to detail in removing references to zookeeper from comments.

admin/bin/qserv-admin.py
andy-slac: OK, thanks for explanation. I'll then remove chunkLevel parameter from createMatchTable() completely and assume it's always partitioned.

PR #150, opened 2015-11-02
Tickets/dm 3945

jgates108:


core/modules/util/Command.h
```diff
@@ -0,0 +1,91 @@
...
+    std::condition_variable _trCV;
+};
+
+/// Base class for commands. Can be used with functions as is or
+/// as a base class when data is needed.
+class Command {
+public:
+    using Ptr = std::shared_ptr<Command>;
+    Command() {};
+    Command(std::function<void(Command*)> func) : _func{func} {}
```
fritzm: Is the "Command *" argument useful/necessary?  Since there is no data and not virtual function API on the base class, a callee would have to downcast this to something more specific to make use of it in any case.  Instead, if necessary, a more strongly typed pointer could be closed over in a lambda or std::bind and the use could have type safety and avoid the cast?

I like the idea of having the default implementation of a glue class like this be a std::function version -- kind of the best of both.

core/modules/util/Command.h
```diff
@@ -0,0 +1,91 @@
...
+#include <mutex>
+
+
+namespace lsst {
+namespace qserv {
+namespace util {
+
+/// Tracker provides an interface for indicating an action is complete.
+///
+class Tracker {
```
fritzm: Where is the implementation of this class?  I would expect to find it in Command.cpp, but there doesn't seem to be that file?

jgates108: Oops. It's in EventThread.cpp. I'll move it to Command.cpp.

core/modules/util/Command.h
jgates108: That's left over from a failed experiment. No point in keeping it around.

core/modules/xrdsvc/SsiService.cc
```diff
@@ -79,12 +80,17 @@ SsiService::SsiService(XrdSsiLogger* log) {
...
     _foreman = wcontrol::Foreman::newForeman(
-        std::make_shared<wsched::BlendScheduler>(
-            std::make_shared<wsched::GroupScheduler>(),
-            std::make_shared<wsched::ScanScheduler>()
-        )
-    );
+            std::make_shared<wsched::BlendScheduler>(
+                std::make_shared<wsched::GroupScheduler>(20, 10),
+                std::make_shared<wsched::ScanScheduler>(20)),
+            poolSize);
```
jgates108: Something to think about here is that, as opposed to just controlling how many threads are running, we can easily change the group and scan schedulers to limit how many chunks are being accessed at least in the sense of Tasks being associated with a chunk. The scan scheduler does one chunk at a time. The group scheduler tries to group up Tasks on the same chunk. It would be pretty easy to change the group scheduler to only have X chunks open at a time.

core/modules/util/EventThread.h
```diff
@@ -0,0 +1,195 @@
...
+        };
+        std::shared_ptr<MsgEnd> cmd = std::make_shared<MsgEnd>();
+        queCmd(cmd);
+    }
+
+protected:
+    virtual void startup() {};  ///< Things to do before handling commands
+    virtual void finishup() {}; ///< things to do when the thread is closing down.
+
+    CommandQueue::Ptr _q{std::make_shared<CommandQueue>()}; ///< Queue of commands.
```
fritzm: Is there a reason that _q is a shared_ptr instead of just a directly embedded member?

jgates108: _q can be set in the constructor. For ThreadPool, all the threads share the same CommandQueue object. All the schedulers are now based on CommandQueue and passed to the worker ThreadPool when it is made in the SsiService constructor.

fritzm: Ah, I see, thx!

core/modules/util/EventThread.h
```diff
@@ -0,0 +1,195 @@
...
+     : EventThread{q}, _threadPool{threadPool} {}
+    ~PoolEventThread();
+
+protected:
+     void finishup() override;
+     std::shared_ptr<ThreadPool> _threadPool;
+};
+
+/// ThreadPool is a variable size pool of threads all fed by the same CommandQueue.
+/// Growing the pool is simple, shrinking the pool is complex. Both operation should
```
fritzm: s/operation/operations

core/modules/util/Command.cc
```diff
@@ -0,0 +1,73 @@
...
+    }
+}
+
+/// If _func is a lambda with a captured shared_ptr to this command,
+/// this function must be called or the lambda will keep this object alive.
+void Command::resetFunc() {
+    setFunc(nullptr);
+}
+
+}}} // namespace
```
fritzm: Didn't we argue for consistency on these close comments on previous jira issue?  :smile:

core/modules/util/EventThread.cc
```diff
@@ -0,0 +1,164 @@
...
+    auto eqTest = [this](){ return _targetThrdCount == _pool.size(); };
+    std::unique_lock<std::mutex> lock(_countMutex);
+    if (millisecs > 0) {
+        _countCV.wait_for(lock, std::chrono::milliseconds(millisecs), eqTest);
+    } else {
+        _countCV.wait(lock, eqTest);
+    }
+}
+
+}}}
```
fritzm: <cough>

core/modules/util/EventThread.h
```diff
@@ -0,0 +1,195 @@
...
+    std::deque<Command::Ptr> _qu{};
+    std::condition_variable  _cv{};
+    std::mutex               _mx{};
+};
+
+/// An event driven thread.
+/// Thread must be started with run(). Stop the thread by calling queEnd().
+class EventThread {
+public:
+    enum { HALT = -1000 };
```
fritzm: Did you prefer "naked" enum rather than "enum class" here on purpose?

core/modules/util/EventThread.cc
```diff
@@ -0,0 +1,164 @@
...
+            cmd->resetFunc(); // Reset _func in case it has a captured Command::Ptr
+        }
+    }
+    finishup();
+}
+
+/// call this to start the thread
+void EventThread::run() {
+    std::thread t{&EventThread::handleCmds, this};
+    _t = move(t);
```
fritzm: prefer std::move spelling for clarity here?

core/modules/util/EventThread.cc
```diff
@@ -0,0 +1,164 @@
...
+    LOGF_DEBUG("PoolEventThread::~PoolEventThread()");
+}
+
+void PoolEventThread::finishup() {
+    //  Need to make sure the thread data doesn't disappear before the thread finishes.
+    PoolEventThread::Ptr pet = _threadPool->release(this);
+    if (pet != nullptr) {
+        auto f = [pet](){ pet->join(); };
+        std::thread t{f};
+        t.detach();
```
fritzm: Haha a detached joiner, slick...  That's sorta just kicking join can down the road though isn't it?  Should we just detach the pool event thread directly instead?

core/modules/util/EventThread.cc
abh3: Hmmm, I though your were not supposed to make std an automatically 
searchable namespace. True?

fritzm: @abh3 I think the rule is "abso not in header files, but you can in inplementation files".  In any case, there's a std::thread right above here, so consistency, etc.?

abh3: OK, I understand that but would prefer consistency everywhere. It's 
incredibly difficult to debug if you don't know where a method was 
defined. Adding std to the mix just complicates things. I'd vote no even 
for implementations. Yeah, I know xroot code violates that but only for 
methods that are common (e.g. cerr etc) that everyone knows come from std.

Andy

core/modules/util/testEventThread.cc
```diff
@@ -0,0 +1,188 @@
...
+        int total{0};
+        void add(int val) { total += val; }
+    };
+
+    /// Queue a sum depending on mutex protection within the EventThread.
+    {
+        EventThread et{};
+        SumUnprotected sum;
+        int total{0};
+        for (int j=1; j<10; j++) {
```
abh3: I have a general problem in this file. Many for loops use hard-coded constants and in many cases they are repeated for the same task. Some the constants seem arbitrary with no explanation. I would suggest defining a const int with the value and explaining what it controls. So, in the future if I want to change it it will change the code everywhere where it needs to change.

core/modules/wcontrol/Foreman.cc
```diff
@@ -38,65 +38,61 @@
...
-void Foreman::_startRunner(wbase::Task::Ptr const& t) {
-    auto f = [this](wbase::Task::Ptr t) {
-        LOGF_DEBUG("Foreman::_startRunner start");
-        RunnerMgr *rm = this->_rManager.get();
-        Runner::Ptr rp = std::make_shared<Runner>(rm, t);
-        (*rp)();
-        LOGF_DEBUG("Foreman::_startRunner end");
+    auto func = [this, task](){
+        proto::TaskMsg const& msg = *task->msg;
+        if(!msg.has_protocol() || msg.protocol() < 2) {
```
abh3: I would use a descriptive const int or enum here. I don't know what protocol you are referring to other than a magic number.

core/modules/wdb/QueryRunner.cc
```diff
@@ -276,7 +279,7 @@ class ChunkResourceRequest {
         : _mgr{mgr}, _msg{msg} {}
 
     ChunkResource getResourceFragment(int i) {
-        wbase::Task::Fragment const& f(_msg.fragment(i));
+        auto const& f(_msg.fragment(i));
```
abh3: I've seen this used several times. I guess I am mystified. The deleted code made some sense because I know the type. The liberal use of auto hides all of that (yes, I am not a fan of over-using auto). Also, I would use something other than single letter objects as it's too easy to loose track of what 'f' really is.

core/modules/wsched/GroupScheduler.cc
```diff
@@ -44,127 +44,109 @@
...
+        _chunkId = task->msg->chunkid();
+    }
+    assert(queTask(task));
 }
 
-void
-GroupScheduler::queueTaskAct(wbase::Task::Ptr incoming) {
-    std::lock_guard<std::mutex> guard(_mutex);
-    _enqueueTask(incoming);
+/// Return true if this GroupQueue accepts this Task.
```
abh3: Given the line below, I think Task should be consistently capitalized one way or the other.

core/modules/wsched/ScanScheduler.cc
```diff
@@ -54,11 +54,10 @@ ChunkDisk* dbgChunkDisk1 = nullptr; ///< A symbol for gdb
 ////////////////////////////////////////////////////////////////////////
 // class ScanScheduler
 ////////////////////////////////////////////////////////////////////////
-ScanScheduler::ScanScheduler()
-    : _maxRunning(32),  // FIXME: set to some multiple of system proc count.
-      _disks(),
-      _logger(LOG_GET(getName())),
-      _mutex()
+ScanScheduler::ScanScheduler(int maxThreads)
+    : _maxThreads{maxThreads},  // FIXME: set to some multiple of system proc count.
```
abh3: The NPROC issue is more complicated than that and it may conflict with max_pid. I'd suggest deciding what the maximum can be at static initialization time and using a (ick) global variable that holds  the absolute maximum you can have. If this is the only place maxThreads makes sense then make the static a member of the class.

core/modules/util/EventThread.cc
jgates108: I'm not really sure why that compiled in the first place. I'm not using 'using std' anywhere. Anyway, it's now std::move.

core/modules/util/EventThread.cc
jgates108: I think I'm probably being overly cautious here.  'pet' is probably the only thing keeping the PoolEventThread object alive after _threadPool->release is called (but it may not be). I don't really expect anything to be called after finishup(), but I'd rather waste a few cpu cycles than segfault as the captured 'pet' will keep the object alive until it joins. I don't expect this code to be called often. I can't move _threadPool->release to the destructor as it will never get called. It's wonky but safe.

core/modules/util/EventThread.cc
jgates108: lol

core/modules/util/EventThread.h
jgates108: The ThreadPool worked out very smoothly until it came time to shut a thread down. Having Command::action() return a specific int is a solution, but I don't like it very much. I've got another idea, that might be a little better. If I made it a class, it would need to be defined before Util::Command, and I really don't like that.

core/modules/wsched/ScanScheduler.cc
jgates108: The maximum number of threads is really decided by the size of the thread pool. maxThreads is meant to be slightly smaller than the size of the thread pool so ScanScheduler doesn't completely crowd out the other schedulers ( really just GroupScheduler so far). I feel this should be set in a configuration file so it can be easily changed for different hardware and we can tweak them to see what gives best performance. Currently, all of these values are set in the SsiService constructor.

core/modules/util/testEventThread.cc
jgates108: The numbers in these unit tests are kind of arbitrary. I'l add some comments explaining them.

core/modules/util/EventThread.cc
fritzm: @abh3 : Yup, I think we're all on the same page here, actually!   Missing off the std:: here was apparently an oversight.  And I think we all also agree that even if using namespace std in an impl, its best to leave off the std:: only for the very common things.

core/modules/util/EventThread.cc
fritzm: Oh, I see.  Maybe a short comment on the motivation here would be good.

core/modules/util/EventThread.h
fritzm: I meant the new C++11 strongly-typed enum instead of a naked int.  They can be forward-declared, too: http://www.cprogramming.com/c++11/c++11-nullptr-strongly-typed-enum-class.html

Also, if the only determination is halt / don't-halt, would just a bool be reasonable?

core/modules/wbase/Task.h
```diff
@@ -68,19 +74,38 @@ class TaskScheduler {
...
+};
+
 /// struct Task defines a query task to be done, containing a TaskMsg
 /// (over-the-wire) additional concrete info related to physical
 /// execution conditions.
 /// Task is non-copyable
 /// Task encapsulates nearly zero logic, aside from:
 /// * constructors
-struct Task {
+struct Task : public util::Command {
```
fritzm: Why "struct" here, and not "class"?  Access is immediately overridden to "public" anyway, and there are private members as well.

core/modules/wbase/Task.h
```diff
@@ -111,13 +137,16 @@ struct Task {
     void setTaskScheduler(TaskScheduler::Ptr const& scheduler) { _taskScheduler = scheduler; }
     friend std::ostream& operator<<(std::ostream& os, Task const& t);
```
fritzm: This operator<< should probably be declared in this file as well at top level?

core/modules/wbase/Task.h
```diff
@@ -68,19 +74,38 @@ class TaskScheduler {
...
+/// forever, dependency on BlendScheduler needs to be re-worked.
+struct IdSet {
+    void add(int id) { std::lock_guard<std::mutex> lock(mx); _ids.insert(id); }
+    void remove(int id) { std::lock_guard<std::mutex> lock(mx); _ids.erase(id); }
+    std::string toString() {
+        std::ostringstream os;
+        std::lock_guard<std::mutex> lock(mx);
+        os << "count=" << _ids.size() << " ";
+        for(auto j: _ids) {
+            os << j << ", ";
```
fritzm: The extra ", " that will appear at the end is the sort of thing that makes me wake up sweating in the middle of the night.

core/modules/wbase/Task.h
```diff
@@ -111,13 +137,16 @@ struct Task {
...
+    static util::Sequential<int> sequence; // for debugging only
+    static IdSet allTSeq; // set of all task sequence numbers that are not complete.
+    int tSeq{-1}; // for debugging only
+
 private:
     std::atomic<bool> _cancelled{false};
     TaskQueryRunner::Ptr _taskQueryRunner;
     std::weak_ptr<TaskScheduler> _taskScheduler;
 };
 using TaskQueue =  std::deque<Task::Ptr>;
```
fritzm: Is this type needed anymore?

core/modules/wbase/Task.h
andy-slac: There is `utils::printable()` in `utils/IterableFormatter.h` which does nice formatting of sequences and maps. I personally only start sweating when I see `toString()` method, this is so Java and I don't speak Java well :smirk_cat:

core/modules/wcontrol/Foreman.h
```diff
@@ -72,38 +74,17 @@ namespace lsst {
...
     virtual void taskCancelled(wbase::Task *task) { return; }
-    virtual void queueTaskAct(wbase::Task::Ptr incoming) = 0;
-    virtual wbase::TaskQueuePtr nopAct(wbase::TaskQueuePtr running) = 0;
-    virtual wbase::TaskQueuePtr newTaskAct(wbase::Task::Ptr incoming,
-                                           wbase::TaskQueuePtr running) = 0;
-    virtual wbase::TaskQueuePtr taskFinishAct(wbase::Task::Ptr finished,
-                                              wbase::TaskQueuePtr running) = 0;
 };
 
 /// Foreman is a pooling thread manager that is pluggable with different scheduling objects.
```
fritzm: Yay!  Forman is very straightforward with this refactor.  This comment is kinda crusty now though: "Foreman is a MsgProcessor that dispatches incoming queries via contained thread pool and scheduler" or some such now?

fritzm: Oh, also, I think the big explanatory comment at the head of the file is also now stale -- would you please revise?

core/modules/wcontrol/Foreman.cc
```diff
@@ -38,65 +38,61 @@
...
     assert(s); // Cannot operate without scheduler.
+
+    LOGF_DEBUG("poolSize=%1%" % poolSize);
+    _pool = util::ThreadPool::newThreadPool(poolSize, _scheduler);
 }
 Foreman::~Foreman() {
     LOGF(_log, LOG_LVL_DEBUG, "Foreman::~Foreman()");
     // FIXME: Cancel and drain runners. It will take significant effort to have xrootd shutdown cleanly
     // and this will never get called until that happens, making this a very low priority item.
     // This should only (get called on shutdown/restart).
```
fritzm: Outdated comment?

fritzm:
Is FifoScheduler.h needed any more?

core/modules/xrdsvc/SsiService.cc
fritzm: Maybe add this observation in a comment in the source?

fritzm:
group/scan/blend schedulers really need some unit tests, unless I missed them...  Would you please file a Jira issue for that so we don't forget?   

We should decide if we want those in place before abh's upcoming shared scan work (for regression/refactor proof) or if we think refactor will be significant if we should just do them later so they don't have to be totally reworked?

fritzm:
Phew, done!  Some really nice work here, John -- thanks!

core/modules/wcontrol/Foreman.cc
jgates108: I don't really know either. This is associated with worker.pb.h which is a generated file. My guess is Result protocol in a comment in proto/worker.proto.

PR #151, opened 2015-11-05
Simplify handling of packed keys (DM-4138)

andy-slac:
Packed keys now have unique names (.packed.json) and they appear on the
same level in CSS tree as the keys the replace. This simplifies
CssAccess code which handles packing/unpacking.

### DIFF
--- a/doc/source/devel/css.rst
+++ b/doc/source/devel/css.rst
@@ -2,48 +2,67 @@
 CSS metadata layout
 ###################
 
-The Central State System (CSS) maintains information about the data
-stored in Qserv. Currently, this consists of the data schema: the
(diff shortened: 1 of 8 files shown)


doc/source/devel/css.rst
```diff
@@ -2,48 +2,67 @@
...
+identically to ``"{"key": "1"}"``.
+
+*************
+CSS interface
+*************
+
+CSS key-value storage is hidden completely behind CSS interface
+(``css/CssAccess`` class). This interface defines all logical operations on
+databases, tables, nodes, etc. and translates those operations into key-value
+structure manipulatons, including packing and upacking of the keys.
```
fjammes: Looks ok for me.

PR #152, opened 2015-11-09
Updated watcher to delete database from chunk registry (DM-4206)

andy-slac:
Watcher now tries to remove database name from chunk registry before
actually dropping the database, this is done for workers only, czar
databases do not have chunk registries.

### DIFF
--- a/admin/python/lsst/qserv/admin/watcherLib.py
+++ b/admin/python/lsst/qserv/admin/watcherLib.py
@@ -133,10 +133,9 @@ class WatcherCss(object):
 
     def getNodes(self):
         """
-        Return the list of nodes (NodeAdmin instances) for all active nodes.
-        Both workers and czars are included in the list.
-
-        @param wmgrSecretFile:  Path to a file with wmgr secret.
+        Return tuple with two lists of nodes (NodeAdmin instances) for all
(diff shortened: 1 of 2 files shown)


admin/python/lsst/qserv/admin/watcherLib.py
```diff
@@ -442,19 +442,32 @@ def dropDb(self, dbName):
...
+                        # Do not restart xrootd
+                        restart = False
+                        wmgr.xrootdUnregisterDb(dbName, restart)
+                        _LOG.info('Removed database %r from chunk registry on node %r', dbName, node.name())
+                    except ServerError as exc:
+                        if exc.code == 409:
+                            _LOG.info('Database %r is not in chunk registry on node %r', dbName, node.name())
+                        else:
+                            raise
                 try:
```
jbecla: What happens if we successfully unregister db, but fail to drop database on all nodes? Perhaps that is for a separate tickets, If there are some flaws like this, let's at least capture in a new story somewhere.

andy-slac: The corresponding database in CSS will me marked with 'FAILED' status. What happens after that still needs to be defined, It's not clear how much auto-recovery we can do in this case. This more or less the same problem as what happens id DROP DATABASE fails in mysql server?

PR #153, opened 2015-11-11
Add support for DROP TABLE (DM-2622)

andy-slac:
Implemented via special subclass of UserQuery returned by
UserQueryFactory.


core/modules/ccontrol/UserQuery.h
```diff
@@ -33,108 +33,52 @@
...
-    std::mutex _killMutex;
-    int _sessionId; ///< External reference number
-    int _sequence; ///< Sequence number for subtask ids
-    std::string _errorExtra; ///< Additional error information
-    mutable std::string _errorExtraCache; ///< Cache so getError can return a ref
+    virtual std::shared_ptr<qdisp::MessageStore> getMessageStore() = 0;
+
 };
 
 }}} // namespace lsst::qserv:ccontrol
```
vaikuntht: The only changes here have been to convert all functions to pure virtual functions. Is the stuff removed not needed anymore or moved to another class? Looks fine to me otherwise.

core/modules/ccontrol/UserQueryDropTable.h
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+     *  @param css:         CSS interface
+     *  @param dbName:      Name of the database where table is
+     *  @param tableName:   Name of the table to drop
+     *  @param resultDbConn: Connection to results database
+     *  @param resultTable: Name of the table for query results
+     *  @param queryMetadata: QMeta interface
+     *  @param qMetaCzarId: Czar ID in QMeta database
+     */
```
vaikuntht: Alignment after ":" for easy readability.

core/modules/ccontrol/UserQueryDropTable.cc
```diff
@@ -0,0 +1,187 @@
...
+        auto statusMap = _css->getTableStatus(_dbName);
+        LOGF(getLogger(), LOG_LVL_DEBUG, "all table status: %s" % util::printable(statusMap));
+        if (statusMap.count(_tableName) != 1) {
+            std::string message = "Unknown table " + _dbName + "." + _tableName;
+            _messageStore->addMessage(-1, 1051, message, MessageSeverity::MSG_ERROR);
+            _qState = ERROR;
+            return;
+        }
+        LOGF(getLogger(), LOG_LVL_DEBUG, "table status: %s" % statusMap[_tableName]);
+        if (statusMap[_tableName] != css::KEY_STATUS_READY) {
```
vaikuntht: Not sure about the difference from the previous if statement, are you checking all tables first and then the particular table that you want to drop here? Wouldn't it be quicker to just check the one you want to drop?

core/modules/ccontrol/UserQueryDropTable.cc
```diff
@@ -0,0 +1,187 @@
...
+    std::string newStatus = css::KEY_STATUS_DROP_PFX + std::to_string(::time(nullptr)) +
+            ":qid=" + std::to_string(qMetaQueryId);
+    LOGF(getLogger(), LOG_LVL_DEBUG, "new table status: %s" % newStatus);
+    try {
+        // TODO: it's better to do it in one atomic operation with
+        // getTableStatus, but CSS API does not have this option yet
+        _css->setTableStatus(_dbName, _tableName, newStatus);
+        _qState = SUCCESS;
+    } catch (css::NoSuchTable const& exc) {
+        // Have it disappeared already?
```
vaikuntht: Have -> Has

core/modules/ccontrol/UserQueryDropTable.cc
```diff
@@ -0,0 +1,187 @@
...
+        _messageStore->addMessage(-1, 1051, message, MessageSeverity::MSG_ERROR);
+        _qState = ERROR;
+    } catch (css::CssError const& exc) {
+        LOGF(getLogger(), LOG_LVL_ERROR, "CSS failure: %s" % exc.what());
+        std::string message = "CSS error: " + std::string(exc.what());
+        _messageStore->addMessage(-1, 1051, message, MessageSeverity::MSG_ERROR);
+        _qState = ERROR;
+    }
+
+    // if failed then update status in QMeta
```
vaikuntht: Does status not need to be updated in QMeta for success?

core/modules/ccontrol/UserQuery.h
```diff
@@ -33,108 +33,52 @@
 
 // System headers
 #include <memory>
-#include <mutex>
 
 // Third-party headers
```
andy-slac: I don't know if there is any specific policy on this, I sometimes leave it there in case I need to add more includes in the future. I don't remember exactly what text I have to put in the comment if I have to add a section and it's annoying to have to search other files for an example.

core/modules/ccontrol/UserQuery.h
andy-slac: Yes, this class was made abstract and there are few implementations in classes `UserQuerySelect` (which is ~identical to old `UserQuery`), `UserQueryDropTable`, and `UserQueryInvalid`

core/modules/ccontrol/UserQueryDropTable.h
andy-slac: OK

core/modules/ccontrol/UserQueryDropTable.cc
andy-slac: First `if` (I guess you mean line 103) is checking that table name is present in CSS metadata, i.e. that table actually exists. This `if` checks its status which is supposed to be "READY", and issues an error if status is anything different. Table may exists but have a status which is not valid for dropping (e.g. "PENDING_DROP").

core/modules/ccontrol/UserQueryDropTable.cc
andy-slac: OK

core/modules/ccontrol/UserQueryDropTable.cc
andy-slac: Success "happens" at a later stage when watcher actually drops table from all databases. The status of this query in CSS is updated then by watcher. At this point the query is still executing.

core/modules/ccontrol/UserQueryFactory.cc
```diff
@@ -194,3 +240,41 @@ UserQueryFactory::Impl::Impl(StringMap const& m) {
 }
 
 }}} // lsst::qserv::ccontrol
+
+
+namespace {
+
+// regex for DROP TABLE [dbname.]table; both table and db names can be in quotes;
```
vaikuntht: Shouldn't it be "[dbname].table" ?

andy-slac: It should be either `DROP TABLE table` or `DROP TABLE db.table`, square brackets mean optional stuff, and dot separates db from table if there is no db then dot should not be there.

core/modules/ccontrol/UserQueryFactory.cc
```diff
@@ -194,3 +240,41 @@ UserQueryFactory::Impl::Impl(StringMap const& m) {
...
+// regex for SELECT *
+// Note that parens around whole string are not part of the regex but raw string literal
+std::regex _selectRe(R"(^select\s+.+$)",
+                     std::regex::ECMAScript | std::regex::icase | std::regex::optimize);
+
+bool isSelect(std::string const& query) {
+    LOGF(_log, LOG_LVL_DEBUG, "isSelect: %s" % query);
+    std::smatch sm;
+    bool match = regex_match(query, sm, _selectRe);
+    if (match) {
```
vaikuntht: You did `bool match = ...` then `if (match ) ...` here but directly use the match statement in the `isDropTable` function above. For consistency maybe change it to the same here and do `if (regex_match ...)`, or vice versa?

vaikuntht: Oh, is that because you want to return `match` here?

core/modules/ccontrol/UserQuerySelect.cc
```diff
@@ -272,25 +287,17 @@ void UserQuery::discard() {
...
 /// Setup merger (for results handling and aggregation)
-void UserQuery::_setupMerger() {
+void UserQuerySelect::_setupMerger() {
     LOG(getLogger(), LOG_LVL_TRACE, "Setup merger");
     _infileMergerConfig->mergeStmt = _qSession->getMergeStmt();
     _infileMerger = std::make_shared<rproc::InfileMerger>(*_infileMergerConfig);
 }
 
-void UserQuery::_setupChunking() {
+void UserQuerySelect::setupChunking() {
```
vaikuntht: All other functions were kept private, but this one is changed from `_setupChunking()` to `setupChunking()` . Any need for that?

core/modules/ccontrol/UserQuerySelect.h
```diff
@@ -0,0 +1,147 @@
...
+ */
+
+#ifndef LSST_QSERV_CCONTROL_USERQUERYSELECT_H
+#define LSST_QSERV_CCONTROL_USERQUERYSELECT_H
+/**
+  * @file
+  *
+  * @brief Umbrella container for user query state
+   *
+  * @author Daniel L. Wang, SLAC
```
vaikuntht: Your name here?

core/modules/ccontrol/UserQuerySelect.h
```diff
@@ -0,0 +1,147 @@
...
+    std::shared_ptr<qmeta::QMeta> _queryMetadata;
+
+    qmeta::CzarId _qMetaCzarId;   ///< Czar ID in QMeta database
+    qmeta::QueryId _qMetaQueryId;   ///< Query ID in QMeta database
+    bool _killed;
+    bool _submitted;                ///< True after submit() is completed
+    std::mutex _killMutex;
+    int _sessionId; ///< External reference number
+    int _sequence; ///< Sequence number for subtask ids
+    std::string _errorExtra; ///< Additional error information
```
vaikuntht: Comment alignments (L135-142) are inconsistent.

core/modules/ccontrol/UserQueryFactory.cc
andy-slac: Well IIRC it was `return regex_match(query, sm, _selectRe);` and then I wanted to add log message and ended up with this code. It can be done the same way above, ti also returns the same value as the one returned from `regex_match()`. I'll change code above to make it look similar.

core/modules/czar/bin/startCzar.py
```diff
@@ -113,15 +103,19 @@ def main():
...
     elif options.testName:
+        # not working
         runNamedTest(options.testName)
         return
     elif options.sanityClient:
+        # not working
         client.runSanityClient()
         return
     elif options.makeIndex:
+        # not working
```
vaikuntht: I'm confused about this file, were these never working or is it because of your changes now?

core/modules/ccontrol/UserQuerySelect.cc
andy-slac: Yes, this method is called by UserQueryFactory. UserQueryFactory was a friend (:yuck:) of UserQuery class so it could call private methods. I killed that friendship and updated UserQuerySelect interface to make this method public.

core/modules/ccontrol/UserQuerySelect.h
andy-slac: This is just a renaming of the UserQuery, I did not change much so authorship is still Daniel's. But you are right, at some point it may become difficult to say who is to blame for the mess, I think we should rely more on git history instead of sticking our names in the source.

core/modules/ccontrol/UserQuerySelect.h
andy-slac: I don't like to do it but I will :smile_cat:

core/modules/proxy/mysqlProxy.lua
```diff
@@ -281,8 +281,7 @@ function queryType()
            string.find(qU, "^LOAD ") or
            string.find(qU, "^CREATE ") or
            string.find(qU, "^ALTER ") or
-           string.find(qU, "^TRUNCATE ") or
-           string.find(qU, "^DROP ") then
+           string.find(qU, "^TRUNCATE ") then
             err.set(ERR_NOT_SUPPORTED,
                     "Sorry, this type of queries is disallowed.")
```
vaikuntht: This is unrelated to the ticket work but it would probably be nice to update this to something nicer like "Sorry, this type of query is not supported."

core/modules/czar/bin/startCzar.py
andy-slac: These were probably working long time ago but not recently. I tried to run these tests before doing my changes and it all crashed. This stuff will disappear anyways and I only added comments there to see what pieces are still functional and need to be ported to C++.

core/modules/proxy/mysqlProxy.lua
andy-slac: I can do it now, no problem, I agree we should apologize nicer :bow:

core/modules/ccontrol/UserQueryInvalid.h
```diff
@@ -0,0 +1,91 @@
...
+public:
+
+    UserQueryInvalid(std::string const& message)
+        : _message(message),
+          _messageStore(std::make_shared<qdisp::MessageStore>()) {}
+
+    UserQueryInvalid(UserQueryInvalid const&) = delete;
+    UserQueryInvalid& operator=(UserQueryInvalid const&) = delete;
+
+    // Accessors
```
vaikuntht: Not so important but in all other header files the functions in this section have been defined as `func() override;` without the empty `{}` at the end.

core/modules/ccontrol/UserQueryFactory.cc
vaikuntht: Makes sense. No changes needed here then.

core/modules/ccontrol/UserQueryInvalid.h
andy-slac: I was lazy and did not want to create separate .cc file for this class so I made all its methods inline (they are all empty so it does not take much space in header). In other classes these methods have some non-trivial implementations and are defined in .cc files.

core/modules/ccontrol/UserQuerySelect.h
vaikuntht: Interesting, Github shows this as a new file and not a "oldfile -> newfile" in the PR so I thought it was completely new code.

andy-slac: Yes, git unfortunately tends to think that file is new if you just copy it and make few simple changes to it. But it still thinks that .cc files are related (UserQuery.cc → UserQuerySelect.cc). I don't think there is an easy workaround for that besides committing right after renaming.

PR #156, opened 2015-11-17
Add header file that describes the shared scan interface.

abh3:
The subject line says it all.


core/modules/shrscan/Director.h
```diff
@@ -0,0 +1,315 @@
+// -*- LSST-C++ -*-
+/*
+ * LSST Data Management System
+ * Copyright 2014-2015 LSST Corporation.
```
andy-slac: 2015

core/modules/shrscan/Director.h
```diff
@@ -0,0 +1,315 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the LSST License Statement and
+ * the GNU General Public License along with this program.  If not,
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+
+#ifndef LSST_QSERV_SSCAN_DIRECTOR_H
```
andy-slac: `LSST_QSERV_SHRSCAN_DIRECTOR_H`

core/modules/shrscan/Director.h
```diff
@@ -0,0 +1,315 @@
...
+#include <vector>
+
+namespace lsst {
+namespace qserv {
+namespace shrscan {
+
+//-----------------------------------------------------------------------------
+//! @brief Describe a required table for a shared scan query.
+//-----------------------------------------------------------------------------
+
```
andy-slac: This comment is for every line from here to EOF - it needs some serious work to correct indentation, all in-class code must be indented, comments included. See 6-5 in https://confluence.lsstcorp.org/pages/viewpage.action?pageId=16908737

core/modules/shrscan/Director.h
```diff
@@ -0,0 +1,315 @@
...
+//-----------------------------------------------------------------------------
+//! Constructor
+//!
+//! @param  tabName   is the name of the table.
+//! @param  lkIndex   whether or not to lock the table's index
+//! @param  lkData    whether or not to lock the table's data
+//-----------------------------------------------------------------------------
+
+            TableInfo(const char *tabName="", bool lkIndex=true, bool lkData=true)
```
andy-slac: Use `std::string const&` instead of `const char*`

core/modules/shrscan/Director.h
```diff
@@ -0,0 +1,315 @@
...
+                     : tableName(tabName), lockIndex(lkIndex), lockData(lkData)
+                     {}
+
+//-----------------------------------------------------------------------------
+//! Destructor
+//-----------------------------------------------------------------------------
+
+           ~TableInfo() {}
```
andy-slac: Explicit destructor is not needed, remove.

core/modules/shrscan/Director.h
```diff
@@ -0,0 +1,315 @@
...
+//! @param  badChunks is a smart pointer to a vector of unprocessed chunks. This
+//!                   includes the chunk that was being processed if Run()
+//!                   returned false. You must take ownership of the shared
+//!                   pointer if you want to persist the vector upon return.
+//! @return canceled  When true, the query was cancelled and badChunks will be a
+//!                   zero length vector. Otherwise, the query completed and
+//!                   badChunks does hold the unprocessed chunk numbers.
+//-----------------------------------------------------------------------------
+
+virtual void endQuery(std::shared_ptr<std::vector<int>> const& badChunks,
```
andy-slac: Why `shared_ptr`? Just use `std::vector<int> const&`, it will be copied if needed?

core/modules/shrscan/Director.h
```diff
@@ -0,0 +1,315 @@
...
+
+virtual bool runQuery(int chunkNum) = 0;
+
+//-----------------------------------------------------------------------------
+//! The following vector should be used to add additional tables that may be
+//! needed by this particular query prior to call Director::addQuery(). These
+//! are in addition to the tables that the director was created with.
+//-----------------------------------------------------------------------------
+
+std::vector<TableInfo> tables; //< Additional tables needed by this query
```
andy-slac: This is some mix of abstract interface ans concrete data. Can we keep everything abstract without assumptions that it needs to have a vector inside it?

core/modules/shrscan/Director.h
```diff
@@ -0,0 +1,315 @@
...
+    unsigned int       numChunks;   //< The number of chunks being handled
+    unsigned int       maxCanRun;   //< Maximum number of parallel queries
+    unsigned int       numQueued;   //< Number waiting to run
+    unsigned int       numRunning;  //< Number actually running
+    unsigned int       numSuspend;  //< Number suspended
+    unsigned int       numErrors;   //< Number of queries that had errors
+    unsigned int       numCancelled;//< Number of queries that were cancelled
+};
+
+virtual void  getStatistics(Statistics &stats) = 0;
```
andy-slac: I'd say return it by value. There are schools in LSST that say that parameters that are modified should be passed by pointer instead of ref. As an alternative I think returning by value is better (if not too expensive but there is also (N)RVO).

core/modules/shrscan/Director.h
```diff
@@ -0,0 +1,315 @@
...
+struct Status {
+    unsigned int   secInQ;    //< Seconds      waiting to run (wall clock)
+    unsigned int   msRunning; //< Milliseconds running so far (wall clock)
+    unsigned int   msIdling;  //< Milliseconds in idle state  (wall clock)
+    unsigned short pctChunks; //< Percentage of chunks processed times 100
+    bool           isRunning; //< When true is actually running
+    bool           isSuspend; //< When true is currently suspended
+};
+
+virtual bool  getStatus(Query const* query, Status &status) = 0;
```
andy-slac: Throw exception if query is not found? And return Status by value.

core/modules/shrscan/Director.h
```diff
@@ -0,0 +1,315 @@
...
+//! @param  dbPath     - filesystem path where database tables reside.
+//! @param  baseTables - The tables that need to remain locked in memory.
+//! @param  chunkList  - The list of chunk numbers to be processed.
+//!
+//! @return false: Initialization failed (e.g. not all chunks were found).
+//! @return true:  Initialization succeeded.
+//-----------------------------------------------------------------------------
+
+virtual bool  init(const std::string dbPath, TableInfo const& baseTables,
+                   std::shared_ptr<std::vector<int>> const& chunkList);
```
andy-slac: Does it have an implementation? If not add `= 0`;

core/modules/shrscan/Director.h
```diff
@@ -0,0 +1,315 @@
...
+};
+
+virtual bool  getStatus(Query const* query, Status &status) = 0;
+
+//-----------------------------------------------------------------------------
+//! @brief Initialize for processing.
+//!
+//! The init() method should be called after construction of the director and
+//! before calling any of it's method. Initialization involves locking the base
+//! tables in memory which may be a lengthy processes.
```
andy-slac: This I do not understand at all. Needs to be described on Confluence probably.

core/modules/shrscan/Director.h
```diff
@@ -0,0 +1,315 @@
...
+//!
+//! @param  bytes  - The maximum number of bytes that may be locked.
+//-----------------------------------------------------------------------------
+
+virtual void  setMaxMemory(unsigned long long bytes) = 0;
+
+//-----------------------------------------------------------------------------
+//! @brief Set the maximum number of queries that be run in parallel.
+//!
+//! The default number of parallel queries is 7.
```
andy-slac: `7` is a nice magical number, I totally approve :stuck_out_tongue:

core/modules/shrscan/Director.h
```diff
@@ -0,0 +1,315 @@
...
+//! queue in FIFO order. If the query is executing its runQuery() method, the
+//! suspension occurs upon return.
+//!
+//! @param  query   - Pointer to the Query object that describes the query.
+//! @param  unlock  - When true, any additional tables required by the query
+//!                   that were locked are unlocked. Otherwise, the tables
+//!                   remain locked which may prevent new queries to run.
+//!
+//! @return false: The query was not found.
+//! @return true:  The query was suspended or scheduled to be suspended.
```
andy-slac: Should probably raise an exception if query is not found.

core/modules/shrscan/Director.h
```diff
@@ -0,0 +1,315 @@
...
+//!                   remain locked which may prevent new queries to run.
+//-----------------------------------------------------------------------------
+
+virtual void  suspendAll(bool unlock=true) = 0;
+
+//-----------------------------------------------------------------------------
+//! @brief Constructor & Destructor
+//-----------------------------------------------------------------------------
+
+              Director() {}
```
andy-slac: I'd probably disable copy constructor and assignment operator.

core/modules/shrscan/Director.h
jgates108: I'm ok with the destructor. The 3 lines saying that it's a destructor are superfluous.

core/modules/shrscan/Director.h
jgates108: If this object is being passed around different threads, I think the shared_ptr is appropriate.

core/modules/shrscan/Director.h
jgates108: Returning a unique_ptr<Statistics> might be better.

andy-slac: well, memory (de)allocation is much more expensive than making a copy of a small bunch of integers.

core/modules/shrscan/Director.h
abh3: I don't like throwing exceptions for things that are not really 
exceptions. In this case, the query could have been cancelled (either by 
itself or someone else) or finished. So, not finding a query is not an 
exception. Returning a true/false value allows the caller to do something 
reasonable in an easy way. I wish there was a standard way of warning of 
not using a returned value which would be more effective here.

Andy

core/modules/shrscan/Director.h
abh3: OK, we have aq conflict here. I don't want to throw an exception on 
something that is not exceptional (see previous mail file). I do need to 
tell you if I returned something. So, I am disposed to leave this as it 
is. It's the only reasonable compromise without contorting the code.

Andy

core/modules/shrscan/Director.h
andy-slac: I guess my reasoning is that after query is cancelled no one is supposed to call `suspendQuery()` for that query, right? But if you say that calling this method is OK even for cancelled queries I'm fine.

core/modules/shrscan/Director.h
```diff
@@ -0,0 +1,315 @@
...
+
+//-----------------------------------------------------------------------------
+//! @brief Set the maximum number of queries that be run in parallel.
+//!
+//! The default number of parallel queries is 7.
+//!
+//! @param  maxNum - The maximum number of queries to run in parallel.
+//-----------------------------------------------------------------------------
+
+virtual void  SetMaxQueries(int qnum) = 0;
```
andy-slac: `set...` instead of `Set...`

core/modules/shrscan/Director.h
andy-slac: @jgates108  Multi-threading argument actually works the opposite way - it is *much* better to avoid unnecessary sharing in multi-threaded programs. If you can copy stuff and make it thread-private you have fewer problems to worry about. Shared pointer does not solve data sharing issue, it is of course thread safe, but it does not protect data that it points to.

core/modules/shrscan/Director.h
andy-slac: Well, I probably read too much "modern C++" stuff, but anyways. With value-like types (non-abstract classes that can be copied/assigned) you only need to declare non-default destructor if the class has some non-trivial resources (pointers, etc.). But in those cases one also has to define non-trivial copy constructor and assignment operator. Basically rule of thumb is that if you need one then you need all three, having only one of them declared raises suspicion that something may be forgotten. Personally in general I prefer not to write code that I don't have to write (unless I'm paid by LOC count :smile: ), and reading code that does nothing is a waste of time too (just as writing long comments about it :angel: ).

core/modules/shrscan/Director.h
jgates108: I was making the assumption that it wouldn't change once created and would not require locking.

core/modules/shrscan/Director.h
jgates108: True.

core/modules/shrscan/Director.h
abh3: Yes, all of these methods are fine. They simply tell you whether they were effective or not. There may be cases where you don't care or if you do, check the return value.

core/modules/shrscan/Director.h
abh3: By that logic, one should avoid copying much of anything as the first thing a copy does is a malloc().

andy-slac: Why is that? If you write:

    Statistics stat = director->getStatistics();

there should not be any single malloc involved (unless you pack bunch of strings inside Statistics class). And if `getStatistics()` is implemented optimally there won't be even single copy happening thanks to NRVO.

abh3: So, let's say I do that I still have a problem. That structure is updated my multiple threads, so it's controlled by a mutex. The only way I can safely return it is to copy it anyway under a mutex control. That may cause NRVO to be disabled. I would have avoided all of those issues if I had done a single copy into the object the caller provided under mutex control. So, returning by value does not solve this particular problem.

andy-slac: It does not solve a problem because there is no problem. If you want to copy something from mutex-protected data then one copy has to happen anyways. And there is no reason to copy it more than once no matter how you do it, NRVO is not going to be disabled just because you have a mutex somewhere.

andy-slac: OK, one more reason why return-by-value is better and I stop trying to convince you :smile: 

In this code:

    Statistics stat;         // 1
    getStatistics(stat);     // 2

There is _always_ a construction happening (in 1) and copy (in 2), there is no way to avoid this two-step process (in you particular case construction does not do anything but in general this is not always true, depends entirely on the class).

In this code:

    Statistics stat = getStatistics(stat);

if method is written optimally and all optimizations are in place there is only one constructor call happening, no additional copies (it can be a copy constructor).

If you care about efficiency then you should think about avoiding unnecessary initialization too.

abh3: OK, you've convinced me :-)

core/modules/shrscan/Director.h
```diff
@@ -0,0 +1,292 @@
...
+        unsigned short     numChunks; //< Number of chunks processed
+        unsigned short     badChunks; //< Number of chunks that were bad
+        bool               isPresent; //< When true it exists (i.e. found)
+        bool               isRunning; //< When true is actually running
+        bool               isSuspend; //< When true is currently suspended
+        bool               isCanceld; //< When true is cancelling
+    };
+
+    virtual Status getStatus(Query const* query) = 0;
+    //-----------------------------------------------------------------------------
```
fritzm: Blank line needed before this comment block (I am adding value!)

core/modules/shrscan/Director.h
```diff
@@ -0,0 +1,292 @@
...
+
+    Director & operator=(const Director&) = delete;
+    Director(const Director&) = delete;
+
+                  Director() {}
+    virtual      ~Director() {}
+};
+
+}}} // namespace lsst:qserv:shrscan
+#endif
```
fritzm: We usually put a comment with the name of the include guard here as well, to make it clear that this #endif belongs to the include guard, e.g.:

```C++
#endif // LSST_QSERV_SHRSCAN_DIRECTOR_H
```

Also, prefer a trailing newline since some compilers bitch/warn about that, believe it or not...

PR #160, opened 2015-12-01
Fix some minor gcc 5.2 & clang warns

fritzm:
Pretty trivial: qMetaCzarId was missing init, and default args for lambdas are verboten...

### DIFF
--- a/core/modules/xrdsvc/SsiSession.cc
+++ b/core/modules/xrdsvc/SsiSession.cc
@@ -69,10 +69,8 @@ void SsiSession::ProcessRequest(XrdSsiRequest* req, unsigned short timeout) {
 
     auto replyChannel = std::make_shared<ReplyChannel>(*this);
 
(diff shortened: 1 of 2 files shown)


core/modules/xrdsvc/SsiSession.cc
```diff
@@ -69,7 +69,7 @@ void SsiSession::ProcessRequest(XrdSsiRequest* req, unsigned short timeout) {
 
     auto replyChannel = std::make_shared<ReplyChannel>(*this);
 
-    auto errorFunc = [this, &req, &replyChannel](std::string const& errStr, bool sendStr=true ) {
+    auto errorFunc = [this, &req, &replyChannel](std::string const& errStr, bool sendStr) {
```
jgates108: I talked with Andy and there's no reason not to have an error message, so it's probably better to eliminate the sendStr argument and anywhere it is false, just make an appropriate error message. I think there's only one and there's already an error message.

PR #162, opened 2015-12-04
Replace XML-RPC with in-process communication (DM-4348)

andy-slac:
This is a rather large bunch of commits. It replaces XML-RPC-based
czar with in-process instance wrapped into small number of Lua methods.

Everything is tested and works, all integration tests are fine but need one
small change in data loader script, see qserv_testdata branch with the 
same name.


admin/templates/configuration/etc/init.d/mysql-proxy
```diff
@@ -70,7 +71,7 @@ if [ "$USER" = "root" ]; then
     proxy_user_option="--user=$proxy_user"
 fi
 
-cmd="DEBUG=${LUA_DEBUG_LEVEL} QSERV_RPC_PORT=${QSERV_RPC_PORT} ${DAEMON} \
+cmd="DEBUG=${LUA_DEBUG_LEVEL} QSERV_RPC_PORT=${QSERV_RPC_PORT} QSERV_CONFIG=${QSERV_CONFIG} ${DAEMON} \
```
fjammes: QSERV_RPC_PORT should be no more required here?

fjammes: Passing configuration trough environment variable isn't nice, we should open a ticket to solve this, isn't it?

admin/templates/configuration/etc/init.d/qserv-czar
```diff
@@ -1,85 +0,0 @@
...
-    status)
-        status_of_proc -p ${pidfile} "$DAEMON" "$NAME" && exit 0 || exit $?
-        ;;
-    *)
-        echo "Usage: $0 {start|stop|restart|reload|status|condrestart|try-restart}"
-        exit 2
-        ;;
-esac
-
-exit $?
```
fjammes: Nice, I like code removal :+1:

admin/templates/configuration/etc/log4cxx.czar.properties
```diff
@@ -3,7 +3,7 @@
 # used by Qserv-czar
 #
 
-log4j.rootLogger=DEBUG, FILE
+log4j.rootLogger=DEBUG, CONSOLE
```
fjammes: Don't you want to log to file? All will go now to mysql-proxy console?

admin/templates/configuration/etc/sysconfig/qserv
```diff
@@ -50,8 +50,6 @@ export LD_LIBRARY_PATH={{LD_LIBRARY_PATH}}
 # Make sure core file is not discarded
 ulimit -c unlimited
 
-# used by qserv-czar
-export QSW_RESULTDIR=${QSERV_RUN_DIR}/qserv-run/tmp
```
fjammes: Hi think this is not in this ticket scope, but this parameter is now obsolete?

core/modules/SConscript
```diff
@@ -66,19 +64,12 @@ shProducts = { "xrdsvc" : {'mods' : """global mysql proto
...
-                                         rproc sg""".split(),
+               "czarProxy" : {'mods' : """ccontrol czar global parser proto proxy qana
+                                          query qdisp qproc rproc sg""".split(),
                              'shmods' : ['css'],
-                             'libs' : """ssl crypto antlr log
-                                      mysqlclient_r protobuf
-                                      XrdClient XrdPosix XrdSsi XrdUtils
-                                      boost_regex boost_thread boost_system
-                                      boost_signals""".split(),
+                             'libs' : """antlr log protobuf XrdSsi boost_regex""".split(),
```
fjammes: I thought this libraries where mandatory for the build, does new gcc version change that?

core/modules/czar/Czar.cc
```diff
@@ -0,0 +1,328 @@
...
+
+    // spawn background thread to wait until query finishes to unlock
+    auto finalizer = [](ccontrol::UserQuery::Ptr uq, MessageTable msgTable) {
+        uq->join();
+        try {
+            msgTable.unlock(uq);
+            if (uq) uq->discard();
+        } catch (std::exception const& exc) {
+            // TODO? if this fails there is no way to notify client, and client
+            // will likely hang because table may still be locked.
```
fjammes: Add a error log message here?

core/modules/czar/Czar.cc
```diff
@@ -0,0 +1,328 @@
...
+        for (auto iter = _clientToQuery.begin(); iter != _clientToQuery.end(); ) {
+            if (iter->second.expired()) {
+                iter = _clientToQuery.erase(iter);
+            } else {
+                ++ iter;
+            }
+        }
+
+        // remember query (weak pointer) in case we want to kill query
+        // TODO: the map grows indefinitely now, will have to do some cleanup
```
fjammes: Why this comment? Previous cleanup step isn't used?

core/modules/proxy/mysqlProxy.lua
```diff
@@ -3,61 +3,13 @@
...
---  * invoke(cleanQueryString, hintString)
---  * cleanQueryString is the query without special hints: objectId
---    related hints are passed through, but bounding boxes are not
---  * hintString: "box", "1,2,11,12", "box", "5,55,6,66",
---    "objectId", "3", "objectId", "5,6,7,8" and so on
 
--------------------------------------------------------------------------------
---                             debug tool                                    --
--------------------------------------------------------------------------------
-local DEBUG = os.getenv('DEBUG') or 0
```
fjammes: Don't you want to keep a native LUA logger for debugging? If yes,  env variable can now be removed from etc/init.d/mysql-proxy and LUA_DEBUG_LEVEL can also be removed.

site_scons/state.py
```diff
@@ -176,6 +176,7 @@ def _setEnvWithDependencies():
             (PathVariable('LOG_LIB', 'log libraries path', os.path.join(env['LOG_DIR'], "lib"), PathVariable.PathIsDir)),
             (PathVariable('PROTOBUF_INC', 'protobuf include path', os.path.join(env['PROTOBUF_DIR'], "include"), PathVariable.PathIsDir)),
             (PathVariable('PROTOBUF_LIB', 'protobuf libraries path', os.path.join(env['PROTOBUF_DIR'], "lib"), PathVariable.PathIsDir)),
+            (PathVariable('LUA_INC', 'lua include path', os.path.join(env['LUA_DIR'], "include"), PathVariable.PathIsDir)),
```
fjammes: Add a comment "Needed to swig C++ to Lua"?

ups/qserv.table
```diff
@@ -14,7 +13,6 @@ setupRequired(partition)
 setupRequired(protobuf)
 setupRequired(scisql)
 setupRequired(swig)
-setupRequired(twisted)
 setupRequired(xrootd)
 
 envPrepend(LD_LIBRARY_PATH, ${PRODUCT_DIR}/lib)
```
fjammes: Nice ticket :+1: Thanks for this interesting review

core/modules/czar/Czar.cc
```diff
@@ -0,0 +1,328 @@
...
+int
+parseKillQuery(std::string const& aQuery) {
+    // the query that proxy passes us is all uppercase and spaces compressed
+    // but it may have trailing space which we strip first
+    std::string query = aQuery;
+    auto pos = query.find_last_not_of(' ');
+    query.erase(pos+1);
+
+    // try to match against one or another form of KILL
+    static const std::string prefixes[] = {"KILL QUERY ", "KILL CONNECTION ", "KILL "};
```
fjammes: Is this parsing technique strong enough?

core/modules/czar/Czar.cc
```diff
@@ -0,0 +1,328 @@
...
+                iter = _clientToQuery.erase(iter);
+            } else {
+                ++ iter;
+            }
+        }
+
+        // remember query (weak pointer) in case we want to kill query
+        // TODO: the map grows indefinitely now, will have to do some cleanup
+        if (not clientId.empty() and threadId >= 0) {
+            ClientThreadId ctId(clientId, threadId);
```
fjammes: Could you please explain me ClientThreadId? In the cluster (where sql client is launched from ccqservbuild and master is ccqserv125) log I can see: "client_dst_name" "master_ip_address:4040" "server_thread_id" "2501". So client_dst_name seems to be the ip:port of mysql-proxy, and shouldn't change. server_thread_id seems to be the thread_id of the mysql-proxy  thread with is connected to mysql, this one could be enough to identify the query, isn't it?

admin/templates/configuration/etc/init.d/mysql-proxy
andy-slac: Indeed, we son't need QSERV_RPC_PORT anymore, will remove it.

Regarding environment - I don't know if there is any other way to pass configuration to Lua script. We could try to implement config file reader in Lua, but config file name itself needs to be passed somehow to Lua level.

admin/templates/configuration/etc/log4cxx.czar.properties
andy-slac: I want to reduce the number of log files produced by mysql-proxy, easiest way to do it now is to log everything to standard output (console) and redirect it to a log file (mysql-proxy-lua.log).

admin/templates/configuration/etc/sysconfig/qserv
andy-slac: It's not used anywhere, I thought I would just remove it as a part of the cleanup.

core/modules/SConscript
andy-slac: This new list is only direct dependencies (libraries needed by 'mods'), there are other libraries that are indirectly used through libqserv_css but they don't need to be included here.

core/modules/czar/Czar.cc
andy-slac: OK

core/modules/czar/Czar.cc
andy-slac: It's leftover from work-in-progress version of the code, I'll remove it, the issue does not exist any more.

core/modules/proxy/mysqlProxy.lua
andy-slac: All logging now goes to log4cxx, there is no native Lua logging anymore, I'll remove LUA_DEBUG_LEVEL.

site_scons/state.py
andy-slac: Well, do we want to comment each of these variables? I don't really want to make LUA_INC look special in any way, we either comment all of them or none.

core/modules/czar/Czar.cc
andy-slac: Yes, proxy guarantees that, see comment few lines above.

core/modules/czar/Czar.cc
andy-slac: This needs some thought, right now it is just a transparent rewrite of the XML-RPC handling of this. In principle you are correct that `threadId` is probably sufficient to identify client in current system, but this depends on your knowledge that "client_dst_name" is never going to change which is a sort of implementation detail. 
In the long term I think we should switch from this czar-internal mapping to using QMeta for this info, I just did not want to do it in this ticket. So let's keep this as it is now and we'll come back to it later anyway when we start implementing async queries.

core/modules/ccontrol/UserQueryFactory.h
```diff
@@ -32,16 +32,18 @@
...
 // System headers
+#include <cstdint>
 #include <memory>
 
 // Third-party headers
 #include "boost/utility.hpp"
 
-// LSST headers
 #include "lsst/log/Log.h"
+// LSST headers
```
fritzm: Comment line got flopped to wrong place?

core/modules/czar/Czar.cc
```diff
@@ -0,0 +1,328 @@
...
+        "");
+    mysqlConfig.dbName = cm.get(
+        "resultdb.db",
+        "Error, resultdb.db not found. Using qservMeta.",
+        "qservResult");
+
+    return mysqlConfig;
+}
+
+}
```
fritzm: comment that we are closing anon. namespace here for clarity?

core/modules/czar/Czar.h
```diff
@@ -0,0 +1,114 @@
...
+     */
+    Czar(std::string const& configPath, std::string const& czarName);
+
+    Czar(Czar const&) = delete;
+    Czar& operator=(Czar const&) = delete;
+
+    /**
+     * Submit query for execution.
+     *
+     * Returns list of strings:
```
fritzm: Would it be more clear to use a struct (and thus have member names) instead of a vector of unnamed strings?   Some reason to use a vector here because of Lua wrappers?

core/modules/czar/Czar.cc
```diff
@@ -0,0 +1,328 @@
...
+        result[0] = "Failed to instantiate query: " + error;
+        return result;
+    }
+
+    // start execution
+    LOGF(_log, LOG_LVL_DEBUG, "submitting new query");
+    uq->submit();
+
+    // spawn background thread to wait until query finishes to unlock
+    auto finalizer = [](ccontrol::UserQuery::Ptr uq, MessageTable msgTable) {
```
fritzm: @jgates108 : would it be reasonable to extend the UserQuery::submit() (and executive beneath) to just take the finalizer as an argument?  Then we could avoid the yucky detached threads below?

fritzm: Maybe add to the comment above that msgTable is copied into this lambda, so future readers will just know that its okay and on purpose?  I had to go look...

jbecla: This request made me think that we definitely should test that code with something that submits multiple concurrent queries. The proxy can block and stop accepting new queries if we mess up thread detaching etc.

core/modules/czar/Czar.cc
```diff
@@ -0,0 +1,328 @@
...
+parseKillQuery(std::string const& aQuery) {
+    // the query that proxy passes us is all uppercase and spaces compressed
+    // but it may have trailing space which we strip first
+    std::string query = aQuery;
+    auto pos = query.find_last_not_of(' ');
+    query.erase(pos+1);
+
+    // try to match against one or another form of KILL
+    static const std::string prefixes[] = {"KILL QUERY ", "KILL CONNECTION ", "KILL "};
+    for (auto& prefix: prefixes) {
```
fritzm: I think, if you `#include <initializer_list>`, you could also just say
```c++
for (std::string const& prefix: {"KILL QUERY ", "KILL CONNECTION", "KILL "}) {
```

core/modules/czar/MessageTable.cc
```diff
@@ -0,0 +1,124 @@
...
+
+std::string const createAndLockTmpl("CREATE TABLE IF NOT EXISTS %1% "
+    "(chunkId INT, code SMALLINT, message CHAR(255), "
+    "severity ENUM ('INFO', 'ERROR'), timeStamp FLOAT)"
+    "ENGINE=MEMORY; LOCK TABLES %1% WRITE;");
+
+std::string const writeTmpl("INSERT INTO %1% (chunkId, code, message, severity, timeStamp) "
+    "VALUES (%2%, %3%, '%4%', '%5%', %6%)");
+
+std::string const unlockTmpl("UNLOCK TABLES");
```
fritzm: Is there a reason we don't unlock just the one specific table here?  If so, add explanatory comment?

core/modules/ccontrol/UserQueryFactory.h
andy-slac: Hmm, I wonder who've done that :innocent:  I blame my cat stepping on keyboard :octocat:

core/modules/czar/Czar.cc
andy-slac: ok

core/modules/czar/Czar.h
andy-slac: I tried to keep it compatible with existing Lua code, but you are right, we should use cleaner approach when possible. I'll try to see how SWIG wraps simple structures in Lua.

core/modules/czar/Czar.cc
andy-slac: Having both `UserQuery::join()` and finalizer probably does not make much sense. We could probably get rid of `join()` if we want UserQuery to handle everything internally, but then finalizer can become a part of UserQuery class itself. Mess. I say keep it as it is for now until we have more problems to solve in that area.
I'll add a comment, and I think it may be clearer if I pass those two objects in lambda capture explicitly by value.

core/modules/czar/Czar.cc
andy-slac: But that would create new copy of `std::string` on every call and iteration? I prefer for those string to be constructed once only (my array is `static`).

core/modules/czar/MessageTable.cc
andy-slac: Mysql can only unlock all locked table at once, there is no syntax to unlock one table. I'll add comment.

core/modules/czar/Czar.cc
fritzm: Ah right -- since the inits are char* -- yup!  I'd bet they'd be small, on the stack, and that compiler would have the underlying char*s as static, tho', so I wouldn't be concerned...  But your way is explicit, readable, and compact as well so I'm fine with it. :-)

core/modules/czar/Czar.cc
fritzm: Yah, my thought was just that submit() would take a function object to be run at completion (like async. socket ops are done in boost::asio).  Then you can still wrap up your finalizer code in a lambda at point of call, but you wouldn't have to spawn/detach additional threads to run them.  But yeah -- for a different commit if we wanted to do that!

core/modules/proxy/czarProxy.cc
```diff
@@ -0,0 +1,109 @@
...
+    static std::map<std::string, std::string> const hints{
+        std::make_pair("client_dst_name", clientId)
+    };
+    return czarInstance(hints).killQuery(query, clientId);
+}
+
+void log(std::string const& loggername, std::string const& level,
+         std::string const& filename, std::string const& funcname,
+         unsigned int lineno, std::string const& message) {
+    lsst::log::Log::log(loggername, log4cxx::Level::toLevel(level),
```
fritzm: Wait -- are you trying to log something here?  Could we find some way to say "log" a few more times on this line to clarify? :smirk_cat:

core/modules/proxy/czarProxy.cc
```diff
@@ -0,0 +1,109 @@
...
+        // use proxy client id (which includes host name and port number of the
+        // proxy-side client connection)
+        return iter->second;
+    } else {
+        return "czar." + std::to_string(getpid());
+    }
+}
+
+lsst::qserv::czar::Czar&
+czarInstance(std::map<std::string, std::string> const& hints) {
```
fritzm: I don't personally like lazy-init'd-singletons.  I strongly prefer having a separate, explicit, init call (which can also return an error code that can be meaningfully handled, can be reliably sequenced with other inits, and will only happen when expected).  Then I'm fine if access without init totally explodes in flames (if you forgot to do it you will find out on the first run fix it right away).

I realize that this flies in the face of the capital-P Patterns people and their almighty Singleton...  Still, they are WRONG :smiley:  What do you think?

fritzm:
Yay!  This is great.  Not the least because it is +1k,-4k (though some lines added are in a .i and I think those are scored at a 10x penalty ;-) ).  Thanks, Andy!

core/modules/proxy/czarProxy.cc
andy-slac: Easily:

    lsst::log::Log::Log::Log::Log::Log::Log::log(..)

should compile without any trouble and it makes totally clear that I really mean it!

core/modules/proxy/czarProxy.cc
andy-slac: Pattern people can never be wrong, we are just using incorrect patterns :repeat: 

I'm generally OK with singletons and lazy init, but this particular case is indeed ugly in that it needs some extra info for initialization which I need to pass every time even if it's only used once. Part of the trouble is mysql-proxy which does not have one-time initialization callback with relevant info. Still I think I can improve this a little bit without too much complication.

andy-slac: @fritzm, please check commit c1994ae (this same PR) to see if you like it better.

core/modules/czar/Czar.h
andy-slac: Implemented in 02f4a42b

core/modules/ccontrol/UserQuerySelect.cc
```diff
@@ -146,11 +146,12 @@ UserQuerySelect::UserQuerySelect(std::shared_ptr<qproc::QuerySession> const& qs,
                                  std::shared_ptr<qproc::SecondaryIndex> const& secondaryIndex,
                                  std::shared_ptr<qmeta::QMeta> const& queryMetadata,
                                  qmeta::CzarId czarId,
+                                 uint64_t taskId,
                                  std::string const& errorExtra)
     :  _qSession(qs), _messageStore(messageStore), _executive(executive),
        _infileMergerConfig(infileMergerConfig), _secondaryIndex(secondaryIndex),
        _queryMetadata(queryMetadata), _qMetaCzarId(czarId), _qMetaQueryId(0),
-       _killed(false), _submitted(false), _sessionId(0), _sequence(0),
+       _killed(false), _submitted(false), _taskId(taskId), _sequence(0),
```
jgates108: I'm wondering if can come up with another name for _taskId (maybe queryId or userQueryId?). The executive breaks user queries into jobs with jobIds, those are passed to the worker where they are tasks. To avoid confusion, it would be good if they all have unique names.

jbecla: yes, userQueryId / chunkQueryId type names would be very clear. ("task" always was a little less so to me)

core/modules/proxy/mysqlProxy.lua
```diff
@@ -341,65 +288,66 @@ function queryProcessing()
...
+    end
+
     -- q  - original query
     -- qU - original query but all uppercase
     --
     --
     local sendToQserv = function(q, qU)
 
-        hintsToPassArr = {} -- Reset hints (it's global)
+        local hintsToPassArr = {} -- Reset hints (it's global)
```
jgates108: Is this comment correct?

core/modules/ccontrol/UserQuerySelect.cc
andy-slac: I renamed id to taskId only because the code which generates this number calls it taskId. I'll change it to userQueryId.

core/modules/proxy/mysqlProxy.lua
andy-slac: Not even close :) I'll remove it.

PR #163, opened 2015-12-05
Tickets/dm 4529

kelseymh:
Four files were modified to address MacOSX clang70 compiler errors:
-- `core/modules/util/EventThread.h,cc`: `uint` ==> `unsigned int`
-- `core/modules/wbase/SendChannel.h,cc`: Add `#include <functional>`
-- `core/modules/wsched/ChunkState.cc`: Add `#include <iostream>`
-- `build/qmeta/qmetaLib_wrap.cc`:  Use `::uint*` (global namespace) for typedefs into `std::`

### DIFF
--- a/core/modules/util/EventThread.h
+++ b/core/modules/util/EventThread.h
@@ -153,7 +153,8 @@ protected:
 class ThreadPool : public std::enable_shared_from_this<ThreadPool> {
 public:
     using Ptr = std::shared_ptr<ThreadPool>;
-    static ThreadPool::Ptr newThreadPool(uint thrdCount, CommandQueue::Ptr const& q) {
+    static ThreadPool::Ptr newThreadPool(unsigned int thrdCount,
+                                         CommandQueue::Ptr const& q) {
         Ptr thp{new ThreadPool{thrdCount, q}}; // private constructor
         thp->_resize();
         return thp;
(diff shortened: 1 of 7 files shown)


timj:
Looks good to me. Can you please redo the commit history so that the reverts aren't there and the messages don't say "Try" (Yoda quote goes here I think). The act of running Jenkins is also not relevant to the commit history.

core/modules/wsched/ChunkState.cc
```diff
@@ -25,6 +25,7 @@
 
 // System headers
 #include <iterator>
+#include <iostream>
```
jbecla: better to swap order, we try to keep includes ordered alphabetically (within each group)

core/modules/util/EventThread.h
```diff
@@ -153,7 +153,8 @@ class PoolEventThread : public EventThread {
 class ThreadPool : public std::enable_shared_from_this<ThreadPool> {
 public:
     using Ptr = std::shared_ptr<ThreadPool>;
-    static ThreadPool::Ptr newThreadPool(uint thrdCount, CommandQueue::Ptr const& q) {
+    static ThreadPool::Ptr newThreadPool(unsigned int thrdCount,
+					 CommandQueue::Ptr const& q) {
```
jbecla: looks slightly misaligned

core/modules/wsched/ChunkState.cc
kelseymh: Done.  I like that convention as well.

core/modules/util/EventThread.h
kelseymh: You're right.  My emacs is still sticking in tabs; I replaced them with spaces and all is well.

timj:
The include file reordering commit and replace tabs with spaces commits should be squashed into the relevant parents since they are both correcting previous issues introduced in this PR.

timj:
Looks good to me now. Thanks.

kelseymh:
Confirmed building on MacOSX (XCode 7) resolves all errors.

PR #164, opened 2015-12-08
minor tweaks related to DM-4348

jbecla:


core/modules/czar/Czar.cc
```diff
@@ -328,4 +328,4 @@ mysqlConfig(lsst::qserv::StringMap const& config) {
     return mysqlConfig;
 }
 
-}
+} // annonymous namespace
```
andy-slac: anonymous (extra n), I's skip "anonymous" altogether

PR #166, opened 2015-12-15
Implement empty chunk cache reset (DM-4383)

andy-slac:
Adding special SQL syntax for czar to reset empty chunk cache
(`FLUSH QSERV_CHUNKS_CACHE [FOR database]`). Wmgr also gets
a special method to reset chunks cache which uses above SQL.

### DIFF
--- a/admin/templates/configuration/etc/qserv-wmgr.cnf
+++ b/admin/templates/configuration/etc/qserv-wmgr.cnf
@@ -34,9 +34,16 @@ DB_PASSWD = ''
 DB_USER_PRIV = '{{MYSQLD_USER}}'
 DB_PASSWD_PRIV = '{{MYSQLD_PASS}}'
 
+# parameters for mysql-proxy connection, these are only useful
+# for czar wmgr instance
+PROXY_HOST = '127.0.0.1'
+PROXY_PORT = {{MYSQL_PROXY_PORT}}
+PROXY_USER = '{{QSERV_USER}}'
+PROXY_PASSWD = ''
+
 # CSS-related parameters
 # CSS_CONFIG is a dictionary as accepted by CssAccess.createFromConfig() method,
-# all values in a dictionary must bestrings even if represent numbers
+# all values in a dictionary must be strings even if they represent numbers
 USE_CSS = True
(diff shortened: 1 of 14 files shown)


fritzm:
Sorry to come in to this discussion to so late.  It makes me a little uncomfortable to extend the SQL grammar for this.  What about just using stored-procedure "CALL" syntax instead on a magic procedure name for this instead of extending "FLUSH"?  Would that be better or worse?

andy-slac:
I can't really say if it's better or worse. It may depend on what we plan to do with `CALL` in general (how do we handle it in proxy/Lua). My general idea is that we'll need to extend (or abuse/repurpose) some of the SQL syntax anyways (like `SHOW PROCESSLIST`) so we can't avoid introducing qserv-specific syntax. This new FLUSH syntax makes it very clear that it's Qserv extension, besides this is sort of internal command that end users are not supposed to even know, we can change it at any time to different syntax and not worry about compatibility. Still I'd be happy to change it to `CALL QSERV_FLUSH_CHUNKS_CACHE('db')` instead if people think it's a better way to extend qserv SQL interface.

ktlim:
It does seem common for other databases to extend SQL syntax by both adding new syntax to existing verbs (like this) and adding new verbs.  Would it make it easier on the parser to just add a "QSERV" verb for all of our extensions that need to be at that level?

andy-slac:
I guess it's mostly an issue of how user-friendly we want it to be (for user-facing queries). Do we want something like `QSERV SELECT ASYNC ...` instead of just `SELECT ASYNC ...`? I think latter is less surprising while still fits the pattern of SQL extensions supported by mysql. From the parser point of view it's certainly easier to have clear identifying token, but I don't think it's a good argument. I'd prefer to have syntax that fits well into existing language structure and has fewer surprises.
And QSERV is not a verb :) Not until we make it as popular as Google so people start saying "I qserved that" :tada:

ktlim:
Definitely not "QSERV SELECT ASYNC" (although at first glance I would kind of prefer that asynchronicity be a property of the connection rather than the query).  As I said: "that need to be at that level" -- i.e. new top-level commands.

PR #171, opened 2016-01-04
U/fjammes/DM-224: Move to MariaDB

fjammes:


admin/python/lsst/qserv/admin/configure.py
```diff
@@ -59,7 +59,7 @@
 ETC = 'etc'
 CLIENT = 'client'
 
-MYSQL = 'mysql'
+MARIADB = 'mariadb'
```
jbecla: What is the name of the process in mariadb? Is it mariadbd or still mysqld? If the latter, I'd definitely keep the mysql names, mariadb is a drop in replacement of mysql, isn't it? I worry otherwise we will never be sure whether to use "mysql" or "mariadb" name.

admin/templates/configuration/tmp/configure/mariadb.sh
```diff
@@ -2,14 +2,14 @@
```
andy-slac: Keep the old name of this file for now? I think ideally we should be able to switch between mysql/mariadb quickly by changing one line in qserv.table file, in reality it could be more complicated but we should try to support both for now until we decide that we do not want to support mysql any more. "mysql" in my view is more generic term than mariadb, think of it as mariadb being one particular implementation of mysql. We still have all those MYSQL_* names everywhere, if we want to do renaming we should probably rename those too?

admin/templates/configuration/tmp/configure/scisql.sh
```diff
@@ -2,30 +2,31 @@
...
 export PATH=${PYTHON_DIR}:${PATH}
-export LD_LIBRARY_PATH={{LD_LIBRARY_PATH}}
+export LD_LIBRARY_PATH='{{LD_LIBRARY_PATH}}'
 
 ${QSERV_RUN_DIR}/etc/init.d/mysqld start
-cd ${SCISQL_DIR}
 echo "-- Deploying sciSQL plugin in MySQL database"
 # password is given on stdin, so that is can't be catched by ps command
 PASSFILE=${QSERV_RUN_DIR}/tmp/pass.txt
 cat <<EOF > ${PASSFILE}
```
andy-slac: Do we really need to create this temporary file, can we just pipe password directly to the script?

admin/templates/configuration/tmp/configure/sql/qserv-worker.sql
```diff
@@ -1,6 +1,7 @@
--- Used by xrootd Qserv plugin:
+CREATE USER IF NOT EXISTS 'qsmaster'@'localhost';
```
andy-slac: Why do you use "CREATE USER IF NOT EXISTS" here and "CREATE OR REPLACE USER" in qserv-czar.sql? Do you depend on the order in which these two scripts are executed?

core/modules/mysql/MySqlConnection.cc
```diff
@@ -200,6 +200,7 @@ MYSQL* MySqlConnection::_connectHelper() {
         janitor.reset(new MySqlThreadJanitor);
     }
     unsigned long clientFlag = CLIENT_MULTI_STATEMENTS;
+	mysql_options( m, MYSQL_OPT_LOCAL_INFILE, 0 );
```
andy-slac: Your editor inserts TABs again, replace with spaces. Spaces are not needed after/before parenthesis.

core/modules/mysql/testLocalInfile.cc
```diff
@@ -48,33 +48,37 @@ class Api {
...
-            assert(cur);
+        mysql_real_connect(&cursor,
+				"localhost", // host
+				"qsmaster", // user
+				"", // pw
+				"", // db
+				0, // port
+				"/home/qserv/qserv-run/git/var/lib/mysql/mysql.sock", // socket
+				0); // client flag
+        if(!&cursor) {
```
andy-slac: This test will never succeed, `&cursor` can never be NULL, it's an address of member variable. You need to return back `MYSQL* cur = mysql_real_connect(...)`, otherwise there is no way to test for failure.

core/modules/mysql/testLocalInfile.cc
```diff
@@ -48,33 +48,37 @@ class Api {
...
-                                        "", // db
-                                        0, // port
-                                        "/u1/local/mysql.sock", // socket
-                                        0); // client flag
-        if(!cur) {
-            std::cout << "Failed to connect to MySQL: Error: "
-                      << mysql_error(cur) << std::endl;
-            assert(cur);
+        mysql_real_connect(&cursor,
+				"localhost", // host
```
andy-slac: TABs here

core/modules/mysql/testLocalInfile.cc
```diff
@@ -48,33 +48,37 @@ class Api {
...
+				"", // db
+				0, // port
+				"/home/qserv/qserv-run/git/var/lib/mysql/mysql.sock", // socket
+				0); // client flag
+        if(!&cursor) {
+            std::cerr << "Failed to connect to MySQL: Error: "
+                      << mysql_error(&cursor) << std::endl;
+            assert(&cursor);
         }
+		int result = mysql_query(&cursor, "show databases;");
```
andy-slac: TABs

core/modules/mysql/testLocalInfile.cc
```diff
@@ -247,9 +251,13 @@ void playRead() {
 void checkDoubleTable() {
     Api a; // Source: will execute "select ..."
     a.connect();
+    if(!a.exec("show databases;")) {
+        std::cerr << "error creating test.twofloats.\n";
```
andy-slac: The message may be misleading, you are not creating database here yet.

ups/qserv.table
```diff
@@ -17,4 +17,5 @@ setupRequired(xrootd)
 
 envPrepend(LD_LIBRARY_PATH, ${PRODUCT_DIR}/lib)
 envPrepend(PATH, ${PRODUCT_DIR}/bin)
+envPrepend(PATH, ${MYSQLCLIENT_DIR}/bin)
```
andy-slac: This does not look right, there is no dependency on `mysqlclient` package in this table file, so MYSQLCLIENT_DIR will not be defined?

core/modules/mysql/testLocalInfile.cc
```diff
@@ -48,33 +48,37 @@ class Api {
 public:
     Api() {
         mysql_init(&cursor);
+		mysql_options( &cursor, MYSQL_OPT_LOCAL_INFILE, 0 );
```
andy-slac: TABs

admin/templates/configuration/tmp/configure/sql/qserv-worker.sql
fjammes: Yes this order matter, but only for mono-node configuration:

- qserv-czar.sql is loaded
- qserv-worker.sql is loaded and `qsmaster` account must not be removed in order to keep  access rights granted by previous sql script.

admin/python/lsst/qserv/admin/configure.py
fjammes: Ok this is reasonable, but please note that mariadb is packaged as "lsst/mariadb" in eups, and this imply creation of MARIADB_DIR env var. I did this not to disrupt lsst stack behaviour.

admin/templates/configuration/tmp/configure/sql/qserv-worker.sql
andy-slac: Why not use "CREATE USER IF NOT EXISTS" in both cases (or DROP USER if you want to get rid of account)? I understand that "OR REPLACE" is a relatively new mariadb extension, we should try to avoid depending on mariadb-specific features for now.

fjammes: :+1: done

PR #172, opened 2016-01-04
Tickets/dm 2699

jgates108:


core/modules/mysql/MySqlConnection.cc
```diff
@@ -148,7 +148,9 @@ MySqlConnection::cancel() {
     }
     // KILL QUERY only, not KILL CONNECTION.
     int threadId = mysql_thread_id(_mysql);
-    std::string killSql = boost::str(boost::format("KILL QUERY ") % threadId);
+    std::ostringstream osKill;
+    osKill << "KILL QUERY " << threadId;
+    std::string killSql = osKill.str();
```
andy-slac: Why not:

    std::string const killSql = "KILL QUERY " + std::to_string(threadId);

?

core/modules/qdisp/Executive.cc
```diff
@@ -208,35 +229,26 @@ void Executive::markCompleted(int jobId, bool success) {
...
     bool alreadyCancelled = _cancelled.exchange(true);
     if(alreadyCancelled) {
-        LOGF_WARN("Executive::squash() already cancelled! refusing.");
+        LOGF_WARN("%1% Executive::squash() already cancelled! refusing." % _id);
         return;
     }
 
-    LOGF_INFO("Trying to cancel all queries...");
+    LOGF_WARN("%1% Executive::squash Trying to cancel all queries..." % _id);
+    std::deque<JobQuery::Ptr> jobsToCancel;
```
andy-slac: Why not `std::vector`, the size of that is known in advance, you can reserve exact length?

core/modules/qdisp/Executive.h
```diff
@@ -143,8 +147,11 @@ class Executive {
     mutable std::mutex _errorsMutex;
 
     std::condition_variable _allJobsComplete;
-    mutable std::mutex _jobsMutex;
+    mutable std::recursive_mutex _jobsMutex;
 
+    // Give this executive a reasonable identifier, to be replaced by a unique id.
+    std::string _id;
+    static util::Sequential<int> _seq;
```
andy-slac: Should we use `std::atomic` instead?

core/modules/qdisp/Executive.cc
```diff
@@ -274,8 +286,8 @@ void Executive::_setup() {
         _xrdSsiService = XrdSsiProviderClient->GetService(eInfo, _config.serviceUrl.c_str()); // Step 1
     }
     if(!_xrdSsiService) {
-        LOGF_ERROR("Error obtaining XrdSsiService in Executive: "
-                   %  getErrorText(eInfo));
+        LOGF_ERROR("%1% Error obtaining XrdSsiService in Executive: "
```
andy-slac: `%2%` is missing

jgates108: Good catch. It was missing before as well.

core/modules/qdisp/JobQuery.cc
```diff
@@ -95,25 +100,38 @@ void JobQuery::provisioningFailed(std::string const& msg, int code) {
 
 /// Cancel response handling. Return true if this is the first time cancel has been called.
 bool JobQuery::cancel() {
+    LOGF_DEBUG("%1% JobQuery::cancel()" % getIdStr());
     if (_cancelled.exchange(true) == false) {
         std::lock_guard<std::recursive_mutex> lock(_rmutex);
-        // Nothing needs to be done for _queryResourcePtr.
-        if (_queryRequestPtr) {
+        // If _queryRequestPtr is not nullptr, then this job has been passed to xrootd and cancellation is complicated.
+        if (_queryRequestPtr != nullptr) {
```
andy-slac: I believe that `if (_queryRequestPtr)` convention is much more common in our code than `if (_queryRequestPtr != nullptr)`.

core/modules/qdisp/JobQuery.cc
```diff
@@ -95,25 +100,38 @@ void JobQuery::provisioningFailed(std::string const& msg, int code) {
...
+            std::ostringstream os;
+            os << getIdStr() <<" cancel before QueryRequest" ;
+            LOGF_DEBUG("%1%" % os.str());
+            getDescription().respHandler()->errorFlush(os.str(), -1);
+            _executive->markCompleted(getIdInt(), false);
         }
         _jobDescription.respHandler()->processCancel();
         return true;
     }
+    LOGF_DEBUG("%1% cancel, skipping, already cancelled.");
```
andy-slac: Needs ` % getIdStr()`

core/modules/qdisp/Executive.h
jgates108: This is really a place holder until we use real unique query ids

core/modules/qdisp/JobQuery.cc
jgates108: I think != nullptr is safer in that it will only compile when used with pointers.

core/modules/qdisp/JobStatus.cc
```diff
@@ -76,55 +76,58 @@ std::ostream& operator<<(std::ostream& os, JobStatus::State const& state) {
...
+        msg = "Error merging result";
         break;
     case JobStatus::COMPLETE:
-         msg = "Complete (success)";
+        msg = "Complete (success)";
+        break;
+    case JobStatus::CANCEL:
+        msg = "CANCEL";
         break;
     default:
```
andy-slac: Maybe move default case outside `switch` so that compiler can warn you when any enum label is missing?

core/modules/qdisp/JobQuery.cc
jgates108: Good catch.

core/modules/qdisp/JobStatus.cc
jgates108: The missing state generates a warning, which could be missed, and then there'd be a segfault. I changed it so the initial value of msg =  "State error (unrecognized)";

core/modules/qdisp/QueryRequest.cc
```diff
@@ -52,14 +52,13 @@ namespace qdisp {
 // QueryRequest
 ////////////////////////////////////////////////////////////////////////
 QueryRequest::QueryRequest( XrdSsiSession* session, std::shared_ptr<JobQuery> const& jobQuery) :
-  _session(session), _jobQuery(jobQuery), _jobDesc(_jobQuery->getDescription()),
-  _jobId(jobQuery->getId()) {
-    LOGF_INFO("New QueryRequest with payload(%1%)" % _jobDesc.payload().size());
-    LOGF_DEBUG("QueryRequest JQ_jobId=%1%" % _jobId);
+  _session{session}, _jobQuery{jobQuery},
+  _jobId{std::string("jobId=" + jobQuery->getIdStr())} {
```
andy-slac: `std::string` is superfluous here

core/modules/qdisp/QueryRequest.cc
```diff
@@ -71,29 +70,48 @@ QueryRequest::~QueryRequest() {
 
 // content of request data
 char* QueryRequest::GetRequest(int& requestLength) {
-    requestLength = _jobDesc.payload().size();
-    LOGF_DEBUG("Requesting, payload size: [%1%]" % requestLength);
+    std::lock_guard<std::mutex> lock(_finishStatusMutex);
+    if (_finishStatus != ACTIVE) {
+        LOGF_WARN("%1% QueryRequest::GetRequest called after job finished (cancelled?)" % _jobId);
+        return const_cast<char*>("");
```
andy-slac: Yuck. Of course we should blame xrootd for this but maybe we can avoid cast, e.g.:

    static char* empty = "";
    return empty;

core/modules/qdisp/QueryRequest.cc
```diff
@@ -146,47 +166,61 @@ bool QueryRequest::_importStream() {
...
     return true;
 }
 
 void QueryRequest::ProcessResponseData(char *buff, int blen, bool last) { // Step 7
-    LOGF_INFO("ProcessResponseData with buflen=%1% %2%" % blen % (last ? "(last)" : "(more)"));
+    LOGF_INFO("%1% ProcessResponseData with buflen=%2% %3%" % _jobId % blen % (last ? "(last)" : "(more)"));
+    // Work with a copy of _jobQuery so it doesn't get reset underneath us by a call to cancel().
+    JobQuery::Ptr jq = _jobQuery;
+    {
+    std::lock_guard<std::mutex> lock(_finishStatusMutex);
```
andy-slac: inconsistent indentation inside block

core/modules/qdisp/QueryRequest.cc
jgates108: The alarm bells from seeing const_cast might be more useful than hiding it. I'm a bit more concerned about requestLength not being set and if that is correct.

core/modules/qdisp/QueryRequest.cc
```diff
@@ -213,6 +252,15 @@ bool QueryRequest::isCancelled() {
...
 void QueryRequest::cleanup() {
+    LOGF_DEBUG("%1% QueryRequest::cleanup()" %_jobId);
+    {
+        std::lock_guard<std::mutex> lock(_finishStatusMutex);
+        if (_finishStatus == ACTIVE) {
+            LOGF_ERROR("%1% QueryRequest::cleanup called before finished" % _jobId);
+            return;
+        }
+    }
+    // These need to be outside the mutex lock, or you could delete the mutex before it is freed.
```
andy-slac: I don't understand this comment, can you elaborate? Does it apply only to `_keepAlive`?

core/modules/qdisp/QueryRequest.h
```diff
@@ -109,34 +109,40 @@ class QueryRequest : public XrdSsiRequest {
     /// Called by xrootd when new data is available.
     virtual void ProcessResponseData(char *buff, int blen, bool last);
 
+    std::string getXrootdErr();
```
andy-slac: I could not find definition for this method anywhere?

core/modules/qdisp/QueryRequest.cc
jgates108: I changed the comment to "These need to be outside the mutex lock, or you could delete _finishStatusMutex before it is unlocked." Resetting either one of _jobQuery or _keepAlive could result in this object being deleted immediately, or nothing happening to the object for a while. It depends on the state of the rest of the system. As long as either one is not null, then this object can't be deleted.

core/modules/util/threadSafe.h
```diff
@@ -73,21 +73,23 @@ class Flag {
 
     /** Sets flag value to 'val' and returns the old value of flag.
      */
-    virtual T set(T val) {
+    virtual T exchange(T val) {
```
andy-slac: For the record - I do not like this class :smile_cat: It should be removed, together with Sequential.

core/modules/qdisp/QueryRequest.cc
jgates108: Comment at line 292, about _keepAlive, might help.

core/modules/qdisp/QueryRequest.h
jgates108: QueryRequest.cc line 351

core/modules/qdisp/QueryRequest.cc
andy-slac: That sound dangerous. If _jobQuery reset causes this object to disappear then using _keepAlive is undefined behavior.

core/modules/qdisp/QueryRequest.h
andy-slac: I think it's different method with (int*) argument (declared below)

core/modules/util/threadSafe.h
jgates108: Flag is serving a specific purpose. There is a case where std::atomic is what is needed most of the time, but you cannot access the underlying locking mechanism, and occasionally you have to have access to the mutex or there's no way prevent a race condition. Flag looks like an atomic, but you can get at the mutex if you need to. The example of this is Executive::_cancelled.

andy-slac: Well, my view of that is that:
- it should not be used as atomic, because it's very inefficient
- in many other cases when we need to have locks for extended periods we are already using standard mutex classes so there is little benefit for having this separate class
- many people (me including) think that recursive mutex is an indication of a bad/inefficient design

core/modules/qdisp/QueryRequest.cc
jgates108: I think this should work. It depends on only being called by _finish or _errorFinish, so I added some comments to that effect.
```
if (_keepAlive != nullptr) {
        _jobQuery.reset();
        _keepAlive.reset();
} else {
        _jobQuery.reset();
}
```

core/modules/qdisp/QueryRequest.h
jgates108: Ah, unused declaration deleted.

core/modules/util/threadSafe.h
jgates108: If I could use atomic, I would use atomic. In at least one instance I cannot. Using the syntax of atomic where applicable is clear and concise.

The class already exists (written before atomic was available).

Recursive mutex design is up for debate. The recursive_mutex could be avoided, but I doubt it is worth the effort.

core/modules/qdisp/QueryRequest.cc
andy-slac: Or something like (though maybe less obvious):

    std::shared_ptr<JobQuery> jq(std::move(_jobQuery));
    std::shared_ptr<QueryRequest> keep(std::move(_keepAlive));
    return;

core/modules/qdisp/Executive.cc
fritzm: Or just copy-assign the whole thing to a local JobMap on the stack and iterate on that?  Then you don't need to either reserve or for/push_back?

core/modules/qdisp/QueryRequest.cc
```diff
@@ -220,29 +268,36 @@ void QueryRequest::cleanup() {
...
+
+    if (!_retried.exchange(true) && !shouldCancel) {
         // There's a slight race condition here. _jobQuery::runJob() creates a
         // new QueryResource object which is used to create a new QueryRequest object
         // which will replace this one in _jobQuery. The replacement could show up
         // before this one's cleanup is called, so this will keep this alive.
+        LOGF_WARN("%1% QueryRequest::_errorFinish retrying" % _jobId);
         _keepAlive = _jobQuery->getQueryRequest(); // shared pointer to this
-        if (_jobQuery->runJob()) {
+        if (!_jobQuery->runJob()) {
```
fritzm: Aha! :-)

core/modules/qdisp/QueryRequest.cc
fritzm: I'd +1 Andy's recommend here -- it seems a lot less cumbersome to me?  Maybe prefix with a one-line comment to clarify:
```C++
// Vacate both _jobQuery and _keepAlive before freeing either
```

core/modules/qdisp/Executive.cc
jgates108: deque is reasonably efficient, it's got a a very small code footprint, it's simple, and I'd be incredibly surprised if a profiler indicated this would be a good focus for optimization.

core/modules/qdisp/QueryRequest.cc
jgates108: changed

core/modules/qdisp/Executive.cc
andy-slac: But why do you need it? Deque is typically used when you need insert/delete from both ends, which you don't do. And I don't agree on efficiency, it's inefficient compared to vector (if you don't need insert/delete at begin()). Vector should be a default container for anything, using any other container type should only be done after profiling :)

PR #174, opened 2016-01-05
core/modules/log cleanup (DM-3037)

jbecla:


fritzm:
The constants are still in the lsst::qserv::log namespace -- should they be moved to lsst::qserv::ccontrol now?  Yay for one less module!

PR #175, opened 2016-01-06
Fix Dockerfile producing Qserv build

fjammes:
- git pull removed (weak if initial clone build fails)
- eupspkg -er build removed
  (strange docker behaviour on PREFIX/lib/python)

### DIFF
--- a/admin/tools/docker/tagged/Dockerfile.tpl
+++ b/admin/tools/docker/tagged/Dockerfile.tpl
@@ -7,12 +7,10 @@ RUN apt-get update && apt-get -y install byobu git vim gdb lsof
 
 USER qserv 
 
-WORKDIR /home/qserv
+ENV QSERV_SRC_DIR /home/qserv/src/qserv
 
-RUN mkdir src && cd src && git clone -b {{GIT_TAG_OPT}} --single-branch https://github.com/LSST/qserv
-WORKDIR /home/qserv/src/qserv
-RUN bash -c ". /qserv/stack/loadLSST.bash && setup -r . -t qserv-dev && eupspkg -er build"
 # Force execution of lines below by changing timestamp
-RUN git pull # Build performed on {{TIMESTAMP}}
+RUN git clone -b {{GIT_TAG_OPT}} --single-branch https://github.com/LSST/qserv $QSERV_SRC_DIR # Build performed on {{TIMESTAMP}}
+WORKDIR $QSERV_SRC_DIR
 RUN bash -c ". /qserv/stack/loadLSST.bash && setup -r . -t qserv-dev && eupspkg -er install"
 RUN bash -c ". /qserv/stack/loadLSST.bash && setup -r . -t qserv-dev && eupspkg -er decl -t qserv-dev"


admin/tools/docker/tagged/Dockerfile.tpl
```diff
@@ -3,16 +3,17 @@ MAINTAINER Fabrice Jammes <fabrice.jammes@in2p3.fr>
 
 USER root
 
-RUN apt-get update && apt-get -y install byobu git vim gdb lsof
+RUN apt-get update
```
fritzm: I would leave these in one line, so the updates will only make a single new layer.

admin/tools/docker/tagged/Dockerfile.tpl
```diff
@@ -3,16 +3,17 @@ MAINTAINER Fabrice Jammes <fabrice.jammes@in2p3.fr>
...
+RUN apt-get -y install byobu git vim gdb lsof
 
 USER qserv 
 
 WORKDIR /home/qserv
 
-RUN mkdir src && cd src && git clone -b {{GIT_TAG_OPT}} --single-branch https://github.com/LSST/qserv
-WORKDIR /home/qserv/src/qserv
-RUN bash -c ". /qserv/stack/loadLSST.bash && setup -r . -t qserv-dev && eupspkg -er build"
+RUN mkdir src
```
fritzm: Same here -- I think these are better as one line, so there's one new layer for add the sources.  If you put the "cd" back, you won't need the extra "WORKDIR" line either, since you lost the "-er build"?

admin/tools/docker/tagged/Dockerfile.tpl
```diff
@@ -3,16 +3,17 @@ MAINTAINER Fabrice Jammes <fabrice.jammes@in2p3.fr>
...
+WORKDIR /home/qserv/src
 # Force execution of lines below by changing timestamp
-RUN git pull # Build performed on {{TIMESTAMP}}
+RUN git clone -b {{GIT_TAG_OPT}} --single-branch https://github.com/LSST/qserv # Build performed on {{TIMESTAMP}}
```
fritzm: Do we really need/want a full git clone (all history, branches, metadata etc.) in our docker images?  Maybe we should try something like this instead, to just get the tip of the source tree:

    git archive --format=tar --remote=<repository URL> {{GIT_TAG_OPT}} | tar xf -

fjammes: It's a good idea but github prevent archiving:
https://help.github.com/articles/can-i-archive-a-repository/

admin/tools/docker/tagged/Dockerfile.tpl
fjammes: ok, I do this. But note it is sometime convenient to have 2 layers for debugging purpose.

admin/tools/docker/tagged/Dockerfile.tpl
fjammes: Good point, I'll do it in one line, please note that git clone can create src/qserv.

admin/tools/docker/tagged/Dockerfile.tpl
ktlim: A tarball can be retrieved through the API: https://developer.github.com/v3/repos/contents/#get-archive-link specifies https://api.github.com/repos/lsst/qserv/tarball/master   Also through the site: https://github.com/lsst/qserv/archive/master.tar.gz

fjammes: Good to know, thanks :-) For now, I'll stay with the standard git command, and I'll move to this API if more speed is needed.

PR #178, opened 2016-01-11
Ticket/dm 4677

jgates108:


core/modules/memman/MemMan.cc
```diff
@@ -0,0 +1,43 @@
+// -*- LSST-C++ -*-
+/*
+ * LSST Data Management System
+ * Copyright 2015 LSST Corporation.
+ *
```
jgates108: 2016

core/modules/memman/MemMan.h
```diff
@@ -0,0 +1,177 @@
+// -*- LSST-C++ -*-
+/*
+ * LSST Data Management System
+ * Copyright 2015 LSST Corporation.
```
jgates108: 2016

core/modules/memman/MemManNone.h
```diff
@@ -0,0 +1,83 @@
+// -*- LSST-C++ -*-
+/*
+ * LSST Data Management System
+ * Copyright 2015 LSST Corporation.
```
jgates108: 2016

core/modules/memman/MemManNone.h
```diff
@@ -0,0 +1,83 @@
...
+                        for (auto it=tables.begin() ; it != tables.end(); it++) {
+                             if (it->theData  == TableInfo::MUSTLOCK
+                             ||  it->theIndex == TableInfo::MUSTLOCK)
+                                {errno = ENOMEM; return 0;}
+                        }
+                        return 1;
+                   }
+
+    virtual bool  unlock(Handle handle) {(void)handle; return true;}
+
```
jgates108:
```
virtual bool unlock(Handle) {return true;}
```
I think is a little cleaner, but this is just place holder code, so not a big deal. Same applies to (void)chunk; above and getStatus below.

abh3: I'd rather leave the (void)x's because, depending on the compilation option, you may get a warning about an unused parameter. This squelches the warning.

core/modules/memman/MemManNone.h
```diff
@@ -0,0 +1,83 @@
...
+// This class define a memory manager implementation that basically does
+// nothing. If a table needs to be locked it says that there is no memory to
+// do so. However, flexible locking is allowed. Eventually, this will be
+// replaced by an actual implementation. For now, this allows testing.
+
+class MemManNone : public MemMan {
+public:
+
+    virtual Handle lock(std::vector<TableInfo> const& tables,
+                        unsigned int chunk) {
```
jgates108: Remove virtual and add override for all functions overriden from MemMan

core/modules/memman/MemMan.h
```diff
@@ -0,0 +1,177 @@
...
+//-----------------------------------------------------------------------------
+//! @brief Describe a table that can be potentially locked in memory.
+//-----------------------------------------------------------------------------
+
+class TableInfo {
+public:
+
+    std::string tableName;    //< Name of the table
+
+    enum LockOptions {
```
fritzm: Use the improved "enum class" here, since we've cut over to c++11?

core/modules/memman/MemMan.h
```diff
@@ -0,0 +1,177 @@
...
+    //! @brief Create a memory manager and initialize for processing.
+    //!
+    //! @param  maxBytes   - Maximum amount of memory that can be used
+    //! @param  dbPath     - Path to directory where the database resides
+    //!
+    //! @return !0: The pointer to the memory manager.
+    //! @return  0: A manager could not be created.
+    //-----------------------------------------------------------------------------
+
+    static MemMan *create(unsigned long long maxBytes,
```
fritzm: Prefer the new c++11 fixed-width integer types for "large" values, unless you have an argument otherwise?

core/modules/memman/MemMan.h
```diff
@@ -0,0 +1,177 @@
...
+                         );
+
+    //-----------------------------------------------------------------------------
+    //! @brief Lock a set of tables in memory for a particular chunk.
+    //!
+    //! @param  tables - Reference to the tables to process.
+    //! @param  chunk  - The chunk number associated with the tables.
+    //!
+    //! @return =0     - Nothing was locked. The errno variable holds the
+    //!                  reason, as follows:
```
fritzm: Will errno return meet the thread-safety needs of this module (whatever they are?)

core/modules/memman/MemMan.h
```diff
@@ -0,0 +1,177 @@
...
+    //!                  EFAULT - filesystem or memory error
+    //!                  ENOENT - a chunk was missing
+    //!                  ENOMEM - insufficient memory to fully satisfy request
+    //! @return !0     - Is the resource handle associated with this request.
+    //-----------------------------------------------------------------------------
+
+    typedef unsigned int Handle;
```
fritzm: I have a slight preference for the new c++11 "using" for these now, for consistency with templated classes.  But I don't think it's policy anywhere yet, and I don't know how the other team members feel?  If you'd rather leave it as a typedef, if we decide to we can catch them all in the future on some Jacek-style repo-wide syntax correction cl :)

core/modules/memman/MemMan.h
```diff
@@ -0,0 +1,177 @@
...
+    //!
+    //! @param  maxBytes   - Maximum amount of memory that can be used
+    //! @param  dbPath     - Path to directory where the database resides
+    //!
+    //! @return !0: The pointer to the memory manager.
+    //! @return  0: A manager could not be created.
+    //-----------------------------------------------------------------------------
+
+    static MemMan *create(unsigned long long maxBytes,
+                          std::string const &dbPath
```
fritzm: Below you have "const& tables", which I think is the dominant style in our code.  I would let this slide, but you must pick one or the other consistently :)

core/modules/memman/MemMan.h
fritzm: Also, un-sized "unsigned int" here implies to me a "small" number.  How many unique handle values will there be?  If its not many, I'm okay with this.  For numbers that are "large", prefer the new c++11 fixed size types.

core/modules/memman/MemMan.h
```diff
@@ -0,0 +1,177 @@
...
+    //!                  EFAULT - filesystem or memory error
+    //!                  ENOENT - a chunk was missing
+    //!                  ENOMEM - insufficient memory to fully satisfy request
+    //! @return !0     - Is the resource handle associated with this request.
+    //-----------------------------------------------------------------------------
+
+    typedef unsigned int Handle;
+
+    virtual Handle lock(std::vector<TableInfo> const& tables,
+                        unsigned int chunk
```
fritzm: It looks like the rest of the qserv code is using plain "int" for chunk id right now, so maybe go with that instead of unsigned?  It actually bothers me that we don't have an explicit typedef or alias for chunk ids -- I will open a new issue for that...

core/modules/memman/MemMan.h
```diff
@@ -0,0 +1,177 @@
...
+
+    //-----------------------------------------------------------------------------
+    //! @brief Obtain statistics about this memory manager.
+    //!
+    //! @return The statistics.
+    //-----------------------------------------------------------------------------
+
+    struct Statistics {
+        unsigned long long bytesLockMax;//< Maximum number of bytes to lock
+        unsigned long long bytesLocked; //< Current number of bytes locked
```
fritzm: See above re. explicitly sized types for "large" numbers.

core/modules/memman/MemMan.h
```diff
@@ -0,0 +1,177 @@
...
+    //!
+    //! @return The query status. If the resource was not found numTables is
+    //!         set to zero.
+    //-----------------------------------------------------------------------------
+
+    struct Status {
+        unsigned long long bytesLock; //< Number of resource bytes locked
+        unsigned int       numTables; //< Number of tables resource has
+        unsigned int       chunk;     //< Chunk number associated with resource
+    };
```
fritzm: Types, per above

core/modules/memman/MemMan.h
```diff
@@ -0,0 +1,177 @@
...
+
+    virtual Status getStatus(Handle handle) = 0;
+
+    //-----------------------------------------------------------------------------
+
+    MemMan & operator=(const MemMan&) = delete;
+    MemMan(const MemMan&) = delete;
+
```
fritzm: zOMG, ampersands all *over* the place! :)

core/modules/memman/MemMan.h
```diff
@@ -0,0 +1,177 @@
...
+//-----------------------------------------------------------------------------
+//! @brief The memory manager.
+//!
+//! The MemMan is an abstract class the defines the interface to the memory
+//! manager that is used to lock database chunks in memory.
+//-----------------------------------------------------------------------------
+
+class MemMan {
+public:
+
```
fritzm: Consider providing
```C++
    using Ptr = std::shared_ptr<MemMan>;
```
as a courtesy/convenience to clients, as done often elsewhere in our source?

core/modules/memman/MemMan.cc
```diff
@@ -0,0 +1,43 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the LSST License Statement and
+ * the GNU General Public License along with this program.  If not,
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+
```
fritzm: Please add explicitly here:
```C++
// Class Header
#include "MemMan.h"
```

core/modules/memman/MemManNone.h
```diff
@@ -0,0 +1,83 @@
...
+                  MemManNone(unsigned long long maxBytes)
+                           {memset(&_myStats, 0, sizeof(_myStats));
+                            _myStats.bytesLockMax = maxBytes;
+                            _myStats.bytesLocked  = maxBytes;
+                            memset(&_status, 0, sizeof(_status));
+                           }
+    virtual      ~MemManNone() {}
+
+private:
+Statistics _myStats;
```
fritzm: indent?

core/modules/memman/MemMan.h
abh3: Agreed, that's a good idea.

core/modules/memman/MemMan.cc
jbecla: preferably with module name in front? 
#include "memman/MemMan.h"

core/modules/memman/MemMan.h
abh3: That's fine, though I'm still conflicted on whether to uniformly use "size_t" which is a standard kernel size type and what MemMan is all about. So, I may replace ull with size_t everywhere.

core/modules/memman/MemMan.h
abh3: Yes, errno is thread-local.

core/modules/memman/MemMan.h
abh3: Yes, I should change it to a longer type as I don't plan on optimizing value usage.

core/modules/memman/MemMan.h
abh3: Yes, I would prefer an explicit typedef for chunk ID. I used unsigned because I don't want to deal with negative numbers. I can change it back to int until we get an official typedef.

core/modules/memman/MemMan.cc
abh3: Ah, OK I see that the rule is to always include the director that the include file comes from. As for explicit includes... is that the rule even though it's obvious you would be including the file multiple times because the explicit dependence requires an implicit include?

core/modules/memman/MemMan.h
smonkewitz: I know it's the LSST convention, but... I question whether it's a good idea. I think things should be called what they are, and obscuring shared_ptr with a broad generic term to save such a small number of characters doesn't seem like a good trade-off to me.

fritzm: Fair enough -- it also saves dragging in the whole shared_ptr header here for what is only a speculative use.  I'm fine leaving it out.

core/modules/memman/MemMan.cc
smonkewitz: I think the intent behind making an implementation file Foo.cc include its header Foo.h first no matter what is that the compiler will then verify that Foo.h has included everything needed to make itself useable by others. Otherwise it's easy to forget to include say `<string>` in Foo.h because Foo.cc includes it before including Foo.h. At that point someone else using Foo.h but not `<string>` will run into compile errors.

fritzm: Including a .h first in its corresponding .cpp ensures the .h properly pulls in all its own dependencies and can stand alone.  This helps flush out "this was only compiling because somebody included that first, which included that which uncluded that, now somebody else changed their header and i'm busted".  Also ensures your header clients won't have to go hunting to see what they have to include first for your thing to work, etc?

fritzm: Uh yeah, what he said :)

core/modules/memman/MemMan.h
jgates108: I don't think it really hides any details and the reduced number of characters adds up pretty quickly. Not a big deal here as shared_ptr isn't used at all in this code.

core/modules/memman/MemMan.cc
abh3: Ah, got it. I was confused as to what include was wanted because of the low level amount of context. Sure!

core/modules/memman/MemMan.h
smonkewitz: To me a raw pointer / unique_ptr / shared_ptr are very different. As a newcomer looking at a Ptr, I wouldn't immediately know which kind of pointer is in use. In that sense, Ptr hides more than just detail to me, and would probably cause a hiccup in my reading of some code. Once you know the convention that Ptr == shared_ptr it's moot, so maybe I'm making too big of a deal over it.

abh3: I totally agree. Why they didn't call it ShrPtr is beyond me but it certainly would have been more descriptive.

ktlim: Because we discouraged the use of raw pointers and very rarely used scoped_ptrs (or auto_ptrs), only shared_ptr needed a typedef.

core/modules/memman/MemFile.h
```diff
@@ -0,0 +1,138 @@
...
+    //!
+    //! @return The filename.
+    //-----------------------------------------------------------------------------
+
+    std::string filePath() {return _fPath;}
+
+    //-----------------------------------------------------------------------------
+    //! @brief Lock database file in memory.
+    //!
+    //! @param  maxBytes- maximum file size       for locking the table file.
```
jgates108: Extra spaces.

core/modules/memman/MemFile.h
```diff
@@ -0,0 +1,138 @@
...
+    //!                   A value of zero skips the check.
+    //! @param  minRefs - minimum reference count for locking the table file.
+    //!
+    //! @return MLResult  When bLocked > 0 this number of bytes locked.
+    //!                   When bLocked = 0 no bytes were locked and retc holds
+    //!                   the reason as follows:
```
jgates108: Missing reasons.

core/modules/memman/MemFile.h
```diff
@@ -0,0 +1,138 @@
...
+    //! @return =0        retc = 0 the file did not meet minRefs restriction;
+    //!                   otherwise, it holds the errno failure code.
+    //-----------------------------------------------------------------------------
+
+    struct MLResult {size_t bLocked;
+                     int    retc;
+                     MLResult() {}
+                     MLResult(size_t lksz, int rc) : bLocked(lksz), retc(rc) {}
+                    };
```
jgates108: size_t should be on a new line indented 4 spaces. other entries should lineup with size_t, end bracket should line up with struct.

core/modules/memman/MemFile.cc
```diff
@@ -0,0 +1,179 @@
...
+
+/******************************************************************************/
+/*                               m e m L o c k                                */
+/******************************************************************************/
+
+MemFile::MLResult MemFile::memLock(size_t maxBytes, int minRefs) {
+
+MemInfo mInfo;
+std::lock_guard<std::mutex> gaurd(cacheMutex);
+
```
jgates108: Indent 4 spaces.

core/modules/memman/MemFile.cc
```diff
@@ -0,0 +1,179 @@
...
+    // If the file is already locked, indicate success
+    //
+    if (_isLocked)
+       {MLResult aokResult(_memInfo.size(), 0);
+        return aokResult;
+       }
```
jgates108: Opening bracket should be on same line as if. Closing bracket should line up with if. This applies to all of the if statements.

core/modules/memman/MemFile.h
```diff
@@ -0,0 +1,138 @@
...
+    //! @param  fPath   - The path to the file.
+    //! @param  mem     - Reference to the associated memory object.
+    //! @param  mInfo   - Initial value of the MemInfo object for the file.
+    //-----------------------------------------------------------------------------
+
+           MemFile(std::string const& fPath, Memory& mem, MemInfo const& minfo)
+                   : _fPath(fPath), _memory(mem), _memInfo(minfo),
+                     _refs(1), _isLocked(0) {}
+
+          ~MemFile() {}
```
jgates108: Constructor and destructor should be indented 4 spaces.

Default values for _refs and _isLocked could be defined with their declarations so they would only be defined in one place if multiple constructors are used.

core/modules/memman/MemFile.h
```diff
@@ -0,0 +1,138 @@
...
+
+std::string _fPath;
+Memory&     _memory;
+MemInfo     _memInfo;
+int         _refs;      // Protected by cacheMutex
+bool        _isLocked;  // Ditto
```
jgates108: Indent 4 spaces.

core/modules/memman/MemFile.h
jgates108: Applies to all struct definitions. They should look like class definitions.

core/modules/memman/MemFile.h
```diff
@@ -0,0 +1,138 @@
...
+#include <string>
+#include <unistd.h>
+
+// Qserv headers
+#include "memman/Memory.h"
+
+// System headers
+#include <atomic>
+#include <string>
+
```
jgates108: All the system headers should be grouped together.

core/modules/memman/MemFile.h
```diff
@@ -0,0 +1,138 @@
...
+//! @brief Description of a memory based file.
+//-----------------------------------------------------------------------------
+
+class MemFile {
+public:
+
+    //-----------------------------------------------------------------------------
+    //! @brief Get file path for this table.
+    //!
+    //! @return The filename.
```
jgates108: File name or path?

core/modules/memman/MemFile.cc
```diff
@@ -0,0 +1,179 @@
...
+    return errResult;
+}
+
+/******************************************************************************/
+/*                              n u m F i l e s                               */
+/******************************************************************************/
+
+uint32_t MemFile::numFiles() {
+
+std::lock_guard<std::mutex> gaurd(cacheMutex);
```
jgates108: All local variable definitions should be indented 4 spaces.

core/modules/memman/MemFile.cc
```diff
@@ -0,0 +1,179 @@
...
+namespace memman {
+
+/******************************************************************************/
+/*                  L o c a l   S t a t i c   O b j e c t s                   */
+/******************************************************************************/
+  
+namespace {
+std::mutex                                 cacheMutex;
+std::unordered_map<std::string, MemFile *> fileCache;
+}
```
jgates108: I dislike local static objects, but I'm having trouble coming up with a legitimate complaint. I'd feel better if these were members of MemMan or some other managing class. They are pretty well buried here and it limits it to one instance of MemManReal, but I can't think of any case where there would be a need for more than one MemMan.

core/modules/memman/MemFile.cc
```diff
@@ -0,0 +1,179 @@
...
+    // return the object as it may be shared.
+    //
+    auto it = fileCache.find(fPath);
+    if (it != fileCache.end())
+       {if (&(it->second->_memory) != &mem)
+           {MFResult errResult(0, EXDEV);
+            return errResult;
+           }
+        it->second->_refs++;
+        MFResult aokResult(&(*(it->second)),0);
```
jgates108: Might help clarity to a line like `auto memFile = it->second;`

core/modules/memman/MemFile.cc
abh3: This is actually quite legitimate. There is no reason for this object being anything other file local. It does not limit the number of memory managers you may have (you may have any number). What it does mean is that a particular file cannot belong to more than one memory manager which is quite reasonable given the kernel restrictions on memory mapping and locking.

core/modules/memman/MemManReal.cc
```diff
@@ -0,0 +1,231 @@
...
+bool        mustLock;
+
+    // Make sure a table was specified
+    //
+    _numLocks++;
+    if (tables.size() == 0)
+       {_numErrors++;
+        errno = EINVAL;
+        return HandleType::INVALID;
+       }
```
jgates108: I'm not sure this empty table check is necessary. If this if statement is deleted this function will return HandleType::ISEMPTY, which should work without causing problems and means the caller wont have to make a separate code path for an empty list.

core/modules/memman/MemFile.cc
jgates108: It works and it's legitimate, but it is buried very deep. In the off chance 2 memory managers are created and one of them, let's say B, releases a file that doesn't belongs to A due to a bug, it could be very difficult to notice this underlying connection between the 2 different objects for people not familiar with the code.

core/modules/memman/MemFileSet.cc
```diff
@@ -0,0 +1,136 @@
...
+
+    // Calculate the number of bytes available at this point. Note that we force
+    // freeBytes to be atleast 1 to make memlock check before it tries locking
+    // the file to avoid a useless memory map operation if it can't be locked.
+    // By the time we get here someone else may have already locked the file.
+    //
+    bytesMax    = _memory.bytesMax();
+    bytesLocked = _memory.bytesLocked();
+    if (bytesMax <= bytesLocked) freeBytes = 1;
+       else freeBytes = bytesMax - bytesLocked;
```
jgates108: size_t freeBytes = std::max(bytesMax - bytesLocked, 1);

core/modules/memman/MemFileSet.cc
```diff
@@ -0,0 +1,136 @@
...
+    // Try locking as many flexible files as we can. We only lock the file table
+    // if the reference count >= 2 to optimize memory usage. At some point we
+    // will place unlocked flex files on a "want to lock" queue. FUTURE!!!
+    //
+    for (auto mfP : _flexFiles) {
+        mlResult = mfP->memLock(freeBytes, 2);
+        if (mlResult.bLocked == 0) continue;
+        _lockBytes += mlResult.bLocked;
+        if (freeBytes > mlResult.bLocked) freeBytes -= mlResult.bLocked;
+           else freeBytes = 1;
```
jgates108: This check shows up a lot. Might be worth adding and inline function to do it.

core/modules/memman/MemManReal.cc
```diff
@@ -0,0 +1,222 @@
...
+
+std::lock_guard<std::mutex> gaurd(hanMutex);
+
+    // Delete all of the file set entries that we own via handle cache. The
+    // file set destructor will unlock any memory that it needs to unlock.
+    //
+    auto it = hanCache.begin();
+
+    while(it != hanCache.end()) {
+         if (it->second->isOwner(_memory)) {
```
jgates108: `for (auto it = hanCache.begin(), end = hanCache.end(); it != end; ) {`

core/modules/memman/MemManReal.cc
```diff
@@ -0,0 +1,222 @@
...
+            delete it->second;
+             hanCache.erase(it++);
```
jgates108: `it = hanCache.erase(it);`

core/modules/memman/Memory.cc
```diff
@@ -0,0 +1,141 @@
...
+MemInfo Memory::memLock(std::string const& fPath) {
+
+MemInfo     mInfo;
+struct stat sBuff;
+int         fdNum;
+
+    // We first open the file. we currently open this R/W because we want to
+    // disable copy on write operations when we memory map the file.
+    //
+    if ((fdNum = open(fPath.c_str(), O_RDWR)) < 0 || fstat(fdNum, &sBuff))
```
jgates108: LSST wants fdNum = open() outside of the if statement.

core/modules/memman/Memory.h
```diff
@@ -0,0 +1,173 @@
...
+    void    memRel(MemInfo& mInfo);
+
+    //-----------------------------------------------------------------------------
+    //! Constructor
+    //!
+    //! @param  dbDir  - Directory path to where managed files reside.
+    //! @param  memSZ  - Size of memory to manage.
+    //-----------------------------------------------------------------------------
+            Memory(std::string const& dbDir, size_t memSZ)
+                  : _dbDir(dbDir), _maxBytes(memSZ), _lokBytes(0) {}
```
jgates108: Might be better to set the default value of  _lokBytes where it is declared.

core/modules/memman/MemFileSet.cc
```diff
@@ -0,0 +1,136 @@
...
+    // if the reference count >= 2 to optimize memory usage. At some point we
+    // will place unlocked flex files on a "want to lock" queue. FUTURE!!!
+    //
+    for (auto mfP : _flexFiles)
+        {mlResult = mfP->memLock(freeBytes, 2);
+         if (mlResult.bLocked == 0) continue;
+         _lockBytes += mlResult.bLocked;
+         if (freeBytes > mlResult.bLocked) freeBytes -= mlResult.bLocked;
+            else freeBytes = 1;
+        }
```
jgates108: Looking at this, I think we want to track how often flex lock fails to lock the file. If it happens a lot, we will want to know.

core/modules/memman/MemFileSet.cc
abh3: Well, actually exactly twice verbatim and once in elaboration. Writing a lambda for something that repeats twice seems like over-kill.

core/modules/memman/MemFileSet.cc
abh3: Hi John,

Did you delete this comment? I can't find it in github inorder to reply to 
it. If you did delete then I assume that's a change of mind on this.

Andy

jgates108: Yeah, I hit enter and then went, right, that isn't going to work for 
unsigned and deleted it.

core/modules/memman/MemManReal.cc
```diff
@@ -0,0 +1,229 @@
...
+
+    // Delete all of the file set entries that we own via handle cache. The
+    // file set destructor will unlock any memory that it needs to unlock.
+    //
+    auto it = hanCache.begin();
+
+    while(it != hanCache.end()) {
+         if (it->second->isOwner(_memory)) {
+            delete it->second;
+            hanCache.erase(it++);
```
jgates108: I had a comment here, but it disappeared. It was `it = hanCache.erase(it);`

core/modules/memman/MemFile.cc
```diff
@@ -0,0 +1,181 @@
...
+std::unordered_map<std::string, MemFile *> fileCache;
+}
+
+/******************************************************************************/
+/*                               m e m L o c k                                */
+/******************************************************************************/
+
+MemFile::MLResult MemFile::memLock(size_t maxBytes, int minRefs) {
+
+    MemInfo mInfo;
```
andy-slac: IF default MemInfo constructor does not have side effects move the declaration to where you first initialize it.

core/modules/memman/MemFile.cc
```diff
@@ -0,0 +1,181 @@
...
+    return errResult;
+}
+
+/******************************************************************************/
+/*                              n u m F i l e s                               */
+/******************************************************************************/
+
+uint32_t MemFile::numFiles() {
+
+    std::lock_guard<std::mutex> gaurd(cacheMutex);
```
andy-slac: Is "g**au**rd" a word? :)

core/modules/memman/MemFile.cc
```diff
@@ -0,0 +1,181 @@
...
+
+/******************************************************************************/
+/*                                o b t a i n                                 */
+/******************************************************************************/
+  
+MemFile::MFResult MemFile::obtain(std::string const& fPath,
+                                  Memory& mem, bool isFlex) {
+
+    MemFile *mfP;
+    MemInfo  mInfo;
```
andy-slac: Same for these two declarations

core/modules/memman/MemFile.cc
```diff
@@ -0,0 +1,181 @@
...
+    // return the object as it may be shared. Note: it->second == MemFile*!
+    //
+    auto it = fileCache.find(fPath);
+    if (it != fileCache.end()) {
+        if (&(it->second->_memory) != &mem) {
+            MFResult errResult(0, EXDEV);
+            return errResult;
+        }
+        it->second->_refs++;
+        MFResult aokResult(&(*(it->second)),0);
```
andy-slac: Is not `&(*(it->second))` the same thing as `it->second`?

core/modules/memman/MemFile.cc
```diff
@@ -0,0 +1,181 @@
...
+    std::lock_guard<std::mutex> gaurd(cacheMutex);
+
+    // First look up if this table already exists in our cache and is using the
+    // the same memory object (error if not). If so, up the reference count and
+    // return the object as it may be shared. Note: it->second == MemFile*!
+    //
+    auto it = fileCache.find(fPath);
+    if (it != fileCache.end()) {
+        if (&(it->second->_memory) != &mem) {
+            MFResult errResult(0, EXDEV);
```
andy-slac: REplace 0 with nullptr.

core/modules/memman/MemFile.cc
```diff
@@ -0,0 +1,181 @@
...
+        it->second->_refs++;
+        MFResult aokResult(&(*(it->second)),0);
+        return aokResult;
+    }
+
+    // Validate the file and get its size
+    //
+    mInfo = mem.fileInfo(fPath);
+    if (!mInfo.isValid()) {
+        MFResult errResult(0, mInfo.errCode());
```
andy-slac: same here

core/modules/memman/MemMan.cc
```diff
@@ -0,0 +1,43 @@
...
+// Qserv Headers
+#include "MemManNone.h"
+#include "MemManReal.h"
```
jgates108: `#include "memman/MemManNone.h"`
`#include "memman/MemManReal.h"`

abh3: OK, I keep forgetting because our compilation option includes '-I.'. So, if the rule is always to add the module directory in the include then perhaps we should disallow the compiler from including things in the current directory, yes?

core/modules/memman/MemFile.cc
```diff
@@ -0,0 +1,181 @@
...
+        return errResult;
+    }
+
+    // Get a new file object and insert it into the map
+    //
+    mfP = new MemFile(fPath, mem, mInfo, isFlex);
+    fileCache.insert({fPath, mfP});
```
andy-slac: I'm staring at this for 10 minutes, I can't figure out how it works. Modern C++ miracles, I guess.

core/modules/memman/MemFile.cc
```diff
@@ -0,0 +1,181 @@
...
+    return aokResult;
+}
+
+/******************************************************************************/
+/*                               r e l e a s e                                */
+/******************************************************************************/
+
+void MemFile::release() {
+
+    std::lock_guard<std::mutex> gaurd(cacheMutex);
```
andy-slac: copy-and-paste works fine with typos :smile:

core/modules/memman/MemFile.h
```diff
@@ -0,0 +1,143 @@
...
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+
+#ifndef LSST_QSERV_MEMMAN_MEMFILE_H
+#define LSST_QSERV_MEMMAN_MEMFILE_H
+
+// Qserv headers
+#include "memman/Memory.h"
+
+// System headers
```
andy-slac: I believe system headers should go first

core/modules/memman/MemFile.h
```diff
@@ -0,0 +1,143 @@
...
+class MemFile {
+public:
+
+    //-----------------------------------------------------------------------------
+    //! @brief Get file path for this table.
+    //!
+    //! @return The file path.
+    //-----------------------------------------------------------------------------
+
+    std::string filePath() {return _fPath;}
```
andy-slac: Return by const ref?

core/modules/memman/MemFile.cc
abh3: I assume you mean the "insert". Well, {} is equivalent to an std::pair in this context. Yes, it's a modern C++ miracle.

core/modules/memman/MemFileSet.cc
```diff
@@ -0,0 +1,136 @@
...
+
+    // Obtain a memory file object for this table and chunk
+    //
+    MemFile::MFResult mfResult = MemFile::obtain(fPath, _memory, !mustLK);
+    if (mfResult.mfP == 0) return mfResult.retc;
+
+    // Add to the appropriate file set
+    //
+    if (mustLK) _lockFiles.push_back(mfResult.mfP);
+       else     _flexFiles.push_back(mfResult.mfP);
```
andy-slac: That hurts my eyes :cry: And does not match any of the LSST recommended styles :smile:

core/modules/memman/MemFileSet.cc
```diff
@@ -0,0 +1,136 @@
...
+
+/******************************************************************************/
+/*                                s t a t u s                                 */
+/******************************************************************************/
+  
+MemMan::Status MemFileSet::status() {
+
+    MemMan::Status myStatus;
+
+    // Fill out status informatiuon and return it.
```
andy-slac: informati**u**on

core/modules/memman/MemFile.h
abh3: I hate returning references to things inside an object that may be later deleted. In this case, I will just remove the method as it's not used anywhere in the code.

core/modules/memman/MemFileSet.cc
abh3: Ah, which part? Don't want you to cry :-)

core/modules/memman/MemFileSet.h
```diff
@@ -0,0 +1,120 @@
...
+ * the GNU General Public License along with this program.  If not,
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+
+#ifndef LSST_QSERV_MEMMAN_MEMFILESET_H
+#define LSST_QSERV_MEMMAN_MEMFILESET_H
+
+// System headers
+#include <string>
+#include <unistd.h>
```
andy-slac: is this for `size_t`? It's defined in <cstddef> in C++.

abh3: Ah, if only where that simple. unistd.h defines it as an unsigned long while cstddef defines it as unsigned int from what I can see. We definitely want the long version. I suppose I could change that to be uint64_t but I get that value from stat() where it is defined size_t and figured I'd standardize on that definition. I don't have feelings one way or another as long as it 64 bits on 64 bit machines,

core/modules/memman/MemFileSet.cc
andy-slac: The style I used to and what is recommended by LSST (https://confluence.lsstcorp.org/pages/viewpage.action?pageId=16908737, clause 6-7):

    if (mustLK) {
        _lockFiles.push_back(mfResult.mfP);
    } else {
        _flexFiles.push_back(mfResult.mfP);
    }

One could drop curlies for one-line body, but I prefer to only do it for for-without-else.

core/modules/memman/MemFileSet.h
andy-slac: This should not happen. On any platform `size_t` is guaranteed to be at least the size of pointer (size_t can be used to index any array/container type and the size of the container could be the whole virtual memory). I'd be very surprised to see different definitions in different headers, it should all come from some kernel-level typedef.

core/modules/memman/MemFile.cc
andy-slac: Right, `insert` is overloaded method with multiple signatures. How does compiler know that construct `{fPath, mfP}` is an `std::pair` and not anything else? Are there any possible surprises if you omit the actual type of that initializer? I guess I'm just to used to always providing explicit `value_type()`. For cases like this I'd probably prefer also-modern `fileCache.emplace(fPath, mfP);`

core/modules/memman/MemFileSet.h
```diff
@@ -0,0 +1,120 @@
...
+
+    ~MemFileSet();
+
+private:
+Memory&               _memory;
+std::vector<MemFile*> _lockFiles;
+std::vector<MemFile*> _flexFiles;
+size_t                _lockBytes;     // Total bytes locked
+int                   _numFiles;
+int                   _chunk;
```
andy-slac: Indent all data members?

core/modules/memman/MemMan.h
```diff
@@ -0,0 +1,183 @@
...
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+
+#ifndef LSST_QSERV_MEMMAN_MEMMAN_H
+#define LSST_QSERV_MEMMAN_MEMMAN_H
+
+// System headers
+#include <memory>
+#include <string>
+#include <vector>
```
andy-slac: Need `<cstdint>` header if you use portable integer types (intNN_t).

I also believe that all those portable integer types should be prefixed with `std::` if you include `<cstdint>` instead of `<stdint.h>` but no one is doing that :frowning:

core/modules/memman/MemManReal.cc
```diff
@@ -0,0 +1,228 @@
...
+namespace qserv {
+namespace memman {
+  
+/******************************************************************************/
+/*                         g e t S t a t i s t i c s                          */
+/******************************************************************************/
+  
+MemMan::Statistics MemManReal::getStatistics() {
+
+Statistics stats;
```
andy-slac: Not worth indenting? :smiling_imp:

core/modules/memman/MemManReal.cc
```diff
@@ -0,0 +1,228 @@
...
+}
+  
+/******************************************************************************/
+/*                                  l o c k                                   */
+/******************************************************************************/
+  
+MemMan::Handle MemManReal::lock(std::vector<TableInfo> const& tables, int chunk) {
+
+int         lockNum, flexNum, retc = 0;
+bool        mustLock;
```
andy-slac: I see a pattern here. :eyes:

core/modules/memman/MemManReal.h
```diff
@@ -0,0 +1,77 @@
...
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+
+#ifndef LSST_QSERV_MEMMAN_MEMMANREAL_H
+#define LSST_QSERV_MEMMAN_MEMMANREAL_H
+
+// System headers
+#include <atomic>
+#include <mutex>
+#include <unordered_map>
```
andy-slac: This header is not needed? But `<vector>`, `<string>`, `<cstdint>` should probably be here.

core/modules/memman/Memory.cc
```diff
@@ -0,0 +1,145 @@
...
+
+MemInfo     fInfo;
+struct stat sBuff;
+
+    // Simply issue a stat() to get the size
+    //
+    if (stat(fPath.c_str(), &sBuff)) fInfo._errCode = errno;
+       else {if (sBuff.st_size > 0) fInfo._memSize = sBuff.st_size;
+                else fInfo._errCode = ESPIPE;
+            }
```
andy-slac: My eyes :sob: 

Is not it the same as:

    if (stat(fPath.c_str(), &sBuff)) {
        fInfo._errCode = errno;
    } else if (sBuff.st_size > 0) {
        fInfo._memSize = sBuff.st_size;
    } else {
        fInfo._errCode = ESPIPE;
    }

core/modules/memman/Memory.cc
```diff
@@ -0,0 +1,145 @@
...
+struct stat sBuff;
+int         fdNum;
+
+    // We first open the file. we currently open this R/W because we want to
+    // disable copy on write operations when we memory map the file.
+    //
+    fdNum = open(fPath.c_str(), O_RDWR);
+    if (fdNum < 0 || fstat(fdNum, &sBuff))
+       {mInfo.setErrCode(errno);
+        return mInfo;
```
andy-slac: Do you want to close fdNum before returning here?

core/modules/memman/MemManReal.h
abh3: Granted, that was a left over before a refactored. As for <vector> and <string> those come from MemMan.h that must be included because the class inherits from it. I don't like sprinkling unnecessary includes across various files.

core/modules/memman/MemFileSet.h
abh3: Well, technically yes but std::size_t is official defined as the type sizeof() returns and I consider that pretty squishy. While in Unix land size_t is defined as the maximum size memory or a file or file offset can have. The latter is more closely aligned with what I am doing. Like I said, I am willing to remove any ambiguity by switching to uint64_t which makes it very clear.

andy-slac: Well, size_t is strictly for memory, file offset does not have anything to do with memory size (you should remember good old days when we had to read >2GB files on 32-bit OSes). For file offsets more portable type should be `off64_t` from `<sys/types.h>` but any 64-bit type should work ok for our purposes.

core/modules/memman/MemManNone.h
```diff
@@ -0,0 +1,82 @@
...
+                    {memset(&_myStats, 0, sizeof(_myStats));
+                     _myStats.bytesLockMax = maxBytes;
+                     _myStats.bytesLocked  = maxBytes;
+                     memset(&_status, 0, sizeof(_status));
+                    }
+         ~MemManNone() override {}
+
+private:
+Statistics _myStats;
+Status     _status;
```
jgates108: Indent all member variables.

core/modules/memman/MemManReal.h
```diff
@@ -0,0 +1,77 @@
...
+
+   ~MemManReal() override {unlockAll();}
+
+private:
+
+Memory           _memory;
+std::atomic_uint _numLocks;
+std::atomic_uint _numErrors;
+uint32_t         _numReqdFiles;  // Under control of hanMutex
+uint32_t         _numFlexFiles;  // Ditto
```
jgates108: Indent all member variables.

core/modules/memman/Memory.h
```diff
@@ -0,0 +1,187 @@
...
+            Memory(std::string const& dbDir, uint64_t memSZ)
+                  : _dbDir(dbDir), _maxBytes(memSZ), _lokBytes(0), _flexNum(0) {}
+
+           ~Memory() {}
+private:
+
+std::string        _dbDir;
+uint64_t           _maxBytes;
+std::atomic_ullong _lokBytes;
+std::atomic_uint   _flexNum;
```
jgates108: Indent all member variables.

core/modules/memman/MemFile.cc
```diff
@@ -0,0 +1,178 @@
...
+
+    // If the file is already locked, indicate success
+    //
+    if (_isLocked) {
+        if (_isFlex) _memory.flexNum(1);
+        MLResult aokResult(_memInfo.size(), 0);
+        return aokResult;
+    }
+
+    // If the file doesn't meet the refcount restrictio, don't lock it
```
fritzm: s/restrictio/restriction

core/modules/memman/MemFile.h
```diff
@@ -0,0 +1,144 @@
...
+class MemFile {
+public:
+
+    //-----------------------------------------------------------------------------
+    //! @brief Get file path for this table.
+    //!
+    //! @return The file path.
+    //-----------------------------------------------------------------------------
+
+//  std::string filePath() {return _fPath;}
```
fritzm: why commented out?

core/modules/memman/MemManNone.h
```diff
@@ -0,0 +1,82 @@
...
+    void  unlockAll() override {}
+
+    Statistics getStatistics() override {return _myStats;}
+
+    Status getStatus(Handle handle) override {(void)handle; return _status;}
+
+    MemManNone & operator=(const MemManNone&) = delete;
+    MemManNone(const MemManNone&) = delete;
+
+          MemManNone(uint64_t maxBytes)
```
fritzm: extra indentation here?

core/modules/memman/MemManNone.h
```diff
@@ -0,0 +1,82 @@
...
+// This class define a memory manager implementation that basically does
+// nothing. If a table needs to be locked it says that there is no memory to
+// do so. However, flexible locking is allowed. Eventually, this will be
+// replaced by an actual implementation. For now, this allows testing.
+
+class MemManNone : public MemMan {
+public:
+
+    Handle lock(std::vector<TableInfo> const& tables, int chunk) override {
+               (void)chunk;
```
fritzm: extra indentation here?

core/modules/memman/MemManNone.h
```diff
@@ -0,0 +1,82 @@
...
+#include <memory.h>
+
+// Qserv Headers
+#include "MemMan.h"
+
+namespace lsst {
+namespace qserv {
+namespace memman {
+
+// This class define a memory manager implementation that basically does
```
fritzm: s/define/defines

core/modules/memman/MemMan.h
```diff
@@ -0,0 +1,184 @@
...
+    //!                  xxxxxx - filesystem or memory error
+    //!                  ENOENT - a chunk was missing
+    //!                  ENOMEM - insufficient memory to fully satisfy request
+    //! @return !0     - Is the resource handle associated with this request.
+    //-----------------------------------------------------------------------------
+
+    using Handle = uint64_t;
+
+    struct HandleType {
+           static const Handle INVALID=0;
```
fritzm: extra indentation?  looks like you've got a 4 vs. 8 thing going on with your editor?

core/modules/memman/MemFileSet.cc
```diff
@@ -0,0 +1,139 @@
...
+    uint64_t bytesLocked, bytesMax, freeBytes;
+
+    // Calculate the number of bytes available at this point. Note that we force
+    // freeBytes to be atleast 1 to make memlock check before it tries locking
+    // the file to avoid a useless memory map operation if it can't be locked.
+    // By the time we get here someone else may have already locked the file.
+    //
+    bytesMax    = _memory.bytesMax();
+    bytesLocked = _memory.bytesLocked();
+    if (bytesMax <= bytesLocked) freeBytes = 1;
```
fritzm: hmm, you use fully line broken if/else above (line 68) but non-broken if/else here and below in the same file?  pick consistent style throughout file?

fritzm: P.S. don't pick the wrong one

core/modules/memman/MemFile.h
```diff
@@ -0,0 +1,144 @@
...
+            bool               isFlex)
+           : _fPath(fPath), _memory(mem), _memInfo(minfo),
+             _refs(1), _isFlex(isFlex) {}
+
+   ~MemFile() {}
+
+    std::string _fPath;
+    Memory&     _memory;
+    MemInfo     _memInfo;
+    int         _refs;               // Protected by cacheMutex
```
fritzm: assuming you wanted those comments aligned

core/modules/memman/MemFile.cc
```diff
@@ -0,0 +1,178 @@
...
+    //
+    if (_refs < minRefs) {
+        MLResult nilResult(0,0);
+        return nilResult;
+    }
+
+    // If a space is wanted, do so now before we attemptto lock the file
```
fritzm: s/attemptto/attempt to/

core/modules/memman/Memory.h
```diff
@@ -0,0 +1,187 @@
...
+
+    void    memRel(MemInfo& mInfo);
+
+    //-----------------------------------------------------------------------------
+    //! Constructor
+    //!
+    //! @param  dbDir  - Directory path to where managed files reside.
+    //! @param  memSZ  - Size of memory to manage.
+    //-----------------------------------------------------------------------------
+            Memory(std::string const& dbDir, uint64_t memSZ)
```
fritzm: extra indents?

core/modules/memman/MemFile.h
abh3: I should have deleted it as no one uses it anyway.

core/modules/memman/MemManNone.h
abh3: I scooted stuff over but I don't know what extra indentation you are referring to. What are the indentation  rules when you define a function body in the header file? Perhaps my (soon to be posted) correction is what you wanted.....

core/modules/memman/MemMan.h
abh3: Not at all since that was 7 vs 4. So, I scooted it over by three. Fell into the xroot format where we try to make things readable.

core/modules/memman/MemFileSet.cc
abh3: I don't think I can satisfy both you and Andy at the same time. Line 68 was preferred by AndyS for readability sake. Generally, simple one liner if/else don't seem to need to be fully bracketed if they read well (in fact fully bracketing simple ones makes harder to read). Line 68 was sufficiently confusing as the items were almost identical to warrant bracketing. I agree with consistency but not at the expense of readability, especially when both forms are generally allowed.

andy-slac: My take on that is that one-liners without brackets are OK, but this is not a one-liner, it's a regular if-else on two separate lines (strangely indented) and it should be bracketed. As it is now it's hard do read (for me). This applies to all if-elses.

PR #181, opened 2016-01-13
Improvements to our scons files (DM-3447)

andy-slac:
Replacing most of the core/module/SConscript with simpler structures and
algorithms. Uses regular shared libraries when building test
applications. Removed separate SConscript.test files, all
module-specific stuff now lives in regular SConscript file. Default
implementation for "standard SConscript" now takes care of thes
applications and unit tests as well.


core/modules/SConscript
```diff
@@ -1,35 +1,159 @@
...
+"""
+This SConscript file defines all targets produced from code in core/modules.
+
+This file is loaded from top-level SConstruct and its main purpose is to
+return tuple with two items - install targets and test targets - former is
+assigned to "install" alias, latter to "test" alias.
+
+The code in core/modules is split into a bunch of modules, modules do most
+of the job definig targets for their own code, this includes:
+  - making shared object (*.os) from C++ code
```
fritzm: typo: .so?

andy-slac: No, it's confusing but correct :) scons compiles code into *.o or *.os depending on whether it's static/shared object. Shared objects are normally collected into shared libs *.so (which are also called shared objects).

PR #183, opened 2016-01-17
Remove unused toString() functions, switched to used operator<< (DM-2452)

jbecla:


core/modules/qana/DuplSelectExprPlugin.cc
```diff
@@ -122,7 +120,7 @@ DuplSelectExprPlugin::getDuplicateSelectErrors(query::SelectStmt const& stmt) co
     if (LOG_CHECK_LVL(_log, LOG_LVL_DEBUG)) {
         std::ostringstream stream;
         selectList.dbgPrint(stream);
-        LOGS(_log, LOG_LVL_DEBUG, "Input stmt:\n" << stream.str());
+        LOGS(_log, LOG_LVL_DEBUG, "Input stmt:\n" << stream);
```
andy-slac: I think we should keep `.str()` here, pushing stream to stream is non-standard, I think we saw this already with LOGS migration

PR #184, opened 2016-01-19
U/fjammes/dm 4295

fjammes:


admin/tools/docker/configured/scripts/wait.sh
```diff
@@ -20,14 +20,19 @@
...
 set -e
 
-DIR=$(cd "$(dirname "$0")"; pwd -P)
-. $DIR/params.sh
+QSERV_RUN_DIR=/qserv/run
+
+
+while ! "$QSERV_RUN_DIR"/bin/qserv-status.sh > /dev/null
+    do
+    sleep 1
```
jbecla: We talking about adding a comment somewhere near a sleep like this one. Was it somewhere else?

admin/tools/docker/shmux/run-multinode-tests.sh
```diff
@@ -7,16 +7,25 @@ set -x
 
 . ./env.sh
 
-shmux -c "docker rm -f qserv; nohup docker run --name qserv --rm --net=host \
+ssh "$MASTER" "docker rm -f qserv; \
+    docker run --detach=true \
+    --name qserv --net=host \
```
fritzm: Since you are using --net=host, I don't think you need the pass-through port mapping below?  I guess it doesn't hurt anything though...

admin/tools/docker/configured/scripts/wait.sh
fjammes: A comment is in shmux/run-multinode-tests.sh where I call this script, I add also one here now.

admin/tools/docker/shmux/run-multinode-tests.sh
fjammes: Good point :+1: thanks

PR #185, opened 2016-01-21
Added lockInMem and scanSpeed for shared scans (DM-4807)

jbecla:


core/modules/css/CssAccess.cc
```diff
@@ -503,7 +503,8 @@ CssAccess::getPartTableParams(std::string const& dbName,
 
     std::vector<std::string> subKeys{"partitioning", "partitioning/subChunks",
         "partitioning/dirDb", "partitioning/dirTable", "partitioning/dirColName", "partitioning/latColName",
-        "partitioning/lonColName", "partitioning/overlap", "partitioning/secIndexColName"};
+        "partitioning/lonColName", "partitioning/overlap", "partitioning/secIndexColName",
+        "partitioning/lockInMem", "partitioning/scanSpeed"};
```
andy-slac: At the KV level I'd suggest to put those parameters into separate group, would simplify modification of those values if we need to do it later. Something like `"sharedScan/lockInMem", "sharedScan/scanSpeed"`.

core/modules/css/CssAccess.cc
```diff
@@ -533,6 +534,18 @@ CssAccess::getPartTableParams(std::string const& dbName,
         throw KeyValueError(tableKey + "/partitioning", "one of the sub-keys is not numeric: " +
                             std::string(exc.what()));
     }
+    auto iter = paramMap.find("partitioning/lockInMem");
+    if (iter != paramMap.end()) {
+        params.lockInMem = std::stoi(paramMap["partitioning/lockInMem"]);
+    } else {
+      params.lockInMem = false;
```
andy-slac: indentation

core/modules/css/CssAccess.cc
```diff
@@ -533,6 +534,18 @@ CssAccess::getPartTableParams(std::string const& dbName,
...
+    }
+    iter = paramMap.find("partitioning/scanSpeed");
+    if (iter != paramMap.end()) {
+        params.scanSpeed = std::stoi(paramMap["partitioning/scanSpeed"]);
+    } else {
+        params.scanSpeed = 0;
+    }
```
andy-slac: I'd put this inside try block above to better protect against non-numeric values.

core/modules/css/CssAccess.cc
```diff
@@ -533,6 +534,18 @@ CssAccess::getPartTableParams(std::string const& dbName,
...
 
     return params;
 }
```
andy-slac: `createTable()` method below needs to be modified to store these new parameters.

admin/python/lsst/qserv/admin/partConfig.py
```diff
@@ -184,6 +184,8 @@ def cssTableOptions(self):
             options['subChunks'] = bool(int(self.get('part.subChunks', 1)))
             options['dirDb'] = self['dirDb']
             options['dirTable'] = self['dirTable']
+            options['lockInMem'] = 'lockInMem' in self ? self['lockInMem'] : False
+            options['scanSpeed'] = 'scanSpeed' in self ? self['scanSpeed'] : 0
```
andy-slac: This looks like C++ except that `self` needs to be replaced with `this` :) This won't compile unfortunately, I'd replace it with more pythonic:

    options['lockInMem'] = self.get('lockInMem', False)
    options['scanSpeed'] = self.get('scanSpeed', 0)

core/modules/css/PartTableParams.h
```diff
@@ -100,6 +104,16 @@ struct PartTableParams {
...
         return res;
     }
+
+    /// Returns true if table should be locked in memory for shared scan
+    bool lockInMemory() const {
+        return lockInMem;
+    }
+
+    /// Returns the prefered speed of shared scan for the table
+    int getScanSpeed() const {
```
andy-slac: These two new methods are not needed, this is a struct and members can be accessed directly.

admin/python/lsst/qserv/admin/partConfig.py
andy-slac: I'd probably also created separate "group" in config file for the shared scan-related options, so the code would be something like:

    options['lockInMem'] = int(self.get('sharedScan.lockInMem', False))
    options['scanSpeed'] = int(self.get('sharedScan.scanSpeed', 0))

and config file would look like:

    sharedScan = {
        lockInMem = 1
        scanSpeed = 3
    }

core/modules/css/CssAccess.cc
jbecla: I'm not too sure, please rethink. We have some code that blindly assumes the prefix is "partitioning", see eg "_storePacked(tableKey + "/partitioning", partMap);", or "throw KeyValueError(tableKey + "/partitioning"", that code would need to be rewritten to distinguish between the two prefixes.

core/modules/css/CssAccess.cc
jbecla: Actually, see my most recent commit. I didn't like repeating a large block of code so I simplified it. I think it is better now, though a bit less efficient, but these functions are marked "deprecated" anyway. Let me know if you are ok with that simplification. I'll squash it all into one commit later.

core/modules/css/CssAccess.cc
jbecla: oh right, done

core/modules/css/PartTableParams.h
jbecla: I only added it because we had similarly trivial accessors there. Ok, I removed them.

core/modules/css/CssAccess.cc
andy-slac: I still think it's better to save them in two separate groups in KV. Partitioning parameters are supposed to be stable, but you may want to modify shared scan parameters. It would be easier to implement updating is shared scan parameters are in their own group. And I'm still a bit unsure that adding shared scan parameters to `PartTableParams` struct is the way to go, we may want to rethink it in the future.

core/modules/css/CssAccess.cc
```diff
@@ -467,74 +467,16 @@ MatchTableParams
...
-    // fill the structure
-    params.dirTable1 = paramMap["dirTable1"];
-    params.dirColName1 = paramMap["dirColName1"];
-    params.dirTable2 = paramMap["dirTable2"];
-    params.dirColName2 = paramMap["dirColName2"];
-    params.flagColName = paramMap["flagColName"];
-
-    return params;
+    TableParams params = getTableParams(dbName, tableName);
+    return params.match;
```
andy-slac: No, please don't do this. Potentially you are reading (or trying to read) much more than you need here but then throw away big fraction of it. This code is supposed to be very efficient (it's executed by czar for every table in a query), we need to avoid possible overheads. The only reason that this separate method exist is to make its implementation more efficient, otherwise everyone could just use `getTableParams()`.

core/modules/css/CssAccess.cc
```diff
@@ -579,8 +522,20 @@ CssAccess::getTableParams(std::string const& dbName, std::string const& tableNam
         if (iter != paramMap.end()) {
             params.partitioning.overlap = std::stod(iter->second);
         }
+        iter = paramMap.find("partitioning/lockInMem");
+        if (iter != paramMap.end()) {
+            params.partitioning.lockInMem = std::stoi(paramMap["partitioning/lockInMem"]);
+        } else {
+            params.partitioning.lockInMem = false;
```
andy-slac: TYhe default value should be set by default constructor, you don't really need it here.

core/modules/css/CssAccess.cc
jbecla: Hmm, now I am confused, sorry :). Confused by your comment "I'm still a bit unsure that adding shared scan parameters to PartTableParams struct is the way to go". I thought that is what you suggested when we talked on Wednesday. (My original idea was to have a dedicated struct for shared scan parameters that'd be like PartTableParams or MatchTableParams, and we'd add it to TableParams). And now you are suggesting moving to a dedicated prefix for shared scans parameters instead of "partitioning", putting config values in their own namespace, and now you are unsure if they should even be in PartTableParams. Are you saying I should try going with my original idea?

core/modules/css/CssAccess.cc
jbecla: Ok, but the comment in the header clearly says "Deprecated: use getTableParams() instead to get a consistent set of all table parameters.", which to me means that these two functions are going away soon. If we want to keep them, then how about we define two private functions

_getPartTableParams(std::string const& dbName, std::string const& tableName, PartTableParams& param);
_getMatchTableParams(std::string const& dbName, std::string const& tableName, MatchTableParams& param);

and then in getTableParams:
TableParams params;
_getPartTableParams(dbName, tableName, params.match);
_getMatchTableParams(dbName, tableName, params.partitioning),


MatchParams mParams;
_getMatchTableParams(dbName, tableName, mParams),

in getMatchTableParams()

and

PartParams pParams;
_getPartTableParams(dbName, tableName, pParams),

in getPartTableParams()

core/modules/css/CssAccess.cc
andy-slac: No, I suggest we stick with current PartTableParams for now, but be prepared to separate them in the future. CSS will evolve and I don't assume that what we do today will be the final solution, it depends a lot on how these parameters are used or updated. I guess what I'm saying let's implement simplest thing today but be prepared to change it when we understand it better. If you want to make separate structure for it it's OK with me but then we need to think about CssAccess interface as well, because that structure is used differently from regular partitioning parameters.

core/modules/css/CssAccess.cc
andy-slac: OK, I have already forgotten that I wanted to deprecate those two methods. Still it does not matter, until we actually remove them I propose to keep things as they are now. Implementing `getTableParams()` in terms of two other methods is incorrect (it runs in two separate queries which is inefficient and potentially inconsistent). Your original idea (implementing two other methods based on `getTableParams()`) is better but has other issues as I mentioned above. 

Actually deprecating those two methods and removing them needs significant changes in czar and I do not know how soon it's going to happen, again, until we are ready to do that I suggest we keep with the current approach.

core/modules/css/CssAccess.cc
```diff
@@ -511,29 +506,29 @@ CssAccess::getPartTableParams(std::string const& dbName,
...
+CssAccess::getScanTableParams(std::string const& dbName,
+                              std::string const& tableName) const {
+    LOGS(_log, LOG_LVL_DEBUG, "getScanTableParams(" << dbName << ", " << tableName << ")");
+    _checkVersion();
+
+    std::string const tableKey = _prefix + "/DBS/" + dbName + "/TABLES/" + tableName;
+
+    ScanTableParams params;
+
+    std::vector<std::string> subKeys{"sharedScan", "sharedScan/lockInMem", "sharedScan/scanSpeed"};
```
andy-slac: "sharedScan" is not needed here, we are only using subkeys from this key. In case of "partitioning" key situation is different, we check the presence of that key to decide whether table is partitioned or not.

core/modules/css/CssAccess.cc
```diff
@@ -626,6 +602,13 @@ CssAccess::createTable(std::string const& dbName,
         _storePacked(tableKey + "/partitioning", partMap);
     }
 
+    // save shared scan info
+    std::map<std::string, std::string> partMap{
+        std::make_pair("lockInMem", std::to_string(int(scanParams.lockInMem))),
+        std::make_pair("scanSpeed", std::to_string(int(scanParams.scanSpeed)))
+    };
+    _storePacked(tableKey + "/sharedScan", partMap);
+
```
andy-slac: This creates `sharedScan` key even for non-partitioned tables which does not make much sense, I propose to move this inside `if` above.

andy-slac: Another thing to consider - maybe we want to save scan parameters only if they are non-default?

PR #193, opened 2016-02-03
Tickets/DM-4931: Use Apple CommonCrypto rather than OpenSSL

timj:
These patches switch the digest creation code from OpenSSL to Apple CommonCrypto on Apple systems. The changes to `StringHash.cc` do not seem to be optimal so I would welcome advice on how to get this to work with the templating rather than writing a forwarding function. The issue is that the API for the Apple versions of `MD5` and `SHA256` seem to have a slightly different signature to the OpenSSL variants.

### DIFF
--- a/core/modules/proto/TaskMsgDigest.cc
+++ b/core/modules/proto/TaskMsgDigest.cc
@@ -27,6 +27,10 @@
 // System headers
 #ifdef __SUNPRO_CC
 #include <sys/md5.h>
+#elif __APPLE__
+#define COMMON_DIGEST_FOR_OPENSSL
+#include <CommonCrypto/CommonDigest.h>
+#define MD5 CC_MD5
 #else // Linux?
 #include <openssl/md5.h>
 #endif
--- a/core/modules/util/StringHash.cc
+++ b/core/modules/util/StringHash.cc
@@ -33,38 +33,33 @@
 #ifdef __APPLE__
 #define COMMON_DIGEST_FOR_OPENSSL
 #include <CommonCrypto/CommonDigest.h>
-
-// On OS X MD5() is CC_MD5() the size argument CC_LONG is a uint32_t
-// and the data argument is a const void *.
-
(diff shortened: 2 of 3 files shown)


core/modules/proto/TaskMsgDigest.cc
```diff
@@ -27,6 +27,10 @@
 // System headers
 #ifdef __SUNPRO_CC
 #include <sys/md5.h>
+#elif __APPLE__
+#define COMMON_DIGEST_FOR_OPENSSL
+#include <CommonCrypto/CommonDigest.h>
+#define MD5 CC_MD5
```
fritzm: Is this #define wanted/needed?  It looks like you define "MD5" as a non-macro below?

timj: Where? Apple do not define `MD5()` so I have to alias it here.

fritzm: Line 44 of StringHash.cc below?  I will pull the branch and look more closely so I can see what's going on...

timj: Right, but this is a different file that also uses `MD5()`.

fritzm: ah, yes

fritzm:
Tim -- I pushed some minor suggested cleanup to StringHash.cc:

* would like to avoid mixing code into our #include section, so moved it down to the anon namespace
* make the explicit static casts on calling the common crypto functions explicit, so they're easier to notice
* minor formatting changes throughout for consistency with current qserv style

I can't build this right now on my El Cap (blowing up in base package right now for some reason), so if you agree to these changes please check it in your build env. to make sure I didn't mess anything up?

Thanks!

fritzm:
Oh, and I'm not concerned about the single-line forwarding functions -- the extra dispatch will be swamped in comparison to the digest functions being called, and will probably be optimized away anyway.

PR #194, opened 2016-02-04
Remove dependency on MySQLdb module (DM-3949)

andy-slac:
Updated wmgr code to use sqlalchemy instead of MySQLdb module.

Same is done for tests/MySqlUdf.py, plus updated that test script for
most recent scisql function names and signatures.

core/examples/obsolete/loader.py still has an import for MySQLdb, we
keep that obsolete script just in case someone wants to look at it but
we do not use it anywhere.

Note that I did not remove mysqlpython from ups table file as we still
depend on MySQLdb as a driver for sqlalchemy. We may switch to some
other driver in the future, then we can replace this dependency.

### DIFF
--- a/core/modules/wmgr/python/config.py
+++ b/core/modules/wmgr/python/config.py
@@ -38,7 +38,6 @@ import tempfile
 from .errors import ExceptionResponse
 from lsst.db.engineFactory import getEngineFromArgs
 from lsst.qserv import css
-import MySQLdb as sql
 
 #----------------------------------
 # Local non-exported definitions --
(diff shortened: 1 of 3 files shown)


core/modules/wmgr/python/config.py
```diff
@@ -122,21 +121,16 @@ def privDbEngine(self):
...
         if self.proxyPort: kwargs['port'] = self.proxyPort
-        if self.proxyUser: kwargs['user'] = self.proxyUser
+        if self.proxyUser: kwargs['username'] = self.proxyUser
         _log.debug('creating new connection %s', kwargs)
-        if self.proxyPasswd: kwargs['passwd'] = self.proxyPasswd
-        conn = sql.connect(**kwargs)
-        return conn
+        if self.proxyPasswd: kwargs['password'] = self.proxyPasswd
+        inst = getEngineFromArgs(**kwargs)
+        return inst
```
jbecla: why not just "return getEngineFromArgs(**kwargs);"?

I wonder if we should mention in the logging something about "password hidden". I can imagine the current version might lead to confusion if someone is debugging connection related issue.

andy-slac: I'll add "(password not shown)" to the message and remove intermediate "inst" variable.

PR #195, opened 2016-02-04
Tickets/DM-5030: Allow Qserv tests to pass on OS X El Capitan

timj:
Minimum patch required to get Qserv to finally build on OS X El Capitan.

The `build/` directory in path was the only wrinkle. In theory we could make more extensive changes to how libraries and binaries are built with `@rpath` support but this seemed like a quick way to get things to work. I'm not sure if I can get the value of `build_dir` from the Scons environment rather than hard-coding it in as `build`.

### DIFF
--- a/site_scons/site_tools/unittest.py
+++ b/site_scons/site_tools/unittest.py
@@ -40,10 +40,12 @@ Example which also checks unit test results:
 """
 
 import os
+import platform
 import shutil
 
 from SCons.Builder import Builder
 from SCons.Errors import StopError
+from SCons.Platform import platform_default
 
 
 class _unitTest(object):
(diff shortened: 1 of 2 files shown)


site_scons/site_tools/unittest.py
```diff
@@ -61,9 +63,30 @@ def __call__(self, target, source, env):
...
+            release_str = platform.release()
+            release_major = int(release_str.split('.')[0])
+            if release_major >= 15:
+                if "LSST_LIBRARY_PATH" in os.environ:
+                    os.environ["DYLD_LIBRARY_PATH"] = os.environ["LSST_LIBRARY_PATH"]
+
+            # We need to tell the tests where the uninstalled Qserv libraries
+            # are located because rpaths aren't (yet?) handled properly
+            # when the libraries and binaries are built on OS X.
+            os.environ["DYLD_LIBRARY_PATH"] = "build/:" + os.environ["DYLD_LIBRARY_PATH"]
```
andy-slac: Do you need to change environment DYLD_LIBRARY_PATH? If it's for `libpathstr` then local variable should be enough. 

I'd replace hard-coded `build` with the value from SCons env:

    os.environ["DYLD_LIBRARY_PATH"] = env['build_dir'] + ":" + os.environ["DYLD_LIBRARY_PATH"]

andy-slac: Also this code will be executed multiple times (once per unit test) so it will extend DYLD_LIBRARY_PATH in the environment every time in case the above `if` does not reset it.

timj:
Thanks for the hints. Patch updated.

PR #196, opened 2016-02-05
Tickets/DM-5022: Modernize Python usage

timj:
Modernize the python code in Qserv to Python 2.7 style. Mainly:

* Print function
* `in` rather than `has_key()`
* Octal numbers use "o"
* List comprehensions rather than map/lambda

### DIFF
--- a/admin/bin/qserv-admin.py
+++ b/admin/bin/qserv-admin.py
@@ -644,7 +644,7 @@ def main():
             # wait for commands and process
             parser.receiveCommands()
         except(KeyboardInterrupt, SystemExit, EOFError):
-            print("")
+            print()
 
 if __name__ == "__main__":
     main()
--- a/admin/bin/qserv-admin.py
+++ b/admin/bin/qserv-admin.py
@@ -32,6 +32,7 @@ Known issues and todos:
  - many commands still need to be implemented
  - need to separate dangerous admin commands like DROP EVERYTHING
 """
+from __future__ import print_function
 
 # standard library imports
 import ConfigParser
@@ -368,7 +369,7 @@ class CommandParser(object):
         """
         Print available commands.
         """
-        print self._supportedCommands
+        print(self._supportedCommands)
 
     def _parseRelease(self, tokens):
         """
@@ -390,7 +391,7 @@ class CommandParser(object):
         if t == 'DATABASES':
             names = self._css.getDbNames()
             for name in names:
-                print name
+                print(name)
         elif t == 'NODES':
             self._parseShowNodes()
         else:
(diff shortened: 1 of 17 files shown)


admin/bin/qserv-admin.py
```diff
@@ -636,14 +637,14 @@ def main():
             try:
                 parser.parse(cmd)
             except (_ToolError, css.CssError) as e:
-                print "ERROR: ", e.__str__()
+                print("ERROR: ", e.__str__())
```
andy-slac: Not related to modernization, but should not use special methods like `__str__` directly, we'll need to make another round of cleanup for those I guess.

admin/bin/qserv-admin.py
```diff
@@ -636,14 +637,14 @@ def main():
...
                 sys.exit(1)
     else:
         try:
             # wait for commands and process
             parser.receiveCommands()
         except(KeyboardInterrupt, SystemExit, EOFError):
-            print ""
+            print("")
```
andy-slac: Is not this the same as `print()`?

admin/bin/qserv-admin.py
timj: I did not spot that. I'll replace them with `str(e)`.

admin/bin/qserv-admin.py
timj: Yes. Will fix.

admin/python/lsst/qserv/admin/nodeMgmt.py
```diff
@@ -123,11 +123,11 @@ def selectDict(self, state=None, nodeType=None):
         """
 
         # if state is a string make a list out of it
-        if isinstance(state, types.StringTypes):
+        if isinstance(state, str):
```
andy-slac: Does not this leave out `unicode` in Python2?

timj: Yes, I had not meant to leave that change in. Sorry. This needs to be dealt properly for python portability (including the option of jut going to python 3).

site_scons/site_tools/unittest.py
```diff
@@ -91,14 +92,14 @@ def __call__(self, target, source, env):
                 shutil.move(out, out + '.failed')
                 msg = '*** Unit test failed, check log file ' + out + '.failed ***'
                 sep = '*' * len(msg)
-                print sep + '\n' + msg + '\n' + sep
+                print(sep + '\n' + msg + '\n' + sep)
```
andy-slac: Should we modernize further:

    print(sep, msg, sep, sep='\n')

Mixing `sep` as argument and keyword argument is confusing, I'd rename `sep` variable to something else

timj: I'm happy to modernize further.

PR #199, opened 2016-02-12
Tickets/dm 2900

fjammes:


admin/tools/docker/work/Dockerfile
```diff
@@ -15,7 +15,8 @@ COPY scripts/*.sh /home/dev/scripts/
 RUN chown -R dev:dev /home/dev/.eups && \
     chown -R dev:dev /home/dev/scripts && \
     chmod g+w /qserv/stack/ && \
-    chmod -R g+w /qserv/stack/Linux64/mysql
+    chmod -R g+w /qserv/stack/Linux64/mysql/*/lib/plugin && \
```
andy-slac: Will this fail if stack does not have mysql package?

fjammes: ok removed, we use mariadb now.

PR #201, opened 2016-02-23
Tickets/dm 4440

fjammes:


core/modules/wbase/Base.h
```diff
@@ -65,20 +65,6 @@ extern std::string const CREATE_DUMMY_SUBCHUNK_SCRIPT;
 void updateResultPath(char const* resultPath=0);
 void clearResultPath();
 
-// Hashing-related
-std::string hashToPath(std::string const& hash);
-std::string hashToResultPath(std::string const& hash);
```
andy-slac: Do you want to remove these two methods from .cc file as well?

core/modules/wdb/QueryRunner.cc
```diff
@@ -135,13 +134,12 @@ bool QueryRunner::runQuery() {
     Release release(_task, this);
 
     if (_task->getCancelled()) {
-        LOGS(_log, LOG_LVL_DEBUG, "runQuery, task was cancelled before it started. "
-             << _task->dbName);
+        LOGF(_log, LOG_LVL_DEBUG, "runQuery, task was cancelled before it started. %1%" % _task->hash);
```
andy-slac: Don't do this, we are removing LOGF macros from log and replacing them with LOGS. Same for the message below.

core/modules/wbase/Base.h
fjammes: :+1: thanks

PR #204, opened 2016-03-04
Fix minor issues in docker procedure

fjammes:
- params.sh was missing at configuration
- startup.py wasn't importing correctly module "utils"
- remove unused parameters in params.sh

### DIFF
--- a/admin/tools/docker/work/scripts/eups-declare.sh
+++ b/admin/tools/docker/work/scripts/eups-declare.sh
@@ -60,10 +60,8 @@ if [ $# -ne 0 ] ; then
     exit 2
 fi
 
-. /qserv/scripts/params.sh
-
-. $INSTALL_DIR/loadLSST.bash
+. /qserv/stack/loadLSST.bash
 
 echo "Setup Qserv from source version"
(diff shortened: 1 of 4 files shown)


admin/tools/docker/work/scripts/eups-declare.sh
```diff
@@ -60,10 +60,8 @@ if [ $# -ne 0 ] ; then
     exit 2
 fi
 
-. /qserv/scripts/params.sh
-
-. $INSTALL_DIR/loadLSST.bash
+. /qserv/stack/loadLSST.bash
```
jgates108: I'm not sure hard coding the directory is the best way to go here.

fjammes: I understand, I was using a parameter, but in practice it never changes (we are inside a Docker container so /qserv/stack might be a good convention), so I go back to hardcoded, which avoid managing a param file between all scripts/containers and ease maintenance (if you change the param file in the release container then you have to generate all child container, which is not easy).

I'll implement a more refined mechanism if have a parameter for stack install dir become useful.

PR #206, opened 2016-03-09
Tickets/dm 4755

jgates108:


core/modules/qdisp/Executive.cc
```diff
@@ -125,7 +129,7 @@ void Executive::add(JobDescription const& jobDesc) {
         return;
     }
     if (_empty.exchange(false)) {
-        LOGS(_log, LOG_LVL_DEBUG, "Flag _empty set to false by jobId=" << jobQuery->getIdStr());
+        LOGS(_log, LOG_LVL_DEBUG, "Flag _empty set to false by QId" << jobQuery->getIdStr());
```
andy-slac: Just checking, this message will appear as:

    Flag _empty set to false by QIdQI=12345:

Is this as intended?

core/modules/qdisp/Executive.cc
```diff
@@ -210,7 +214,7 @@ void Executive::markCompleted(int jobId, bool success) {
                 throw Bug(msg);
             }
         }
-        LOGS(_log, LOG_LVL_ERROR, "Executive: error executing jobId=" << jobId
+        LOGS(_log, LOG_LVL_ERROR, "Executive: error executing QId" << jobId
```
andy-slac: Is `jobId` a job number (as opposed to query ID)? Should it be prefixed with `QId`?

core/modules/qdisp/Executive.cc
```diff
@@ -306,7 +310,7 @@ void Executive::_setup() {
   */
 bool Executive::_track(int jobId, std::shared_ptr<JobQuery> const& r) {
     assert(r);
-    LOGS(_log, LOG_LVL_DEBUG, "Attempt to TRACK jobId="
+    LOGS(_log, LOG_LVL_DEBUG, "Attempt to TRACK QId"
          << _id << "_" << jobId << " to Executive ("<< this <<") tracked jobs");
```
andy-slac: Use `_idStr` instead of `_id`?

core/modules/qdisp/Executive.h
```diff
@@ -91,7 +93,9 @@ class Executive {
 
     bool getEmpty() { return _empty; }
 
-    std::string& getId() { return _id; }
+    void setQueryId(qmeta::QueryId id);
+    qmeta::QueryId getId() { return _id; }
```
andy-slac: Make method `const`?

core/modules/qdisp/Executive.h
```diff
@@ -91,7 +93,9 @@ class Executive {
...
+    std::string& getIdStr() { return _idStr; }
```
andy-slac: Should it be a const method returning const reference?

andy-slac:
One general comment - I noticed that you make a strings from QeryId+JobId in several places, I'd probably introduced a separate class to encapsulate that, something like:

    struct JobId {
        uint64_t qId;
        int jobId;
    }
    friend ostream& operator<<(ostream& out, JobId const& jobId) {
        return "QI=" << jobId.qId << ":" << jobId.jobId << ";";
    }

and probably added `operator<()` to so that you could use it with `std::set` without doing another string conversion.

core/modules/qdisp/Executive.cc
jgates108: I went through and replaced all of them with proper queryId and jobIds and got rid of the places where the executive was identified by `<< this <<` in log messages

PR #212, opened 2016-03-29
Fix MySQL connection close handling

vaikuntht:
If the MySQL pointer was allocated by `mysql_init()`, and you plan to reuse the connection descriptor, then it is important to set it to `NULL` after `mysql_close()`, otherwise the next init will think it's already allocated, then the next call for `mysql_real_connect()` will cause a SEGFAULT. This is what was observed and I've added a new function to handle closing this connection properly.

### DIFF
--- a/core/modules/sql/SqlConnection.cc
+++ b/core/modules/sql/SqlConnection.cc
@@ -126,14 +126,18 @@ SqlConnection::~SqlConnection() {
 
 bool
 SqlConnection::connectToDb(SqlErrorObject& errObj) {
-    assert(_connection.get());
-    if (_connection->connected()) return true;
+    if (_connection->connected()) {
+        int rc = mysql_ping(_connection->getMySql());
+        if (rc == 0) return true;
+        _connection->closeMySqlConn();
+    }
+
     if (!_connection->connect()) {
         _setErrorObject(errObj);
         return false;
-    } else {
-        return true;
     }
+
+    return true;
 }
 
 bool
--- a/core/modules/mysql/MySqlConnection.cc
+++ b/core/modules/mysql/MySqlConnection.cc
@@ -94,14 +94,21 @@ MySqlConnection::~MySqlConnection() {
             while((row = mysql_fetch_row(_mysql_res))); // Drain results.
             _mysql_res = nullptr;
         }
-        mysql_close(_mysql);
+        closeMySqlConn();
     }
+}
+
+void
+MySqlConnection::closeMySqlConn() {
+    // Close mysql connection and set deallocated pointer to null
+    mysql_close(_mysql);
+    _mysql = nullptr;
 }
 
 bool
 MySqlConnection::connect() {
     // Cleanup garbage
-    if (_mysql) { mysql_close(_mysql); }
+    if (_mysql) { closeMySqlConn(); }
     _isConnected = false;
     // Make myself a thread
     _mysql = _connectHelper();
--- a/core/modules/mysql/MySqlConnection.h
+++ b/core/modules/mysql/MySqlConnection.h
@@ -55,6 +55,7 @@ public:
 
     ~MySqlConnection();
 
+    void closeMySqlConn();
     bool connect();
 
     bool connected() const { return _isConnected; }


core/modules/sql/SqlConnection.cc
```diff
@@ -126,14 +127,29 @@ SqlConnection::~SqlConnection() {
 
 bool
 SqlConnection::connectToDb(SqlErrorObject& errObj) {
-    assert(_connection.get());
-    if (_connection->connected()) return true;
+    assert(_connection != nullptr);
+    static int lastChecked = 0;
+    time_t currentTime = time(0);
```
jgates108: lastChecked and currentTime don't seem to serve a purpose. I think they can be deleted.

core/modules/sql/SqlConnection.cc
```diff
@@ -27,6 +27,7 @@
 // System headers
 #include <cstdio>
 #include <cstddef>
+#include <ctime>
```
jgates108: It might be better to use std::chrono instead of ctime, but I don't think you need either here.

core/modules/sql/SqlConnection.cc
```diff
@@ -126,14 +127,29 @@ SqlConnection::~SqlConnection() {
 
 bool
 SqlConnection::connectToDb(SqlErrorObject& errObj) {
-    assert(_connection.get());
-    if (_connection->connected()) return true;
+    assert(_connection != nullptr);
+    static int lastChecked = 0;
```
andy-z: Should this be per-connection variable (member) and not global?

core/modules/sql/SqlConnection.cc
```diff
@@ -126,14 +127,29 @@ SqlConnection::~SqlConnection() {
...
+
+    if (_connection->connected()) {
+        int rc = mysql_ping(_connection->getMySql());
+        if (rc == 0) {
+            if (currentTime < lastChecked+60) return true;
+            lastChecked = currentTime;
+            return true;
+        } else {
+            _connection->closeMySqlConn();
+            _connection->connect();
```
andy-z: `_connection->connect();` is called again below, so it will be called twice in this case.

core/modules/sql/SqlConnection.cc
```diff
@@ -126,14 +127,29 @@ SqlConnection::~SqlConnection() {
...
-    assert(_connection.get());
-    if (_connection->connected()) return true;
+    assert(_connection != nullptr);
+    static int lastChecked = 0;
+    time_t currentTime = time(0);
+
+    if (_connection->connected()) {
+        int rc = mysql_ping(_connection->getMySql());
+        if (rc == 0) {
+            if (currentTime < lastChecked+60) return true;
```
andy-z: Is 60 a magic number? What does this logic try to protect from - disconnects due to inactivity timeout (which should be longer than 60 seconds) or random server disconnects (then this logic does not really work)?

andy-z: Sorry, ignore this comment, I'm confused by trying to guess intention here. Code as it is written does not do much useful, as John says currentTime and lastChecked can be removed. My first guess was that you want to limit ping frequency, but ping is done every time anyways.

andy-slac: andy-z is my cat, sorry for the noise :smiley_cat:

core/modules/sql/SqlConnection.cc
jgates108: I thought I commented on this. Anyway, all the instances of lastChecked can go, as I don't think there's any point in rate limiting connection attempts. And, as Andy says, lastChecked should not be static but a member variable if you do want to rate limit something. Sharing the variable across all class instances is probably not what you want to do.

vaikuntht: I agree, it should not be static but a per-connection variable. More about this below.

core/modules/sql/SqlConnection.cc
vaikuntht: The intention was not to limit ping-check rate but to make sure the connection actually existed, as `_connection->connected()` just reads off the value of a saved class variable (bool) initiated when the connection was first made. A "timeout" would've made it so that variable would still say `true` even after the connection had timed out. 

As it is applied here I agree it doesn't do anything since I convoluted the multiple commits from my original test branch. I'll make sure it's fine to just get rid of the timing stuff entirely and clean up.

core/modules/sql/SqlConnection.cc
vaikuntht: Thanks for catching that, I'll remove this call.

PR #213, opened 2016-03-31
Tickets/dm 5307

jgates108:


core/modules/ccontrol/UserQuerySelect.cc
```diff
@@ -248,11 +247,7 @@ QueryState UserQuerySelect::join() {
     bool successful = _executive->join(); // Wait for all data
```
andy-slac: Will this return `false` if the query was only partially submitted (`_executive->getCancelled()` was true in the above loop)?

core/modules/util/InstanceCount.cc
```diff
@@ -0,0 +1,92 @@
...
+    if (iter != _instances.end()) {
+        iter->second -= 1;
+        LOGS(_log, LOG_LVL_DEBUG, "~InstanceCount " << iter->first << "=" << iter->second << " : " << *this);
+    } else {
+        LOGS(_log, LOG_LVL_ERROR, "~InstanceCount " << _className << " was not found! : " << *this);
+    }
+}
+
+
+InstanceCount& InstanceCount::operator=(InstanceCount const& o) {
```
andy-slac: Default implementation would do the same?

core/modules/util/InstanceCount.h
```diff
@@ -0,0 +1,42 @@
...
+    InstanceCount& operator=(InstanceCount const& o);
+
+    int getCount(); //< Return the number of instances of _className.
+
+    friend std::ostream& operator<<(std::ostream &out, InstanceCount const& instanceCount);
+
+private:
+    std::string _className; //< Names of the of which this is a member.
+    static std::map<std::string, int> _instances; //< Map of instances per class name.
+    static std::recursive_mutex _mx; //< Protects _instances.
```
andy-slac: Does it have to be recursive?

core/modules/util/testEventThread.cc
```diff
@@ -184,4 +183,38 @@ BOOST_AUTO_TEST_CASE(EventThreadTest) {
...
+
+BOOST_AUTO_TEST_CASE(InstanceCountTest) {
+
+    class CA {
+    public:
+        InstanceCount instanceCount{"CA"};
+    };
+
+    class CB {
+       public:
```
andy-slac: Indentation. Use `struct` instead of `class` to save one `public:` line?

core/modules/ccontrol/UserQuerySelect.cc
jgates108: As long as 'uq->submit();` precedes `uq->join();`, the not submitted case would never be reached and isn't terribly meaningful. The the query was cancelled and why is probably the important information, but we aren't sending that back to the user as we usually have some valid rows to send back.

PR #216, opened 2016-04-08
tickets/DM-5753: XCode 7.3 is broken

timj:
Provides a workaround on OSX for the bug linking indirect libraries on OS X. Since I was messing with that code there is a second commit that cleans up the `-install_name` option and also switches that to use `@rpath` (which seems to be the recommended thing to do on OS X).


fritzm:
LGTM, thanks!

PR #217, opened 2016-04-12
Improve install documentation

achbal:
- use of qserv-dev tag
- explicit use of QSERV_RUN_DIR
- stop qserv after mono-node test


fritzm:
Looks like these commits were merged by @fjammes on 4/15, so closing this PR now.  Thanks @achbal!

PR #220, opened 2016-04-28
Tickets/dm 5514

jgates108:


core/modules/memman/Memory.cc
```diff
@@ -148,7 +148,7 @@ void Memory::memRel(MemInfo& mInfo) {
     //
     if (mInfo._memSize > 0 && mInfo._memAddr != MAP_FAILED) {
         munmap(mInfo._memAddr, mInfo._memSize);
-        if (_lokBytes < mInfo._memSize) {
+        if ( mInfo._memSize < _lokBytes) {
```
andy-slac: No space after paren

core/modules/wsched/ChunkDisk.cc
```diff
@@ -139,6 +139,17 @@ bool ChunkDisk::ready(bool useFlexibleLock) {
...
 /// Return true if this disk is ready to provide a Task from its queue.
 bool ChunkDisk::_ready(bool useFlexibleLock) {
+    auto logMemManRes =
+        [this](bool newVal, std::string const& msg, std::vector<memman::TableInfo> const& tblVect) {
+        if (setResourceStarved(newVal)) {
+            std::string str;
+            for (auto const& tblInfo:tblVect) {
+                str += tblInfo.tableName + " ";
+            }
+            LOGS(_log, LOG_LVL_DEBUG, "ready memMan " << msg << " - " << str);
```
andy-slac: Use `util::printable` instead of explicit loop?

core/modules/wsched/SchedulerBase.cc
```diff
@@ -65,4 +65,74 @@ void SchedulerBase::setPriorityDefault() {
 }
 
 
+int SchedulerBase::_incrCountForUserQuery(uint64_t queryId) {
+    std::pair<uint64_t, int> entry(queryId, 0);
+    std::lock_guard<std::mutex> lock(_countsMutex);
+    auto result = _userQueryCounts.insert(entry);
+    return ++result.first->second;
```
andy-slac: I think four lines above could be replaced with:

    std::lock_guard<std::mutex> lock(_countsMutex);
    return ++_userQueryCounts[queryId];

core/modules/wsched/SchedulerBase.cc
```diff
@@ -65,4 +65,74 @@ void SchedulerBase::setPriorityDefault() {
...
+
+
+int SchedulerBase::_decrCountForUserQuery(uint64_t queryId) {
+    std::lock_guard<std::mutex> lock(_countsMutex);
+    // Decrement the count for this user query and remove the entry if count is 0.
+    int count = 0;
+    auto iter = _userQueryCounts.find(queryId);
+    if (iter != _userQueryCounts.end()) {
+        count = --(iter->second);
+        if (iter->second <= 0) {
```
andy-slac: Nitpick - use `count` instead of `iter->second` here.

core/modules/wsched/SchedulerBase.cc
```diff
@@ -65,4 +65,74 @@ void SchedulerBase::setPriorityDefault() {
...
+    std::lock_guard<std::mutex> lock(_countsMutex);
+    return _userQueryCounts.size();
+}
+
+
+void SchedulerBase::_incrChunkTaskCount(int chunkId) {
+    std::pair<int, int> entry(chunkId, 0);
+    std::lock_guard<std::mutex> lock(_countsMutex);
+    auto result = _chunkTasks.insert(entry);
+    result.first->second += 1;
```
andy-slac: Same thing here.

core/modules/wsched/SchedulerBase.cc
```diff
@@ -65,4 +65,74 @@ void SchedulerBase::setPriorityDefault() {
...
+}
+
+
+void SchedulerBase::_decrChunkTaskCount(int chunkId) {
+    // Decrement the count for this user query and remove the entry if count is 0.
+    std::lock_guard<std::mutex> lock(_countsMutex);
+    auto iter = _chunkTasks.find(chunkId);
+    if (iter != _chunkTasks.end()) {
+        iter->second -= 1;
```
andy-slac: Why `-= 1` here and `--` above?

core/modules/wsched/SchedulerBase.h
```diff
@@ -105,6 +122,17 @@ class SchedulerBase : public wcontrol::Scheduler {
     int _priorityNext; ///< Priority to use starting with the next chunk.
 
     std::atomic<int> _inFlight{0}; //< Number of Tasks running.
+
+private:
+    /// The true purpose of _userQuerycount is to track how many different UserQuery's are on the queue.
+    /// Number of Tasks for each UserQuery in the queue.
+    std::map<uint64_t, int> _userQueryCounts;
```
andy-slac: WE should define typedef for QueryId, this becomes too widespread.

core/modules/xrdsvc/SsiService.cc
```diff
@@ -99,8 +99,10 @@ SsiService::SsiService(XrdSsiLogger* log) {
         // Default to 1 gigabyte
         uint64_t memManSizeMb = config.getInt("QSW_MEMMAN_MB", 1000);
         std::string memManLocation = config.getString("QSW_MEMMAN_LOCATION");
-        LOGS(_log, LOG_LVL_DEBUG, "Using MemManReal with memManSizeMb=" << memManSizeMb);
-        memMan = std::shared_ptr<memman::MemMan>(memman::MemMan::create(memManSizeMb*1000000, memManLocation));
+        LOGS(_log, LOG_LVL_DEBUG, "Using MemManReal with memManSizeMb=" << memManSizeMb
+                << " location=" << memManLocation);
+        memMan =
+            std::shared_ptr<memman::MemMan>(memman::MemMan::create(memManSizeMb*1000000, memManLocation));
```
andy-slac: maybe:

    memMan.reset(memman::MemMan::create(memManSizeMb*1000000, memManLocation));

core/modules/memman/MemFile.cc
```diff
@@ -107,6 +107,21 @@ MemFile::MLResult MemFile::memLock() {
         return nilResult;
     }
 
+    // TODO: Find a better solution for systems where mmap is not viable, which
+    // manifests as no bytes being locked but bytes are reserved.
+    // Fake mmap as the configuration isn't working properly.
+    // At this point, there is enough free space, it was not a flex table,
+    // but mmap failed. So,this will fake it being a flexilock file and
+    // reserve space for it.
+    if (_memInfo.size() < freeBytes) {
```
abh3: I can't say I am happy with the change to this logic. It really violates what MUSTLOCK means and makes for rather strange side-effects (at least they are strange to me). For instance, subsequent FLEXIBLE locks on this file are effectively upgraded to MUSTLOCK status. It also means there is no way you can really tell that a MUSTLOCK really locked anything at all.

jgates108: I'm not happy with it either. That's why it starts out with todo. However, without this change, memman doesn't work well on the cluster at all (queries on large tables go from 6 hours to 8 hours). I think we will need something like this for use on systems where the workers don't have enough memory and/or can't be easily configured so that mmap works.

core/modules/wsched/ChunkDisk.cc
jgates108: It would be nice. There is currently no << operator for TableInfo and I'm not sure where to put it as memman.h defines an interface and intentionally has no implementation. Maybe AndyH has something to say about it.

core/modules/wsched/SchedulerBase.cc
jgates108: I think I'm being paranoid about the first entry being 0. I looked around a little but couldn't verify that it would certainly  use 0 if there was no entry for queryId.

core/modules/wsched/SchedulerBase.cc
jgates108: Can I blame your cat?

core/modules/wsched/SchedulerBase.cc
andy-slac: If `map[key]` has to add new value to the map it's guaranteed to be zero-initialized, one of the rare cases in C++ when behavior is intuitive :smile_cat:

core/modules/wsched/SchedulerBase.cc
andy-slac: That privilege is mine, sorry :)

core/modules/wsched/SchedulerBase.h
jgates108: Agreed. Any thoughts on where it should be defined? It shows up on both the czar and the worker and it would be good to define it once for both. There's QueryId in qmeta/types.h, but including that in worker coded causes issues.

andy-slac: `global` module already has `intTypes.h` and `stringTypes.h` (though I do not like this separation), maybe `intTypes.h` is a good place for this typedef?

jgates108: I don't understand the separation either. I'll move it to intTypes.h and see what happens.

core/modules/wsched/ChunkDisk.cc
andy-slac: Sorry, I messed that it prints only one attribute. Ignore it.

core/modules/wsched/SchedulerBase.h
jgates108: I'm going to put QueryID in lsst:qserv namespace. Putting qmeta in in global feels wrong.

andy-slac: Yes, definitely. I guess we'll need another ticket to migrate all uses of qmeta::QueryId and uint64_t used for the same purpose to this new typedef.

jgates108: I changed them on the worker side, but qmeta:QueryId is needed for swig and python, so we need a ticket for removing that and cleaning up.

core/modules/wsched/ChunkDisk.cc
abh3: Just tell me what you want the "<<" operator to do. It's not clear from 
this comment.

Andy

core/modules/wsched/SchedulerBase.h
abh3: Is this the global query ID we all keep talking about? If so, we need to 
talk on how to pass it to the SSI layer so we can keep track of the query 
there as well.

Andy

core/modules/wsched/ChunkDisk.cc
andy-slac: Andy, just ignore this, not worth spending time on this.

core/modules/memman/MemFile.cc
abh3: After some thought, I think the appropriate change is to add "NEEDLOCK" 
attribute that would work like the modified code, intermediate between 
MUSTLOCK and FLEXLOCK.

Andy

jgates108: That would be fine with me.

PR #221, opened 2016-05-06
Added code to allow a thread to leave the thread pool.

jgates108:


core/modules/util/EventThread.cc
```diff
@@ -46,40 +46,98 @@ namespace util {
...
+        _currentCommand = nullptr;
     }
     finishup();
 }
 
+
+/// Ensure that commandFinish is only called once.
+void EventThread::callCommandFinish(Command::Ptr const& cmd) {
+    if (_commandFinishCalled.exchange(true) == false) {
+        _q->commandFinish(cmd);
```
andy-slac: Isn't `std::call_once` doing the same with ~less typing (I never used it myself though)

core/modules/util/EventThread.h
```diff
@@ -113,25 +114,37 @@ class EventThread : public CmdData {
...
         std::shared_ptr<MsgEnd> cmd = std::make_shared<MsgEnd>();
         queCmd(cmd);
     }
 
+    Command* getCurrentCommand() const { return _currentCommand; }
+
 protected:
     virtual void startup() {};  ///< Things to do before handling commands
     virtual void finishup() {}; ///< things to do when the thread is closing down.
+    virtual void specialActions(Command::Ptr const& cmd) {}; ///< Things to do before running a command.
```
andy-slac: extraneous semicolon

core/modules/util/EventThread.h
```diff
@@ -141,11 +154,32 @@ class PoolEventThread : public EventThread {
...
 };
 
+
+/// A Command that is aware that it is running as part of a PoolEventThread,
+// which allows it to tell the pool to take special actions.
+class CommandThreadPool : public CommandTracked {
+public:
+    using Ptr = std::shared_ptr<CommandThreadPool>;
+
+    CommandThreadPool() {};
```
andy-slac: extraneous semicolon

core/modules/util/EventThread.cc
jgates108: I'll try it out and see how it goes.

jgates108: Using std::call_once uses about the same amount of code. The interesting thing is that std::call_once only considers the function called if it does not throw an exception. If the function throws an exception, std::call_once will allow it be called until it does not throw an exception, and that's not what I'm looking for here.

abh3:
This was actually pushed to master on May 11th and then orphaned.

PR #223, opened 2016-05-11
Add MySQL account dedicated to monitoring

fjammes:
- Refine MySQL account managements in configuration
  procedure.
- Remove useless parameter 'version' from configuration
- Hard-code MySQL root user parameter
- Add privileges to monitoring account
  * Remove unused access to mysql database for Qserv
    account.
  * Rename parameters for MySQL accounts
- Improve SQL configuration files templates
  * Replace hard-coded 'qsmaster' value with variable


admin/python/lsst/qserv/admin/configure.py
```diff
@@ -212,42 +212,43 @@ def _get_template_params():
...
+            'QSERV_UNIX_USER': getpass.getuser(),
+            'SCISQL_DIR': scisql_dir,
+            'WMGR_PASSWORD': random_string(string.ascii_letters + string.digits + string.punctuation, 23),
+            'WMGR_PORT': config['wmgr']['port'],
+            'WMGR_SECRET_KEY': random_string(string.ascii_letters + string.digits, 57),
+            'WMGR_USER_NAME': random_string(string.ascii_letters + string.digits, 12),
+            'XROOTD_ADMIN_DIR': os.path.join(config['qserv']['qserv_run_dir'], 'tmp'),
+            'XROOTD_DIR': config['xrootd']['base_dir'],
+            'XROOTD_MANAGER_HOST': config['qserv']['master'],
+            'XROOTD_PORT': config['xrootd']['xrootd_port'],
```
andy-slac: Better not to mix whitespace change with content change, it's very hard to see what has changed here.

admin/templates/configuration/tmp/configure/scisql.sh
```diff
@@ -1,13 +1,12 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
```
andy-slac: I think `#!/bin/sh` is portable enough, no need to use `env` here.

timj: Yes. That's more portable than env.

PR #224, opened 2016-05-13
Tickets/dm 5909

jgates108:


admin/tools/docker/3_build-git-image.sh
```diff
@@ -60,10 +60,10 @@ if [ -z "$GIT_REF" -a -z "$SRC_DIR" ]; then
...
     usage
     exit 2
-elif [ -n "$GIT_REF" -a -n "$SRC_DIR" ]; then
-    echo "ERROR: must not provide both branch/tag and local source directory"
-    usage
-    exit 2
+#elif [ -n "$GIT_REF" -a -n "$SRC_DIR" ]; then
+#    echo "ERROR: must not provide both branch/tag and local source directory"
+#    usage
+#    exit 2
```
jgates108: I'll uncomment this before merging.

core/modules/util/EventThread.h
```diff
@@ -207,6 +207,8 @@ class CommandThreadPool : public CommandTracked {
     CommandThreadPool() {}
     CommandThreadPool(std::function<void(CmdData*)> func) : Command{func} {}
 
+    PoolEventThread* getPoolEventThread() { return _poolEventThread; } //< Not thread safe.
+
     friend class PoolEventThread; // Only class that should use _poolEventThread.
```
andy-slac: This `friend` isn't needed any more if you add method above?

core/modules/rproc/InfileMerger.cc
```diff
@@ -280,11 +169,17 @@ bool InfileMerger::merge(std::shared_ptr<proto::WorkerResponse> response) {
             return false;
         }
     }
-    return _importResponse(response);
+
+    // Nothing to do if size is zero.
+    if (response->result.row_size() != 0) {
```
andy-slac: :nitpick: (damn, no emoticon for nitpick) I'd probably move this check into `_merge()` to make it more generic.

core/modules/util/EventThread.h
jgates108: The comment is incorrect. I'll fix it.

core/modules/rproc/InfileMerger.cc
jgates108: I'm removing _merge in DM-5910 and putting the code here as there's no longer any need to keep it separate with the end of the Mgr class.

PR #225, opened 2016-05-18
Add yaml file for monitoring qserv/deps versions

fjammes:
- Improve Docker image generation, building from
  git repository instead of archive
- Use pkgautoversion during build/install
  to computer and add version number to qserv-meta.conf
- Add MYSQLPROXY_DIR parameter to template engine
- Update travis-ci procedure to retrieve correct git
  branch number.
- Add QSERV_VERSION variable to template engine


site_scons/state.py
```diff
@@ -74,6 +75,14 @@ def _getBinPath(binName, msg=None):
         log.debug("Found %s here %s" % (binName, binFullPath))
         return binFullPath
 
+def _getEupsVersion():
+    binName = 'pkgautoversion'
+    try:
+        version = subprocess.check_output(['pkgautoversion'])
```
andy-slac: The output from the command may have some newlines and/or spaces, may be better to strip leading/trailing stuff.

site_scons/state.py
```diff
@@ -74,6 +75,14 @@ def _getBinPath(binName, msg=None):
...
+    except:
```
andy-slac: It's better to check for specific exception types if you can. Currently this will catch everything, even syntax errors in the above line.

PR #229, opened 2016-05-23
Manage ulimit memlock in containers

fjammes:


admin/tools/docker/deployment/in2p3/env.example.sh
```diff
@@ -22,6 +22,8 @@ HOST_DATA_DIR=/qserv/data
 # Log directory location on docker host, optional
 HOST_LOG_DIR=/qserv/log
 
+# ulimit memory lock setting, in bytes, optional
+ULIMIT_MEMLOCK=10485760
```
abh3: Ehr, well 10 megabytes is pretty much useless. What is the problem of setting it to 75% of the real memory size. That should be at least tens of gigabytes.

fjammes: My mistake, just  changed it to  10737418240 bytes (i.e. 10GB), should be fine now.

PR #230, opened 2016-05-26
Quote library path string on OS X

timj:
It may include a space.


timj:
This is a trivial fix for OS X only.

PR #232, opened 2016-06-02
Added code to replace heap with map of ChunkTasks.

jgates108:


core/modules/wsched/ChunkDisk.cc
```diff
@@ -89,7 +89,7 @@ wbase::Task::Ptr ChunkDisk::MinHeap::pop() {
 }
 
 
-void ChunkDisk::enqueue(wbase::Task::Ptr const& a) {
+void ChunkDisk::queTask(wbase::Task::Ptr const& a) {
```
fjammes: Rename to queueTask()?

core/modules/wsched/ChunkDisk.h
```diff
@@ -40,30 +40,32 @@
...
+#include "wsched/ChunkTaskCollection.h"
 
 namespace lsst {
 namespace qserv {
 namespace wsched {
 
 /// Limits Tasks to running when resources are available.
 /// TODO: DM-4943 Maybe merge this class into ScanScheduler.
-class ChunkDisk {
+class ChunkDisk : public ChunkTaskCollection {
```
fjammes: Don't you add some inline doc for class an public methods?

core/modules/wsched/ChunkTaskCollection.h
```diff
@@ -0,0 +1,57 @@
...
+
+namespace lsst {
+namespace qserv {
+namespace wsched {
+
+/// Class defines an interface to store Tasks related to chunks in an ordered manner.
+// The only derived classes are expected to be ChunkDisk and ChunkTaskList.
+// With luck, one of them will prove to be superior and the other will be destroyed, thus
+// this class will no longer be needed.
+class ChunkTaskCollection {
```
fjammes: Inline doc?

core/modules/wsched/ChunkDisk.h
jgates108: I expect to delete this class, but I wanted to keep it around for testing purposes to make sure that ChunkTaskList is better. ChunkTaskCollection should probably go away at the same time as it is only intended to make it easy to switch between ChunkDisk and ChunkClassList. I'd rather avoid writing documentation for code that I believe is doomed.

core/modules/wsched/ChunkDisk.cc
jgates108: There are several other functions named queCmd, queTask, etc. I've been using 'que' as flag to indicate that the function is placing a Command on an EventThread in hopes that it is clear that something is being scheduled to happen later, likely on a different thread.

ktlim: I think we would prefer to have these all spelled out with "queue" as a full word instead of the odd abbreviation "que".

core/modules/wsched/ChunkDisk.h
fjammes: Might we test and add other implementations in the future, if yes we could keep and document this interface?

core/modules/wsched/ChunkTaskList.cc
```diff
@@ -0,0 +1,387 @@
...
+namespace wsched {
+
+
+void ChunkTaskList::queTask(wbase::Task::Ptr const& task) {
+    std::lock_guard<std::mutex> lg(_mapMx);
+    int chunkId = task->getChunkId();
+    auto iter = _chunkMap.find(chunkId);
+    if (iter == _chunkMap.end()) {
+        iter = _insertChunkTask(chunkId);
+    }
```
fjammes: This check could be performed in_insertChunkTask to avoid test on res.second and simplify alittle bit the code?

core/modules/wsched/ChunkTaskList.cc
```diff
@@ -0,0 +1,387 @@
+// -*- LSST-C++ -*-
```
fjammes: It would be nice to have a detailed documentation with a few schema once this code will be stable, maybe in a future ticket?

core/modules/wsched/ChunkTaskList.cc
```diff
@@ -0,0 +1,393 @@
...
+    std::lock_guard<std::mutex> lock(_mapMx);
+    return _ready(useFlexibleLock);
+}
+
+/// Precondition: _queueMutex must be locked
+/// If true is returned, _readyChunk will point to a chunk with a Task that is ready to run.
+/// This function starts checking at the _activeChunk and only looks to the next chunk
+/// if there are no tasks to run on the current chunk.
+/// The _activeChunk advances when all of its Tasks have completed.
+/// @return true if this object is ready to provide a Task from its queue. _readyChunk != nullptr.
```
fjammes: What does mean _readyChunk != nullptr?

core/modules/wsched/ChunkTaskList.cc
```diff
@@ -0,0 +1,393 @@
...
+
+/// @return true if this object is ready to provide a Task from its queue.
+bool ChunkTaskList::ready(bool useFlexibleLock) {
+    std::lock_guard<std::mutex> lock(_mapMx);
+    return _ready(useFlexibleLock);
+}
+
+/// Precondition: _queueMutex must be locked
+/// If true is returned, _readyChunk will point to a chunk with a Task that is ready to run.
+/// This function starts checking at the _activeChunk and only looks to the next chunk
```
fjammes: What happens if false is returned?

core/modules/wsched/ChunkTaskList.cc
```diff
@@ -0,0 +1,393 @@
...
+        // _inFlightTasks must be empty as readyToAdvance was true.
+        if (_activeChunk->second->empty()) {
+            if (newActive == _activeChunk) {
+                newActive = _chunkMap.end();
+            }
+            _chunkMap.erase(_activeChunk);
+        }
+
+        _activeChunk = newActive;
+        if (newActive == _chunkMap.end()) {
```
fjammes: What does happen here? No more chunk remaining?

core/modules/wsched/ChunkTaskList.cc
```diff
@@ -0,0 +1,393 @@
...
+    while (chunkState != ChunkTasks::ReadyState::READY
+           && chunkState != ChunkTasks::ReadyState::NO_RESOURCES) {
+        ++iter;
+        if (iter == _chunkMap.end()) {
+            iter = _chunkMap.begin();
+        }
+        if (iter == _activeChunk) {
+            return false;
+        }
+        chunkState = iter->second->ready(useFlexibleLock);
```
fjammes: Could we replace it by testing if _chunkMap size is equals 1? If yes this might be shorter?

core/modules/wsched/ChunkTaskList.cc
```diff
@@ -0,0 +1,393 @@
...
+    _readyChunk = iter->second;
+    return true;
+}
+
+
+wbase::Task::Ptr ChunkTaskList::getTask(bool useFlexibleLock) {
+    std::lock_guard<std::mutex> lock(_mapMx);
+    if (_readyChunk == nullptr) {
+        // Attempt to set _readyChunk.
+        _ready(useFlexibleLock);
```
fjammes: Don't you use _ready return value?

core/modules/wsched/ChunkTaskList.cc
```diff
@@ -0,0 +1,393 @@
...
+// to change priority.
+bool ChunkTaskList::nextTaskDifferentChunkId() {
+    if (_activeChunk == _chunkMap.end()) {
+        return true;
+    }
+    return _activeChunk->second->readyToAdvance();
+}
+
+
+void ChunkTaskList::taskComplete(wbase::Task::Ptr const& task) {
```
fjammes: Inline doc?

core/modules/wsched/ChunkTaskList.cc
```diff
@@ -0,0 +1,393 @@
...
+    auto iter = _chunkMap.find(task->getChunkId());
+    if (iter != _chunkMap.end()) {
+        iter->second->taskComplete(task);
+    }
+}
+
+
+bool ChunkTaskList::setResourceStarved(bool starved) {
```
fjammes: Inline doc?

core/modules/wsched/ChunkTaskList.cc
```diff
@@ -0,0 +1,393 @@
...
+    bool ret = _resourceStarved;
+    _resourceStarved = starved;
+    return ret;
+}
+
+
+void ChunkTasks::SlowTableHeap::push(wbase::Task::Ptr const& task) {
```
fjammes: Inline doc?

fjammes: Inline doc?

core/modules/wsched/ChunkTaskList.cc
```diff
@@ -0,0 +1,393 @@
...
+    _tasks.push_back(task);
+    std::push_heap(_tasks.begin(), _tasks.end(), compareFunc);
+}
+
+
+wbase::Task::Ptr ChunkTasks::SlowTableHeap::pop() {
```
fjammes: Inline doc?

core/modules/wsched/ChunkTaskList.cc
fjammes: ChunkTasks definition is quite huge and could be moved to ChunkTasks.cc?

core/modules/wsched/ChunkTaskList.cc
```diff
@@ -0,0 +1,393 @@
...
+LOG_LOGGER _log = LOG_GET("lsst.qserv.wsched.ChunkTaskList");
+}
+
+namespace lsst {
+namespace qserv {
+namespace wsched {
+
+
+/// Queue a Task with other tasks on the same chunk.
+void ChunkTaskList::queTask(wbase::Task::Ptr const& task) {
```
fjammes: ChunkTaskList and ChunkTasks seems a bit redundant but does not represent the same object. Could we add a more meaningful name to the first one?

core/modules/wsched/ChunkTaskList.cc
jgates108: I was trying to say that if this function returns true, _readyChunk will point to the chunk with a Task to run. If it returns false, _readyChunk will be nullptr. I'll fix the comment.

core/modules/wsched/ChunkTaskList.cc
jgates108: Yes, this would only happen if _chunkMap is empty.

core/modules/wsched/ChunkTaskList.cc
```diff
@@ -0,0 +1,393 @@
...
+    auto task = _activeTasks.top();
+    if (!task->hasMemHandle()) {
+        memman::TableInfo::LockType lckOptTbl = memman::TableInfo::LockType::MUSTLOCK;
+        memman::TableInfo::LockType lckOptIdx = memman::TableInfo::LockType::NOLOCK;
+        if (useFlexibleLock) lckOptTbl = memman::TableInfo::LockType::FLEXIBLE;
+        auto scanInfo = task-> getScanInfo();
+        auto chunkId = task->getChunkId();
+        if (chunkId != _chunkId) {
+            // This would slow things down badly, but the system would survive.
+            LOGS(_log, LOG_LVL_ERROR, "ChunkTasks " << _chunkId << " got task for chunk " << chunkId
```
fjammes: Shouldn't this be check when task is added to ChunkTasks?

core/modules/wsched/ChunkTaskList.cc
jgates108: I'm not sure what you mean. There could easily be one or thousands of ChunkTasks in _chunkMap, but we might not be able to run anything from the ChunkTask in iter because there isn't enough memory for the next Task or all its Tasks are already in flight. 
In the case of queries on the Object table (a relatively small table), it would be easy to wrap all the way around and have all the Tasks for 10 chunks in flight at once, so this needs to stop checking if it loops back to the _activeChunk.

core/modules/wsched/ChunkTaskList.cc
```diff
@@ -0,0 +1,393 @@
...
+bool ChunkTasks::readyToAdvance() {
+    return _activeTasks.empty() && _inFlightTasks.empty();
+}
+
+
+/// @Return true if a Task is ready to be run.
+// ChunkTasks does not have its own mutex and depends on its owner for thread safety.
+// If a Task is ready to be run, _readyTask will not be nullptr.
+ChunkTasks::ReadyState ChunkTasks::ready(bool useFlexibleLock) {
+    auto logMemManRes =
```
fjammes: Why not a documented private method to simplify code?

core/modules/wsched/ChunkTaskList.cc
```diff
@@ -0,0 +1,393 @@
...
+
+int ChunkTaskList::getActiveChunkId() {
+    if (_activeChunk == _chunkMap.end()) {
+        return -1;
+    }
+    return _activeChunk->second->getChunkId();
+}
+
+
+/// @return old value of _resourceStarved.
```
fjammes: If this is used only for logging purpose, maybe you could add it in documentation?

core/modules/wsched/ChunkTaskList.cc
```diff
@@ -0,0 +1,393 @@
...
+bool ChunkTasks::setResourceStarved(bool starved){
```
fjammes: Could it be a private method?

core/modules/wsched/ChunkTaskList.h
```diff
@@ -0,0 +1,160 @@
...
+
+namespace lsst {
+namespace qserv {
+namespace wsched {
+
+
+/// A class to store Tasks for a specific chunk.
+//  Tasks are normally placed on _activeTasks, but will be added
+//  to _pendingTasks when this is the active chunk.
+class ChunkTasks {
```
fjammes: Could have its own .h file?

core/modules/wsched/ChunkTaskList.h
```diff
@@ -0,0 +1,160 @@
...
+        void heapify() {
+            std::make_heap(_tasks.begin(), _tasks.end(), compareFunc);
+        }
+
+        std::vector<wbase::Task::Ptr> _tasks;
+    };
+
+private:
+    int _chunkId;          ///< Chunk Id for all Tasks in this instance.
+    bool _active{false};   ///< True when this is the active chunk.
```
fjammes: What does mean "active" please?

core/modules/wsched/ChunkTaskList.cc
jgates108: There are places where ready() is called outside this class where returning a bool is more generic and appropriate. ready() could be called many times before a single getTask() is called, and getTask() should be the only thing that is interested in _readyChunk.  _ready() does the heavy lifting, getTask() just returns the result and clears _readyTask.

core/modules/wsched/ChunkTaskList.h
```diff
@@ -0,0 +1,160 @@
...
+    bool _resourceStarved; ///< True when advancement is prevented by lack of memory.
+    wbase::Task::Ptr              _readyTask{nullptr}; ///< Task that is ready to run with memory reserved.
+    SlowTableHeap                 _activeTasks;        ///< All Tasks must be put on this before they can run.
```
fjammes: 'run' and 'in Flight' have the same meaning? If yes maybe this could be clarified?

core/modules/wsched/ChunkTaskList.h
```diff
@@ -0,0 +1,160 @@
...
+    class SlowTableHeap {
+    public:
+        // Using a greater than comparison function results in a minimum value heap.
+        static bool compareFunc(wbase::Task::Ptr const& x, wbase::Task::Ptr const& y) {
+            if(!x || !y) { return false; }
+            // compare scanInfo (slower scans first)
+            int siComp = x->getScanInfo().compareTables(y->getScanInfo());
+            return siComp < 0;
+        };
+        void push(wbase::Task::Ptr const& task);
```
fjammes: Inline doc for all methods below?

core/modules/wsched/ChunkTaskList.h
```diff
@@ -0,0 +1,160 @@
...
+
+    /// This must be std::map to maintain valid iterators.
+    /// Only erase() will invalidate and iterator with std::map.
+    using ChunkMap = std::map<int, ChunkTasks::Ptr>;
+
+    enum {READY, NOT_READY, NO_RESOURCES};
+
+    ChunkTaskList(memman::MemMan::Ptr const& memMan) : _memMan{memMan} {}
+    ChunkTaskList(ChunkTaskList const&) = delete;
+    ChunkTaskList& operator=(ChunkTaskList const&) = delete;
```
fjammes: Inline doc for all methods below?

core/modules/wsched/ScanScheduler.cc
```diff
@@ -59,7 +60,8 @@ ScanScheduler::ScanScheduler(std::string const& name, int maxThreads, int maxRes
                              memman::MemMan::Ptr const& memMan, int minRating, int maxRating)
     : SchedulerBase{name, maxThreads, maxReserve, priority},
       _memMan{memMan}, _minRating{minRating}, _maxRating{maxRating} {
-    _disk = std::make_shared<ChunkDisk>(_memMan);
+    //_disk = std::make_shared<ChunkDisk>(_memMan); // keeping for testing.
+    _disk = std::make_shared<ChunkTaskList>(_memMan);
```
fjammes: is _disk still a consistent name?

core/modules/wsched/ChunkTaskList.cc
jgates108: This is the only method that would ever call it, so it would just make the header file more complicated.

core/modules/wsched/ChunkTaskList.h
jgates108: I've been basing "in flight" on Daniel's past code comments. "in flight" basically means running, but to avoid race conditions, a Task is considered "in flight" as soon as getTask() pulls it off the heap.

core/modules/wsched/ChunkTaskList.h
jgates108: The comments would be redundant for most of these do exactly what the function name says they will do. In my opinion, such comments just make the header harder to read. Specifics to the implementation are more appropriate in the cc files. 

I did add descriptions to some of these in ChunkTaskCollection.

core/modules/wsched/ChunkTaskList.cc
jgates108: A creating a ticket to handle it in the future would probably be good. There's a significant amount of flux still in the implementation, but the basic design is the same. Group tasks by chunkId and table so that several tasks can run while only loading the table once.

core/modules/wsched/ChunkTaskList.cc
jgates108: ChunkTasksQueue?

core/modules/wsched/ChunkTaskList.cc
abh3: Perhaps then use a lambda for this.

core/modules/wsched/ChunkTaskList.cc
```diff
@@ -0,0 +1,393 @@
...
+    if (!res.second) {
+        LOGS(_log, LOG_LVL_DEBUG, "ChunkTaskList::_insertChunkTask entry already existed " << chunkId);
+    }
+    return res.first;
+}
+
+/// @return true if this object is ready to provide a Task from its queue.
+bool ChunkTaskList::ready(bool useFlexibleLock) {
+    std::lock_guard<std::mutex> lock(_mapMx);
+    return _ready(useFlexibleLock);
```
abh3: Would be good to mention complexity of _ready. That way one can assess how long this lock will be held.

abh3: Also, I see mapNX being locked here and there but not everywhere when you look at the map. So, I don't understand what this mutex is trying to protect. An explanation of how and when the mutex nees to be locked would be helpful.

core/modules/wsched/ChunkTaskList.cc
```diff
@@ -0,0 +1,393 @@
...
+        return nullptr;
+    }
+    auto task = _tasks.front();
+    std::pop_heap(_tasks.begin(), _tasks.end(), compareFunc);
+    _tasks.pop_back();
+    return task;
+}
+
+
+/// Queue new Tasks to be run, ordered with the slowest tables first.
```
abh3: What is considered a slow task and who specifies that?

core/modules/wsched/ChunkTaskList.cc
```diff
@@ -0,0 +1,393 @@
...
+}
+
+
+/// @return true if active AND pending are empty.
+bool ChunkTasks::empty() const {
+    return _activeTasks.empty() && _pendingTasks.empty();
+}
+
+
+/// This is ready to advance when _activeTasks is empty and no Tasks are in flight.
```
abh3: Does this mean that once a task becomes active you can't add any more? What avoids always running a single task just because you advanced something too quickly?

core/modules/wsched/ChunkTaskList.cc
```diff
@@ -0,0 +1,393 @@
...
+        return ChunkTasks::ReadyState::NOT_READY;
+    }
+    // Calling this function doesn't get expensive until it gets here. Luckily,
+    // after this point it will return READY or NO_RESOURCES, and ChunkTaskList::_ready
+    // will not examine any further chunks upon seeing those results.
+    auto task = _activeTasks.top();
+    if (!task->hasMemHandle()) {
+        memman::TableInfo::LockType lckOptTbl = memman::TableInfo::LockType::MUSTLOCK;
+        memman::TableInfo::LockType lckOptIdx = memman::TableInfo::LockType::NOLOCK;
+        if (useFlexibleLock) lckOptTbl = memman::TableInfo::LockType::FLEXIBLE;
```
abh3: Could you make this an if-else or even ?: instead of setting once and then resetting it later. Sort of odd.

core/modules/wsched/ChunkTaskList.cc
abh3: I agree. Anyway, how could this ever happen anyway?

core/modules/wsched/ChunkTaskList.cc
```diff
@@ -0,0 +1,393 @@
...
+            }
+        }
+        task->setMemHandle(handle);
+        logMemManRes(false, task->getIdStr() + " got handle", tblVect);
+    }
+
+    // There is a Task to run at this point, pull it off the heap to avoid confusion.
+    auto popped = _activeTasks.pop();
+    _readyTask = task;
+    if (popped != task) {
```
abh3: Hmm, how could this ever happen? I guess I'm always suspicious when I see a throw. I would recommend an alternate recovery action other than killing the whole server (I didn't see a catch, was there one?). Perhaps just abandon the task and keep going.

core/modules/wsched/ScanScheduler.h
```diff
@@ -82,7 +83,7 @@ class ScanScheduler : public SchedulerBase {
 
 private:
     bool _ready();
-    std::shared_ptr<ChunkDisk> _disk; //< Constrains access to files.
+    std::shared_ptr<ChunkTaskCollection> _disk; //< Constrains access to files.
```
abh3: I'd rather you use a different name. The structure is important enough as it is. Using the same name for two different things makes difficult to keep the context in your head (well, actually impossible).

core/modules/wsched/ChunkTaskList.cc
jgates108: Thanks, there were a couple of calls I missed.

core/modules/wsched/ChunkTaskList.cc
jgates108: The tables are rated by hand when the when the system is setup. Only shared scan tables must have a rating. The czar passes the rating of all shared scan tables in the query to the worker.

core/modules/wsched/ChunkTaskList.cc
jgates108: Only chunks become active. Tasks are run. Once a chunk becomes active, all new Tasks for that chunk are placed on the _pending list. Once all the Tasks on the active chunk's _active heap are run, the next chunk becomes active and all the Tasks in _pending are moved to the _active heap.

jgates108: I've been adding comments to the code but I haven't pushed them to master yet.

core/modules/wsched/ChunkTaskList.cc
```diff
@@ -0,0 +1,393 @@
...
+            LOGS(_log, LOG_LVL_ERROR, "ChunkTasks " << _chunkId << " got task for chunk " << chunkId
+                    << " " << task->getIdStr());
+        }
+        std::vector<memman::TableInfo> tblVect;
+        for (auto const& tbl : scanInfo.infoTables) {
+            memman::TableInfo ti(tbl.db + "/" + tbl.table, lckOptTbl, lckOptIdx);
+            tblVect.push_back(ti);
+        }
+        // If tblVect is empty, we should get the empty handle
+        memman::MemMan::Handle handle = _memMan->lock(tblVect, chunkId);
```
jgates108: A key part of ready() is that it says a Task is "ready" to run. A critical part of that decision is that memMan has memory for the shared scan tables available, or they are already locked. The critical thing is that mmap has reserved space for the file. If all Tasks assume they'll be able to get mmap space, there will be dozens of tasks running, failing to get space and being scrubbed and added back to the scheduler, while another scheduler missed it's shot as all threads were used. It would be very difficult to make any order out of that.
 
One thing I have noticed, is that if mmap works, mlock always works as long as things are configured properly mmap is a fast call. The mlock call and mmap call could be split into separate functions. ready() would call mmap and the running Task would call mlock. 

The key part of the configuration is that the amount of memory MemMan believes it can lock is less than or equal to what the system will allow it to lock.

Conversely, in the other ticket DM-6518, if mlock fails, it is easy to pass the failure to the running Task via CommandMlock. Either way, the running Task knows mlock didn't work and appropriate action can be taken,
whatever that may be.

PR #235, opened 2016-06-14
Provision Openstack infrastructure for Qserv

achbal:
- Use cloud-init technology
- Create a customized CentOS7-based snapshot
  containing Docker
- Use overlay storage backend for Docker
- Boot Qserv master and worker nodes
- Add user ssh key inside nodes
- Add public floating ip to master node
- Wait for cloud-init completion and ssh startup
- Update /etc/hosts on each node
- Add test script
- Add documentation

### DIFF
--- a/admin/tools/provision/LSST-openrc.sh.example
+++ b/admin/tools/provision/LSST-openrc.sh.example
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+# To use an OpenStack cloud you need to authenticate against the Identity
+# service named keystone, which returns a **Token** and **Service Catalog**.
+# The catalog contains the endpoints for all services the user/tenant has
+# access to - such as Compute, Image Service, Identity, Object Storage, Block
+# Storage, and Networking (code-named nova, glance, keystone, swift,
+# cinder, and neutron).
+#
+# *NOTE*: Using the 2.0 *Identity API* does not necessarily mean any other
+# OpenStack API is version 2.0. For example, your cloud provider may implement
+# Image API v1.1, Block Storage API v2, and Compute API v2.0. OS_AUTH_URL is
+# only for the Identity API served through keystone.
+export OS_AUTH_URL=http://nebula.ncsa.illinois.edu:5000/v2.0
+
+# With the addition of Keystone we have standardized on the term **tenant**
+# as the entity that owns the resources.
+export OS_TENANT_ID=8c1ba1e0b84d486fbe7a665c30030113
+export OS_TENANT_NAME="LSST"
+export OS_PROJECT_NAME="LSST"
+
+# In addition to the owning entity (tenant), OpenStack stores the entity
+# performing the action as the **user**.
+export OS_USERNAME="lsst-aaoualid"
+
+# With Keystone you pass the keystone password.
+echo "Please enter your OpenStack Password: "
+read -sr OS_PASSWORD_INPUT
+export OS_PASSWORD=$OS_PASSWORD_INPUT
+
+# If your configuration has multiple regions, we set that information here.
+# OS_REGION_NAME is optional and only valid in certain environments.
+export OS_REGION_NAME="RegionOne"
+# Don't leave a blank variable, unset it if it was empty
+if [ -z "$OS_REGION_NAME" ]; then unset OS_REGION_NAME; fi
--- a/admin/tools/provision/README.rst
+++ b/admin/tools/provision/README.rst
@@ -1,0 +1,57 @@
+**************************************
+Provision Qserv using Docker+Openstack
+**************************************
+
+Pre-requisite
+-------------
+
+* Install system dependencies
+
+.. code-block:: bash
+
+   sudo apt-get install python-dev python-pip
+
+   # Install the OpenStack client
+   sudo pip install python-openstackclient
+
+   # Install the nova client
+   sudo pip install python-novaclient
+
+* Download Openstack RC file: http://docs.openstack.org/user-guide/common/cli_set_environment_variables_using_openstack_rc.html
+
+* Install shmux to run multi node tests
+see http://web.taranis.org/shmux/
+
+
+Provision Qserv & run multinode tests
+-------------------------------------
+   
+* Clone Qserv repository and set Openstack environment and parameters:
+
+.. code-block:: bash
+
+   SRC_DIR=${HOME}/src
+   mkdir ${SRC_DIR}
+   cd ${SRC_DIR}
+   git clone https://github.com/lsst/qserv.git
+   cd qserv/admin/tools/provision
+
+   # Source Openstack RC file
+   # This is an example for NCSA 
+   . ./LSST-openrc.sh
+
+   # Update the configuration file which contains instance parameters
+   # Add special tuning if needed
+   mv ncsa.conf.example ncsa.conf
+
+Update :program:`test.sh` top parameters to set the configuration files above, the snapshot name and the number of servers to boot.
+
+* Create customized image, provision openstack cluster and run Qserv multinodes tests:
+
+.. warning::
+   Make sure that run-multinode-test doesn't fail in order to save ~/.ssh/config
+   your old ~/.ssh/config is in ~/.ssh/config.backup
+
+.. code-block:: bash
+
+   ./test.sh
--- a/admin/tools/provision/README.rst
+++ b/admin/tools/provision/README.rst
@@ -20,7 +20,7 @@ Pre-requisite
 * Download Openstack RC file: http://docs.openstack.org/user-guide/common/cli_set_environment_variables_using_openstack_rc.html
 
 * Install shmux to run multi node tests
-  * see http://web.taranis.org/shmux/
+   see http://web.taranis.org/shmux/
 
 
 Provision Qserv & run multinode tests
--- a/admin/tools/provision/README.rst
+++ b/admin/tools/provision/README.rst
@@ -20,7 +20,7 @@ Pre-requisite
 * Download Openstack RC file: http://docs.openstack.org/user-guide/common/cli_set_environment_variables_using_openstack_rc.html
 
 * Install shmux to run multi node tests
-see http://web.taranis.org/shmux/
+  * see http://web.taranis.org/shmux/
 
 
 Provision Qserv & run multinode tests
--- a/admin/tools/provision/cloudmanager.py
+++ b/admin/tools/provision/cloudmanager.py
@@ -102,7 +102,6 @@ class CloudManager(object):
 
         with open(config_file_name, 'r') as config_file:
             config.readfp(config_file)
-        config_file.close()
 
         self._creds = get_nova_creds()
         logging.debug("Openstack user: %s", self._creds['username'])
--- a/admin/tools/provision/cloudmanager.py
+++ b/admin/tools/provision/cloudmanager.py
@@ -91,11 +91,11 @@ class CloudManager(object):
         Constructor parse all arguments
 
         @param config_file_name define the configuration file containing cloud parameters
-        @param used_image_key base image or snapshot
+        @param used_image_key Add a key to choose the appropriate image (base image or snapshot)
         @param add_ssh_key Add ssh key only while launching instances in provision qserv.
         """
 
-        logging.debug("Use configuration file: {}".format(config_file_name))
+        logging.debug("Use configuration file: %s", config_file_name)
 
         config = ConfigParser.ConfigParser({'net-id': None,
                                             'ssh_security_group': None})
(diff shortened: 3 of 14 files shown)


admin/tools/provision/LSST-openrc.sh.example
```diff
@@ -0,0 +1,36 @@
+
```
andy-slac: There should not be empty line here

admin/tools/provision/README.rst
```diff
@@ -0,0 +1,57 @@
...
+   # Install the OpenStack client
+   sudo pip install python-openstackclient
+
+   # Install the nova client
+   sudo pip install python-novaclient
+
+* Download Openstack RC file: http://docs.openstack.org/user-guide/common/cli_set_environment_variables_using_openstack_rc.html
+
+* Install shmux to run multi node tests
+see http://web.taranis.org/shmux/
```
andy-slac: This line should probably be a part of the list item (previous line, separated by colon)?

admin/tools/provision/cloudmanager.py
```diff
@@ -0,0 +1,311 @@
...
+
+class CloudManager(object):
+    """Application class for common definitions for creation of snapshot and provision qserv"""
+
+    def __init__(self, config_file_name, used_image_key=BASE_IMAGE_KEY, add_ssh_key=False):
+        """
+        Constructor parse all arguments
+
+        @param add_ssh_key Add ssh key only while launching instances in provision qserv.
+        @param go_for_snapshot find new_image until it's created
```
andy-slac: There is no `go_for_snapshot` parameter, and `config_file_name` and `used_image_key` description should be here too.

admin/tools/provision/cloudmanager.py
```diff
@@ -0,0 +1,311 @@
...
+
+        config = ConfigParser.ConfigParser({'net-id': None,
+                                            'ssh_security_group': None})
+
+        try:
+            config_file = open(config_file_name, 'r')
+            try:
+                config.readfp(config_file)
+            finally:
+                config_file.close()
```
andy-slac: There is a better way to do this using context manager in modern Python, five lines above can be replaced with:

    with open(config_file_name, 'r') as config_file:
        config.readfp(config_file)

or even

    config.readfp(open(config_file_name, 'r'))

to rely on Python garbage collector to close the file.

admin/tools/provision/cloudmanager.py
```diff
@@ -0,0 +1,311 @@
...
+        base_image_name = config.get('openstack', used_image_key)
+        self.snapshot_name = config.get('openstack', 'snapshot_name')
+        self.image = self.nova.images.find(name=base_image_name)
+
+        flavor_name = config.get('openstack', 'flavor_name')
+        self.flavor = self.nova.flavors.find(name=flavor_name)
+
+        self.network_name = config.get('openstack', 'network_name')
+        if config.get('openstack', 'net-id'):
+            unicode_net_id = unicode(config.get('openstack', 'net-id'))
```
andy-slac: Why do you need explicit `unicode` here and not in other places? This will likely break with python3.

admin/tools/provision/cloudmanager.py
```diff
@@ -0,0 +1,311 @@
...
+        self.key_filename = '~/.ssh/id_rsa'
+
+
+
+    def get_nb_servers(self):
+        """
+        Get number of servers to boot
+        """
+        args = None
+        return args.nbServers
```
andy-slac: Not sure what this means. `args` is `None`, this will crash

admin/tools/provision/cloudmanager.py
```diff
@@ -0,0 +1,311 @@
...
+
+    def nova_image_create(self, instance):
+        """
+        Create an openstack image containing Docker
+        """
+        logging.info("Creating Qserv image")
```
andy-slac: Log image name here too?

admin/tools/provision/cloudmanager.py
```diff
@@ -0,0 +1,311 @@
...
+                                            key_name=self.key,
+                                            nics=self.nics)
+        # Poll at 5 second intervals, until the status is no longer 'BUILD'
+        status = instance.status
+        while status == 'BUILD':
+            time.sleep(5)
+            instance.get()
+            status = instance.status
+        logging.info("status: {}".format(status))
+        logging.info("Instance {} is active".format(instance_name))
```
andy-slac: Could this fail and return status which means something other than active? If yes then this last message will be confusing.

admin/tools/provision/cloudmanager.py
```diff
@@ -0,0 +1,311 @@
...
+    def detect_end_cloud_config(self, instance):
+        """
+        Add clean wait for cloud-init completion
+        """
+        check_word = "---SYSTEM READY FOR SNAPSHOT---"
+        end_word = None
+        while not end_word:
+            time.sleep(10)
+            output = instance.get_console_output()
+            logging.debug("console output: {}".format(output))
```
andy-slac: For `logging` please use lazy formatting:

    logging.debug("console output: %s", output)

which avoids unnecessary formatting in case the output is discarded. This applies to all logging messages in this module.

admin/tools/provision/cloudmanager.py
timj: Can't we just write admin scripts like these in python 3 and forget about python 2?

admin/tools/provision/cloudmanager.py
```diff
@@ -0,0 +1,311 @@
...
+            self.nova.keypairs.delete(key=self.key)
+
+        with open(os.path.expanduser(self.key_filename + ".pub")) as fpubkey:
+            self.nova.keypairs.create(name=self.key, public_key=fpubkey.read())
+
+    def get_floating_ip(self):
+        """
+        Allocate floating ip address to project
+        """
+        i = 0
```
andy-slac: Move this line right before `while`

admin/tools/provision/cloudmanager.py
```diff
@@ -0,0 +1,311 @@
...
+        floating_ips = self.nova.floating_ips.list()
+        floating_ip = None
+        floating_ip_pool = self.nova.floating_ip_pools.list()[0].name
+
+        # Check for available public ip address in project
+        while i < len(floating_ips) and floating_ip is None:
+            if floating_ips[i].instance_id is None:
+                floating_ip = floating_ips[i]
+                logging.debug('Available floating ip found {}'.format(floating_ip))
+            i += 1
```
andy-slac: Using indices is a bit fragile and ugly, I think more idiomatic way to do it is like:

    for ip in self.nova.floating_ips.list():
        if ip.instance_id is None:
            floating_ip = ip
            logging.debug('Available floating ip found %s', floating_ip)
            break

admin/tools/provision/cloudmanager.py
```diff
@@ -0,0 +1,311 @@
...
+        # Check for available public ip adress in ext-net pool
+        if floating_ip is None:
+            try:
+                logging.debug("Use floating ip pool: {}".format(floating_ip_pool))
+                floating_ip = self.nova.floating_ips.create(floating_ip_pool)
+            except novaclient.exceptions.Forbidden as exc:
+                logging.fatal("Unable to retrieve public IP: {0}".format(exc))
+                sys.exit(1)
+
+        return floating_ip
```
andy-slac: I think there may be a race condition in this class. Here you return floating IP which is available at this moment, but by the time you assign it to a new instance this IP may become used by some other instance. The chance is probably small but it's non-zero.

admin/tools/provision/cloudmanager.py
```diff
@@ -0,0 +1,311 @@
...
+        f.write(ssh_config_extract)
+        f.close()
+
+    def check_ssh_up(self, instances):
+        """
+        Check if the ssh service started
+        """
+        for instance in instances:
+            cmd = ['ssh', '-t', '-F', './ssh_config', instance.name, 'true']
+            success = False
```
andy-slac: `success` is not used, only assigned.

admin/tools/provision/cloudmanager.py
```diff
@@ -0,0 +1,311 @@
...
+            while not success:
```
andy-slac: This looks like an infinite loop in case something wrong happens for a particular host. Maybe some timeout could be useful here (but then you need to decide what to do in case of failure)?

admin/tools/provision/create-image.py
```diff
@@ -0,0 +1,108 @@
...
+# Imports for other modules --
+# ----------------------------
+import cloudmanager
+
+# -----------------------
+# Exported definitions --
+# -----------------------
+def get_cloudconfig():
+    """
+    Write cloud init file
```
andy-slac: Comment is confusing, this method does not write anything to a file.

admin/tools/provision/create-image.py
```diff
@@ -0,0 +1,108 @@
...
+                echo "---CLOUD-INIT-DETECT RUNNING---"
+              done
+              sync
+              fsfreeze -f / && read x; fsfreeze -u /
+              echo "---SYSTEM READY FOR SNAPSHOT---") &
+
+        package_upgrade: true
+        package_reboot_if_required: true
+        timezone: Europe/Paris
+        '''
```
andy-slac: This looks like YAML document, is it OK to heave leading spaces for YAML?

admin/tools/provision/provision-qserv.py
```diff
@@ -0,0 +1,140 @@
...
+# Imports for other modules --
+# ----------------------------
+import cloudmanager
+
+# -----------------------
+# Exported definitions --
+# -----------------------
+def get_cloudconfig():
+    """
+    Write cloud init file
```
andy-slac: Same comments here

admin/tools/provision/provision-qserv.py
```diff
@@ -0,0 +1,140 @@
...
+        for instance in instances:
+            cloudManager.detect_end_cloud_config(instance)
+
+        cloudManager.check_ssh_up(instances)
+
+        cloudManager.update_etc_hosts(instances)
+
+        logging.debug("SUCCESS: Qserv Openstack cluster is up")
+
+    except Exception as exc:
```
andy-slac: `try` block is very long, I'd move its contents in a separate method (e.g. `main()`)

admin/tools/provision/test.sh
```diff
@@ -0,0 +1,80 @@
+ #!/bin/sh
```
andy-slac: Space before `#`

admin/tools/provision/cloudmanager.py
andy-slac: Do you mean python3 for admin scripts and python2 for all others? That would be a mess, I'm not sure I like it. Abandoning python2 is a way to go but it needs more work.

timj: I'm saying you have to start somewhere. Qserv are essentially free to switch to python3 whenever they want (currently we don't have a "python3" package to allow you to check that a python3 is available so there's a bit of a setup overhead: hence admin scripts being an easy way to start).

admin/tools/provision/test.sh
```diff
@@ -0,0 +1,80 @@
...
+# Take a snapshot
+python create-image.py -f "$CONF_FILE" -vv
+
+# Execute provision-qserv with an input cloud conf file
+python provision-qserv.py -f "$CONF_FILE" -n "$NB_SERVERS" -vv
+
+# Warning : if  multinode tests failed save your ~/.ssh/config
+# your old ~/.ssh/config is in ~/.ssh/config.backup
+
+cp ~/.ssh/config ~/.ssh/config.backup
```
achbal: TODO: check for file existence

admin/tools/provision/provision-qserv.py
```diff
@@ -0,0 +1,150 @@
...
+    # Find a floating ip address for gateway
+    floating_ip = cloudManager.get_floating_ip()
+    if not floating_ip:
+        logging.critical("Unable to add public ip to Qserv gateway")
+        sys.exit(1)
+    logging.info("Add floating ip ({0}) to {1}".format(floating_ip,
+                                                       gateway_instance.name))
+    gateway_instance.add_floating_ip(floating_ip)
+    try:
+        gateway_instance.add_floating_ip(floating_ip)
```
andy-slac: Why are there two calls to `add_floating_ip()`?

admin/tools/provision/cloudmanager.py
```diff
@@ -0,0 +1,302 @@
...
+        """
+
+        logging.debug("Use configuration file: %s", config_file_name)
+
+        config = ConfigParser.ConfigParser({'net-id': None,
+                                            'ssh_security_group': None})
+
+        with open(config_file_name, 'r') as config_file:
+            config.readfp(config_file)
+        config_file.close()
```
andy-slac: I don't think this line is needed, `with` will close the file after use.

PR #237, opened 2016-07-05
Added QueryId strings to start and end of user queries.

jgates108:


core/modules/ccontrol/UserQuery.h
```diff
@@ -85,6 +86,11 @@ class UserQuery {
     /// @return ORDER BY part of SELECT statement to be executed by proxy
     virtual std::string getProxyOrderBy() = 0;
 
+    /// @return this query's QueryId string. Many query types do not have valid Id numbers.
+    virtual std::string getQueryIdString() {
```
andy-slac: Make it `const`?

core/modules/ccontrol/UserQuerySelect.cc
```diff
@@ -162,7 +162,7 @@ std::string UserQuerySelect::getError() const {
 
 /// Attempt to kill in progress.
 void UserQuerySelect::kill() {
-    LOGS(_log, LOG_LVL_DEBUG, "UserQuerySelect kill");
+    LOGS(_log, LOG_LVL_DEBUG, _queryIdStr << " UserQuerySelect kill");
```
andy-slac: I'd probably use `getQueryIdString()` instead of `_queryIdStr ` to log it. If someone subclasses `UserQuerySelect` and overrides `getQueryIdString()` you probably want log output be consistent here and in the clients of this class.

core/modules/czar/Czar.cc
```diff
@@ -126,18 +126,19 @@ Czar::submitQuery(std::string const& query,
...
+        result.errorMessage = queryIdStr + " Failed to instantiate query: " + error;
         return result;
     }
 
     // spawn background thread to wait until query finishes to unlock,
     // note that lambda stores copies of uq and msgTable.
-    auto finalizer = [uq, msgTable]() mutable {
-        LOGS(_log, LOG_LVL_DEBUG, "submitting new query");
+    auto finalizer = [uq, msgTable, queryIdStr]() mutable {
+        LOGS(_log, LOG_LVL_DEBUG, queryIdStr << " submitting new query");
```
andy-slac: You don't need extra argument to lambda if you use `uq->getQueryIdString()`.

core/modules/qmeta/QMetaMysql.cc
```diff
@@ -296,9 +297,9 @@ QMetaMysql::registerQuery(QInfo const& qInfo,
         query += _conn.escapeString(itr->second);
         query += "')";
 
-        LOGS(_log, LOG_LVL_DEBUG, "Executing query: " << query);
+        LOGS(_log, LOG_LVL_DEBUG, qIdStr << " Executing query: " << query);
```
andy-slac: I'm not sure you want query ID in this message, this will be printed even before "UserQuery registered" message in UserQuerySelect and I thought you insisted on "UserQuery registered" being the very first message with query ID.

jgates108: "Executing query:" should be fine as a first entry in the file. I'll edit the comment in jira. Vaikunth just really needs a reliable entry for the query starting.

PR #238, opened 2016-07-05
Refactor MemMan to be more compatible with scheduler contraints.

abh3:


core/modules/memman/MemManReal.cc
```diff
@@ -162,7 +214,7 @@ MemMan::Handle MemManReal::lock(std::vector<TableInfo> const& tables, int chunk)
        // (with global lock held) update statistics, generate a file handle,
        // add it to the handle cache, and return the handle.
        //
```
jgates108: Please replace "lock" in the comments for this function with "map" or something similar.

core/modules/memman/MemMan.h
```diff
@@ -97,7 +97,39 @@ class MemMan {
...
+    //! had memory reserved are locked but nothing fails if they cannot be.
+    //!
+    //! @param  handle - Handle returned by prepare() given a set of tables.
+    //! @param  strict - When false, if all the required tables are not locked
+    //!                  (i.e. return != 0) the handle is automatically deleted
+    //!                  and any partially locked resources are released.
+    //!                  Otherwise, the handle remains valid and any partially
+    //!                  locked resources remain locked. The handle must be
+    //!                  manually released by calling unlock().
+    //!
```
jgates108: Maybe it's just me, but strict does the opposite of what I expect. Maybe rename it tolerant?

core/modules/memman/MemFile.cc
```diff
@@ -48,85 +47,91 @@ std::unordered_map<std::string, MemFile*> fileCache;
...
-            }
-            MLResult nilResult(0,0);
-            return nilResult;
-        } else {
-            MLResult bigResult(0, ENOMEM);
-            return bigResult;
+    // Lock this table in memory if possible. If not, simulate an ENOMEM.
+    //
+    if (!_isMapped) rc = ENOMEM;
+        else {
```
jgates108: indentation

core/modules/memman/MemFile.cc
```diff
@@ -191,32 +196,42 @@ MemFile::MFResult MemFile::obtain(std::string const& fPath,
...
+
+         // Decrease the reference count. If there are still references, return
+         //
+         _refs--;
+         if (_refs > 0) return;
+
+         // Remove the object from our cache
+         //
+         auto it = fileCache.find(_fPath);
+         if (it != fileCache.end()) fileCache.erase(it);
```
andy-slac: `fileCache.erase(_fPath);` will do the same as above two lines

core/modules/memman/MemMan.h
```diff
@@ -195,6 +213,10 @@ class MemMan {
 
                   MemMan() {}
     virtual      ~MemMan() {}
+
+protected:
+
+static uint64_t lockLimit;
```
andy-slac: indent

core/modules/memman/MemManReal.cc
```diff
@@ -113,7 +117,55 @@ MemMan::Status MemManReal::getStatus(Handle handle) {
 /*                                  l o c k                                   */
 /******************************************************************************/
   
-MemMan::Handle MemManReal::lock(std::vector<TableInfo> const& tables, int chunk) {
+int MemManReal::lock(MemMan::Handle handle, bool strict) {
+
+    MemFileSet* fsP;
```
andy-slac: Safer to initialize it with `nullptr`.

core/modules/memman/MemManReal.cc
```diff
@@ -180,52 +232,38 @@ MemMan::Handle MemManReal::lock(std::vector<TableInfo> const& tables, int chunk)
...
-}
-  
 /******************************************************************************/
 /*                                u n l o c k                                 */
 /******************************************************************************/
 
 bool MemManReal::unlock(Handle handle) {
 
-    std::lock_guard<std::mutex> guard(hanMutex);
+    MemFileSet* fsP;
```
andy-slac: Same `nullptr` init.

core/modules/memman/Memory.h
```diff
@@ -43,7 +41,7 @@ namespace memman {
 
 class MemInfo {
 public:
-    friend class Memory;
+friend class Memory;
```
andy-slac: It was fine before.

core/modules/memman/Memory.h
```diff
@@ -79,72 +77,40 @@ class MemInfo {
...
     MemInfo() : _memAddr((void *)-1), _memSize(0) {}
    ~MemInfo() {}
 
 private:
 
-   union {void  *_memAddr; int _errCode;};
-   uint64_t      _memSize;  //!< If contains 0 then _errCode is valid.
-   CommandMlock::Ptr _cmdMlock; //< Tracks the completion of mlock call.
+union {void  *_memAddr; int _errCode;};
+uint64_t      _memSize;  //!< If contains 0 then _errCode is valid.
```
andy-slac: Should be indented

core/modules/memman/Memory.h
```diff
@@ -222,11 +189,40 @@ class Memory {
...
+    struct MemStats {
+        uint64_t bytesMax;       //!< Maximum number of bytes being managed
+        uint64_t bytesReserved;  //!< Number of bytes reserved
+        uint64_t bytesLocked;    //!< Number of bytes locked
+        uint32_t numMapErrors;   //!< Number of mmap()  calls that failed
+        uint32_t numLokErrors;   //!< Number of mlock() calls that failed
+        uint32_t numFlexFiles;   //!< Number of Flexible files encountered
+    };
+
+    MemStats Statistics() {
```
andy-slac: Method names should start with lower case.

core/modules/memman/MemFileSet.h
```diff
@@ -76,27 +78,57 @@ class MemFileSet {
...
+    //! @param dolok If true, obtains a mutex on this object to serialize access.
+    //!              Otherwise, the lock is released.
+    //!
+    //! @return true  Normal return.
+    //! @return false lktst was true and serialization mutex was already locked.
+    //-----------------------------------------------------------------------------
+
+    bool serialize(bool dolok) {
+        if (dolok) {
+            _setMutex.lock();
```
andy-slac: Check that mutex is not locked before locking?

core/modules/memman/MemManReal.cc
```diff
@@ -180,52 +232,38 @@ MemMan::Handle MemManReal::lock(std::vector<TableInfo> const& tables, int chunk)
...
+    }
 
-    // Delete the file set and remove it from the map
+    // We must protect unlock() from lock(). We do this by obtaining the
+    // file set lock. It will be unlocked by the destructor.
     //
-    delete it->second;
-    hanCache.erase(it);
+    fsP->serialize(true);
+    delete fsP;
```
andy-slac: `unique_ptr` is probably better than manual memory management.

core/modules/memman/Memory.cc
```diff
@@ -159,16 +158,17 @@ MemInfo Memory::memLock(std::string const& fPath, bool isFlex) {
...
     if (mInfo._memSize > 0 && mInfo._memAddr != MAP_FAILED) {
         munmap(mInfo._memAddr, mInfo._memSize);
-        if (mInfo._memSize < _lokBytes) {
-            _lokBytes -= mInfo._memSize;
-        } else {
-            _lokBytes = 0;
+        if (islkd) {
+            _memMutex.lock();
+            if (_lokBytes > mInfo._memSize) _lokBytes -= mInfo._memSize;
+                else _lokBytes = 0;
```
andy-slac: Strange indentation

core/modules/memman/MemMan.h
abh3: It may be just you. "strict=false" -> tolerant. So, strict is the correct 
word here, you just have to interpret it in the context of its value.

Andy

core/modules/memman/MemManReal.cc
```diff
@@ -113,7 +117,55 @@ MemMan::Status MemManReal::getStatus(Handle handle) {
...
+         }
+         fsP = it->second;
+         fsP->serialize(true);
+         _numLocks++;
+    }
+
+    // Perform the lock and then drop the file set lock.
+    //
+    rc = fsP->lockAll(strict);
+    fsP->serialize(false);
```
jgates108: fsP->serialize() seems overly complicated. 

If fsp was a shared_ptr and hanCache used shared_ptr, you wouldn't need to call delete in MemManReal::unlock(), and wouldn't need to call serialize(false) in MemFileSet::~MemFileSet() as it wouldn't be deleted until everyone was done with it.

The code here could be changed to the following and MemFileSet::serialize(bool) can go away. 
```
     {
        std::unique_lock<std::mutex> hanGuard(hanMutex);
        ///  ... find handle in hanCache ...
        fsP = it->second;
        std::lock_guard<std::mutex> fspGuard(fsP->getMutex());  // return a reference to _setmutex
        _numlocks++;
        hanGuard.unlock();
        rc = fsP->lockAll(strict);
     }
```  
The main advantage here being there's no way you can get out of that block without unlocking the mutexes, nor can the MemFileSet destructor accidentally be called when some other thread has locked _setMutex.

core/modules/memman/MemFileSet.h
```diff
@@ -76,27 +78,57 @@ class MemFileSet {
...
+    //!        this method should be called with a common mutex locked.
+    //!
+    //! @param dolok If true, obtains a mutex on this object to serialize access.
+    //!              Otherwise, the lock is released.
+    //!
+    //! @return true  Normal return.
+    //! @return false lktst was true and serialization mutex was already locked.
+    //-----------------------------------------------------------------------------
+
+    bool serialize(bool dolok) {
```
jgates108: The code can be re-written using shared pointers and lock guards so this function isn't needed.

core/modules/memman/MemManReal.cc
abh3: Well, simplicity may be in the eye of the beholder, That particular lock 
is used to gaurd against multiple operations on the same file set that 
also happens to includes delete. Yes, I could go around replacing simple 
locks everywhere with shared pointers and additional gaurd locks. I 
actually find that more complicated and not very obvious on what the 
control flow is. Also, it would require substantial lock reorganization to 
obtain the same amount of parallelsm I wanted to maintain.

abh3:
This was already merged but orphaned on July 5th.

PR #241, opened 2016-07-12
Enabled largeResultPoolSize.

jgates108:


admin/templates/configuration/etc/qserv-czar.cnf
```diff
@@ -43,7 +43,7 @@ emptyChunkListFile = {{QSERV_DATA_DIR}}/qserv/emptyChunks.txt
 
 #[tuning]
```
andy-slac: Do you want to uncomment this line too?

PR #244, opened 2016-07-29
Made changes to work with newsest version of xrootd.

jgates108:


core/modules/qdisp/QueryResource.h
```diff
@@ -72,10 +72,12 @@ class QueryResource : public XrdSsiService::Resource {
     friend class QueryResourceDebug;
 
 private:
+    char const* _getRname(std::string const& rName);
     XrdSsiSession* _xrdSsiSession {nullptr}; ///< unowned, do not delete.
     std::shared_ptr<JobQuery> _jobQuery;
     std::string const _jobIdStr; ///< for debugging only
     util::InstanceCount _instC{"QueryResource"};
+    char* _rNameHolder;
```
fritzm: Why not just make this holder a std::string?   Then you could just assign instead of strdup in _getRname, and the cleanup in the dtor would be automatic?

fritzm: Okay, I tried it, and I see -- the order of init thing you were talking about.  :-)

So, your solution depends on the fact that pointer members are not initialized.  That's a little gross?  Personally, I find the original const cast less gross, but YMMV, and we are bike shedding now.

In the end, the situation that the xrootd API requires me to derive from a base class which must be ctor-init'd with a raw char*, and also requires I manage the lifetime of that char* externally, is no so great -- it results in this code which multiple people have tried to remove and then had to grudgingly restore.  The subtlety of what is required to make this work correctly has wasted engineer-days of time in the project so far -- myself, Serge, John, and now you.  And we're all pretty good at this sort of thing!

I think there's a reasonable argument here for using lifetime-managed objects (std::string instead of raw char*).  For your consideration.

For now, let's put the strudup back in and move ahead!

PR #245, opened 2016-08-02
DM-6972: Updates for new newinstall.sh location

fritzm:
Change scripts and docs to curl newinstall.sh from new location, as documented at https://pipelines.lsst.io/install/newinstall.html (plus some trailing whitespace removal courtesy of sublimetext)

Note that the corresponding change to admin/tools//docker/latest/scripts/install-stack.sh is done as part of a separate ticket (DM-6922).

### DIFF
--- a/admin/tools/qserv-install.sh
+++ b/admin/tools/qserv-install.sh
@@ -7,7 +7,7 @@ DIR=$(cd "$(dirname "$0")"; pwd -P)
 
 # Default values below may be overidden by cmd-line options
 MODE="internet mode"
-NEWINSTALL_URL="https://sw.lsstcorp.org/eupspkg/newinstall.sh"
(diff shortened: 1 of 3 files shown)


admin/tools/qserv-install.sh
```diff
@@ -7,7 +7,7 @@ DIR=$(cd "$(dirname "$0")"; pwd -P)
 
 # Default values below may be overidden by cmd-line options
 MODE="internet mode"
-NEWINSTALL_URL="https://sw.lsstcorp.org/eupspkg/newinstall.sh"
+NEWINSTALL_URL="https://raw.githubusercontent.com/lsst/lsst/12.0/scripts/newinstall.sh"
```
vaikuntht: Will the new one always be at this location or will it require constant updating? I wonder why the "12.0" is there.

PR #246, opened 2016-08-11
Minor fixes for issues flagged by Eclipse codan

fritzm:
Some uninitialized members missing virtual dtors

### DIFF
--- a/core/modules/memman/MemFile.h
+++ b/core/modules/memman/MemFile.h
@@ -50,7 +50,7 @@ public:
     struct MLResult {
         uint64_t bLocked;
         int      retc;
-        MLResult() {}
+        MLResult() : bLocked(0), retc(0) {}
         MLResult(uint64_t lksz, int rc) : bLocked(lksz), retc(rc) {}
     };
 
@@ -77,7 +77,7 @@ public:
     struct MFResult {
         MemFile* mfP;
         int      retc;
-        MFResult() {}
+        MFResult() : mfP(nullptr), retc(0) {}
         MFResult(MemFile* mfp, int rc) : mfP(mfp), retc(rc) {}
     };
 
--- a/core/modules/wsched/ChunkTasksQueue.h
+++ b/core/modules/wsched/ChunkTasksQueue.h
@@ -97,9 +97,9 @@ public:
     };
 
 private:
-    int _chunkId;          ///< Chunk Id for all Tasks in this instance.
-    bool _active{false};   ///< True when this is the active chunk.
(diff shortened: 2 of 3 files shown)


core/modules/memman/MemFile.h
```diff
@@ -50,7 +50,7 @@ class MemFile {
     struct MLResult {
         uint64_t bLocked;
         int      retc;
-        MLResult() {}
+        MLResult() : bLocked{false}, retc(0) {}
```
abh3: Ah, false is the wrong type here. Yeah it will work but it's inconsistent, use 0 instead.

fritzm: Yup, thanks for catching this -- I got thrown by the hungarian!

core/modules/wsched/ChunkTasksQueue.h
```diff
@@ -97,9 +97,9 @@ class ChunkTasks {
     };
 
 private:
-    int _chunkId;          ///< Chunk Id for all Tasks in this instance.
-    bool _active{false};   ///< True when this is the active chunk.
-    bool _resourceStarved; ///< True when advancement is prevented by lack of memory.
+    int _chunkId;                    ///< Chunk Id for all Tasks in this instance.
+    bool _active{false};            ///< True when this is the active chunk.
+    bool _resourceStarved{false};   ///< True when advancement is prevented by lack of memory.
```
abh3: Out of curiosity, do we have any standards on ho members should be initialized. I've now see people use all the variants allowed in C++11. Perhaps some guidance on which forms are better than others and in what contexts they should be used.

core/modules/memman/MemFile.h
abh3: Oh yes, Travis didn’t like your code at all. Why is that?

core/modules/wsched/ChunkTasksQueue.h
fritzm: No, we don't (have guidelines) that I am aware of.  There is one hard rule so far: do not use bracket-style ctor init for member variables that are references -- this is known to cause crashes under older gccs due to documented compiler bug.

Beyond that, John is a big fan of bracket init.  I think most new qserv code will be written in that style.  Since there are not any hard rules (beyond the one above) I default to letting file-by-file consistency win.  Since this file had paren-init style I opted for that.

core/modules/memman/MemFile.h
fritzm: I am not sure.  Travis seems like its spinning up two builds per pr right now, and the first always fails.  The second usually works, but in this case did not.  Investigating...

PR #247, opened 2016-08-12
Use eupspkg to build Qserv inside container

fjammes:
Enable parallel build.


jgates108:
I'm trying to run it on my desktop.  './3_build-git-image.sh  -R tickets/DM-6444' starts building something, but './3_build-git-image.sh -j4 -R tickets/DM-6444' just shows me the usage help message and exits. 

The help message should probably include the -j option and it would be nice if it gave and example of the correct form for -R option, which would be good to have in 4_build-configured-images.sh. I'm always forgetting "-" or "_" instead of "/" when I haven't run it in a while.

fjammes:
You don't have to provide -j option, maximum number of processors (but less than 32) will always be used, as defined by eupspkg default behaviour.

I've added an example for -R option.

PR #257, opened 2016-10-13
Renamed Backend SQLBackend and made it more visibile.

jgates108:


core/modules/wdb/ChunkResource.h
```diff
@@ -191,8 +189,7 @@ class ChunkResourceMgr {
     using Ptr = std::shared_ptr<ChunkResourceMgr>;
 
     /// Factory
-    static Ptr newMgr(mysql::MySqlConfig const& c);
-    static Ptr newFakeMgr();
+    static Ptr newMgr(SQLBackend::Ptr const& backend);
```
iagaponenko: I have a long question about a role of the Backend in the chunk manager. First of all why should this class be called SQLBackend instead of just Backend? Using the 'SQL' prefix implies possible plans for having other types of backends, which in my view wouldn't be such a bad idea, for example for unifying the real and testing branches of the code. Imagine this design based on the polymorphic model if the backend:

Backend: an abstract base class (an interface) which defines a technology-independent contract for the operations with temporary tables in the backend.

SQLBackend: derives from Backend in order to implement the real operations with the database tabs.

FakeBackend: derives from Backend in order to _simulate_ operations with the database.

In that case you may end up with a single (unified) static constructor for the class. Something like this:
   ChunkResourceMgr::newMgr(Backend::Ptr& iface);

This design would allow to avoid matching-and-mixing different behavior within the implementation of class ChunkResourceMgr and move this into Backend. In some sense the Backend interface will be playing a role of some kind of backend policy or so.

Does this make a sense to you?

iagaponenko: I see it. By the way, why did you remove "newFakeMgr"? Was it meant for the unit-testing of the class?

jgates108: I added the FakeBackend class for unit tests.

PR #259, opened 2016-10-21
Tickets/dm 7139

fjammes:


admin/tools/docker/deployment/swarm/manager/env-docker.sh
```diff
@@ -0,0 +1,27 @@
+# Configuration file sourced by Swarm scripts
+# Allow to customize Docker container execution
+
+# VERSION can be a git ticket branch but with _ instead of /
+# example: u_fjammes_DM-4295
+VERSION=tickets_DM-7139
```
jgates108: Change this to a comment and add VERSION=dev before pushing to master?

fjammes:
Hi @jgates108 , thanks for your review. 
Yvan will install tomorrow Docker 1.12 which is required for swarm to work. Then env-infrastructure.sh needs to be created and we can use Qserv masters as manager nodes in the short term, then launching 'setup-and-test.sh' should do the job.
This setup is not documented yet, sorry.
Here's an example for env-infrastructure.sh, generated for openstack setup, which might help (please use cluster first 25 node):
```shell
# Parameters related to Openstack instructure
# WARN: automatically generated by provisionning script, do not edit

SWARM_NODE="lsst-fabricejammes-qserv-swarm"

# Used by shmux
HOSTNAME_TPL="lsst-fabricejammes-qserv-"
WORKER_LAST_ID="2"

printf -v MASTER "%s0" "$HOSTNAME_TPL"

for i in $(seq 1 "$WORKER_LAST_ID");
do
    printf -v WORKERS "%s %s%s" "$WORKERS" "$HOSTNAME_TPL" "$i"
done
```

admin/tools/docker/deployment/swarm/manager/env-docker.sh
fjammes: I have to use tickets_DM-7139 here, because dev container do not support swarm configuration yet and test fails.

PR #260, opened 2016-10-25
Tickets/dm 7436

fjammes:


admin/tools/qserv-install.sh
```diff
@@ -7,8 +7,8 @@ DIR=$(cd "$(dirname "$0")"; pwd -P)
 
 # Default values below may be overidden by cmd-line options
 MODE="internet mode"
-NEWINSTALL_URL="https://raw.githubusercontent.com/lsst/lsst/master/scripts/newinstall.sh"
-VERSION="-t qserv"
+NEWINSTALL_URL="https://raw.githubusercontent.com/lsst/lsst/12.0/scripts/newinstall.sh"
```
vaikuntht: This is a hard coded link, is it the right one to use? I thought ``master`` was the latest one.

fjammes: LSST documentation recommend using the release link:
https://pipelines.lsst.io/install/newinstall.html
master could be broken as it is development version, isn't it?

timj: Master is not meant to be broken as such. LSST policy is that master is always fine. Whether you want the version associated with a specific release or the current "best" is a different question. For example, 12.0 does not support python 3 but master does.

fjammes: Thanks for your review :-)

PR #262, opened 2016-10-28
Extend a configuration of the deployment startup script to ensure

iagaponenko:
the containers will inherit timezone settings of their host machines.

### DIFF
--- a/admin/tools/docker/deployment/parallel/run.sh
+++ b/admin/tools/docker/deployment/parallel/run.sh
@@ -20,6 +20,17 @@ if [ -n "$ULIMIT_MEMLOCK" ]; then
     ULIMIT_OPT="--ulimit memlock=$ULIMIT_MEMLOCK"
 fi
 
+# Capture the local timezone as it needs to be passed down
+# into the containers.
+SET_CONTAINER_TIMEZONE=false
+LOCALTIME=$(readlink /etc/localtime)
+if [ -n "$LOCALTIME" ]; then
+    CONTAINER_TIMEZONE=${LOCALTIME#/usr/share/zoneinfo/}
+    if [ -n "$CONTAINER_TIMEZONE" ]; then
+        SET_CONTAINER_TIMEZONE=true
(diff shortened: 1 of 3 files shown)


admin/tools/docker/deployment/parallel/run.sh
```diff
@@ -51,6 +51,7 @@ docker run --detach=true \
     -e "QSERV_MASTER=$MASTER" \
     $DATA_VOLUME_OPT \
     $LOG_VOLUME_OPT \
+    -v /etc/localtime:/etc/localtime:ro \
```
iagaponenko: Hi Fabrice

Thank you very much for the suggestion! It makes a lot of sense to me. I’ve made proper modifications to the code. Could you review them please?

It’s in :
https://github.com/lsst/qserv/pull/262/files

I have a question to you – how do I build the container out of this? Will it be built automatically? I would certainly want to pull this container into PDAC to give it a try.

Thanks,
Igor

ktlim: The preference is for all machine activities to be in UTC, but if that is not feasible, local timezone is allowed.  All logs must have the timezone in each timestamp.  See DM-1203.

iagaponenko: The procedure has been improved (look at the 'Files Changes' tab of this pull request page) to allow more control in setting both 'timezone' and 'local time'. What's presently implemented is:
- the default timezone of a container is set to Etc/UTC
- when the container starts it will check if the deployment procedure wants it to switch to some other timezone. And if so the script will make proper adjustment. Otherwise it will leave the default timezone as it is.
- the 'run.sh' will obtain and pass the local timezone info down to the container

In principle this may be further improved by adding a configuration option into env.sh. This file is supposed to be tuned to address site-specific requirements (such as a location of a temporary folder, the data (MySQL/MariaDB) directory of Qserv, a location within a host system where Qserv would place its log files, etc.).

PR #263, opened 2016-11-01
Change to allow logging to be routed to log4cxx

abh3:
This pull request depends on recent changes in the xrdssi xrootd-github branch.


core/modules/SConscript
```diff
@@ -194,8 +194,12 @@ shlibs = {}
 shlibs["qserv_common"] = dict(mods="""global memman proto mysql sql util""",
                               libs="""log protobuf mysqlclient_r boost_thread crypto""")
 
+# library implementing xrootd logging intercept (worker side)
+shlibs["xrdlog"] = dict(mods="""xrdlog""",
+                        libs="""log XrdSsiLib""")
+
 # library implementing xrootd services (worker side)
-shlibs["xrdsvc"] = dict(mods="""wbase wcontrol wconfig wdb wlog wpublish wsched xrdsvc""",
+shlibs["xrdsvc"] = dict(mods="""wbase wcontrol wconfig wdb wpublish wsched xrdsvc""",
```
andy-slac: It looks like with you changes `wlog` will no longer be needed. Can you open a JIRA ticket to delete it at some point later so we don't forget to get rid of deadwood. Or maybe even drop it now if you feel confident.

abh3: You know, I was thinking about that. Maybe we really want to keep it in some form just in case we scuttle the log4cxx back end and use the xrootd logging system that already include file rotation and the like. In that case, it would be prudent to keep this.

core/modules/czar/CzarConfig.cc
```diff
@@ -38,6 +41,22 @@ namespace {
 
 LOG_LOGGER _log = LOG_GET("lsst.qserv.czar.CzarConfig");
 
+void QservLogger(struct timeval const& mtime,
```
andy-slac: This function is almost identical to one XrdLogging.cc. As it does not have direct dependency on xrootd we can probably move it on a common place, for example `util` module (maybe find it a better name).

core/modules/czar/CzarConfig.cc
```diff
@@ -38,6 +41,22 @@ namespace {
...
+                 unsigned long         tID,
+                 const char*           msg,
+                 int                   mlen) {
+
+    static log4cxx::spi::LocationInfo xrdLoc("client", "<xrdssi>", 0);
+    static LOG_LOGGER myLog = LOG_GET("lsst.qserv.xrdssi.msgs");
+
+    if (myLog.isInfoEnabled()) {
+        std::string theMsg(msg, mlen);
```
andy-slac: Looking at the log it seems that message comes from xrootd with a trailing newline character. To avoid empty lines in the log can you strip that newline character here?

core/modules/xrdlog/XrdLogging.cc
```diff
@@ -0,0 +1,74 @@
...
+
+// Qserv headers
+
+/******************************************************************************/
+/*                L o g g i n g   I n t e r c e p t   H o o k                 */
+/******************************************************************************/
+  
+namespace {
+
+const char* origin;
```
andy-slac: This could also be made a static inside `QservLogger` function. But if you are going to move `QservLogger` to some other location you will probably need to repackage it in some way.

admin/templates/configuration/etc/log4cxx.worker.properties
```diff
@@ -2,9 +2,9 @@
 # Configuration file for log4cxx
 #
 
-log4j.rootLogger=DEBUG, XRDLOG
+log4j.rootLogger=INFO, CONSOLE
```
abh3: OK, that was the initial recommendation. If I change INFO to DEBUG, then I should change the test whether or not those messages go through. I assume the call will be isDebugEnabled() (at least that is what I get from log4cxx docs). It would be good if you confirm.

core/modules/czar/CzarConfig.cc
abh3: Ah, which function are we talking about? QservLogger? Well, if you look closely they are not quite the same as one looks for a server context while the other assume a client context and the messages reflect that. I suppose I could come up with some fancy way of distinguishing that but it would be straightforward. Is it really that annoying that an eight line function is "duplicated" in two places?

core/modules/czar/CzarConfig.cc
abh3: OK, I can do that. John tells me that all the logging frameworks add the new line.

core/modules/xrdlog/XrdLogging.cc
abh3: Well, if I move it to another place it will likely will have to become a class of its own so the origin can be properly set by either the czar or the logging plugin. It certainly cannot be anonymous anymore.

core/modules/czar/CzarConfig.cc
andy-slac: Yes, it is very annoying, we should try to avoid repeating code.`util` is actually not a perfect place for that code, the library that we build from that package already has too many dependencies which are not needed for your xrootd plugin. LEt's say we keep it as it is now and then think about it at next refactoring.

core/modules/czar/CzarConfig.cc
andy-slac: Yes, indeed. Now you should remember to edit two copies of the same function 😏

core/modules/xrdlog/XrdLogging.cc
andy-slac: It should be anonymous or hidden in some way, instead you have to provide API for changing it, never expose globals to your clients.

core/modules/xrdlog/XrdLogging.cc
```diff
@@ -0,0 +1,75 @@
...
+                 unsigned long         tID,
+                 const char*           msg,
+                 int                   mlen) {
+
+    static log4cxx::spi::LocationInfo xrdLoc(origin, "<xrdssi>", 0);
+    static LOG_LOGGER myLog = LOG_GET("lsst.qserv.xrdssi.msgs");
+
+    if (myLog.isInfoEnabled()) {
+        std::string theMsg(msg, mlen);
+        if (mlen && theMsg[mlen-1] == '\n') theMsg.erase(mlen-1, 1);
```
andy-slac: I'd probably do slightly more efficient variant:

    while (mlen && msg[mlen-1] == '\n') --mlen;  // strip all trailing newlines
    std::string theMsg(msg, mlen);

but `erase()` should work as well 😄

PR #268, opened 2016-11-18
Moved query registration outside of the UserQuerySelect constructor.

jgates108:


core/modules/ccontrol/UserQuerySelect.cc
```diff
@@ -396,6 +391,7 @@ void UserQuerySelect::_qMetaRegister()
     // register query, save its ID
     _qMetaQueryId = _queryMetadata->registerQuery(qInfo, tableNames);
     _queryIdStr = QueryIdHelper::makeIdStr(_qMetaQueryId);
+    _resultTable += "result_" + std::to_string(_qMetaQueryId);
```
jgates108: += not right here

PR #270, opened 2016-11-23
support for weekly docker builds via jenkins

jhoblitt:


fritzm:
@jhoblitt The parent ticket for this PR is closed; is it still desired for testing purposes?  If so we could rebase it and get it in; if not please go ahead and close the PR, thx!

PR #274, opened 2016-11-28
Improve cmsd configuration

fjammes:


admin/templates/configuration/etc/lsp.cf
```diff
@@ -16,8 +16,10 @@ if named master
     # Use manager mode
     all.role manager
 
-    # Enable support for multiple writable file copies
-    cms.space mwfiles linger 0
+    # Specify that no significant free space is required on servers
+    # Indeed urrent configuration doesn't expect to be dynamically
```
vaikuntht: "urrent" -> "current"

PR #275, opened 2016-11-29
Tickets/dm 8001

iagaponenko:


admin/python/lsst/qserv/admin/dataLoader.py
```diff
@@ -765,7 +771,7 @@ def _makeIndex(self, database, table):
         self._log.info('Generating index %r.%r', self.indexDb, metaTable)
 
         # try to delete existing table first
-        self.czarWmgr.dropTable(self.indexDb, metaTable, mustExist=False)
+        #self.czarWmgr.dropTable(self.indexDb, metaTable, mustExist=False)
```
andy-slac: Can you add a comment why you commented-out this line so that we don't forget it in the future? Or if it's truly not needed better to remove it completely.

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -568,9 +568,11 @@ std::string KvInterfaceImplMySql::dumpKV() {
         if (row[1].first == nullptr or *row[1].first == '\0') {
             result += "\\N";
         } else {
+/*
             if (strchr(row[1].first, '\n') != nullptr) {
                 throw CssError("KvInterfaceImplMySql::dumpKV - value contains newline");
             }
+*/
```
andy-slac: I think this exception was here because dump is not readable by our tools if any value field contains newline. Do you use dumps for anything? If we need dumps we should use better format or maybe just escape newline characters (`\n`) on output

admin/python/lsst/qserv/admin/dataLoader.py
andy-slac: I'm somewhat worried about duplicate records if table is not dropped. If you do incremental data uploads and update secondary index every time then it will try to insert duplicate records into secondary index. With our current implementation duplicate records are ignored, but this is most likely just a happy coincidence and not a part of a design (design was to drop table to avoid trouble). We should keep this in mind when we update our data loading procedures.

admin/bin/qserv-admin.py
```diff
@@ -306,6 +313,27 @@ def _parseCreateNode(self, tokens):
...
+        Throws css.CssError, _ToolError.
+        """
+        if len(tokens) != 3:
+            raise _IllegalCommandError("Unexpected number of arguments.")
+
+        (dbTbName, chunkIdStr, nodeName) = tokens
+        if '.' not in dbTbName:
+            raise _IllegalCommandError("Invalid argument '{0}', should be <dbName>.<tbName>".format(dbTbName))
+
+        (dbName, tbName) = dbTbName.split('.')
```
iagaponenko: This was a "cut-and-paste" pattern found in 4 other places within the same file. I chose not to refactor the application, and just leave it all as is. Ideally, there database/table name translation should be handled in a dedicated function/method in which all possible syntax errors would be taken care of. Thank you for bringing this up!

admin/bin/qserv-admin.py
```diff
@@ -340,9 +368,33 @@ def _parseDelete(self, tokens):
...
 
+    def _deleteChunk(self, tokens):
+        """
+        Subparser - handles all DELETE CHUNK requests.
+        Throws css.CssError, _ToolError.
+        """
+        if len(tokens) != 2:
+            raise _IllegalCommandError("unexpected number of arguments")
+
+        (dbTbName,chunkIdStr) = tokens
```
iagaponenko: I will correct this!

admin/bin/qserv-admin.py
```diff
@@ -169,6 +169,9 @@ def __init__(self, connInfo):
     RESTORE <inFile>;
     SHOW DATABASES;
     SHOW NODES;
+    CREATE CHUNK <dbName>.<tableName> <chunk> <nodeName>;
+    DELETE CHUNK <dbName>.<tableName> <chunk> <nodeName>;
```
iagaponenko: You're right! The node name should not be there (in the 'DELETE CHUNK' operation). I will remove it.

admin/bin/qserv-admin.py
```diff
@@ -408,6 +462,27 @@ def _parseShowNodes(self):
...
+        """
+
+        if len(tokens) != 1:
+            raise _IllegalCommandError("Unexpected number of arguments.")
+
+        (dbTbName,) = tokens
+        if '.' not in dbTbName:
+            raise _IllegalCommandError("Invalid argument '{0}', should be <dbName>.<tbName>".format(dbTbName))
+
+        (dbName, tbName) = dbTbName.split('.')
```
iagaponenko: I will fix this.

core/modules/css/KvInterfaceImplMySql.cc
iagaponenko: I don't think the output of the command is meant to be parsed or used anywhere beyond the usual human-driven debugging process. I've searched all potential uses of the command within the package, and I found no single use of it.

A problem here is that the original code crashed when attempting to dump table schemas loaded into CSS by the qserv-data-load.py tool.

admin/python/lsst/qserv/admin/dataLoader.py
iagaponenko: And this is exactly a problem with the distributed (parallel) data loading. Each instance of the loader will always attempt to drop the table before attempting to load it with its own portion of the index. This was not a problem for the various integration tests which were relying on a single run of the loader tool.

Perhaps we should consider controlling this behavior through an option passed to the loader script and passing this flag into the code?

core/modules/css/KvInterfaceImplMySql.cc
andy-slac: Well, the output is not very human-friendly so it has to be for something else :) I think qserv-admin was supposed to have a RESTORE command which would read DUMP output, but it was never implemented.

admin/python/lsst/qserv/admin/dataLoader.py
andy-slac: I'm OK with dropping DROP TABLE if it works both for you and integration tests. What I'm saying is that it works by pure accident now and we should try not to break this accidental feature (or try to make it more official).

core/modules/css/KvInterfaceImplMySql.cc
andy-slac: I don't remember any outstanding tickets for qserv-admin. We should improve it of course to support all the features that we need. We don't care about RESTORE for now, maybe we should wait until there is a use case for that.

iagaponenko: Agreed.

admin/python/lsst/qserv/admin/dataLoader.py
iagaponenko: It turns out the data loader script already has this option:

-e, --delete-tables   If specified then existing tables in database will be
                        deleted if they exist, this includes both data and
                        metadata.

Should I just rely on it here? I can propagate this flag into the code which makes the secondary index.

PR #278, opened 2016-12-06
Refactored some qana items.

jgates108:


core/modules/query/QueryContext.cc
```diff
@@ -31,19 +31,122 @@
...
+
+    if (rc) { return colNames; }
+    MYSQL_RES* result = mysql_use_result(mysqlConn.getMySql());
+
+    auto resSchema = mysql::SchemaFactory::newFromResult(result);
+    std::string resSchemaStr;
+    for(auto i=resSchema.columns.begin(), e=resSchema.columns.end(); i != e; ++i) {
+        resSchemaStr += i->name + ", ";
+    }
+    LOGS(_log, LOG_LVL_DEBUG, "tableSchema resSchema=" << resSchemaStr);
```
andy-slac: If above code is only used for debugging can you move it inside `if (LOG_CHECK_LVL(_log, LOG_LVL_DEBUG))`?

core/modules/query/QueryContext.cc
```diff
@@ -31,19 +31,122 @@
...
+
+    MYSQL_ROW row;
+    int j = 0;
+    while ((row = mysql_fetch_row(result))) {
+        auto lengths = mysql_fetch_lengths(result);
+        colNames.emplace_back(row[0], lengths[0]);
+        LOGS(_log, LOG_LVL_DEBUG, "tableSchema row[" << j << "]=" << colNames[j]);
+        ++j;
+    }
```
andy-slac: You need `mysql_free_result(result)` after you finished looking at MYSQL_RES.

core/modules/query/QueryContext.cc
```diff
@@ -52,35 +155,65 @@ QueryContext::resolve(std::shared_ptr<ColumnRef> cr) {
...
     DbTablePair p;
-    if (cr->table.empty()) { // No db or table: choose first resolver pair
-        p = resolverTables[0];
-        // TODO: We can be fancy and check the column name against the
-        // schema for the entries on the resolverTables, and choose
-        // the matching entry.
+    if (cr->table.empty()) { // No db or table.
+        if (columnToTablesMap.empty()) {
+            // This should only be the case if no connection to the database could be made, which
+            // should only happen in unit tests. This should get through the unit tests and have
```
andy-slac: "which should only happen in unit tests"

Or if connection to database fails for any other reason?

core/modules/query/QueryContext.cc
```diff
@@ -52,35 +155,65 @@ QueryContext::resolve(std::shared_ptr<ColumnRef> cr) {
...
     } else if (cr->db.empty()) { // Table, but not alias.
         // Match against resolver stack
-        for(DbTableVector::const_iterator i=resolverTables.begin(),
-                e=resolverTables.end();
-            i != e; ++i) {
+        DbTableVector::const_iterator i=resolverTables.begin(), e=resolverTables.end();
+        for(; i != e; ++i) {
             if (i->table == cr->table) {
                 p = *i;
                 break;
```
andy-slac: Could there be more than one table with the same name (different databases)?

core/modules/qana/QservRestrictorPlugin.cc
```diff
@@ -77,18 +76,29 @@ namespace qana {
...
     }
-    query::DbTablePair p = context.resolve(cr);
-    cr->table = p.table;
-    cr->db = p.db;
-    return cr;
+    std::string column = cr->column;
+    query::DbTableSet set = context.resolve(cr);
+    for (auto const& dbTblPair : set) {
+        cr = std::make_shared<query::ColumnRef>(dbTblPair.db, dbTblPair.table, column);
+        columnRefs.push_back(cr);
```
jgates108: How about emplace_back?

core/modules/qana/PostPlugin.h
```diff
@@ -0,0 +1,68 @@
...
+    virtual void applyPhysical(QueryPlugin::Plan& plan, query::QueryContext&);
+
+    int _limit;
+    std::shared_ptr<query::OrderByClause> _orderBy;
+};
+
+}}} // namespace lsst::qserv::qana
+
+
+#endif // LSST_QSERV_QANA_POSTPLUGIN_H
```
iagaponenko: If I understand correctly LSST C++ Style Guide does require the default contructor to be coded even if it's trivial. See: https://developer.lsst.io/coding/cpp_style_guide.html#constructors-and-destructors

core/modules/query/QueryContext.cc
jgates108: If the connection to the database fails, this should be the same database that the czar is using to test the user queries before sending them to workers, so the user query should fail and perhaps the czar should be terminated at that point if it can't connect. The worst case scenario here is that a query that only needs to go to one worker is sent to all workers. I think this is a slightly better option than eliminating the unit tests. Other options, such as  creating artificial  columnToTablesMap for the unit tests, is probably worthy of another ticket.

core/modules/query/QueryContext.cc
```diff
@@ -31,19 +31,122 @@
...
+std::vector<std::string> QueryContext::getTableSchema(std::string const& dbName,
+                                                      std::string const& tableName) {
+    // Check if it is in our list of table schemas
+    DbTablePair dTPair(dbName, tableName);
+
+    // Get the table schema from the local database.
+    std::vector<std::string> colNames;
+    mysql::MySqlConnection mysqlConn{mysqlSchemaConfig};
+    if (!mysqlConn.connected()) {
+        LOGS(_log, LOG_LVL_ERROR, "QueryContext::getTableSchema failed to connect to schema database!");
```
jgates108: In the one place where this function is called, I'd just catch the exception and throw a warning due to unit tests, which is related to the other comment below.

core/modules/qana/QservRestrictorPlugin.cc
andy-slac: You could do it too:

    columnRefs.emplace_back(new query::ColumnRef(dbTblPair.db, dbTblPair.table, column));

but it's not very exception-safe, object will leak if `emplace_back` throws.

core/modules/query/QueryContext.cc
andy-slac: I'm not proposing to eliminate unit tests, I'm just saying that comment is not entirely correct, there are cases when connection to database may fail and `columnToTablesMap` will be left empty. I feel that we don't handle this case quite correctly yet, right now it's a silent failure which I do not quite like.

core/modules/query/QueryContext.cc
andy-slac: But what if this fails outside unit test? I think unit test failures should be handled separately, I'd prefer that in production this failure was not suppressed but resulted in user-visible error.

PR #279, opened 2016-12-16
Tickets/dm 8680

fjammes:


jhoblitt:
This looks OK to me but I haven't tried running it.

PR #280, opened 2016-12-19
fix quoting preventing variable interpolation

jhoblitt:


fjammes:
Hi @jhoblitt, thanks for this fix, OK to merge :-)

PR #281, opened 2017-01-09
Tickets/dm 8196

fjammes:


admin/tools/docker/deployment/parallel/run-large-scale-tests.sh
```diff
@@ -0,0 +1,17 @@
...
+# @author Fabrice Jammes SLAC/IN2P3
+
+set -e
+
+DIR=$(cd "$(dirname "$0")"; pwd -P)
+. "${DIR}/env.sh"
+
+TEST_DIR="/tmp/qserv_testscale"
+rm -rf "$TEST_DIR"
+GIT_REF="tickets/DM-8196"
```
fjammes: Ok with all your remarks, thanks! About separation of qserv and deployment code, I think we could target that in the middle-term, but I still performs some update to configuration code so this is convenient to keep it together in the short-term.

admin/tools/provision/cloudmanager.py
```diff
@@ -441,6 +523,21 @@ def build_cloudconfig(self):
...
+  permissions: "0544"
+  owner: "root"
+  content: |
+    #!/bin/sh
+    set -e
+    while [ ! -b /dev/vdb1 ] ;
+    do
+      sleep 2
+      echo "---WAITING FOR CINDER VOLUME---"
+    done
```
fjammes: I keep it like that for know, because this code should be handle soon using openstack-heat configuration files.

PR #282, opened 2017-01-13
Made changes to prevent race conditions from booting user queries.

jgates108:


core/modules/wbase/Task.h
```diff
@@ -151,6 +151,8 @@ class Task : public util::CommandThreadPool {
     void setMemHandle(memman::MemMan::Handle handle) { _memHandle = handle; }
     void setMemMan(memman::MemMan::Ptr const& memMan) { _memMan = memMan; }
     void waitForMemMan();
+    bool getSafeToMoveRunning() { return _safeToMoveRunning; }
+    void setSafeToMoveRunning(bool val) { _safeToMoveRunning = val; }
```
andy-slac: Is `setSafeToMoveRunning()` only meant for testing? Can you add a comment which says so?

core/modules/wpublish/QueriesAndChunks.cc
```diff
@@ -462,17 +463,32 @@ QueriesAndChunks::removeQueryFrom(QueryId const& qId,
...
+            auto taskSched = task->getTaskScheduler();
+            if (taskSched != nullptr && (taskSched == sched || sched == nullptr)) {
+                // removeTask will only return the task if it was removed from the
+                // queue for its scheduler.
+                auto removedTask = taskSched->removeTask(task, moveRunning);
+                if (removedTask != nullptr) {
+                    removedList.push_back(removedTask);
+                } else {
+                    taskListNotRemoved.push_back(task);
+                }
```
andy-slac: Does `removeTaks(task)` always return `task` if not `nullptr`? The code in the `if` sort of implies that but it's somewhat confusing me to think about what happens if `task` and `removedTask` are not the same. Would it be simpler if `removeTaks(task)` just returned bool?

core/modules/wpublish/QueriesAndChunks.h
```diff
@@ -73,6 +74,7 @@ class QueryStatistics {
     int _tasksCompleted{0};
     int _tasksRunning{0};
     int _tasksBooted{0}; ///< Number of Tasks booted for being too slow.
+    std::atomic<bool> _queryBooted{false}; ///< True when the entire query booted.
```
andy-slac: Does it need to be atomic? (My brain only runs single thread today :repeat_one:)

jgates108: It really should be, at it's written by one thread and read several others.

core/modules/wpublish/QueriesAndChunks.cc
```diff
@@ -462,17 +463,32 @@ QueriesAndChunks::removeQueryFrom(QueryId const& qId,
...
             }
         }
-    }
+        taskList = taskListNotRemoved;
```
jgates108: This is a place where I really would expect the optimizer to do that without being explicit.

PR #283, opened 2017-01-15
Improve database connection management (DM-8990)

andy-slac:
Get rid of sqlalchemy connection pooling completely to avoid issues with
connections being left open for longer than we need. Also update dbMgr
code to use context management ("with") for transactions.

### DIFF
--- a/core/modules/wmgr/python/config.py
+++ b/core/modules/wmgr/python/config.py
@@ -38,6 +38,7 @@ import tempfile
 from .errors import ExceptionResponse
 from lsst.db.engineFactory import getEngineFromArgs
 from lsst.qserv import css
+from sqlalchemy.pool import NullPool
 
 #----------------------------------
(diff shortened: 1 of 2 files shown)


core/modules/wmgr/python/config.py
```diff
@@ -96,41 +97,60 @@ def __init__(self, appConfig):
...
-        kwargs['query'] = dict(local_infile=1)
+        """ Return database engine.
+
+        Standard Pool class causes delays in releasing connections, so we
+        disable connection pooling by using NullPool class. Another reason to
+        disable pool is that MySQL connections cannot be shared across many
+        threads.
+        """
+        if self._db is not None:
+            return self._db
```
andy-slac: Should not be any, engine is a sort of a factory for connections, if the connection pool is disabled then there is no state which depends on MySQL.

PR #284, opened 2017-01-17
Implement user-friendly template customization (DM-2580)

fjammes:


admin/tools/docker/deployment/parallel/env.example.sh
```diff
@@ -13,12 +13,19 @@ BRANCH=dev
 # `docker run` settings
 # =====================
 
+# Customized configuration templates directory location
+# on docker host, optional
+HOST_CUSTOM_DIR=/qserv/custom
+
```
jgates108: Am I correct in believing that /qserv/custom would normally be an empty directory and any configuration files found in this directory will be used by the container?

Does the container actively use this directory or does it just copy the contents to the real configuration directory when the container starts?

It might useful to add a bit to this comment and the location of the HOW-TO file.

fjammes: OK, I've added that. Thanks :-)

PR #285, opened 2017-01-23
Replace mariadb 'sock' with 'socket'

fjammes:


fjammes:
Build fails because CI do not support dependencies change, and qserv_testdata has been changed.

PR #286, opened 2017-01-24
Tickets/dm 8629

fjammes:


admin/templates/configuration/etc/init.d/xrootd
```diff
@@ -74,12 +74,11 @@ start_service() {
     pidfile=$(pid_file_name $prog)
     lockfile=${QSERV_RUN_DIR}/var/lock/subsys/${prog}
     BINARY=${XROOTD_DIR}/bin/${prog}
-#   logfile=${LOG_DIR}/${prog}.log
```
vaikuntht: Not sure why this was removed, or, does it need to be replaced with something else?

fjammes: It was commented, so no more used.
Its current value is: 
logfile=@libXrdSsiLog.so

PR #288, opened 2017-02-01
catching ServerError exception code 409 when attempting to create an …

iagaponenko:
The modification is needed to allow loading chunk tables in parallel by multiple client loaders run on many hosts.

### DIFF
--- a/admin/python/lsst/qserv/admin/dataLoader.py
+++ b/admin/python/lsst/qserv/admin/dataLoader.py
@@ -626,16 +626,16 @@ class DataLoader(object):
 
         for name, wmgr in connections:
 
-            self._log.info('Make dummy chunk table for %r', name)
+            self._log.info('Make dummy chunk table for %r', table)
 
             # just make regular chunk with special ID, do not load any data
             try:
                 wmgr.createChunk(database, table, 1234567890, overlap=self.partOptions.isSubChunked)
             except ServerError as exc:
                 if exc.code == 409:
-                    self._log.info('Dummy chunk 1234567890 exists for table %r', name)
+                    self._log.info('Dummy chunk 1234567890 exists for table %r', table)
                 else:
-                    self._log.critical('Failed to create dummy chunk 1234567890 for table %r', name)
+                    self._log.critical('Failed to create dummy chunk 1234567890 for table %r', table)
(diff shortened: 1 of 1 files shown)


admin/python/lsst/qserv/admin/dataLoader.py
```diff
@@ -582,8 +582,14 @@ def _loadChunkedData(self, database, table):
...
-                    wmgr.createChunk(database, table, chunkId, overlap=self.partOptions.isSubChunked)
-                    self.createdChunks.add(chunkId)
+                    try:
+                        wmgr.createChunk(database, table, chunkId, overlap=self.partOptions.isSubChunked)
+                        self.createdChunks.add(chunkId)
+                    except ServerError as exc:
+                        if exc.code == 409:
+                            self._log.info('Chunk %r exists for table %r', chunkId, table)
+                        else:
+                            self._log.critical('Failed to create chunk %r for table %r', chunkId, table)
```
andy-slac: Do you want to re-throw exception here? I think there is no point of continuing if chunk has not been created. Same thing applies below.

iagaponenko: Actually - yes, I would like just to post an information message that the chunk already exists (because some other parallel process has created it) and continue as if noting bad has happened. Note that code 409 is reported when the table already exists!

admin/python/lsst/qserv/admin/dataLoader.py
```diff
@@ -621,8 +629,14 @@ def _createDummyChunk(self, database, table):
             self._log.info('Make dummy chunk table for %r', name)
 
             # just make regular chunk with special ID, do not load any data
-            wmgr.createChunk(database, table, 1234567890, overlap=self.partOptions.isSubChunked)
-
+            try:
+                wmgr.createChunk(database, table, 1234567890, overlap=self.partOptions.isSubChunked)
+            except ServerError as exc:
+                if exc.code == 409:
+                    self._log.info('Dummy chunk 1234567890 exists for table %r', name)
```
andy-slac: I think `name` is not the name of the table here but a name of the worker, the message may be confusing if you mix "table" and worker name. Same for the message below.

PR #290, opened 2017-02-08
Removed Executive::_reapRequests to stop race condition.

jgates108:


core/modules/qdisp/Executive.cc
```diff
@@ -353,35 +357,34 @@ void Executive::_unTrack(int jobId) {
...
-        } else {
-            ++iter;
-        }
+
+/// _incompleteJobsMutex must be held before calling this function.
+/// @return: a string containing a list of incomplete jobs containing up to 'maxToList' jobs.
+///          If maxToList is less than 0, all jobs are printed
+std::string Executive::_getIncompleteJobsString(int maxToList) {
+    std::ostringstream os;
+    int size = _incompleteJobs.size();
```
iagaponenko: Is there any need to cache it here? The only client of  'size' is at line 378.

core/modules/qdisp/Executive.cc
```diff
@@ -353,35 +357,34 @@ void Executive::_unTrack(int jobId) {
...
+    int c = 0;
+    if (maxToList < 0) maxToList = size;
```
iagaponenko: Since you're already holding a mutex '_incompleteJobsMutex ' preventing the list of incomplete jobs to change then this could become:

  if (maxToList <0) maxToList = _incompleteJobs.size();

core/modules/qdisp/Executive.cc
```diff
@@ -353,35 +357,34 @@ void Executive::_unTrack(int jobId) {
...
+    os << "_incompleteJobs listing first" << maxToList << " of size=" << _incompleteJobs.size() << " ";
+    for(auto j = _incompleteJobs.begin(), e = _incompleteJobs.end(); j != e && c < maxToList; ++j) {
+        os << j->first << " ";
+        ++c;
```
iagaponenko: This should go into the increment part of the 'for' loop:

for(auto j = ... ; j != e && c < maxToList; ++j, ++c)

PR #291, opened 2017-02-11
Added code to xrootd PRD_hold feature.

jgates108:


core/modules/czar/Czar.cc
```diff
@@ -52,6 +57,13 @@ namespace lsst {
 namespace qserv {
 namespace czar {
 
+Czar::Ptr Czar::_czarPtr;
+
+Czar::Ptr Czar::createCzar(std::string const& configPath, std::string const& czarName) {
+    _czarPtr.reset(new Czar(configPath, czarName));
+    return _czarPtr;
+}
+
```
andy-slac: I hate singletons, could we avoid changing czar class into singleton? Also you changes small number of classes to depend on czar which is additional dependency I do not like either :) Having `LargeResultMgr` instance a member of czar looks like a right scope, can you just pass it to the objects that need it via parameters?

core/modules/qdisp/LargeResultMgr.h
```diff
@@ -0,0 +1,81 @@
...
+ * the GNU General Public License along with this program.  If not,
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ *
+ */
+
+#ifndef LSST_QSERV_QDISP_LARGERESULTMGR_H
+#define LSST_QSERV_QDISP_LARGERESULTMGR_H
+
+// System headers
+#include <atomic>
```
andy-slac: atomics are not used in this header?

core/modules/qdisp/LargeResultMgr.h
```diff
@@ -0,0 +1,81 @@
...
+#include <memory>
+#include <mutex>
```
andy-slac: `#include <string>` is needed

core/modules/qdisp/LargeResultMgr.cc
```diff
@@ -0,0 +1,142 @@
...
+}
+
+namespace lsst {
+namespace qserv {
+namespace qdisp {
+
+
+/// Set the value of the xrootd semaphore for tracking large response blocks to start.
+void LargeResultMgr::_setup() {
+    std::lock_guard<std::mutex> lck(_mtx);
```
andy-slac: nitpick - the code that runs in constructor does not need locking.

core/modules/czar/Czar.cc
```diff
@@ -68,8 +80,15 @@ Czar::Czar(std::string const& configPath, std::string const& czarName)
...
-    rproc::InfileMerger::setLargeResultPoolSize(largeResultPoolSize);
+    int largeResultConcurrent = _czarConfig.getLargeResultConcurrentMerges();
+    LOGS(_log, LOG_LVL_INFO, "config largeResultConcurrent=" << largeResultConcurrent);
+    _largeResultMgr = std::make_shared<qdisp::LargeResultMgr>(largeResultConcurrent);
+
+    int xrootdCBThreadsMax = _czarConfig.getXrootdCBThreadsMax();
+    int xrootdCBThreadsInit = _czarConfig.getXrootdCBThreadsInit();
+    LOGS(_log, LOG_LVL_INFO, "config xrootdCBThreadsMax=" << xrootdCBThreadsMax);
+    LOGS(_log, LOG_LVL_INFO, "config xrootdCBThreadsInit=" << xrootdCBThreadsInit);
+    XrdSsiProviderClient->SetCBThreads(xrootdCBThreadsMax, xrootdCBThreadsInit);
```
abh3: Well, xrootdCBThreadsInit is really not descriptive of what that parameter does. The first argument is the maximum number of threads to use for callbacks. The second parameters is the number of threads to set aside to handle network events. Just my thing here.

core/modules/qdisp/LargeResultMgr.cc
```diff
@@ -0,0 +1,142 @@
...
+    for (int j = 0; j < _runningCountMax; ++j) {
+        _post("initialization");
+    }
+}
+
+
+/// Check if any large result blocks can be run.
```
andy-slac: Will this do something else in the future, right now it does not seem to check anything?

core/modules/qdisp/LargeResultMgr.cc
```diff
@@ -0,0 +1,142 @@
...
+    if (delayAll == _delayAllPosts) return;
+    _delayAllPosts = delayAll;
+    LOGS(_log, LOG_LVL_INFO, " change delayAll=" << _delayAllPosts);
+    if (!_delayAllPosts) {
+        _freeDelayedPosts();
+    }
+}
+
+
+/// Increment xrootd's large result semaphore to allow another large result to run.
```
abh3: Just a nit pick here. mentions of xrootd should be replaced by SSI. Note that the publicframework does not have a stick of xrootd code in it. Indeed, the semaphore is specific to SSI and xrootd does not get involved whatsoever here.

core/modules/wdb/QueryRunner.cc
```diff
@@ -233,8 +233,15 @@ bool QueryRunner::_fillRows(MYSQL_RES* result, int numFields, uint& rowCount, si
...
+        unsigned int szLimit = proto::ProtoHeaderWrap::PROTOBUFFER_DESIRED_LIMIT;
+
+        // Use small blocks until it is considered a large result.
+        if (!_largeResult) {
+            szLimit = std::min(szLimit, _initialBlockSize);
+        }
+
         // Each element needs to be mysql-sanitized
-        if (tSize > proto::ProtoHeaderWrap::PROTOBUFFER_DESIRED_LIMIT) {
+        if (tSize > szLimit) {
```
abh3: How is szLimit set? Is it always guaranteed to be less than the protobuffer hard limit?

core/modules/qdisp/LargeResultMgr.cc
abh3: I think it does, it calls _post() to set the initial hold semaphore value.

andy-slac: I was thinking about `startBlock()` method which only increments `_blockCount`, and `_blockCount` value is not used anywhere as far as I can tell.

jgates108: I use block count when debugging and I think it has future value in that it would be a useful measure to provide to an external scheduler/monitor to indicate how busy this system is with large results.

andy-slac: Could you add this as a comment to where you use/set block count?

PR #294, opened 2017-03-03
Added debug messages.

jgates108:


core/modules/qproc/SecondaryIndex.cc
```diff
@@ -129,39 +129,49 @@ class MySqlBackend : public SecondaryIndex::Backend {
...
 
-
         LOGS(_log, LOG_LVL_TRACE, "params: " << util::printable(params));
 
-        std::string const& db = params[0];
-        std::string const& table = params[1];
-        std::string const& key_column = params[2];
+        auto iter = params.begin();
+        std::string const& db = *iter; // params[0]
+        ++iter;
```
iagaponenko: This is not strictly require. But these two lines could be combined into just:

std::string const& db= *(iter++);  // param[0]

core/modules/qproc/SecondaryIndex.cc
```diff
@@ -129,39 +129,49 @@ class MySqlBackend : public SecondaryIndex::Backend {
...
         } else if (query_type == QueryType::BETWEEN) {
             if (params.size() != 5) {
                 throw Bug("Incorrect parameters for bounded secondary index lookup ");
             }
-            char const BETWEEN_LOOKUP_SQL_TEMPLATE[] = "SELECT %s, %s FROM %s WHERE %s BETWEEN %s AND %s";
-            sql = (boost::format(BETWEEN_LOOKUP_SQL_TEMPLATE) % CHUNK_COLUMN
-                                                              % SUB_CHUNK_COLUMN % index_table % key_column % params[3]
-                                                              % params[4]).str();
+            std::string const& par3 = *iter;
+            ++iter;
```
iagaponenko: Same suggestion for moving the iterator increment into the previous line

PR #302, opened 2017-04-04
Add qhttp module (embeddable C++11 HTTP server)

fritzm:


core/modules/qhttp/AjaxEndpoint.h
```diff
@@ -0,0 +1,63 @@
...
+
+class Server;
+
+class AjaxEndpoint
+{
+public:
+
+    using Ptr = std::shared_ptr<AjaxEndpoint>;
+
+    static Ptr add(Server& server, std::string const& path);
```
ktlim: Just wondering: why is this public if `Server` is a friend and this looks like it is intended only to be called by that class (in `addAjaxEndpoint`)?

core/modules/qhttp/README.md
```diff
@@ -0,0 +1,109 @@
+# qhttp
+
+This is a somewhat bare-bones C++ HTTP server, inspired by things like Express (http://expressjs.com/) and 
+Martini (https://github.com/go-martini/martini).
+
+The intended use-case is embedding a light-duty HTTP server within C++ back-end services, to conveniently
+provide utility services (rest APIs, served status pages, small browser applications, etc.) for diagnostics
```
ktlim: "ReST", officially.

core/modules/qhttp/README.md
```diff
@@ -0,0 +1,109 @@
...
+* Express and Martini style "middlewares" were not implemented to keep complexity low.  This could be added 
+at a later time if desired.  For now, perceived-to-be-commonly-used functionalities are wired directly into 
+the server.
+
+#### Features currently supported
+
+* Static content serving out of file system directories, with file-extension-based automatic Content-Type
+detection.
+
+* Installable handers taking Express and Martini style Request and Response objects, and URL path
```
ktlim: "handlers"

core/modules/qhttp/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+* Installable handers taking Express and Martini style Request and Response objects, and URL path 
+specifiers with wildcarding and parameter capture, for conveniently implementing REST services.
+
+* AJAX endpoint helpers: push a JSON string to the server-side endpoint at any time, and all currently 
+pending clients will be updated.
+
+* HTTP 1.1 persistent connections.
+
+* Can piggy-pack on existing single-threaded asio::io_service instance from hosting application if desired.
```
ktlim: "piggy-back"

core/modules/qhttp/Server.cc
```diff
@@ -0,0 +1,198 @@
...
+            }
+        }
+    );
+}
+
+
+void Server::_readRequest(std::shared_ptr<ip::tcp::socket> socket)
+{
+    auto timer = std::make_shared<asio::steady_timer>(_io_service);
+    timer->expires_from_now(std::chrono::seconds(300));
```
ktlim: Hard-coded?  Shouldn't this be at least a `#define`?

core/modules/qhttp/Response.cc
```diff
@@ -0,0 +1,182 @@
...
+
+
+void Response::sendStatus(unsigned int status)
+{
+    this->status = status;
+    std::string statusStr = responseStringsByCode[status];
+    std::ostringstream entStr;
+    entStr << "<html>" << std::endl;
+    entStr << "<head><title>" << status << " " << statusStr << "</title></head>" << std::endl;
+    entStr << "<body bgcolor=\"white\">" << std::endl;
```
ktlim: > The `<body>` `bgcolor` attribute is not supported in HTML5. Use CSS instead.  CSS syntax: `<body style="background-color:#E6E6FA">`

core/modules/qhttp/Server.cc
iagaponenko: 300 seconds is default timeout for HTTP requests in Apache since 1.2, Mozilla Firefox browsers, and many other Web applications.
I would argue this value should be configurable either via the constructor of the class, or via some other mechanism.

core/modules/qhttp/StaticContent.cc
```diff
@@ -0,0 +1,101 @@
...
+    auto absPath = absolute(path);
+
+    // Walk existing part of the path, then call canonical() on it.  This removes any
+    // symlinks, ".", or ".." from existing part, correctly handling the fiddly details
+    // about ".." in the possible presence of symlinks.
+    auto it = absPath.begin();
+    for (; exists(result / *it) && it != absPath.end(); ++it) {
+        result /= *it;
+    }
+    result = canonical(result);
```
andy-slac: Is `canonical` the same thing as `fs::canonical`?

core/modules/qhttp/Response.cc
iagaponenko: There should not be any styling within the tags. All has to be done via CSS stylesheets.
My second comment is that the very first line of each HTML5-complient document must be:
<!DOCTYPE html>
<html lang="en">
...

iagaponenko:
To Andy's question about error handling in ASIO. According to the documentation (http://www.boost.org/doc/libs/1_63_0/doc/html/boost_asio.html) the library provides 2 options for handling errors.
...
6. The I/O object throws an exception of type boost::system::system_error if the operation failed. If the code to initiate the operation had instead been written as:

  boost::system::error_code ec;
  socket.connect(server_endpoint, ec);

then the error_code variable ec would be set to the result of the operation, and no exception would be thrown.

When an asynchronous operation is used, a different sequence of events occurs.

1. Your program initiates the connect operation by calling the I/O object:

  socket.async_connect(server_endpoint, your_completion_handler);

where your_completion_handler is a function or function object with the signature:

  void your_completion_handler(const boost::system::error_code& ec);

core/modules/qhttp/Path.cc
```diff
@@ -0,0 +1,175 @@
...
+    static boost::regex patternRe{
+        // Match escaped characters that would otherwise appear in future matches.
+        // This allows the user to escape special characters that won't transform.
+        "(\\\\.)|"
+        // Match Express-style parameters and un-named parameters with a prefix
+        // and optional suffixes. Matches appear as:
+        // "/:test(\\d+)?" => [ "/", "test", "\d+", -----, "?", --- ]
+        // "/route(\\d+)"  => [ ---, ------, -----, "\d+", ---, --- ]
+        // "/*"            => [ "/", ------, -----, -----, ---, "*" ]
+        "([\\/.])?(?:(?:\\:(\\w+)(?:\\(((?:\\\\.|[^()])+)\\))?|\\(((?:\\\\.|[^()])+)\\))([+*?])?|(\\*))"
```
iagaponenko: Yes, it would be nice to see the raw string and translate it if needed (to make C++ happy) before passing into boost::regex. I also think this method requires an extensive documentation for the regular expression and the way it's used.

fritzm:
Thanks very much for the first pass reviewing, guys -- lots of helpful catches and suggestions!  I agree with almost everything said; a couple require a little thought on my part.  I'll implement all the easy ones and update the PR tomorrow to cut down on the noise.

core/modules/qhttp/Request.cc
```diff
@@ -0,0 +1,83 @@
...
+        while(getline(content, line) && boost::regex_match(line, headerMatch, headerRe)) {
+            header[headerMatch[1]] = headerMatch[2];
+        }
+    }
+}
+
+
+void Request::_parseUri()
+{
+    static boost::regex targetRe{"([^\\?#]*)(?:\\?([^#]*))?"}; // e.g. "path[?query]"
```
fritzm: Yes.

core/modules/qhttp/AjaxEndpoint.h
fritzm: "friend" stuff was a leftover from earlier in development.  Gone now.

core/modules/qhttp/Request.cc
andy-slac: Inside square brackets most special character loose their special meaning and do not need escaping. I'm not sure what escape actually does here, it could be adding a literal backslash to a character class.

iagaponenko: That is not quite correct. The expression would match any character except:

   ^-]\

Backslash is also used to represent special characters:

  \W
  \r
  \n

In the remaining cases the backslash is ignored, and the expression would escape '?'. In this case the back slash is not needed. I think it can safely removed.

iagaponenko: Sorry about:   \W
It shouldn't be used inside the square brackets. I meant to say:  \s

PR #303, opened 2017-04-18
Changed to make queries that use a small number of chunks interactive.

jgates108:


core/modules/wsched/BlendScheduler.cc
```diff
@@ -141,7 +141,13 @@ void BlendScheduler::queCmd(util::Command::Ptr const& cmd) {
     // Check for scan tables
     SchedulerBase::Ptr s{nullptr};
     auto const& scanTables = task->getScanInfo().infoTables;
-    if (scanTables.size() > 0) {
+    bool interactive = task->getScanInteractive();
+    if (scanTables.size() <= 0 || interactive) {
```
andy-slac: Does this give a warning about comparing signed and unsigned? I assume `size()` cannot return negative number, so this should be equivalent to `scanTables.size() == 0` or `scanTables.empty()`, I think latter is preferred and more idiomatic.

core/modules/qproc/QuerySession.h
```diff
@@ -215,6 +217,10 @@ class QuerySession {
     ChunkSpecVector _chunks; ///< Chunk coverage
     std::shared_ptr<QueryPluginPtrVector> _plugins; ///< Analysis plugin chain
 
+    /// Maximum number of chunks in an interactive query. TODO: Maybe put in config file.
```
andy-slac: I think in the long tern the decision should be based on query cost estimate (if we ever get that done) which by itself could be configured from config file. Maybe it's worth to add rudimentary "cost estimate" thing and its configuration (in a new ticket).

core/modules/wsched/BlendScheduler.cc
jgates108: No warning generated. Changed to ==.

core/modules/qproc/QuerySession.h
jgates108: Created DM-10273.

jgates108:
I added scaninteractive, jobid, and queryid to the unit tests.

PR #304, opened 2017-04-19
[DM-10263] Add libcurl to qserv eups table file

fritzm:
This allows us to get the libcurl package tagged and included in
the qserv:dev base containers, so Travis CI can work during
branch development of features using libcurl.


fritzm:
Passed jenkins on Centos6 and Centos7: https://ci.lsst.codes/job/stack-os-matrix/23234/
Travis fail is expected -- catch-22 with qserv:dev container, which this PR should fix

PR #308, opened 2017-04-27
Added instrumentation.

jgates108:


core/modules/wdb/ChunkResource.cc
```diff
@@ -307,96 +311,81 @@ class ChunkEntry {
...
 
-////////////////////////////////////////////////////////////////////////
-// ChunkResourceMgr
-////////////////////////////////////////////////////////////////////////
+ChunkResourceMgr::Map& ChunkResourceMgr::_getMap(std::string const& db) {
+    DbMap::iterator it = _dbMap.find(db);
+    if (it == _dbMap.end()) {
+        DbMap::value_type v(db, Map());
+        _dbMap.insert(v);
+        it = _dbMap.find(db);
```
andy-slac: `insert()` returns iterator (and bool) so you do not need`find()`. 
An the whole method is more or less equivalent to `return _dbMap[db];`

core/modules/wdb/QueryRunner.cc
```diff
@@ -340,23 +342,35 @@ class ChunkResourceRequest {
...
     ChunkResource getResourceFragment(int i) {
         proto::TaskMsg_Fragment const& fragment(_msg.fragment(i));
+        LOGS(_log, LOG_LVL_DEBUG, "fragment i=" << i);
         if (!fragment.has_subchunks()) {
-            StringVector tables;
+            DbTableSet dbTbls;
             for (auto const& scanTbl : _msg.scantable()) {
-                tables.push_back(scanTbl.db() + "." + scanTbl.table());
+                DbTable dbT(scanTbl.db(), scanTbl.table());
+                dbTbls.emplace(dbT);
```
andy-slac: I think you can do `dbTbls.emplace(scanTbl.db(), scanTbl.table())`, saves a line of code and is probably more efficient.

core/modules/wdb/QueryRunner.cc
```diff
@@ -340,23 +342,35 @@ class ChunkResourceRequest {
...
+            return _mgr->acquire(_msg.db(), _msg.chunkid(), dbTbls);
         }
 
         std::string db;
         proto::TaskMsg_Subchunk const& sc = fragment.subchunks();
-        StringVector tables(sc.table().begin(),
-                            sc.table().end());
+        DbTableSet dbTableSet;
+        for (int j=0; j < sc.dbtbl_size(); j++) {
+            dbTableSet.emplace(DbTable(sc.dbtbl(j).db(), sc.dbtbl(j).tbl()));
```
andy-slac: Here too I think you do not need to construct `DbTable`.

core/modules/global/DbTable.h
```diff
@@ -0,0 +1,61 @@
...
+#include <string>
+
+
+namespace lsst {
+namespace qserv {
+
+
+struct DbTable {
+    DbTable(std::string const& db_, std::string const& table_) : db(db_), table(table_) {}
+    DbTable(DbTable const& dbtbl) = default;
```
jgates108: The default operator= and/or constructor are being used. If someone, in the future, comes along and adds some elements to the class that make the defaults inappropriate, I'm hoping this is a good hint that they really need to define them.

core/modules/wdb/ChunkResource.cc
```diff
@@ -257,46 +263,44 @@ class ChunkEntry {
...
         // Release subChunkId for the right table
-        for(ScTableVector::const_iterator i=needed.begin(), e=needed.end();
-            i != e; ++i) {
-            SubChunkMap& scm = _tableMap[i->table];
-            int last = scm[i->subChunkId];
-            scm[i->subChunkId] = last - 1;
+        for(auto const& elem : needed) {
+            SubChunkMap& scm = _tableMap[elem.dbTable];
+            int last = scm[elem.subChunkId];
+            scm[elem.subChunkId] = last - 1;
```
jgates108: That would be better... I'm really hoping that's an artifact or the original code.

core/modules/proto/testProtocol.cc
```diff
@@ -172,7 +173,7 @@ BOOST_AUTO_TEST_CASE(MsgBuffer) {
 BOOST_AUTO_TEST_CASE(ProtoHashDigest) {
     std::unique_ptr<lsst::qserv::proto::TaskMsg> t1(makeTaskMsg());
     std::string hash = hashTaskMsg(*t1);
-    std::string expected = "f3890e99c399ab196c164eccf2e99369";
+    std::string expected = "b3ecc6a56205c90642e61943e44c7f14";
```
jgates108: This is not a good test, in my opinion. The hash changes every time the protocol changes and needs to be replaced. I'm not sure what a better test is. Perhaps check that running the hash twice for t1 produces the same result, and does not equal the hash of a slightly different TaskMsg?

fritzm: ^ that sounds like a better idea to me

iagaponenko: Could you use the 'get' methods of the message class and compare returned values against expected ones?

PR #314, opened 2017-05-15
Implement support for query listing (DM-2840)

andy-slac:
Add special UserQuery subclass which dumps info from QMeta tables. To
simplify implementation this adds two special views to QMeta database
which more or less map to the output of "SHOW PROCESSLIST" or
INFORMATION_SCHEMA.PROCESSLIST pseudo-table with some extensions.
Depending on query new UserQuery class queries one of these views and
sends output to a result table.


core/modules/ccontrol/UserQueryFactory.cc
```diff
@@ -40,18 +40,23 @@
 #include "ccontrol/UserQueryDrop.h"
 #include "ccontrol/UserQueryFlushChunksCache.h"
 #include "ccontrol/UserQueryInvalid.h"
+#include "ccontrol/UserQueryProcessList.h"
```
andy-slac: My smart editor folds copyright comment. I wish we got rid of all of that already.

core/modules/ccontrol/UserQueryProcessList.cc
```diff
@@ -0,0 +1,242 @@
...
+    _query = "SELECT Id, User, Host, db, Command, Time, State, ";
+    _query += full ? "Info, " : "SUBSTRING(Info FROM 1 FOR 100) Info, ";
+    // These are non-standard but they need to be there because they appear in WHERE
+    _query += "CzarId, Submitted, Completed";
+    _query += " FROM ShowProcessList";
+
+    // only show stuff for current czar and not too old
+    _query += " WHERE CzarId = " + std::to_string(qMetaCzarId) + " AND "
+            "(Completed IS NULL OR Completed > NOW() - INTERVAL 3 DAY)";
+
```
andy-slac: Idea was that `SHOW PROCESSLIST` shows you what you can retrieve from this particular czar. To see stuff from all czars you have to use `SELECT ... FROM INFORMATION_SCHEMA.PROCESSLIST`. We may revisit this when we have better idea how we are going to work with multiple czars.

PR #315, opened 2017-05-25
Added a check to cancel user queries with very large results.

jgates108:


core/modules/rproc/InfileMerger.cc
```diff
@@ -251,6 +267,56 @@ bool InfileMerger::_applySqlLocal(std::string const& sql) {
...
+            LOGS(_log, LOG_LVL_ERROR, "InfileMerger error: " << _error.getMsg());
+            return 0;
+        }
+        LOGS(_log, LOG_LVL_DEBUG, _queryIdStr <<"InfileMerger " << (void*) this << " connected to db");
+    }
+    if (not _sqlConn->runQuery(tableSizeSql, results, errObj)) {
+        _error = util::Error(errObj.errNo(), "error getting size sql: " + errObj.printErrMsg(),
+                       util::ErrorCode::MYSQLEXEC);
+        LOGS(_log, LOG_LVL_ERROR, _queryIdStr << "result table size error: " << _error.getMsg());
+        return false;
```
andy-slac: `return 0;`?

core/modules/rproc/InfileMerger.cc
```diff
@@ -251,6 +267,56 @@ bool InfileMerger::_applySqlLocal(std::string const& sql) {
...
+        if (not _sqlConn->connectToDb(errObj)) {
+            _error = util::Error(errObj.errNo(),
+                     _queryIdStr + "Error connecting to db: " + errObj.printErrMsg(),
+                     util::ErrorCode::MYSQLCONNECT);
+            _sqlConn.reset();
+            LOGS(_log, LOG_LVL_ERROR, "InfileMerger error: " << _error.getMsg());
+            return 0;
+        }
+        LOGS(_log, LOG_LVL_DEBUG, _queryIdStr <<"InfileMerger " << (void*) this << " connected to db");
+    }
```
andy-slac: Looks like this whole `if `is copied from `InfileMerger::_applySqlLocal`, maybe move tit into separate method to avoid duplication?

core/modules/rproc/InfileMerger.h
```diff
@@ -156,10 +157,18 @@ class InfileMerger {
...
     mysql::MySqlConnection _mysqlConn;
+
     std::mutex _mysqlMutex;
     lsst::qserv::mysql::LocalInfile::Mgr _infileMgr;
+
+    std::string _queryIdStr{"QI=?"}; ///< Unknown until results start coming back from workers.
+
+    int _sizeCheckRowCount{0}; ///< Number of rows read since last size check.
+    int _checkSizeEveryXRows{1000}; ///< Check the size of the result table after every x number of rows.
+    size_t _maxResultTableSizeMB{5000}; ///< Max result table size.
```
iagaponenko: Just a trivial comment - I've been told that we shouldn't be using this type of initialization for primitive data type :-)

core/modules/rproc/InfileMerger.cc
```diff
@@ -251,6 +267,56 @@ bool InfileMerger::_applySqlLocal(std::string const& sql) {
     return true;
 }
 
+
+size_t InfileMerger::_getResultTableSizeMB() {
+    std::string resultDb = _config.mySqlConfig.dbName;
```
andy-slac: `resultDb` is not used

core/modules/rproc/InfileMerger.cc
```diff
@@ -251,6 +267,56 @@ bool InfileMerger::_applySqlLocal(std::string const& sql) {
...
+        LOGS(_log, LOG_LVL_ERROR, _queryIdStr << "result table size no rows returned " << _mergeTable);
+        return 0;
+    }
+    auto& row = *iter;
+    std::string tbName = row[0].first;
+    std::string tbSize = row[1].first;
+    size_t sz;
+    std::stringstream ss;
+    ss << tbSize;
+    ss >> sz;
```
andy-slac: This is sort of inefficient (streams are very heavy), I'd try `std::stoul()` instead.

core/modules/czar/CzarConfig.cc
```diff
@@ -68,7 +68,9 @@ CzarConfig::CzarConfig(util::ConfigStore const& configStore)
     : _mySqlResultConfig(configStore.get("resultdb.user", "qsmaster"),
             configStore.getRequired("resultdb.passwd"),
             configStore.getRequired("resultdb.host"), configStore.getInt("resultdb.port"),
-            configStore.getRequired("resultdb.unix_socket"), configStore.get("resultdb.db","qservResult")),
+            configStore.getRequired("resultdb.unix_socket"),
+            configStore.get("resultdb.db","qservResult"),
+            configStore.getInt("resultdb.maxtablesize", 5001)),
```
iagaponenko: Is 5001 the default value?
A reason why I'm asking is that there is some other default value of 5000 in another file. Could it this be streamlined a bit via some common macro definition, or some utility class with defaults? Otherwise we'll see a propagation of hard-wired limits for the same parameter throughout the code.

admin/templates/configuration/etc/qserv-czar.cnf
```diff
@@ -23,6 +23,8 @@ unix_socket = {{MYSQLD_SOCK}}
 host =
 user= {{MYSQLD_USER_QSERV}}
 port = 0
+# maximum user query result size in MB
+maxtablesize = 5500
```
iagaponenko: Even though the comment indicates this parameter's units are MB, the usual practice is to either include '_mb' into the name of the parameter:

  maxtablesize_mb = 5500

or to require units to be provided with the numbers:

  maxtablesize = 5500000000  # bytes
  maxtablesize = 5500000  # KB
  maxtablesize = 5500MB
  maxtablesize = 5GB
  ...

My personal choice would be the last one. Unfortunately this will require you to fix some of your code to avoid using "MB" in the names of variables.

core/modules/rproc/InfileMerger.h
```diff
@@ -156,10 +157,18 @@ class InfileMerger {
...
 
     mysql::MySqlConnection _mysqlConn;
+
     std::mutex _mysqlMutex;
     lsst::qserv::mysql::LocalInfile::Mgr _infileMgr;
+
+    std::string _queryIdStr{"QI=?"}; ///< Unknown until results start coming back from workers.
+
+    int _sizeCheckRowCount{0}; ///< Number of rows read since last size check.
+    int _checkSizeEveryXRows{1000}; ///< Check the size of the result table after every x number of rows.
```
andy-slac: I'd probably use much higher number for that. I don't think we need very high precision for cutoff, checking every 100MB is probably OK. With record sizes at roughly 1KB it should be OK to check every 100,000 rows. Would be interesting to see if we can guess result row size, and increase check interval it further for short rows to reduce overhead.

core/modules/mysql/MySqlConfig.h
```diff
@@ -46,34 +46,37 @@ class MySqlConfig {
...
      *
      * @throws std::runtime_error: if checkValid is true and some parameters are invalid
      */
     MySqlConfig(std::string const& username, std::string const& password,
                 std::string const& hostname,
                 unsigned int const port,
                 std::string const& socket,
-                std::string const& db = "");
+                std::string const& db = "",
+                size_t maxresultsize = 0);
```
iagaponenko: This is another default which may be confusing. See the relevant comment on this subject I made earlier. I would strongly suggest encapsulating all default into one utility class/struct and make the rest of the C++ code to pull default from there instead of using hard-wired numbers throughout the code.

core/modules/rproc/InfileMerger.h
andy-slac: Another idea is to count protobuf sizes as a very rough estimate of result size, this is not exact but totally avoids multiple trips to database.

core/modules/czar/CzarConfig.cc
jgates108: The 5001 was arbitrary so I could tell the difference between using this default value and the other and pin down where the configuration value was failing to be passed.  In practice, the difference between 5001 and 5000 is not going to make a significant difference in what happens when the software runs. That being said, I'll probably switch it back to 5000.

core/modules/mysql/MySqlConfig.h
jgates108: I'm not terribly thrilled with this. maxresultsize only means something for resultDb, which is why it defaults to zero here. It might better to create a child configuration class just for the result database.

core/modules/czar/CzarConfig.cc
iagaponenko: Got it. This makes a sense.

core/modules/rproc/InfileMerger.h
jgates108: I thought about that. To get a decent approximation, you have to know the types.  Things that are the same size in text can have very different sizes on disk. This looks like significant work for little gain when we can just ask the db how big the table is. It also means we are being consistent with what we tell the users (i.e the table really was bigger than 5GB).

I think a simple thing that can be done is not make the first check until a few million rows have been read and then start checking every 50000 rows.

core/modules/rproc/InfileMerger.h
jgates108: Why? I know there can issues with curly braces in constructor initialization lists, but being able to set the initial values in one place (instead of having to set it in every constructor) is a definite win.

PR #318, opened 2017-06-15
Do not use libcrypto on macOS

timj:
It is not required. Crypto is part of libSystem.

### DIFF
--- a/core/modules/SConscript
+++ b/core/modules/SConscript
@@ -62,6 +62,7 @@ from collections import defaultdict
 import os
 
 from SCons.Script import *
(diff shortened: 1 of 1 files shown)


fritzm:
Looks good to me

timj:
Is that Travis test good enough or is more testing required? In particular, I think this adds ssl back to `libqserv_common`. I think that might be fine but a test might be good. This PR also depends on lsst/log#28

timj:
I've redone the patch to separate out ssl from crypto.

fritzm:
Thanks, Tim!

PR #320, opened 2017-06-21
Add support for SUBMIT to qserv parser (DM-3245)

andy-slac:
This is the first step in implementing async queries, Qserv now accepts
"SUBMIT SELECT" queries, parses them and passes the async flag to
UserQuerySelect class. The rest of the logic will be implemented in the
next steps, for now SUBMIT SELECT is essentially the same as SELECT
(except that query is marked ASYNC in QMeta).


jgates108:
It looks good to me.

PR #321, opened 2017-06-26
Tickets/dm 10970

jgates108:


core/modules/qdisp/JobQuery.cc
```diff
@@ -82,8 +82,8 @@ bool JobQuery::runJob() {
 
         auto qr = std::make_shared<QueryResource>(shared_from_this());
         std::lock_guard<std::recursive_mutex> lock(_rmutex);
-        if (_jobDescription->getAttemptCount() <= _getMaxAttempts()) {
-            bool okCount = _jobDescription->incrAttemptCount();
+        if (_jobDescription->getAttemptCount() < _getMaxAttempts()) {
```
iagaponenko: The condition check in the if statement relies upon a limit which is disconnected from a test made inside method JobDescription::incrAttemptCountScrubResults() relying upon this macro definition:

    MAX_JOB_ATTEMPTS

core/modules/qdisp/JobQuery.cc
```diff
@@ -102,7 +102,7 @@ bool JobQuery::runJob() {
         // only if the executive has not already been cancelled. The cancellation
         // procedure changes significantly once the executive calls xrootd's Provision().
         bool success = executive->xrdSsiProvision(_queryResourcePtr, qr);
```
iagaponenko: Is this really needed instead of just:

  // The only way this can fail is if the user query is cancelled.
  if (executive->xrdSsiProvision(_queryResourcePtr, qr)) return true;

First of all you have fewer lines of code (more observable code).
Secondly you don't need to repeat the method name within the comment (fewer things to take care of should you be doing code refactoring)

core/modules/rproc/InfileMerger.h
```diff
@@ -92,6 +93,29 @@ class InfileMergerConfig {
     std::shared_ptr<query::SelectStmt> mergeStmt;
 };
 
+
+/// This class is used to remove invalid rows from cancelled job attempts.
+class InvalidJobAttemptMgr {
+public:
+    InvalidJobAttemptMgr() {};
```
andy-slac: Extraneous semicolon

core/modules/rproc/InfileMerger.cc
```diff
@@ -185,19 +196,21 @@ bool InfileMerger::merge(std::shared_ptr<proto::WorkerResponse> response) {
 
     bool ret = false;
     // Add columns to rows in virtFile.
-    int resultJobId = response->result.jobid() * MAX_JOB_ATTEMPTS;
-    int attemptCount = response->result.attemptcount(); //
-    if (attemptCount >= MAX_JOB_ATTEMPTS) {
-        LOGS(_log, LOG_LVL_ERROR, queryIdJobStr << " Canceling query attemptCount too large at " << attemptCount);
-        return false;
+    int resultJobId = makeJobIdAttempt(response->result.jobid(), response->result.attemptcount());
```
iagaponenko: Could the JobID calculation moved into a method inside the result object:

  response->result.jobId();

...unless this is a PROTOBUF object.

core/modules/rproc/testInvalidJobAttemptMgr.cc
```diff
@@ -0,0 +1,166 @@
...
+#include "rproc/InfileMerger.h"
+
+// System Headers
+#include <set>
+
+// LSST headers
+#include "lsst/log/Log.h"
+
+// Boost unit test header
+#define BOOST_TEST_MODULE ProtoRowBuffer_1
```
andy-slac: Change the name to `InvalidJobAttemptMgr_1`?

core/modules/ccontrol/MergingHandler.cc
```diff
@@ -198,6 +198,11 @@ bool MergingHandler::reset() {
     return true;
 }
 
+
+bool MergingHandler::scrubResults(int jobId, int attempt) {
+    return _infileMerger->scrubResults(jobId, attempt);
```
jgates108: I don't think there's any way I can know that at this point. Later on, it checks if the table has been created before trying to drop any rows. It also adds it to the set of jobAttemps that should not be added to the result table. The idea being, if job 20 attempt 0 had to be restarted as job 20 attempt 1, no rows from job 20 attempt 0 should be accepted if a worker comes back from a network outage.

core/modules/qdisp/JobDescription.cc
```diff
@@ -59,7 +61,10 @@ JobDescription::JobDescription(QueryId qId, int jobId, ResourceUnit const& resou
 }
 
 
-bool JobDescription::incrAttemptCount() {
+bool JobDescription::incrAttemptCountScrubResults() {
+    if (_attemptCount >= 0) {
+        _respHandler->scrubResults(_jobId, _attemptCount);
+    }
     ++_attemptCount;
     if (_attemptCount > MAX_JOB_ATTEMPTS) {
```
jgates108: For simplicity, I'm mergin the jobId and attempt number into a single int, so it can be one number in the table. I don't think retrying a job 100 times is reasonable. I suspect 5 or 10 is more likely. Using a value of 100 makes it a lot easier for a human to read the combined id. There should be a comment at the definition of MAX_JOB_ATTEMPTS.

core/modules/ccontrol/MergingHandler.cc
jgates108: I could keep a set of all jobAttempts that had been added to the table, and only try to delete rows for those jobAttempts that had been added.

PR #324, opened 2017-07-06
Tickets/dm 10971

jgates108:


admin/tools/docker/deployment/parallel/run.sh
```diff
@@ -63,13 +63,20 @@ else
     echo "No existing ${CONTAINER_NAME} container on any nodes"
 fi
 
+USE_JEMALLOC='/qserv/stack/stack/current/Linux64/jemalloc/master-g708711ce34/lib/libjemalloc.so'
```
andy-slac: No idea how docker scripts work, but if you managed to do normal EUPS setup then this could probably be changed to `$JEMALLOC_DIR/lib/libjemalloc.so`

core/modules/SConscript
```diff
@@ -199,7 +199,7 @@ if platform_default() == "darwin":
 
 # library used by other shared libs
 shlibs["qserv_common"] = dict(mods="""global memman proto mysql sql util""",
-                              libs="""log protobuf mysqlclient_r """ +
+                              libs="""log protobuf mysqlclient_r  """ +
```
andy-slac: Why extra space?

core/modules/ccontrol/MergingHandler.cc
```diff
@@ -56,6 +56,10 @@ LOG_LOGGER _log = LOG_GET("lsst.qserv.ccontrol.MergingHandler");
 namespace lsst {
 namespace qserv {
 namespace ccontrol {
+
+std::atomic<long long int> MergeBuffer::_totalBytes{0};
```
andy-slac: `std::int64_t` (and include `<csdtin>`)? Or is it 128 bit?

core/modules/ccontrol/MergingHandler.h
```diff
@@ -45,6 +46,39 @@ namespace lsst {
...
+    void setTargetSize(int sz);
+    void resizeToTargetSize();
+    void zero(); ///< Set buffer size and _targetSize to zero, ensure memory is freed.
+
+
+private:
+    void _resize(int sz);
+
+    std::string _id;
+    std::shared_ptr<bufType> _buff;
```
andy-slac: Looking at how it is used I do not this that this buffer is actually shared with anything. You could probably avoid `shared_ptr` overhead by declaring it `unique_ptr` and returning `bufType*`  or `bufType&` from `getBuffer()`?

admin/tools/docker/deployment/parallel/run.sh
jgates108: These scripts run outside of the eups setup environment, which leaves JEMALLOC_DIR undefined.

andy-slac: Would it be better to define LD_PRELOAD in `etc/init.d` scripts then?

jgates108: I think that is a better long term solution. We should talk to Fabrice about it. The trade off here is that this is still experimental and it's very easy to edit the run.sh file and switch between mallocs for testing. If it goes in the init.d script, we have to rebuild the container. There's probably a simple way to get the best of both worlds.

iagaponenko: I think a better long-term solution would be to extend *setup* to allow an option for setting required environment variables. Something along these lines:

setup --enable=jemalloc

core/modules/ccontrol/MergingHandler.h
```diff
@@ -45,6 +46,39 @@ namespace lsst {
 namespace qserv {
 namespace ccontrol {
 
+/// A class to delay creating the buffer until it is requested
+/// by xrootd SSI. The size of the buffer that will be needed is
+/// set using setTargetSize(int sz), and the buffer of that size is
+/// created by calling resizeToTargetSize(). When a buffer is no longer
+/// needed, zero() should be called to free the memory.
+class MergeBuffer {
```
iagaponenko: I have two general comments on this class:
- methods need to be documented
- and LSST C++ Style Guide (see 5.6, 5-27) requires the copy constructor to be explicitly declared.

core/modules/ccontrol/MergingHandler.h
```diff
@@ -45,6 +46,39 @@ namespace lsst {
...
+        _id = "MBI=" + std::to_string(_sequence++);
+        zero();
+    }
+    virtual ~MergeBuffer();
+
+    std::shared_ptr<bufType> getBuffer();
+    size_t getSize();
+    size_t getTargetSize() { return _targetSize; }
+    void setTargetSize(int sz);
+    void resizeToTargetSize();
```
iagaponenko: I don't think this should be the public method because the only use of the method is just before calling *getBuffer()*. You may consider doing an implicit resize as needed at a time when the pointer is requested. Besides, combining these two operations will make the code safer.

core/modules/ccontrol/MergingHandler.h
```diff
@@ -45,6 +46,39 @@ namespace lsst {
...
+    void zero(); ///< Set buffer size and _targetSize to zero, ensure memory is freed.
```
iagaponenko: The method should have a bit more meaningful name like:

 void resetTargetSize()

Or even better, you may simplify things even further by eliminating that method altogether and adding the default value to the following method instead:

  void setTargetSize(int sz=0);

core/modules/ccontrol/MergingHandler.h
jgates108: 5-27 says a default constructor must be provided, and implicit copy constructors should be avoided.

This class should not be copied, so I deleted the copy copy constructor and operator=.

core/modules/ccontrol/MergingHandler.h
jgates108: There's a lot of code that access the vector<char> underlying buffer directly, and any resize could invalidate the buffer. I don't want to risk an accidental resize, so simple getBuffer() never calls resize.

PR #326, opened 2017-07-18
Add support for disconnected queries (DM-11190)

andy-slac:
SUBMIT query now sends query for execution and returns immediately with the result containing just query ID and name of the result table. Retrieving of the actual result of the disconnected query will be implemented later.

Because QMeta database schema needs update for this new feature I have also implemented generic framework for schema migration and added migration of QMeta schema from version 0 to version 1. When deploying this new code on existing installation one will have to run migration script for QMeta.

### DIFF
--- a/core/modules/ccontrol/UserQuery.h
+++ b/core/modules/ccontrol/UserQuery.h
@@ -80,17 +80,31 @@ public:
     // Delegate objects
     virtual std::shared_ptr<qdisp::MessageStore> getMessageStore() = 0;
 
+    /// This method should disappear when we start supporting results
+    /// in locations other than MySQL tables. We'll switch to getResultLocation()
+    /// at that point.
     /// @return Name of the result table for this query, can be empty
-    virtual std::string getResultTableName() = 0;
+    virtual std::string getResultTableName() const { return std::string(); }
+
+    /// Result location could be something like "table:table_name" or
+    /// "file:/path/to/file.csv".
+    /// @return Result location for this query, can be empty
+    virtual std::string getResultLocation() const { return std::string(); }
 
     /// @return ORDER BY part of SELECT statement to be executed by proxy
-    virtual std::string getProxyOrderBy() = 0;
+    virtual std::string getProxyOrderBy() const { return std::string(); }
+
+    /// @return this query's QueryId.
+    virtual QueryId getQueryId() const { return QueryId(0); }
 
     /// @return this query's QueryId string. Many query types do not have valid Id numbers.
     virtual std::string getQueryIdString() const {
         // return a string indicating this query has no QueryId.
         return QueryIdHelper::makeIdStr(0, true);
     }
+
+    /// @return True if query is async query
+    virtual bool isAsync() const { return false; }
 };
 
 }}} // namespace lsst::qserv:ccontrol
--- a/core/modules/ccontrol/UserQuerySelect.cc
+++ b/core/modules/ccontrol/UserQuerySelect.cc
@@ -68,6 +68,9 @@
 #include <chrono>
 #include <memory>
 
+// Third-party headers
+#include <boost/algorithm/string/replace.hpp>
+
 // LSST headers
 #include "lsst/log/Log.h"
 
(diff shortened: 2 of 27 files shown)


core/modules/czar/Czar.cc
```diff
@@ -172,33 +181,37 @@ Czar::submitQuery(std::string const& query,
     std::thread finalThread(finalizer);
     finalThread.detach();
 
-    {
-        std::lock_guard<std::mutex> lock(_mutex);
+    // update/cleanup query map
+    _updateQueryHistory(clientId, threadId, uq);
```
iagaponenko: Was it safe to remove the lock guard here?

core/modules/czar/MessageTable.cc
```diff
@@ -68,6 +70,19 @@ MessageTable::MessageTable(std::string const& tableName,
       _sqlConn(std::make_shared<sql::SqlConnection>(resultConfig)) {
 }
 
+// Create the table, do not lock
+void
+MessageTable::create() {
```
iagaponenko: Apart from minor changes in the logged messages the only difference between MessageTable::create() and MessageTable::lock() is in the query template string. Could this be just one method with the optional parameter "bool lock=true" instead? Otherwise there is a great (may be not so great) deal of code duplication with the introduction of the additional method.

core/modules/ccontrol/UserQuery.h
```diff
@@ -80,17 +80,31 @@ class UserQuery {
...
+    /// in locations other than MySQL tables. We'll switch to getResultLocation()
+    /// at that point.
     /// @return Name of the result table for this query, can be empty
-    virtual std::string getResultTableName() = 0;
+    virtual std::string getResultTableName() const { return std::string(); }
+
+    /// Result location could be something like "table:table_name" or
+    /// "file:/path/to/file.csv".
+    /// @return Result location for this query, can be empty
+    virtual std::string getResultLocation() const { return std::string(); }
```
andy-slac: This is not SELECT-specific, any query that returns data should provide location for its results. SHOW PROCESSLIST is one example of non-SELECT, there may be others in the future. `UserQuery` is only interface that we have today to provide all info to Czar, you cannot remove methods from this interface without breaking czar.

core/modules/ccontrol/UserQuery.h
```diff
@@ -80,17 +80,31 @@ class UserQuery {
...
 
     /// @return ORDER BY part of SELECT statement to be executed by proxy
-    virtual std::string getProxyOrderBy() = 0;
+    virtual std::string getProxyOrderBy() const { return std::string(); }
+
+    /// @return this query's QueryId.
+    virtual QueryId getQueryId() const { return QueryId(0); }
```
andy-slac: Same response - this is the only interface that we have now, no other place to stick this method without redesigning the whole thing.

core/modules/ccontrol/UserQuerySelect.cc
```diff
@@ -379,7 +382,7 @@ void UserQuerySelect::setupChunking() {
 }
 
 // register query in qmeta database
-void UserQuerySelect::qMetaRegister()
+void UserQuerySelect::qMetaRegister(std::string const& resultLocation, std::string const& msgTableName)
```
andy-slac: The factory code is a big mess, I'd like to clean it up at some point, but it needs significant refactoring and I do not want to touch it now. The method cannot be called more than once, but can be called zero times I guess this is the main reason why it is outside of constructor (constructor does not have enough info to decide whether to call it or not).

core/modules/czar/Czar.cc
andy-slac: That code was moved into `_updateQueryHistory()` method.

core/modules/czar/MessageTable.cc
andy-slac: From API point of view these methods should be separate, the client code knows which method to call at compile time, there is no run-time decision to whether lock or not. I don't care too much about code duplication, it's not that terrible, that function only does absolutely minimal job to execute the query, there is no heavy logic there.

PR #327, opened 2017-07-21
Fixes for Python 3 SCons

timj:
Work in progress.

### DIFF
--- a/core/modules/qproc/SConscript
+++ b/core/modules/qproc/SConscript
@@ -11,11 +11,11 @@ standardModule(env, test_libs='log4cxx')
 # ./admin/bin/qserv-admin.py < admin/examples/testMap_generateMap
 def charArrayGenerator(filename):
     bSize = 70
-    with open(filename,"rb") as f:
+    with open(filename, "r") as f:
         while True:
             block = f.read(10)
             if not block: return
-            yield ",".join(map(lambda s: hex(ord(s)), block))
+            yield ",".join(hex(ord(s)) for s in block)
 
 ## Scons wants args=(target, source, env)
 ## target: headerFile
@@ -29,7 +29,7 @@ def buildHeader(target, source, env):
     length = os.stat(blobFile).st_size
     targetFile = open(str(target[0]), "w") # coerce SCons.File to string
     targetFile.write(headerPfx)
-    lines = ",\n".join(map(lambda s:s, charArrayGenerator(blobFile)))
+    lines = ",\n".join(s for s in charArrayGenerator(blobFile))
     targetFile.write(lines + "\n")
     targetFile.write(headerSfx + "\n")
     targetFile.write("size_t %s_length = %d;\n" % (arrayName, length))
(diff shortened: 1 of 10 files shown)


timj:
Whoever is tasked with fixing the admin scripts for Python 3. Many of the fixes here are also relevant for the admin directory.

* Use `.decode()` after calling `check_output()`.
* The `string` fix.

timj:
These fixes seem to allow me to build Qserv using SCons 3. The fixes should be safe with SCons 2. I'm not entirely sure why the PR is failing Travis tests.

This PR does not attempt to fix the admin scripts to work with Python 3. These need to be fixed by the end of the year at the latest. It doesn't seem like it will be a lot of work to port them but since I have no idea how to run the admin scripts I am not going to attempt it.

timj:
I should add that I've only tested this in macOS. Builds fine and fails in the http test in the expected manner -  @fritzm knows about this one.

```
Running 7 test cases...
core/modules/qhttp/testqhttp.cc:382: error: in "request_timeout": check ec == boost::asio::error::eof has failed [system:54 != 2]

*** 1 failure is detected in the test module "qhttp"
```

core/modules/qproc/SConscript
```diff
@@ -6,38 +6,42 @@ import os
...
 standardModule(env, test_libs='log4cxx')
 
+
 # build kvmap.h
 # If you need to rebuild the map, use:
 # ./admin/bin/qserv-admin.py < admin/examples/testMap_generateMap
 def charArrayGenerator(filename):
     bSize = 70
-    with open(filename,"rb") as f:
+    with open(filename, "r") as f:
```
andy-slac: I'm not sure this is the right thing to do, I think the intent here is to make the stream of bytes rather than chars, it gets translated into C++ char[]. And it probably does not matter in practice as we don't have Unicode in input file.

timj: I pondered this. The problem is that python 3 `ord()` takes characters and not bytes, implying that you are actually wanting to convert characters to their hex equivalents. I think this is because for bytes, the iterator returns the integer as it goes. So if you keep byte mode you need to change the `hex(ord(x))` to straight `hex(x)`.

I can change it to do that if you prefer but I don't think that works on python2. The current code works on both (Qserv can drop python3 as soon as you get the admin scripts converted).

andy-slac: Ah, OK, I need to learn more about `bytes` 👶 Let's keep your changes now but maybe you can add TODO comment saying that this may break in presence of UNICODE in input file?

timj: I've added a comment. Also removed that `bSize` variable that was not used anywhere.

timj:
What's the Qserv merge policy? Do I go ahead or are tests needed (Travis randomly passed -- seems flaky; and these changes are just to the build system).

andy-slac:
@fritzm should know more about policies, but I think this should be OK to merge as long as it builds and all normal unit tests pass OK (this is a part of the build).

PR #329, opened 2017-08-01
Implement returning results of disconnected queries/jobs (DM-11191)

andy-slac:
New UserQuery subclass (UserQueryAsyncResult) handles most of the job.
It finds necessary info in QMeta and when query is finished it pretends
that its own result table is the result table of original query. Message
table of original query needs to be copied to a new message table.
Presently result can be returned only once (proxy drops result table),
in the future we'll need better lifetime management for results.

### DIFF
--- a/core/modules/ccontrol/UserQueryAsyncResult.cc
+++ b/core/modules/ccontrol/UserQueryAsyncResult.cc
@@ -1,0 +1,214 @@
+/*
+ * LSST Data Management System
+ * Copyright 2017 AURA/LSST.
+ *
+ * This product includes software developed by the
+ * LSST Project (http://www.lsst.org/).
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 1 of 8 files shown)


core/modules/ccontrol/UserQueryAsyncResult.cc
```diff
@@ -0,0 +1,214 @@
...
+    LOGS(_log, LOG_LVL_DEBUG, "UserQueryAsyncResult: QID=" << queryId);
+
+    // get query info from QMeta
+    try {
+        _qInfo = qMeta->getQueryInfo(queryId);
+        LOGS(_log, LOG_LVL_DEBUG, "found QMeta record: czar=" << _qInfo.czarId()
+             << " status=" << _qInfo.queryStatus() << " resultLoc=" << _qInfo.resultLocation()
+             << " msgTableName=" << _qInfo.msgTableName());
+    } catch (qmeta::QueryIdError const& exc) {
+        std::string message = "No job found for ID=" + std::to_string(queryId);
```
andy-slac: This message is what users will see, I want to keep that consistent with "SHOW PROCESSLIST" column name. If you prefer I can write "QID" to the log, but that probably does not even matter — if ID is not found then it never existed and there should not be any QID in the log with the same number.

PR #330, opened 2017-08-02
Tickets/dm 11172

jgates108:


core/modules/qdisp/JobQuery.cc
```diff
@@ -136,12 +138,19 @@ bool JobQuery::cancel() {
         std::lock_guard<std::recursive_mutex> lock(_rmutex);
         // If _queryRequestPtr is not nullptr, then this job has been passed to xrootd and
         // cancellation is complicated.
+        bool cancelled = false;
         if (_queryRequestPtr != nullptr) {
             LOGS_DEBUG(getIdStr() << " cancel QueryRequest in progress");
-            _queryRequestPtr->cancel();
-        } else {
+            if (_queryRequestPtr->cancel()) {
+                LOGS_DEBUG(getIdStr() << " cancelled by QueryRequest");
```
andy-slac: Why is it `_idStr` in some places and `getIdStr()` in other?
And there s also `LOGS_DEBUG()` vs `LOGS(_log, LOG_LVL_DEBUG, ...)`.

core/modules/qdisp/QueryRequest.h
```diff
@@ -70,6 +70,29 @@ class RequestError : public std::exception {
...
+        _largeResultMgr(largeResultMgr), _jobIdStr(jobId) {}
+    ~LargeResultSafety();
+
+    void startBlock();
+    bool finishBlock();
+
+private:
+    qdisp::LargeResultMgr::Ptr _largeResultMgr;
+    bool _startBlockCalled{false}; ///< True if _largeResultMgr->startBlock() called.
+    std::mutex _blockMtx; ///< Protects _largeStartBlockCalled.
```
andy-slac: `_largeStartBlockCalled` -> `_startBlockCalled`?

core/modules/rproc/InfileMerger.cc
```diff
@@ -284,7 +286,11 @@ bool InfileMerger::finalize() {
         std::string sqlDropCol = std::string("ALTER TABLE ") + _mergeTable
                                + " DROP COLUMN " +  _jobIdColName;
         LOGS(_log, LOG_LVL_DEBUG, "Removing w/" << sqlDropCol);
+        auto begin = std::chrono::system_clock::now();
         finalizeOk = _applySqlLocal(sqlDropCol);
+        auto end = std::chrono::system_clock::now();
+        auto diff = std::chrono::duration_cast<std::chrono::microseconds>(end - begin);
```
andy-slac: Don't we have special class for that? `util/Timer` probably.

core/modules/qdisp/JobQuery.cc
jgates108: Consistency is good, do you have a preference?

andy-slac: Re LOGS vs LOGS_DEBUG - I think our preferred way is to use specific logger instead of dumping it all to a root logger, so `LOGS(_log, ...)` wins here. No opinion on `_idStr` vs `getIdStr()`.

core/modules/qdisp/QueryRequest.h
```diff
@@ -70,6 +70,29 @@ class RequestError : public std::exception {
     std::string s;
 };
 
+
+/// It is extremely important that _largeResultMgr->finishBlock() is called exactly once
+/// for every time _largeResultMgr->startBlock() is called, otherwise the semaphore would
+/// vary in value, disastrous if it gets used up or grows large. This class is meant to
+/// ensure that it doesn't happen.
+class LargeResultSafety {
```
iagaponenko: Could you please ad the copy constructor and assignment operators?

core/modules/rproc/InfileMerger.cc
iagaponenko: I've noticed that the performance of some calls made to _applySqlLocal(*) within the compilation unit are measured in this one, meanwhile some others aren't. Wouldn't it make a sense to put this performance measurement (And diagnostic printout) into the method itself to avoid polluting surrounding code?

jgates108: Good point.

PR #332, opened 2017-08-05
Fix flakiness in qhttp timeout test

fritzm:


timj:
Builds for me on my iMac now. Thanks.

fritzm:
I think the Mac fix is good, but it looks like it still failed on one of the Jenkins builds... :-/

fritzm:
Merged

PR #336, opened 2017-08-18
Migrate Qserv to Python3 (DM-11550)

andy-slac:
Standard Python3 migration stuff:

- run futurize -1
- run autopep8
- run futurize -2
- few manual fixes/improvements to bring in line with LSST standards

Integration tests run successfully with both python2/3 now.


admin/tools/provision/cloudmanager.py
```diff
@@ -299,7 +305,8 @@ def _build_instance_name(self, instance_id):
         :param instance_id: instance id
         :return:            instance name
         """
-        if isinstance(instance_id, unicode):
+        # TODO: not sure if the logic here is doing what it is supposed to do
+        if isinstance(instance_id, basestring):
```
timj: I think you want to say if it's *not* `bytes` then encode it to `bytes`.

andy-slac:
@timj, I forgot `python_future`, it is probably there already due to indirect dependencies but I'll fix it, thanks!

admin/tools/provision/cloudmanager.py
andy-slac: To me it seems that real intent here is to remove "bad" characters from `instance_id`, but I'm not sure that conversion to bytes is wanted. @fjammes should probably comment on this, I'll ask him to review this code.

andy-slac: To me it seems that real intent here is to remove "bad" characters from `instance_id`, but I'm not sure that conversion to bytes is wanted. @fjammes should probably comment on this, I'll ask him to review this code.

timj: ok. If the intent is to remove non-ascii characters then an `encode().decode()` should be enough without the test of "is a string" (you know it's a string).

fjammes: The goal here is to be sure that instance_id is pure ascii.

timj: I don't think this is right then. This will trigger for all strings and will encode them in bytes, the instance name below will then include a `b''` in the output because that's what happens when you format a byte array into a string. To strip non-ASCII from a string and still have a string you need to do:
```python
instance_id = instance_id.encode('ascii', 'ignore').decode('ascii')
```
The `if` statement can be removed since it seems that bytes are not expected by this routine and there is no harm in always doing the encode/decode on python2. See below to show this working:
```
>>> x = 'a\xdfbc'
>>> x
'aßbc'
>>> x.encode('ascii','ignore').decode('ascii')
'abc'
```

andy-slac: My biggest problem with all of that is that we probably don't even want it to 'ascii', I think here we need character set that is usable in instance name, and I don't think that all ASCII codes are acceptable there (`\n\r\t\v\a` are all ASCII characters but you probably do not want them in a name).

I would suggest instead of checking for ASCII define a set of valid characters and check against that, something like:

    # strip invalid characters from instance ID
    instance_id = re.sub("[^-_.a-zA-Z0-9]", "", instance_id)

But I'm not very comfortable even with that, I think this code does not even have to exist, we should require valid instance_id and let nova call fail if client fails to provide a valid name. Trying to cleanup someone else's mess will just create more problems - like what happen if after cleanup names start to collide with each other.

So my new proposal is to remove this check completely and let it fail if something wrong is passed as an argument. @fjammes, do you agree with that?

andy-slac: With no comment from Fabrice, I'm going to implement `re`-based filtering and let Fabrice decide later what he wants to do with that.

PR #337, opened 2017-08-19
Fix flaky qhttp ajax unit test (DM-11615)

fritzm:
It seems that C++11 std::thread spawn and sleep times vary widely
under load on some of the Jenkins build slaves.  This change adds
some synchronization between event generation and handler threads
within the qhttp ajax unit test in order to remove sensitivity to
event thread spawn/sleep times.

### DIFF
--- a/core/modules/qhttp/testqhttp.cc
+++ b/core/modules/qhttp/testqhttp.cc
@@ -623,48 +623,64 @@ BOOST_FIXTURE_TEST_CASE(ajax, QhttpFixture)
     BOOST_TEST(n3 == 0);
 
     //
-    //----- Start a thread that will push an update to the first ajax endpoint every 100ms, then run the
-    //      libcurl multi in this thread for at most 225ms.  When we get control back, check via counters
+    //----- Start a thread that will push an two updates to the first ajax endpoint, separated by 100ms.
+    //      Run the libcurl multi in this thread for at least 25ms after the last update.  Check via counters
     //      that both installed handlers for the first endpoint have run twice, and that the handler for
     //      the second endpoint has not been run erroneously.
     //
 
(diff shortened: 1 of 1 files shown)


core/modules/qhttp/testqhttp.cc
```diff
@@ -623,48 +623,64 @@ BOOST_FIXTURE_TEST_CASE(ajax, QhttpFixture)
...
+    std::thread t2([&ajax1, &done2]() {
+        std::this_thread::sleep_for(std::chrono::milliseconds{100});
+        ajax1->update("1");
+        std::this_thread::sleep_for(std::chrono::milliseconds{100});
+        ajax1->update("1");
+        done2 = true;
     });
 
-    m.perform(225);
+    std::atomic<bool> done3{false};
```
fritzm: Hell if I know :-)  Now I don't know why I used `atomic_bool`, and will go look up the differences (if any) between them and choose the best one consistently

fritzm:
>If I understand the code there is still some time sensitivity in that it expects that everything will be finished under 25ms.

Yes, max 25ms lag between the fine update of AJAX endpoints on the server and completion of the processing of the corresponding HTTP responses on the client.  At least the timeout is one-sided on the client side now.  I thought about fully synchronizing the test, but in the end a client timeout is required anyway just to make sure the unit test doesn't hang forever.

core/modules/qhttp/testqhttp.cc
fritzm: Update: a little reading seems to indicate that `atomic_bool` was added for forward compatibility with an upcoming C standards proposal.  `atomic<bool>` is apparently the more C++-ish thing, so I'll go with that consistently.

PR #338, opened 2017-08-24
Tickets/dm 11524

jgates108:


core/modules/rproc/InfileMerger.cc
```diff
@@ -298,16 +311,32 @@ bool InfileMerger::isFinished() const {
...
-    std::string sqlDelRows = std::string("DELETE FROM ") + _mergeTable
-                             + " WHERE " +  _jobIdColName + "=" + std::to_string(jobIdAttempt);
-    LOGS(_log, LOG_LVL_DEBUG, "Deleting invalid rows w/" << sqlDelRows);
-    bool ok = _applySqlLocal(sqlDelRows, "deleteInvalidRows");
-    if (!ok) {
-        LOGS(_log, LOG_LVL_ERROR, "Failed to drop columns w/" << sqlDelRows);
-        return false;
+bool InfileMerger::_deleteInvalidRows(std::set<int> const& jobIdAttempts) {
+    // delete several rows at a time
+    unsigned int maxSize = 950000; /// default 1mb limit
```
iagaponenko: Why is this magic limit (effectively for a total number of invalid jobs to be deleted at once) here? This is perhaps an attempt to deal with some "future" problem which will never occur. From my point of view it just reduces the overall clarity of the code. Would it be possible to refactor the code to set a limit on a number of jobs to be removed with a single DELETE statement and then add another loop to iterate over all jobs while it's all done.

And if there is a concern to run into the "run-away" scenario then reporting each iteration into the log stream may help to detect & debug the problem.

core/modules/rproc/InfileMerger.h
```diff
@@ -107,7 +107,7 @@ class InfileMergerConfig {
 class InvalidJobAttemptMgr {
 public:
     InvalidJobAttemptMgr() {}
-    void setDeleteFunc(std::function<bool(int)> func) {_deleteFunc = func; }
+    void setDeleteFunc(std::function<bool(std::set<int> const& jobAttempts)> func) {_deleteFunc = func; }
```
iagaponenko: Since this definition (of a function) appears on code at least 3 times (the first time is here, the second time is in a definition of the data member, and the third time is in the .CC file) then I recommend introducing a typedef:

  typedef std::function<bool(std::set\<int\> const&)> deleteFuncType;

This will make the code a bit easier to read.

core/modules/rproc/InfileMerger.cc
```diff
@@ -214,19 +215,27 @@ bool InfileMerger::merge(std::shared_ptr<proto::WorkerResponse> response) {
...
+                os << queryIdJobStr << " cancelling queryResult table " << _mergeTable;
+                if (invalidResult) {
+                    os << " failed to delete invalid rows.";
+                } else {
+                    os << " too large at " << tSize << "MB max allowed=" << _maxResultTableSizeMB;
+                }
+                LOGS(_log, LOG_LVL_ERROR, os.str());
+                _error = util::Error(-1, os.str(), -1);
+                return false;
+            }
```
andy-slac: I'm thinking that we can probably avoid this entirely and only do cleanup when we return result. Presumably disk size is (almost) infinite or at least is much larger than a limit on a single result size. If we need to worry about free space on disk (and I think we will need to worry about that at some point when we implement result caching) we'll need a separate mechanism to keep result sizes under control, but it will probably take into account sizes of all results in cache.

core/modules/rproc/InfileMerger.cc
```diff
@@ -298,16 +311,32 @@ bool InfileMerger::isFinished() const {
...
+                invalidStr += ",";
+            }
+            invalidStr += std::to_string(*iter);
+            ++iter;
+        }
+        std::string sqlDelRows = std::string("DELETE FROM ") + _mergeTable
+                + " WHERE " + _jobIdColName + " IN (" + invalidStr + ")";
+        bool ok = _applySqlLocal(sqlDelRows, "deleteInvalidRows");
+        if (!ok) {
+            LOGS(_log, LOG_LVL_ERROR, "Failed to drop columns w/" << sqlDelRows);
```
andy-slac: columns?

core/modules/rproc/InfileMerger.cc
```diff
@@ -298,16 +311,32 @@ bool InfileMerger::isFinished() const {
...
+    auto end = jobIdAttempts.end();
+    while (iter != end) {
+        bool first = true;
+        std::string invalidStr;
+        while (invalidStr.size() < maxSize && iter != end) {
+            if (!first) {
+                invalidStr += ",";
+            }
+            invalidStr += std::to_string(*iter);
+            ++iter;
```
andy-slac: `first = false` ?

core/modules/rproc/InfileMerger.cc
andy-slac: OK, thinking again with clear head after morning coffee - I think we really should get rid of cleanup at this point. Even if you remove rows from cancelled chunk you will have to re-run query on the same chunk anyways, and that will bring the same or larger number of rows again. I think it's safe to say that if partial result is larger than limit then final result will be larger still and just stop here, no cleanup needed.

jgates108: It's very possible that chunks in question have several invalid rows from failed attempts and all the rows from a complete query in the result table. The query doesn't need to be run again, as all the results are already there. So that would reduce the size of the result table, possibly significantly when hundreds of chunks are involved. In large results, it is easy for hundreds of chunks to be affected even if only one worker goes down.

The delete SQL query is only run when there are elements in _invalidJAWithRows, and that is cleared after the delete query since all the invalid rows should be gone.  So, if the delete query runs here, it wont run again at the end unless there are more failures.

Last, this really shouldn't run very often. It is only run if the result table is too large and the delete query will only run if it knows that there are invalid rows in the table.

core/modules/rproc/InfileMerger.cc
jgates108: doh

PR #341, opened 2017-09-13
Fix proxy blocking during query killing (DM-11655)

andy-slac:
Run query killing method in a separate detached thread and return control
to proxy immediately to avoid proxy blocking in single-threaded LUA
code.
    
Additionally fixed race condition between query killing and normal query
completion which accessed shared data without synchronization.

### DIFF
--- a/core/modules/ccontrol/UserQuerySelect.cc
+++ b/core/modules/ccontrol/UserQuerySelect.cc
@@ -169,7 +169,12 @@ void UserQuerySelect::kill() {
     if (!_killed) {
         _killed = true;
         try {
-            _executive->squash();
+            // make a copy of executive pointer to keep it alive and avoid race
+            // with pointer being reset in discard() method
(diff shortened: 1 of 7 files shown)


core/modules/ccontrol/UserQuerySelect.cc
```diff
@@ -169,7 +169,12 @@ void UserQuerySelect::kill() {
     if (!_killed) {
         _killed = true;
         try {
-            _executive->squash();
+            // make a copy of executive pointer to keep it alive and avoid race
+            // with pointer being reset in discard() method
+            std::shared_ptr<qdisp::Executive> exec = _executive;
```
jgates108: Could use auto here.

andy-slac: I wanted to make it very explicit that a copy of a shared pointer is needed here :)

PR #342, opened 2017-09-18
Use host data directory on master

fjammes:


admin/tools/docker/deployment/kubernetes/env.example.sh
```diff
@@ -5,6 +5,7 @@
 # VERSION can be a git ticket branch but with _ instead of /
 # example: tickets_DM-7139, or dev
 VERSION=dev
+MARIADB_VERSION=10.1.25
```
iagaponenko: What's the planned use for this variable, and why should we depend in these scripts on a very specific version of the database server? I think this info could be obtained from the LSST Stack.

admin/tools/docker/deployment/kubernetes/admin/remove-dir.sh
```diff
@@ -1,21 +1,26 @@
...
-# Remove Qserv data from host nodes
+# Empty given directory on host nodes
 
 # @author Fabrice Jammes SLAC/IN2P3
 
 set -e
 set -x
 
+# Directory to empty
+DATA_DIR="/qserv/log"
```
iagaponenko: Is this not a mistake? The variable is pointing to a folder where the log files are supposed too be placed.

admin/tools/docker/deployment/kubernetes/env.example.sh
fjammes: The final goal is to remove dependency between qserv and mariadb in eups, so that qserv build do no more require mysql build and install, but only mysql client library. It will produce a lighter container for Qserv.
That's why I now run mariadb in a separate container, I currently use official mariadb container, but we could decide to use a eups build of mariadb in the middle term to unify version management. You're perfectly right, we have to discuss that.

admin/tools/docker/deployment/kubernetes/admin/remove-dir.sh
fjammes: 👍 , I'll add a parameter to the script. I use it to remove data or logs files, mostly for running integration tests on openstack ...

PR #344, opened 2017-09-22
Add support for query killing by query ID (DM-11729)

andy-slac:
Implemented new SQL command "CANCEL NNN" which takes query Id and tries to kill running query with given ID.

### DIFF
--- a/core/modules/czar/Czar.cc
+++ b/core/modules/czar/Czar.cc
@@ -58,8 +58,10 @@ std::string const createAsyncResultTmpl("CREATE TABLE IF NOT EXISTS %1% "
 
 LOG_LOGGER _log = LOG_GET("lsst.qserv.czar.Czar");
 
-// parse KILL query, return thread ID or -1
-int parseKillQuery(std::string const& query);
+// Parse KILL/CANCEL query.
+// Returns false on parsing failure. On success for KILL queries threadId will
+// be non-zero, for CANCEL query queryId will be non-zero.
+bool parseKillQuery(std::string const& aQuery, int& threadId, lsst::qserv::QueryId& queryId);
 
 } // anonymous namespace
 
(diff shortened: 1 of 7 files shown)


core/modules/czar/Czar.cc
```diff
@@ -229,58 +227,102 @@ Czar::killQuery(std::string const& query, std::string const& clientId) {
...
+                          ccontrol::UserQuery::Ptr const& uq) {
+
+    std::lock_guard<std::mutex> lock(_mutex);
+
+    // first cleanup client query maps from completed queries
+    _cleanupQueryHistoryLocked();
 
     // remember query (weak pointer) in case we want to kill query
+    if (uq->getQueryId() != QueryId(0)) {
+        _idToQuery.insert(std::make_pair(uq->getQueryId(), uq));
```
jgates108: Would it be better to make the weak pointer for uq before the make_pair?

core/modules/czar/Czar.cc
```diff
@@ -229,58 +227,102 @@ Czar::killQuery(std::string const& query, std::string const& clientId) {
...
+        throw std::runtime_error("Query has already finished: " + query);
     }
 }
 
 void
-Czar::_updateQueryHistory(std::string const& clientId,
-                          int threadId,
-                          ccontrol::UserQuery::Ptr const& uq) {
+Czar::_cleanupQueryHistoryLocked() {
+    // mutex must be locked
```
jgates108: _mutex would be clearer

core/modules/czar/Czar.cc
andy-slac: In theory it does not matter, in practice I'd prefer to write Python-like:

    _idToQuery[uq->getQueryId()] = uq;

and not to worry about nonsensical conversions (but it's less efficient in C++). I think there is a reason why weak_ptr conversion constructor is implicit so it should be OK to use with this sort of pattern.

PR #351, opened 2017-10-24
Update newinstall URL 13.0->14.0

fritzm:


fritzm:
So imma go ahead and merge this in hopes of fixing our container and travis builds... it's not going to make anything *worse*, anyway...

PR #355, opened 2017-11-21
Tickets/dm 10515

n8pease:
Work on this ticket is not complete but has reached a point where it would be good to get a checkin from developers more experienced with the qserv code. I'll put more comments in the ticket about the current state of things.

### DIFF
--- a/core/modules/qana/QueryMapping.cc
+++ b/core/modules/qana/QueryMapping.cc
@@ -117,12 +117,6 @@ public:
 
         for(auto i=_tupleDeque.begin(); i != _tupleDeque.end(); ++i) {
             newE->s = replace(newE->s, i->pat, i->tgt);
-            if (i->param == QueryMapping::SUBCHUNK) {
-                // Remember that we mapped a subchunk,
-                    // so we know to iterate over subchunks.
-                    // Or... the plugins could signal that subchunks
-                    // are needed somehow. FIXME.
-            }
         }
         return newE;
     }
--- a/core/modules/qana/QueryMapping.cc
+++ b/core/modules/qana/QueryMapping.cc
@@ -21,15 +21,6 @@
  * see <http://www.lsstcorp.org/LegalNotices/>.
  */
 
-/**
-  * @file
-  *
-  * @brief Implementation of QueryMapping. Local implementations of:
-  * class MapTuple
-  * class Mapping : public QueryTemplate::EntryMapping
-  *
-  * @author Daniel L. Wang, SLAC
-  */
 
 // Class header
 #include "qana/QueryMapping.h"
@@ -61,16 +52,6 @@ namespace lsst {
 namespace qserv {
 namespace qana {
 
-class MapTuple {
-public:
-    MapTuple(std::string const& pattern,
-             std::string const& target,
-             QueryMapping::Parameter p)
-        : pat(pattern), tgt(target), param(p) {}
-    std::string pat;
-    std::string tgt;
-    QueryMapping::Parameter param;
-};
 
 std::string const replace(std::string const& s, std::string const& pat, std::string const& value) {
     std::string result;
@@ -90,8 +71,6 @@ std::string const replace(std::string const& s, std::string const& pat, std::str
 
 class Mapping : public query::QueryTemplate::EntryMapping {
 public:
-    typedef std::deque<int> IntDeque;
-    typedef std::deque<MapTuple> TupleDeque;
 
     Mapping(QueryMapping::ParameterMap const& m, qproc::ChunkSpec const& s)
         : _subChunks(s.subChunks.begin(), s.subChunks.end()) {
@@ -115,8 +94,8 @@ public:
         // FIXME see if this works
         //if (!e.isDynamic()) {return newE; }
 
-        for(auto i=_tupleDeque.begin(); i != _tupleDeque.end(); ++i) {
-            newE->s = replace(newE->s, i->pat, i->tgt);
+        for(auto i=_replaceItems.begin(); i != _replaceItems.end(); ++i) {
+            newE->s = replace(newE->s, i->pattern, i->target);
         }
         return newE;
     }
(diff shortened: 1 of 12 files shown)


core/modules/qproc/QuerySession.cc
```diff
@@ -379,20 +379,9 @@ std::vector<std::string> QuerySession::_buildChunkQueries(query::QueryTemplate::
...
-        }
-    } else { // subchunked:
-        ChunkSpecSingle::Vector sVector = ChunkSpecSingle::makeVector(chunkSpec);
-        for(auto& chunkStr : sVector) {
-            for(auto& qTemplate : queryTemplates) {
-                std::string str = _context->queryMapping->apply(chunkStr, qTemplate);
-                LOGS(_log, LOG_LVL_DEBUG, "adding query " << str);
-                chunkQueries.push_back(str);
-            }
-        }
```
n8pease: since we don't create subchunk query strings in the master anymore, the `else` part is no longer necessary.

core/modules/qana/QueryMapping.cc
```diff
@@ -116,6 +116,10 @@ class Mapping : public query::QueryTemplate::EntryMapping {
         //if (!e.isDynamic()) {return newE; }
 
         for(auto i=_tupleDeque.begin(); i != _tupleDeque.end(); ++i) {
+            if (i->param == QueryMapping::SUBCHUNK) {
+                // don't do subchunk replacement here; the worker will take care of it.
+                continue;
+            }
             newE->s = replace(newE->s, i->pat, i->tgt);
```
n8pease: Checking for the subchunk identifier string here works, but IMO this is the main area that still needs more work: I think the more correct fix is to remove the subchunk from `_tupleDeck` entirely. I need to do a couple days' work to understand the classes involved including `QuerySession`, `ChunkSpec`, `ChunkSpecFragmenter`, and design the subchunk string handling out of those classes.

core/modules/qproc/QuerySession.cc
```diff
@@ -377,7 +377,6 @@ std::vector<std::string> QuerySession::_buildChunkQueries(query::QueryTemplate::
     if (!_context->queryMapping) {
         throw QueryProcessingBug("Missing QueryMapping in _context");
     }
-    qana::QueryMapping const& queryMapping = *_context->queryMapping;
```
n8pease: this line was always there, but was unused.

core/modules/wdb/QueryRunner.cc
```diff
@@ -398,11 +400,31 @@ bool QueryRunner::_dispatchChannel() {
...
+                    for (auto subchunkIdItr = fragment.subchunks().id().begin();
+                            subchunkIdItr != fragment.subchunks().id().end();
+                            ++subchunkIdItr) {
+                        std::string queryStr = fragment.query(qi);
+                        for (auto loc = queryStr.find(SUBCHUNK_TAG);
+                                loc != std::string::npos;
+                                loc = queryStr.find(SUBCHUNK_TAG)) {
+                            queryStr.replace(loc, SUBCHUNK_TAG.length(),
+                                             std::to_string(*subchunkIdItr));
+                        }
```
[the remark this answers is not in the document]
jgates108: I think there's some desire to keep boost use to a minimum, but wrapping this up in a function or using boost would make it easier to read.

core/modules/wdb/QueryRunner.cc
```diff
@@ -398,11 +400,31 @@ bool QueryRunner::_dispatchChannel() {
...
+                    queries.push_back(fragment.query(qi));
+                }
+            }
+
             ChunkResource cr(req.getResourceFragment(i));
             // Use query fragment as-is, funnel results.
-            for(int qi=0, qe=fragment.query_size(); qi != qe; ++qi) {
-                LOGS(_log, LOG_LVL_DEBUG, "running fragment=" << fragment.query(qi));
-                MYSQL_RES* res = _primeResult(fragment.query(qi)); // This runs the SQL query.
+            for (const auto& query : queries) {
```
jgates108: Normally we use auto const&.

jgates108:
It looks good and it's a reasonable place to make the substitutions. Andy has a point about the substitution strings. There should probably be some thought about replacing both of them with versions that are safe.

core/modules/qproc/QuerySession.cc
```diff
@@ -377,22 +377,10 @@ std::vector<std::string> QuerySession::_buildChunkQueries(query::QueryTemplate::
...
-        for(auto& chunkStr : sVector) {
-            for(auto& qTemplate : queryTemplates) {
-                std::string str = _context->queryMapping->apply(chunkStr, qTemplate);
-                LOGS(_log, LOG_LVL_DEBUG, "adding query " << str);
-                chunkQueries.push_back(str);
-            }
-        }
+    for(auto tupleIter=queryTemplates.begin(), e=queryTemplates.end(); tupleIter != e; ++tupleIter) {
+        std::string str = _context->queryMapping->apply(chunkSpec, *tupleIter);
+        chunkQueries.push_back(str);
```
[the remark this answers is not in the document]
andy-slac: I'm not sure about that. My hunch is that it should not be allowed to do that sort of optimizations but I'm not an rvalue guru.

core/modules/wdb/QueryRunner.cc
```diff
@@ -398,11 +401,23 @@ bool QueryRunner::_dispatchChannel() {
                 break;
             }
             proto::TaskMsg_Fragment const& fragment(m.fragment(i));
+            std::vector<std::string> queries;
+            for (const std::string queryStr: fragment.query()) {
+                if (fragment.has_subchunks() && false == fragment.subchunks().id().empty()) {
+                    for (auto subchunkId : fragment.subchunks().id()) {
+                        std::string s(queryStr);
+                        boost::algorithm::replace_all(s, SUBCHUNK_TAG, std::to_string(subchunkId));
```
jgates108: This looks like it does the same thing as replace(std::string const& s, std::string const& pat, std::string const& value) in qana/QueryMapping.cc. We should probably use the same function in both places.

PR #357, opened 2017-11-28
Use per-table ovelaps in RelationGraph (DM-1693)

andy-slac:
Updated RelationGraph code to use per-table overlap values from CSS.
Minor refactoring and simplification. There is still an issue with match
tables, angular separation for them comes from database-global overlap
parameter which is not right. Will work on that in next ticket.

Tested and works OK, integration test needed small fix in one query.

### DIFF
--- a/core/modules/qana/RelationGraph.cc
+++ b/core/modules/qana/RelationGraph.cc
@@ -732,7 +732,7 @@ RelationGraph::RelationGraph(QueryContext const& ctx,
             "passed to RelationGraph constructor.");
     }
     // Create a graph for the left-most table in a join sequence.
-    RelationGraph g(*tr, pool.get(ctx, tr->getDb(), tr->getTable()), overlap);
+    RelationGraph g(*tr, pool.get(tr->getDb(), tr->getTable()), overlap);
     // Process remaining tables in the JOIN sequence. Note that joins are
     // left-associative in the absence of parentheses, i.e. "A JOIN B JOIN C"
     // is equivalent to "(A JOIN B) JOIN C", and that relation graphs are
--- a/core/modules/qana/RelationGraph.cc
+++ b/core/modules/qana/RelationGraph.cc
@@ -245,7 +245,7 @@ size_t addEqEdge(std::string const& ca,
     }
     TableInfo const& ta = *(a->info);
     TableInfo const& tb = *(b->info);
-    LOGS(_log, LOG_LVL_DEBUG, "addEqEdge a=" << a->info->dump() << " b=" << b->info->dump());
+    LOGS(_log, LOG_LVL_DEBUG, "addEqEdge a=" << *a->info << " b=" << *b->info);
     if (ta.isEqPredAdmissible(tb, ca, cb, outer)) {
         // Add a pair of Edge objects, a → b and b → a.
         LOGS(_log, LOG_LVL_DEBUG, "addEqEdge true for (" << ca << "," << cb << ")");
@@ -498,7 +498,7 @@ size_t RelationGraph::_addWhereEqEdges(BoolTerm::Ptr where)
 /// `_addSpEdges` creates a graph edge for each admissible top-level spatial
 /// predicate extracted from the given boolean term. The number of admissible
 /// predicates is returned.
-size_t RelationGraph::_addSpEdges(BoolTerm::Ptr bt, double overlap)
+size_t RelationGraph::_addSpEdges(BoolTerm::Ptr bt)
 {
     size_t numEdges = 0;
     bt = findFirstNonTrivialChild(bt);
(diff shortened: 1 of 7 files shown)


andy-slac:
Travis fails because it does not pick up `qserv_testdata` update (same branch name). Will be OK once we merge both packages.

fritzm:
For Travis to be fixed, we also need to eups tag, eups publish, and then rebuild and push base containers.  Let me know when this hits master and I'll take care of all that.

core/modules/qana/RelationGraph.cc
```diff
@@ -524,16 +524,16 @@ size_t RelationGraph::_addSpEdges(BoolTerm::Ptr bt, double overlap)
     // Try to extract a scisql_angSep() call and a numeric constant
     // from the comparison predicate.
     FuncExpr::Ptr fe;
-    double x = std::numeric_limits<double>::quiet_NaN();
+    double angSep = std::numeric_limits<double>::quiet_NaN();
```
n8pease: andy+1 for meaningful variable names!

core/modules/qana/RelationGraph.cc
```diff
@@ -668,26 +667,29 @@ void RelationGraph::_fuse(JoinRef::Type joinType,
...
 
     if (!info) {
         return;
     } else if (info->kind != TableInfo::MATCH) {
+        LOGS(_log, LOG_LVL_DEBUG, "RG: non-match table tr=\"" << tr << "\" info=" << *info);
         _vertices.push_back(Vertex(tr, info));
         ColumnVertexMap m(_vertices.front());
         _map.swap(m);
     } else {
+        double matchAngSep = static_cast<MatchTableInfo const&>(*info).angSep;
```
n8pease: I think a polymorphic ctor would be cleaner?

andy-slac: Agree, I'd like to cleanup TableInfo and try to avoid casts, that is something for future tickets.

core/modules/qana/RelationGraph.cc
```diff
@@ -927,31 +923,53 @@ RelationGraph::RelationGraph(QueryContext const& ctx,
...
     // Build a graph for the first entry in the from list
-    RelationGraph g(ctx, refs.front(), overlap, pool);
+    RelationGraph g(refs.front(), pool);
     // "SELECT ... FROM A, B, C, ..." is equivalent to
     // "SELECT ... FROM ((A CROSS JOIN B) CROSS JOIN C) ..."
     typedef TableRefList::const_iterator Iter;
     for (Iter i = ++refs.begin(), e = refs.end(); i != e; ++i) {
-        RelationGraph tmp(ctx, *i, overlap, pool);
-        g._fuse(JoinRef::CROSS, false, JoinSpec::Ptr(), overlap, tmp);
+        RelationGraph tmp(*i, pool);
```
n8pease: I think you could make this a range-based for loop while you're here, if you feel like it.

andy-slac: I thought about it but decided to leave it for a larger refactoring later, there are many loops which can be replaced with range `for`.

core/modules/qana/TableInfo.h
```diff
@@ -275,10 +269,12 @@ struct MatchTableInfo : TableInfo {
     /// `fk` is the pair of names for the foreign key columns referencing
     /// `director.first->pk` and `director.second->pk`.
     std::pair<std::string, std::string> fk;
+    double angSep; ///< allowed angular separation between objects in director tables
 
-    MatchTableInfo(std::string const& db, std::string const& t) :
+    MatchTableInfo(std::string const& db, std::string const& t, double sep) :
         TableInfo(db, t, MATCH),
-        director(static_cast<DirTableInfo*>(nullptr), static_cast<DirTableInfo*>(nullptr)) {}
+        director(static_cast<DirTableInfo*>(nullptr), static_cast<DirTableInfo*>(nullptr)),
```
n8pease: I see this is old code, but you shouldn't have to static_cast the nullptr.

core/modules/qana/TableInfoPool.cc
```diff
@@ -116,8 +115,11 @@ TableInfo const* TableInfoPool::get(query::QueryContext const& ctx,
             throw InvalidTableError(db_ + "." + table + " is a director "
                                     "table, but cannot be sub-chunked!");
         }
-        std::unique_ptr<DirTableInfo> p(new DirTableInfo(db_, table));
-        std::vector<std::string> v = css.getPartTableParams(db, table).partitionCols();
+        css::StripingParams dbStriping = _css.getDbStriping(db_);
+        // use per-table or per-database overlap value
+        double overlap = partParam.overlap != 0.0 ? partParam.overlap : dbStriping.overlap;
+        std::unique_ptr<DirTableInfo> p(new DirTableInfo(db_, table, overlap));
```
n8pease: a name more meaningful than 'p' would be 👍 👍 IMO

core/modules/qana/RelationGraph.cc
n8pease: 👍

core/modules/qana/RelationGraph.cc
n8pease: 👍

PR #358, opened 2017-12-08
Tickets/dm 11932

fjammes:


admin/tools/docker/deployment/kubernetes/kubectl/admin/example/run-case01-test.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+# Launch Qserv multinode tests on Swarm cluster
```
fritzm: not using Swarm anymore -- maybe update comment to avoid confusion?

fjammes: :+1:

PR #359, opened 2017-12-13
Change CSS dump format to JSON, add option for subtree dump (DM-1899)

andy-slac:
KVI implementations changed to dump KV data in JSON format instead of
previously used TSV format (which was broken). Loading from JSON is
also implemented (for in-memory implementation). TSV files that we had
before are replaced with either JSON dump or C++ header which has JSON
wrapped into C++ raw string. Scons builder to translate TSV into C++
header is not needed any more and was removed.

DUMP command in qserv-admin was extended to support sub-tree dumping,
e.g. "DUMP /DBS", "DUMP" without arguments is now the same as "DUMP /".


core/modules/css/KvInterfaceImplMem.cc
```diff
@@ -136,21 +143,25 @@ KvInterfaceImplMem::set(string const& key, string const& value) {
         throw ReadonlyCss();
     }
 
+    // stored root key is an empty string, but allow "/" too
+    std::string const path(key == "/" ? "" : key);
```
jgates108: Probably worth it to have a initialization function for this as this is getting repeated a bit, or maybe a class with constructor and normalization as those see some repetition as well.

core/modules/css/KvInterfaceImplMem.cc
```diff
@@ -79,14 +82,18 @@ namespace css {
...
-    : _readOnly(readOnly) {
+    : _readOnly(false) {
     _init(mapStream);
+    // read-only should only be set after _init
+    _readOnly = readOnly;
 }
 
 KvInterfaceImplMem::KvInterfaceImplMem(string const& filename, bool readOnly)
-    : _readOnly(readOnly) {
+    : _readOnly(false) {
```
jgates108: I think it would be better to set _readOnly{false}; in the header file. It might be possible to call the other constructor directly, but I'm not sure if : KvInterfaceImplMem(std::ifstream f(filename), readOnly) will work.

core/modules/css/KvInterfaceImplMem.cc
```diff
@@ -257,38 +279,40 @@ KvInterfaceImplMem::deleteKey(string const& key) {
...
-            result += "\\N";
-        } else {
-            if (pair.second.find('\n') != std::string::npos) {
-                throw CssError("KvInterfaceImplMem::dumpKV - value contains newline");
-            }
-            result += pair.second;
+        if (pair.first.empty()) {
+            // skip root key, it is not useful and not supposed to have any data
+            continue;
+        }
```
jgates108: It might be nice to see the root element when looking at the output so you know it is starting with the root element.

admin/bin/qserv-admin.py
```diff
@@ -383,7 +386,7 @@ def _parseDrop(self, tokens):
             dbName, tbName = CommandParser._parseDbTable(tokens[1])
             self._css.dropTable(dbName, tbName)
         elif t == 'EVERYTHING':
-            raise _NotImplementedError(t)
+            self._css.getKvI().deleteKey("/")
```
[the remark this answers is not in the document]
andy-slac: Don't use it then :smile_cat: 
I do not think it's more dangerous then other DROP commands, plus EVERYTHING is hard to type accidentally.

core/modules/css/KvInterfaceImplMem.cc
andy-slac: I'll move _readOnly init to header; as for delegating constructor - that would not work, you cannot bind temporary to non-const reference.

core/modules/css/KvInterfaceImplMem.cc
andy-slac: Good suggestion, I'll add a function to normalize the path.

core/modules/css/KvInterfaceImplMem.cc
andy-slac: Our data model for CSS/KV does not need root element, it is specified entirely in terms of its leaf paths. I can expect some implementation don't have it all (I suspect that in KvInterfaceImplMem we can drop root element without any change in behavior). So I prefer to keep it invisible in the DUMP output, there was a code in mysql implementation which already done that, I have just extended it to in-memory KVI (and fixed other small inconsistencies).

PR #361, opened 2017-12-20
ATTIC: Move openstack provisionning script to qserv_deploy

fjammes:


ktlim:
We're in theory supposed to be following https://developer.lsst.io/processes/transferring_code.html

fritzm:
Wow, that (developer guide on transferring code) seems a bit overcomplicated.  I'd like to let this one past, since I doubt anybody by Fabrice has ever used, maintained, or even read this code (which is indeed one main reason why it is leaving the main repot).  But we'll conform to the dev guide for future code movement.

PR #365, opened 2018-01-20
Check error condition returned from InfileMerger::finalize (DM-13242)

andy-slac:
UserQuerySelect did not check for errors returned from InfileMerger
which in some rare cases resulted in re-use of the leftover result
tables. Now it returns error condition to user with generic error
message.


core/modules/ccontrol/UserQuerySelect.cc
```diff
@@ -284,8 +284,21 @@ void UserQuerySelect::submit() {
...
+        successful = false;
+        LOGS(_log, LOG_LVL_ERROR, getQueryIdString() << " InfileMerger::finalize failed");
+        // Error: 1105 SQLSTATE: HY000 (ER_UNKNOWN_ERROR) Message: Unknown error
+        _messageStore->addMessage(-1, 1105, "Failure while merging result",
+                MessageSeverity::MSG_ERROR);
+    }
+    try {
+        _discardMerger();
+    } catch (std::exception const& exc) {
+        // exception here means error in qserv logic, we do now want to leak
```
n8pease: now->not (right?)

andy-slac: right, not :smile_cat: will fix

PR #371, opened 2018-02-06
Eliminate SsiRequest_ReplyChannel for simplicity.

abh3:
Ready for review.


core/modules/xrdsvc/SsiRequest.cc
```diff
@@ -154,11 +153,11 @@ void SsiRequest::execute(XrdSsiRequest& req) {
             // and after the call to BindRequest.
             wbase::WorkerCommand::Ptr command;
             switch (workerCmdMsg.cmd()) {
-                case proto::WorkerCmdMsg::RELOAD_CHUNK_LIST:
-                    command = std::make_shared<wcontrol::ReloadChunkListCommand>(
-                                    std::make_shared<ReplyChannel>(shared_from_this()));
+                case proto::WorkerCmdMsg::RELOAD_CHUNK_LIST: {
+                    auto sC = std::make_shared<wbase::SendChannel>(shared_from_this());
```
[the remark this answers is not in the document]
abh3: I'd rather not for two reasons: 1) The case statement is rather complicated and warrants allocating sC before the actual call for clarity, and 2) there are numerous cases where the shared pointer needn't be allocated at all.

core/modules/xrdsvc/SsiRequest.cc
```diff
@@ -211,4 +210,63 @@ void SsiRequest::Finished(XrdSsiRequest& req, XrdSsiRespInfo const& rinfo, bool
...
+bool SsiRequest::replyStream(char const* buf, int bufLen, bool last) {
+    // Create a streaming object if not already created.
+    LOGS(_log, LOG_LVL_DEBUG, "replyStream, checking stream " << (void *) _stream
+         << " len=" << bufLen << " last=" << last);
+    if (!_stream) {
+       _stream = new ChannelStream();
+       SetResponse(_stream);
+    } else if (_stream->closed()) {
+
+        return false;
```
jgates108: Log an error here? Seems like something went sideways if it wound up here.

core/modules/xrdsvc/SsiRequest.cc
```diff
@@ -211,4 +210,63 @@ void SsiRequest::Finished(XrdSsiRequest& req, XrdSsiRespInfo const& rinfo, bool
...
+        LOGS(_log, LOG_LVL_ERROR, "DANGER: Couldn't post error response " << msg);
+        return false;
+    }
+    return true;
+}
+
+bool SsiRequest::replyFile(int fd, long long fSize) {
+    util::Timer t;
+    t.start();
+    Status s = SetResponse(fSize, fd);
```
jgates108: "Status status" would be better than "Status s". Oddly, I don't care about "Timer t". I guess, since t isn't logically important to the function.

PR #372, opened 2018-02-10
fixed a bug in the schema migration script

iagaponenko:


iagaponenko:
Since the fix was rather trivial this pull request can be self-reviewed and merged with master.

PR #375, opened 2018-02-16
extedended main SConscript to support building executables of C++ cod…

iagaponenko:
e, added a demo application which tests if versions of the Gpoogle Protobuf headers and libraries match


core/modules/tests/SConscript
```diff
@@ -2,4 +2,40 @@
...
+    target = source[:-3]
+    module_build_targets += env.Program(
+        target=target,
+        source=[source],
+        LIBS  =["qserv_common",
+                "util",
+                "protobuf",
+                "log",
+                "log4cxx"])
+    exclude.append(f)
```
andy-slac: We should try to make this generic and move to `standardModule()`

core/modules/tests/SConscript
```diff
@@ -2,4 +2,40 @@
...
+# Harvest special binary products - files starting with the package's name
+# followed by underscore:
+#
+#   tests_<something>.cc
+#
+# Then make a Scons target for each of those to be returned by the calling
+# SConscript. Also add the source file into a list of sources to be excluded
+# from the shared library product.
+
+env.Append(LIBPATH=['$build_dir'] + env['LIBPATH'])
```
andy-slac: That changes global environment which you don't want to do.

core/modules/tests/SConscript
```diff
@@ -2,4 +2,40 @@
...
+                "log4cxx"])
+    exclude.append(f)
+
+# Initiate the standard sequence of actions for this module by excluding
+# the above discovered bibary sources
+
+standardModule(env, exclude=exclude, unit_tests="")
+
+Return('module_build_targets')
+s
```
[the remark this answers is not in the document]
iagaponenko: Oops! I'll fix this.

core/modules/tests/SConscript
```diff
@@ -2,4 +2,40 @@
...
+                "log",
+                "log4cxx"])
+    exclude.append(f)
+
+# Initiate the standard sequence of actions for this module by excluding
+# the above discovered bibary sources
+
+standardModule(env, exclude=exclude, unit_tests="")
+
+Return('module_build_targets')
```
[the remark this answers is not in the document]
iagaponenko: I thought that a module's SConscript is a function which returns a value of a locally defined (within the called module) variable which is inspected by the main SConscript. That (returned) list is used on each iteration of the main SConscript's look over modules. Are you suggesting I should inject that variable into the global 'env'?

core/modules/tests/SConscript
iagaponenko: Should I then pass this as a parameter to the Program target when doing?
 env.Program(...LIBPATH=['$build_dir'] + env['LIBPATH']...)

core/modules/tests/SConscript
iagaponenko: Hence, the idea is to add an optional parameter to "standardModule" to allow passing a list of module's binary products (along with their dependencies), someting like:
standardModule(binaries=[{"tests_app.cc": ["qserv_common","util","log4cxx"]}, {"tests_somethingelse.cc":[...]),,,],

core/modules/tests/SConscript
iagaponenko: Okay, I've deleted a line of code which modified LIBPATH in the global environment and passed it to each env.Program() invocation as a parameter. And this works.

iagaponenko:
@andy-slac okay, I've made an improved version of the build system as per your suggestion. Could you have a look at it please? Thanks!

core/modules/SConscript
```diff
@@ -95,6 +95,14 @@ def standardModule(env, exclude=None, module=None, test_libs=None, unit_tests=No
                 then use this parameter to pass list (or space-separated
                 string) of application names. Can be empty string or list to
                 not run any unit tests.
+    @param bin_cc_targets: a dictionary defining whgich binary products (not
```
andy-slac: whgich -> which
Not shure `targets` has a meaning here. The key in dict is a name of the source, not target?

core/modules/SConscript
```diff
@@ -104,13 +112,23 @@ def standardModule(env, exclude=None, module=None, test_libs=None, unit_tests=No
         module = os.path.basename(os.getcwd())
     state.log.debug('standardModule: module = %s path = %s' % (module, path))
 
+    # using two optional lists provided by the nmodule to build names of
+    # *.cc files to be excluded from being used to build module's
+    # library or unit tests
+    exclude_cc_files = []
+    if exclude is not None:
+        exclude_cc_files = exclude
```
andy-slac: `exclude_cc_files += exclude`, otherwise you are potentially modifying `exclude` too

core/modules/SConscript
```diff
@@ -104,13 +112,23 @@ def standardModule(env, exclude=None, module=None, test_libs=None, unit_tests=No
...
+    if bin_cc_targets is not None:
+        for f in bin_cc_targets: exclude_cc_files.append(f)
```
andy-slac: `exclude_cc_files += bin_cc_targets.keys()`?

core/modules/SConscript
```diff
@@ -95,6 +95,14 @@ def standardModule(env, exclude=None, module=None, test_libs=None, unit_tests=No
                 then use this parameter to pass list (or space-separated
                 string) of application names. Can be empty string or list to
                 not run any unit tests.
+    @param bin_cc_targets: a dictionary defining whgich binary products (not
+                the unit tests should be built out of C/C++ files. Each key
```
andy-slac: Opening parenthesis on previous line is not closed, I guess it should go after "tests".

core/modules/SConscript
```diff
@@ -163,6 +181,18 @@ def standardModule(env, exclude=None, module=None, test_libs=None, unit_tests=No
     py_installed = env.Install("$prefix/bin", py_scripts)
     build_data['install'] += py_installed
 
+    # inject binary targets built of the optionally provided C/C++ files
+    if bin_cc_targets is not None:
+        for f,dependencies in bin_cc_targets.items():
+            source = os.path.basename(str(f))
```
andy-slac: Why do you need basename for `source`?

core/modules/SConscript
```diff
@@ -163,6 +181,18 @@ def standardModule(env, exclude=None, module=None, test_libs=None, unit_tests=No
...
+            target = source[:-3]
+            p = env.Program(
+                target=target,
+                source=[source],
+                LIBS=dependencies,
+                LIBPATH=['$build_dir'] + env['LIBPATH'])
```
andy-slac: Looks a bit  complicated, I'd think this should work:

    for f, dependencies in bin_cc_targets.items():
        p = env.Program(f, LIBS=dependencies, LIBPATH=['$build_dir'] + env['LIBPATH'])

core/modules/tests/SConscript
```diff
@@ -2,4 +2,23 @@
...
+#   tests_<something>.cc
+#
+# Then make a Scons target for each of those to be returned by the calling
+# SConscript. Also add the source file into a list of sources to be excluded
+# from the shared library product.
+
+bin_cc_targets = {}
+path = "."
+for f in env.Glob(os.path.join(path, "tests_*.cc"), source=True, strings=True):
+    bin_cc_targets[f] = ["qserv_common","util","protobuf","log","log4cxx"]
```
andy-slac: Don't overdo it. Normally all "test*.cc" files are (unit) tests and we don't want their binaries installed so don't try to make _all_ test_*.cc into true binaries. I'd probably give it different name, not starting with "test".

I believe you can also write:

    bin_cc_targets[f] = "qserv_common util protobuf log log4cxx"

iagaponenko: Perhaps we should come up with some naming convention for the binaries. Something like: qserv-<module>-<whatever>.cc. For instance my only app module tests should be renamed from "tests_protobuf_version.cc" into "qserv-tests-protobuf.cc". And I agree about avoiding using "test*" in apps names". I just wanted to see if my implementation of standardModule would work. And thanks for the advice to replace a list of dependencies with a string. I'll give it a try.

core/modules/SConscript
iagaponenko: I will give it a try, and if it works then I'll make mods to the code.

core/modules/SConscript
iagaponenko: Agreed. No need. I will eliminate it.

core/modules/SConscript
iagaponenko: ok

core/modules/SConscript
iagaponenko: Agreed. My Python is rusty :-)

core/modules/SConscript
iagaponenko: Agreed

core/modules/SConscript
iagaponenko: I'll come up with a better name

iagaponenko:
@andy-slac thank you for reviewing the code and making good suggestions! I've just committed a few more modification to the code as per your suggestions.

PR #380, opened 2018-03-06
Remove SECRET_KEY from wmgr conf

fjammes:


fjammes:
Are you sure it is used yet? I cannot find any reference in Qserv code to this parameter:

{code}
➜  qserv git:(master) grep -r SECRET_KEY *
admin/templates/configuration/etc/qserv-wmgr.cnf:SECRET_KEY = '{{WMGR_SECRET_KEY}}'
admin/python/lsst/qserv/admin/configure.py:                'WMGR_SECRET_KEY': random_string(string.ascii_letters + string.digits, 57),
core/modules/wmgr/test/test.cfg:SECRET_KEY = 'd2d69c1353ddfaeb56bc44504de039270c339eb8811d8af65ee9a01f59b325472b'
{code}

andy-slac:
@fjammes, I think I mentioned it on JIRA ticket - SECRET_KEY is used by Flask (our web framework) directly. This configuration file is not only for Qserv, it is also passed to Flask so we can add some Flask parameters there.

fjammes:
Ok I discard this commit.

PR #382, opened 2018-04-03
Fix dirtily sql migration script

fjammes:


iagaponenko:
I agree with the proposed change.

PR #386, opened 2018-04-08
Fix integration test worker crash for MariaDB 10.2

fritzm:
Under MariaDB 10.2.14, the MYSQL_FIELD def_length field is sometimes coming back garbage (bignum), causing the std::string ctor below to blow up.

MySQL docs say this field and its partner are not set unless this MYSQL_FIELD comes back from a call to mysql_list_fields(), which is not the case in this call chain.  Perhaps this field happened to be init'd or harmless in previous versions of MariaDB, but is seems no more...

With the unsafe code simply ellided per below, the integration tests pass.  Do we need this code?  Was it ever doing anything correct/useful?


fritzm:
(Oh, and in case you are wondering, in the crashing case, the NO_DEFAULT_VALUE_FLAG is *not* set 😡)

PR #390, opened 2018-04-19
Tickets/dm 13749

n8pease:


core/modules/ccontrol/UserQueryFactory.cc
```diff
@@ -154,28 +156,49 @@ UserQueryFactory::newUserQuery(std::string const& aQuery,
...
-            bool theyMatch(*a4stmt == *stmt);
-            LOGS(_log, LOG_LVL_DEBUG, "they match:" << theyMatch);
+        if (a4stmt && a2stmt) {
+            bool theyMatch(*a4stmt == *a2stmt);
+            LOGS(_log, LOG_LVL_DEBUG, "antlr v2 and antlr4 generated queries match:" << theyMatch);
+        }
+
+        // TEMP while developing the antlr4 query generation; decide which generated select statment to use
+        // to process the query:
+#if USE_ANTLR4_QUERY
```
iagaponenko: Would it make a sense to avoid making a2stmt if this is defined? The #ifdef could be moved up to line 159.

core/modules/parser/QSMySqlListener.cc
```diff
@@ -53,8 +53,25 @@ namespace {
...
 
+std::string getQueryString(antlr4::ParserRuleContext * ctx) {
+    return ctx->getStart()->getInputStream()->getText(
+           antlr4::misc::Interval(ctx->getStart()->getStartIndex(), ctx->getStop()->getStopIndex()));
 }
 
+template <typename T>
+std::string getTypeName() {
+    int status;
+    return abi::__cxa_demangle(typeid(T).name(),0,0,&status);
```
iagaponenko: Is this really cross-compiler portable? My understanding is that this is a specific feature of the GCC family of compilers.

core/modules/parser/QSMySqlListener.cc
```diff
@@ -53,8 +53,25 @@ namespace {
...
+}
+
+template <typename T>
+std::string getTypeName(T obj) {
+    int status;
+    return abi::__cxa_demangle(typeid(obj).name(),0,0,&status);
```
iagaponenko: Same comment regarding the compiler portability

core/modules/parser/QSMySqlListener.cc
```diff
@@ -378,18 +395,27 @@ class RootAdapter :
...
+    {}
 
     shared_ptr<query::SelectStmt>& getSelectStatement() { return _selectStatement; }
 
     void handleDmlStatement(shared_ptr<query::SelectStmt>& selectStatement) override {
         _selectStatement = selectStatement;
     }
 
-    void onExit() override {}
+    void onEnter(QSMySqlParser::RootContext * ctx) {
```
iagaponenko: Extra space in 'RootContext *'?

core/modules/parser/QSMySqlListener.cc
```diff
@@ -378,18 +395,27 @@ class RootAdapter :
...
+    }
+
+    void onExit() override {
+        CHECK_EXECUTION_CONDITION(_selectStatement != nullptr, "Could not parse query.", _ctx);
+    }
 
 private:
     shared_ptr<query::SelectStmt> _selectStatement;
+    QSMySqlParser::RootContext * _ctx;
 };
```
iagaponenko: Same comment on the extra space before '*'. Otherwise it looks like a multiplication :-)

core/modules/parser/QSMySqlListener.cc
```diff
@@ -837,11 +856,13 @@ class BinaryComparasionPredicateAdapter :
         public ComparisonOperatorCBH {
 public:
     BinaryComparasionPredicateAdapter(shared_ptr<BinaryComparasionPredicateCBH>& parent,
-                                      antlr4::ParserRuleContext* ctx)
-    : AdapterT(parent) {}
+            QSMySqlParser::BinaryComparasionPredicateContext * ctx)
```
iagaponenko: Extra space before '*'

core/modules/parser/QSMySqlListener.cc
```diff
@@ -90,13 +107,13 @@ void QSMySqlListener::exit##NAME(QSMySqlParser::NAME##Context * ctx) {\
 } \
 
 
-#define CHECK_EXECUTION_CONDITION(CONDITION, MESSAGE) \
+#define CHECK_EXECUTION_CONDITION(CONDITION, MESSAGE, CTX) \
```
iagaponenko: I would suggest renaming this macro into what it actually does:
ASSERT_EXECUTION_CONDITION

core/modules/parser/QSMySqlListener.cc
```diff
@@ -882,6 +903,7 @@ class BinaryComparasionPredicateAdapter :
     shared_ptr<query::ValueExpr> _left;
     string _comparison;
     shared_ptr<query::ValueExpr> _right;
+    QSMySqlParser::BinaryComparasionPredicateContext * _ctx;
```
iagaponenko: Extra space before '*'

core/modules/parser/QSMySqlListener.cc
```diff
@@ -852,13 +873,13 @@ class BinaryComparasionPredicateAdapter :
         } else if (_right == nullptr) {
             _right = valueExpr;
         } else {
-            CHECK_EXECUTION_CONDITION(false, "left and right values must be set only once.")
+            CHECK_EXECUTION_CONDITION(false, "left and right values must be set only once.", _ctx)
         }
     }
 
     void onExit() {
         CHECK_EXECUTION_CONDITION(_left != nullptr && _right != nullptr,
```
iagaponenko: Perhaps just: _left && _right ? The shared_ptr already has this operator:
http://en.cppreference.com/w/cpp/memory/shared_ptr/operator_bool

core/modules/parser/QSMySqlListener.cc
```diff
@@ -890,14 +912,16 @@ class ComparisonOperatorAdapter :
...
+    {}
 
     void onExit() override {
-        lockedParent()->handleComparisonOperator(_comparisonOperatorCtx->getText());
+        lockedParent()->handleComparisonOperator(_ctx->getText());
     }
 
 private:
-    QSMySqlParser::ComparisonOperatorContext * _comparisonOperatorCtx;
+    QSMySqlParser::ComparisonOperatorContext * _ctx;
```
iagaponenko: Extra space before '*'

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1049,12 +1074,14 @@ class UdfFunctionCallAdapter :
 public:
     UdfFunctionCallAdapter(shared_ptr<UdfFunctionCallCBH>& parent,
                            QSMySqlParser::UdfFunctionCallContext * ctx)
```
iagaponenko: Extra space before '*'

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1077,6 +1104,7 @@ class UdfFunctionCallAdapter :
 private:
     shared_ptr<query::ValueExpr> _args;
     string _functionName;
+    QSMySqlParser::UdfFunctionCallContext * _ctx;
```
iagaponenko: Extra space before '*'

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1237,25 +1260,25 @@ class BetweenPredicateAdapter :
...
         if (childCtx == _ctx->max) {
-            CHECK_EXECUTION_CONDITION(nullptr == _max, "max should be set exactly once.");
+            CHECK_EXECUTION_CONDITION(nullptr == _max, "max should be set exactly once.", _ctx);
             _max = valueExpr;
             return;
         }
     }
 
     void onExit() {
         CHECK_EXECUTION_CONDITION(nullptr != _val && nullptr != _min && nullptr != _max,
```
iagaponenko: Perhaps just: _val && _min && _max

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1237,25 +1260,25 @@ class BetweenPredicateAdapter :
     void handleExpressionAtomPredicate(shared_ptr<query::ValueExpr>& valueExpr,
             antlr4::ParserRuleContext* childCtx) override {
         if (childCtx == _ctx->val) {
-            CHECK_EXECUTION_CONDITION(nullptr == _val, "val should be set exactly once.");
+            CHECK_EXECUTION_CONDITION(nullptr == _val, "val should be set exactly once.", _ctx);
```
iagaponenko: !_val instead of the comparison?

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1275,14 +1298,19 @@ class UnaryExpressionAtomAdapter :
...
     // ConstantExpressionAtomCBH
     void handleConstantExpressionAtom(const string& text) override {
-        CHECK_EXECUTION_CONDITION(false, "todo");
+        CHECK_EXECUTION_CONDITION(false, "can't handle ConstantExpressionAtom" << text, _ctx);
     }
 
     void onExit() override {}
+
+private:
+    QSMySqlParser::UnaryExpressionAtomContext * _ctx;
```
iagaponenko: Extra space before '*'?

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1293,7 +1321,9 @@ class MathExpressionAtomAdapter :
 public:
     MathExpressionAtomAdapter(shared_ptr<MathExpressionAtomCBH> parent,
                               QSMySqlParser::MathExpressionAtomContext * ctx)
```
iagaponenko: Extra space before '*'?

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1293,7 +1321,9 @@ class MathExpressionAtomAdapter :
 public:
     MathExpressionAtomAdapter(shared_ptr<MathExpressionAtomCBH> parent,
```
iagaponenko: const& parent?

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1293,7 +1321,9 @@ class MathExpressionAtomAdapter :
...
                               QSMySqlParser::MathExpressionAtomContext * ctx)
-    : AdapterT(parent) {}
+    : AdapterT(parent)
+    , _ctx(ctx)
+    {}
 
     void handleFunctionCallExpressionAtom(shared_ptr<query::FuncExpr> funcExpr) override {
```
iagaponenko: const& funcExpr?

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1303,13 +1333,13 @@ class MathExpressionAtomAdapter :
         // need to make:
         // FactorOp(MINUS, factor:ValueFactor(type:Function, funcExpr)
         CHECK_EXECUTION_CONDITION(MathOperatorCBH::SUBTRACT == operatorType,
-                "Currenty only subtract is supported.");
+                "Currenty only subtract is supported.", _ctx);
         _op = query::ValueExpr::MINUS;
     }
 
     void onExit() override {
         CHECK_EXECUTION_CONDITION(query::ValueExpr::NONE != _op && nullptr != _left && nullptr != _right,
```
iagaponenko: query::ValueExpr::NONE != _op && _left && _right

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1323,13 +1353,14 @@ class MathExpressionAtomAdapter :
...
         } else {
-            CHECK_EXECUTION_CONDITION(false, "left and right are both already set:" << *this);
+            CHECK_EXECUTION_CONDITION(false, "left and right are both already set:" << *this, _ctx);
         }
     }
 
     query::ValueExpr::Op _op {query::ValueExpr::NONE};
     shared_ptr<query::FuncExpr> _left;
     shared_ptr<query::FuncExpr> _right;
+    QSMySqlParser::MathExpressionAtomContext * _ctx;
```
iagaponenko: Extra space before *

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1347,25 +1378,25 @@ class FunctionCallExpressionAtomAdapter :
 public:
     FunctionCallExpressionAtomAdapter(shared_ptr<FunctionCallExpressionAtomCBH> parent,
                                       QSMySqlParser::FunctionCallExpressionAtomContext * ctx)
```
iagaponenko: Extra space before *

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1347,25 +1378,25 @@ class FunctionCallExpressionAtomAdapter :
 public:
     FunctionCallExpressionAtomAdapter(shared_ptr<FunctionCallExpressionAtomCBH> parent,
```
iagaponenko: const& parent

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1347,25 +1378,25 @@ class FunctionCallExpressionAtomAdapter :
...
                                       QSMySqlParser::FunctionCallExpressionAtomContext * ctx)
-    : AdapterT(parent) {}
+    : AdapterT(parent)
+    , _ctx(ctx)
+    {}
 
     void handleUdfFunctionCall(shared_ptr<query::FuncExpr> funcExpr) override {
```
iagaponenko: const& funcExpr

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1347,25 +1378,25 @@ class FunctionCallExpressionAtomAdapter :
...
     void onExit() {
-//        auto valueFactor = query::ValueFactor::newFuncFactor(_funcExpr);
-//        auto valueExpr = make_shared<query::ValueExpr>();
-//        ValueExprFactory::addValueFactor(valueExpr, valueFactor);
         lockedParent()->handleFunctionCallExpressionAtom(_funcExpr);
     }
 
 private:
     shared_ptr<query::FuncExpr> _funcExpr;
+    QSMySqlParser::FunctionCallExpressionAtomContext * _ctx;
```
iagaponenko: Extra space before *

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1374,11 +1405,16 @@ class UnaryOperatorAdapter :
 public:
     UnaryOperatorAdapter(shared_ptr<UnaryOperatorCBH> parent,
                          QSMySqlParser::UnaryOperatorContext * ctx)
```
iagaponenko: Extra space before *

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1374,11 +1405,16 @@ class UnaryOperatorAdapter :
 public:
     UnaryOperatorAdapter(shared_ptr<UnaryOperatorCBH> parent,
```
iagaponenko: const& parent

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1374,11 +1405,16 @@ class UnaryOperatorAdapter :
...
+    , _ctx(ctx)
+    {}
 
     void onExit() override {
-        CHECK_EXECUTION_CONDITION(false, "todo");
+        CHECK_EXECUTION_CONDITION(false, "UnaryOperatorAdapter is not yet handling anything.", _ctx);
     }
+
+private:
+    QSMySqlParser::UnaryOperatorContext * _ctx;
```
iagaponenko: Extra space before *

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1394,7 +1430,7 @@ class LogicalOperatorAdapter :
         if (_ctx->AND() != nullptr) {
```
iagaponenko: if (_xtc->AND()) {

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1441,15 +1475,10 @@ shared_ptr<query::SelectStmt> QSMySqlListener::getSelectStatement() const {
 template<typename ParentCBH, typename ChildAdapter, typename Context>
 shared_ptr<ChildAdapter> QSMySqlListener::pushAdapterStack(Context * ctx) {
```
iagaponenko: Extra space before *

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1441,15 +1475,10 @@ shared_ptr<query::SelectStmt> QSMySqlListener::getSelectStatement() const {
...
     auto p = dynamic_pointer_cast<ParentCBH>(_adapterStack.top());
```
iagaponenko: Normally I would always put 'const' for variables which are not meant to be modified later:
auto const p = ...

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1458,7 +1487,7 @@ shared_ptr<ChildAdapter> QSMySqlListener::pushAdapterStack(Context * ctx) {
 
 
 template<typename ChildAdapter>
-void QSMySqlListener::popAdapterStack() {
+void QSMySqlListener::popAdapterStack(antlr4::ParserRuleContext * ctx) {
```
iagaponenko: Extra space before *

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1467,27 +1496,28 @@ void QSMySqlListener::popAdapterStack() {
...
+        ", Actual: " << getTypeName(adapterPtr) <<
+        ", Are there out of order or unhandled listener exits?",
+        ctx);
 }
 
 
 // QSMySqlListener class methods
 
 
 void QSMySqlListener::enterRoot(QSMySqlParser::RootContext * ctx) {
```
iagaponenko: Extra space before *

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1467,27 +1496,28 @@ void QSMySqlListener::popAdapterStack() {
...
-    // root is pushed by the ctor (and popped by the dtor)
+    CHECK_EXECUTION_CONDITION(_adapterStack.empty(), "RootAdatper should be the first entry on the stack.",
+            ctx);
+    _rootAdapter = make_shared<RootAdapter>();
+    _adapterStack.push(_rootAdapter);
+    _rootAdapter->onEnter(ctx);
 }
 
 
 void QSMySqlListener::exitRoot(QSMySqlParser::RootContext * ctx) {
```
iagaponenko: Extra space before *

core/modules/parser/QSMySqlListener.h
```diff
@@ -1658,7 +1658,7 @@ class QSMySqlListener : public QSMySqlParserBaseListener {
     std::shared_ptr<ChildAdapter> pushAdapterStack(Context * ctx);
 
     template<typename ChildAdapter>
-    void popAdapterStack();
+    void popAdapterStack(antlr4::ParserRuleContext * ctx);
```
iagaponenko: Extra space before *

core/modules/parser/QSMySqlListener.cc
n8pease: My understanding is that it's not totally portable. It is *very* useful. I encapsulated it so that if we encounter problems with other compilers it can be worked around. As far as I know there's not a universal solution to get the class name, ya?

core/modules/parser/QSMySqlListener.cc
n8pease: I think I was following a pattern from elsewhere in the qserv/query code. It's definitely not my instinct to put spaces on either side, but it *is* my instinct to make my code look like other code around me. But I'm not sure why I did it here....

core/modules/parser/QSMySqlListener.cc
n8pease: naming is the hardest part! good suggestion.

core/modules/parser/QSMySqlListener.cc
iagaponenko: Agreed on all counts.

core/modules/parser/QSMySqlListener.cc
n8pease: 👍

n8pease: oh, but in this case `p` is passed into the ctor of the ChildAdapter, it becomes the child's parent, and receives callbacks which may change state; so it is not supposed to be const here.

core/modules/parser/QSMySqlListener.cc
n8pease: yeah; my preference when given the option is usually the verbose route.

core/modules/ccontrol/UserQueryFactory.cc
n8pease: for the time being it makes sense to generate (and log) both statements, because I often compare them to each other when problems arise.

core/modules/parser/QSMySqlListener.cc
iagaponenko: Adding const to the variable means that no one else can change that variable to point to a different object. The proposed earlier 'const'-ness is not 'viral' (it won't propagate down to an object referenced by the smart pointer). For the later you would need to:
auto p = dynamic_pointer_cast<ParentCBH const>
which I didn't suggest :-)

n8pease:
merged via antlr4 integration branch.

PR #397, opened 2018-05-25
fix async integration tests

n8pease:


n8pease:
the ticket branch was in a bad state when I created this. closing & trying again.

PR #399, opened 2018-05-31
Tickets/dm 14560

jgates108:


abh3:
So, has this solved the stalling problem?

jgates108:
It managed to replace the stalling problem with mysql server problem on the czar.

PR #401, opened 2018-06-20
modify the ProtoRowBuffer escaper to handle 2 slashes properly

n8pease:


n8pease:
what I gather from the docs ([here](https://dev.mysql.com/doc/refman/8.0/en/load-data.html)) is that ENCLOSED BY chars don't need to be escaped, as long as they are not followed by the FIELD TERMINATED BY character:

> For input, the ENCLOSED BY character, if present, is stripped from the ends of field values. (This is true regardless of whether OPTIONALLY is specified; OPTIONALLY has no effect on input interpretation.) Occurrences of the ENCLOSED BY character preceded by the ESCAPED BY character are interpreted as part of the current field value.
>
> If the field begins with the ENCLOSED BY character, instances of that character are recognized as terminating a field value only if followed by the field or line TERMINATED BY sequence.

our TERMINATED BY character seems to be `\t` since we don't specify it:

```
std::string formLoadInfile(std::string const& table,
                           std::string const& virtFile) {
    auto sql = "LOAD DATA LOCAL INFILE '" + virtFile + "' INTO TABLE " + table
            + " FIELDS ENCLOSED BY '\\\''";
    LOGS(getLogger(), LOG_LVL_TRACE, "Load query: " << sql);
    return sql;
}
```

>If you specify no FIELDS or LINES clause, the defaults are the same as if you had written this:
> FIELDS TERMINATED BY '\t' ENCLOSED BY '' ESCAPED BY '\\'
> LINES TERMINATED BY '\n' STARTING BY ''

It is not be possible to get into the situation where `'\t` happens inside a field because we escape the tab character, so any instance of that string inside a field would become `'\\t` in `local_infile_read`.

AFAIK a unit test wouldn't verify anything except that we don't escape the `'` char inside fields, verifying good behavior from mysql would require an integration test I think? There are single quotes inside our test data, I can work on making sure one of those rows gets exercised. Finding a quote followed by a tab is challenging me at the moment, but I can keep digging on it.

n8pease:
To create a test to verify that a blob that contains our FIELDS ENCLOSED BY char (`'`) followed by the field separation char (`\t`) gets escaped properly in LocalInfile, and so does not spuriously end the field in the middle of a blob:

I found a blobField that contains `\'\t` having `sourceId=29759322768015696`, using the query `SELECT * FROM Source WHERE Source.blobField LIKE '%\'\t%';`  (I did this while connected to the MariaDb instance, this query does not parse in qserv)

I can use that info to write a query that *does* parse in qserv;
```SELECT blobField FROM Source WHERE sourceId=29759322768015696;``` 

I can add an integration test that contains this query (which will verify that the returned binary blob (or the ascii version of it? hmm...) is the same in both qserv & MariaDb)

I temporarily added a log statement to further verify that what I think is happening is actually happening; I did see that go by once I added the test
```[2018-06-29T11:21:42.036-0700] [LWP:12429] DEBUG rproc.ProtoRowBuffer (core/modules/rproc/ProtoRowBuffer.h:75) - single quote followed by tab in: ^D^B[2018-06-29T11:21:42.036-0700] [LWP:12429] DEBUG mysql.LocalInfile (core/modules/mysql/LocalInfile.cc:169) - QI=3083:4;rowBuffer=ProtoRowBuffer schema((Name=jobId0,colType=INT(9):3)(Name=blobField,colType=BLOB:252)) Row 0('400'~9~'~4~~2~\0~23~\0\0\0~2~\0~11~\0~-126~\0first_fieldsecond_field~-45~~-67~C~-92~t~-17~~-64~?~-97~$D~-80~'\t~-79~?')```

File: core/modules/rproc/ProtoRowBuffer.h [hunk 1]
```diff
@@ -75,8 +75,20 @@ class ProtoRowBuffer : public mysql::RowBuffer {
               case '\r':   *destI++ = '\\'; *destI++ = 'r'; break;
               case '\t':   *destI++ = '\\'; *destI++ = 't'; break;
               case '\032': *destI++ = '\\'; *destI++ = 'Z'; break;
+              case '\\': {
+                  auto nextI = i + 1;
```
andy-slac: nitpick - add `const` to declaration if you don't want to modify it.

n8pease:
this is merged to master.

PR #402, opened 2018-06-21
Migrate containers to Centos7 + Devtoolset6

fritzm:
This should address currently busted Travis CI runs, restore buildability of containers off master (sphgeom pybind11 compiler compatibility problem), and pave the way for integration of Antler4 on master.  Notes from the git commit log:

- Switch base container to lsstsqre/centos:7-stackbase-devtoolset-6
- Remove installation of stack prereqs already in base image
- Use yum instead of apt for remaining package installs
- Add package rh-git29 to meet git version dependency for newinstall and eupspkg
- Add package initscripts (and update qserv-functions) for getpidofproc
- Add -l arg to "RUN bash" lines throughout dockerfiles to pick up scl enables
- Replace deprecated "MAINTAINER" in dockerfiles with now-preferred "LABEL maintainer="
- Remove graphviz from installed debug/utility packages
- Check for existence of /etc/localtime and /etc/timezone before attempting to chown
- Fix Debian-specific "\n" expansion problem in run-multinode-tests.sh

### DIFF
--- a/admin/tools/docker/latest/Dockerfile
+++ b/admin/tools/docker/latest/Dockerfile
@@ -1,40 +1,19 @@
-FROM debian:jessie
-MAINTAINER Fabrice Jammes <fabrice.jammes@in2p3.fr>
+FROM lsstsqre/centos:7-stackbase-devtoolset-6
+LABEL maintainer="Fabrice Jammes <fabrice.jammes@in2p3.fr>"
 
 RUN sh -c "echo \"BUILD ID: $(date '+%Y%m%d_%H%M%S')\" > /BUILD_ID"
 
-# Start with this long step not to re-run it on
-# each Dockerfile update
-RUN echo "deb http://ftp.debian.org/debian jessie-backports main" >> /etc/apt/sources.list
-RUN apt-get -y update && \
-    apt-get -y install apt-utils && \
-    apt-get -y upgrade && \
-    apt-get -y clean
-
-# Install LSST stack dependencies
+# Provide newer git for newinstall and eupspkg
 #
-RUN apt-get -y install bison \
-        bzip2 \
-        cmake \
-        curl \
-        flex \
-        g++ \
-        gettext \
-        git \
-        libbz2-dev \
-        libglib2.0-dev \
-        libpthread-workqueue-dev \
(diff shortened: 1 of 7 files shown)


admin/tools/docker/latest/Dockerfile
```diff
@@ -1,40 +1,19 @@
-FROM debian:jessie
-MAINTAINER Fabrice Jammes <fabrice.jammes@in2p3.fr>
+FROM lsstsqre/centos:7-stackbase-devtoolset-6
```
iagaponenko: What's special about this version of the base container, an why not to use the standard one like centos/devtoolset-6-toolchain-centos7? Does the LSST one comes with extra RPMs required by the LSST Stack?

admin/tools/docker/latest/Dockerfile
```diff
@@ -47,22 +26,26 @@ ARG EUPS_TAG
 # Install development and debugging tools
 #
 RUN if [ "$EUPS_TAG" = "qserv-dev" ] ; then \
-    apt-get -y install byobu dnsutils \
-        gdb graphviz lsof \
-        net-tools vim ; \
+        yum install -y \
```
iagaponenko: Ideally, it would be also nice to include sysstat (for performance monitoring) to have tools like "iostat" inside the container. Some host deployments may not have those tools by default. Though, I don't think it's terribly critical here.

admin/tools/docker/latest/Dockerfile
fritzm: Yes — devtoolset six and all platform-specific prereqs are cooked into this container.  It is maintained by SQuaRE as stack dependencies/tooling evolve, and is the official base container for all the non-qserv Jenkins container builds.

The hope is by using the same base container as the stack folk, we avoid inadvertent compiler-specific breakage as happened with the sphgeom thing recently.  I think this will be important until Qserv and stack/eups part ways.

admin/tools/docker/latest/Dockerfile
fritzm: Systat (and vim) are already in the base image here, so we’re covered!

PR #404, opened 2018-07-11
Add -Wextra option for C++ compilation (DM-15084)

andy-slac:
I had to disable "unused parameter" warning as there are too many of
those and they also appear in external packages.

### DIFF
--- a/site_scons/site_tools/compiler.py
+++ b/site_scons/site_tools/compiler.py
@@ -84,7 +84,8 @@ def generate(env):
         # clang on OS x
 
         # Increase compiler strictness
-        env.Append(CCFLAGS=['-pedantic', '-Wall', '-Wno-variadic-macros'])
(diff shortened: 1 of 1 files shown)


timj:
`sconsUtils` uses `-isystem` to mute warnings from external packages.

andy-slac:
I think one potential issue with `-isystem` is that scons, AFAIK, does not scan "system headers" as dependencies. This is of course very beneficial for faster builds but can also be a problem if one switches to different version of external package. We need to think about it.

fritzm:
👎 on `-isystem` -- I want to know when my compiler is struggling with outdated libraries, not just plug my ears and pretend its all okay

timj:
That only helps if you intend to fix upstream and hope they take your fixes. This was discussed in [RFC-246](https://jira.lsstcorp.org/browse/RFC-246).

PR #406, opened 2018-07-18
DM-15156: fix case01/queries/1012_orderByClause.sql.FIXME

n8pease:


core/modules/qana/PostPlugin.cc
```diff
@@ -157,47 +157,49 @@ PostPlugin::applyPhysical(QueryPlugin::Plan& plan,
...
-                if (!orderByDb.empty() || !orderByTbl.empty()) {
-                    if (!orderByTbl.empty()) orderByTbl += ".";
-                    std::string msg = "ORDER BY argument should not include database or table for "
-                        + orderByDb + " " + orderByTbl + orderByCol + ".\n Remove qualifiers or "
-                        "use an alias.\n  ex:'SELECT a.val AS aVal ... ORDER BY aVal";
-                    LOGS(_log, LOG_LVL_WARN, msg);
-                    throw AnalysisError(msg);
-                }
+                query::ColumnRef::Vector orderByColumnRefVec;
+                factorOp.factor->findColumnRefs(orderByColumnRefVec);
```
n8pease: this is the key change; `getColumnRef` returns the ColumnRef value of the ValueFactor, *if* the ValueFactor is of type column ref. `findColumnRefs` recursively gets all the column refs used by the value factor (e.g. if the ValueFactor is of type Function and the func uses one or more ColumnRef arguments, it will return all the column ref arguments used by the function).

core/modules/parser/QSMySqlListener.cc
```diff
@@ -2317,6 +2318,11 @@ class FunctionCallExpressionAtomAdapter :
         _funcExpr = funcExpr;
     }
 
+    void handleScalarFunctionCall(shared_ptr<query::FuncExpr> const & funcExpr) {
```
iagaponenko: missing override

core/modules/parser/QSMySqlListener.cc
```diff
@@ -2304,7 +2304,8 @@ class MathExpressionAtomAdapter :
 
 class FunctionCallExpressionAtomAdapter :
         public AdapterT<FunctionCallExpressionAtomCBH>,
-        public UdfFunctionCallCBH {
+        public UdfFunctionCallCBH,
+		public ScalarFunctionCallCBH {
```
iagaponenko: bad tab?

n8pease:
I found a couple issues requiring some significant work. closing this PR for now.

PR #410, opened 2018-08-08
Tickets/dm 15320

jgates108:


admin/templates/configuration/etc/xrdssi.cnf
```diff
@@ -29,8 +29,8 @@ location = {{QSERV_DATA_DIR}}/mysql
 [scheduler]
 
 # Thread pool size
-# thread_pool_size = 10
-thread_pool_size = 20
+# thread_pool_size = 20
+thread_pool_size = 12
```
abh3: I've seen these number change a lot. They shouldn't be constants but come from a config file.

admin/templates/configuration/etc/xrdssi.cnf
```diff
@@ -44,11 +44,11 @@ group_size = 10
...
 # priority_snail = -20
-# priority_slow = 2
-priority_slow = 4
+priority_slow = 2
+# priority_slow = 4
 # priority_med = 3
-# priority_fast = 4
-priority_fast = 2
+priority_fast = 4
+# priority_fast = 2
```
abh3: OK, all of these values should be speciable via a config. Please don't use envars -- there are too many constants here.

core/modules/memman/MemFile.cc
```diff
@@ -46,58 +55,81 @@ std::unordered_map<std::string, MemFile*> fileCache;
...
+    int rc = 0;
+    {
+        std::lock_guard<std::mutex> guard(_fileMutex);
+
+        // If the file is already locked, indicate success
+        //
+        if (_isLocked) {
+            MLResult aokResult(_memInfo.size(), _memInfo.mlockTime(), 0);
+            return aokResult;
+        }
```
abh3: I'm usually skeptical of lock within locks because it may point to a misunderstanding of what the lock is supposed to do and ultimately leads to deadlocks. So, why the change here?

core/modules/memman/MemFile.cc
```diff
@@ -46,58 +55,81 @@ std::unordered_map<std::string, MemFile*> fileCache;
...
 
+
+
 /******************************************************************************/
 /*                                m e m M a p                                 */
 /******************************************************************************/
 
 int MemFile::memMap() {
-
+    //
```
abh3: Odd little uninformative comment.

core/modules/memman/MemFile.h
```diff
@@ -138,7 +140,10 @@ class MemFile {
     bool        _isMapped   = false;   // Protected by _fileMutex
     bool        _isReserved = false;   // Ditto
     bool        _isLocked   = false;   // Ditto
```
abh3: So, _fileMutex protects _isMapped while _mlockFileMutex protects_ isLocked. Sounds like limited scope here can only one mutex be used for both or one mutex and an atomic?

core/modules/memman/MemManReal.cc
```diff
@@ -91,26 +92,20 @@ MemMan::Statistics MemManReal::getStatistics() {
...
+       std::lock_guard<std::mutex> lg(hanMutex);
        auto it = hanCache.find(handle);
        if (it != hanCache.end() && it->second->isOwner(_memory)) {
-          status = it->second->status();
-          hanMutex.unlock();
-          return status;
+          return it->second->status();
        }
-       hanMutex.unlock();
      }
```
abh3: Simply editorial -- lot's of braces for one line statements, eh?

core/modules/qdisp/QdispPool.h
```diff
@@ -152,8 +151,8 @@ class QdispPool {
 
 private:
     /// The default priority queue is meant for pool control commands.
-    PriorityQueue::Ptr _prQueue = std::make_shared<PriorityQueue>(100,1); // default (lowest) priority.
-    util::ThreadPool::Ptr _pool{util::ThreadPool::newThreadPool(30, _prQueue)};
+    PriorityQueue::Ptr _prQueue = std::make_shared<PriorityQueue>(100, 1, 5); // default (lowest) priority.
+    util::ThreadPool::Ptr _pool{util::ThreadPool::newThreadPool(1200, _prQueue)};
 };
```
abh3: OK what's the difference between comment with three slashes and ones with only two?

core/modules/util/Timer.cc
```diff
@@ -58,4 +67,65 @@ std::ostream & operator<<(std::ostream & os, Timer const & timer) {
...
+    timeHeld.start();
+}
+
+LockGuardTimed::~LockGuardTimed() {
+    _mtx.unlock();
+    timeHeld.stop();
+    LOGS(_log, LOG_LVL_DEBUG, "lockTime " << _note << " toLock=" << timeToLock.getElapsed() <<
+                              " held=" << timeHeld.getElapsed());
+}
+
```
abh3: You might want to consider simply doing a timed mutex. Then you can use it in more than just a "gaurd" context.

core/modules/wsched/testSchedulers.cc
```diff
@@ -473,7 +477,6 @@ BOOST_AUTO_TEST_CASE(BlendScheduleTest) {
 
     BOOST_CHECK(f.blend->ready() == false);
     BOOST_CHECK(f.blend->calcAvailableTheads() == 5);
-
```
abh3: I sort of like this white space separation.

admin/templates/configuration/etc/xrdssi.cnf
iagaponenko: This change is already in the configuration file :-) ..unless I misinterpreted your comment.

admin/templates/configuration/etc/xrdssi.cnf
iagaponenko: This change is already in the configuration file :-) ..unless I misinterpreted your comment.

admin/templates/configuration/etc/xrdssi.cnf
abh3: Ah, I misinterpreted the file type. It is a config file, OK.

core/modules/memman/MemFile.cc
jgates108: The addition of _mlockFileMutex is to reduce the time _fileMutex is held. There are times when the call to memLock can take more than a minute, and having _fileMutex held blocks calls to MemFile::memMap, which causes serious issues with wsched. This is especially the case when a particular file has already been mapped to memory and it is stuck on _fileMutex waiting for an unrelated mlock call.

core/modules/memman/MemFile.h
jgates108: If _fileMutex gets held for a long period of time, it wedges the scheduler when it checks to see if a file has been mapped. This is terribly painful when a file that has already been mapped is blocked waiting for a long file lock, and this in turn blocks any other task (including interactive tasks) from being started. 

It looks like having 2 threads working their way through MemFile::memLock() at the same time would have some dangerous race conditions, so _mlockFileMutex keeps that from happening.

core/modules/memman/MemManReal.cc
jgates108: Perhaps, I like having them whenever the entire if statement doesn't fit on 1 line.

core/modules/util/Timer.cc
jgates108: The point of this is just have something easy to drop into the code to replace a regular lock_guard and see how long it is taking to acquire the lock and how long it is held. Trying to add other functionality (say like that of unique_lock) made it get much more complicated. I'd rather leave it at this until there's a demonstrated need.

core/modules/qdisp/QueryRequest.cc
```diff
@@ -66,41 +67,43 @@ class QueryRequest::AskForResponseDataCmd : public PriorityCommand {
     void action(util::CmdData *data) override {
         // If everything is ok, call GetResponseData to have XrdSsi ask the worker for the data.
         {
+            util::InstanceCount _ic{"1-act-AskForResponseDataCmd"};
```
iagaponenko: Why this local variable starts with underscore? I think this naming convention is reserved for the non-public member variables of classes and structs.

core/modules/qdisp/QueryRequest.cc
```diff
@@ -109,14 +112,17 @@ class QueryRequest::AskForResponseDataCmd : public PriorityCommand {
         // If more data needs to be sent, _processData will make a new AskForResponseDataCmd
         // object and queue it.
         {
+            util::InstanceCount _ic{"3-act-AskForResponseDataCmd&&&"};
```
iagaponenko: Same comment on the name of the local variable which should not start with underscore.

core/modules/qdisp/QueryRequest.cc
```diff
@@ -66,41 +67,43 @@ class QueryRequest::AskForResponseDataCmd : public PriorityCommand {
...
             std::vector<char>& buffer = jq->getDescription()->respHandler()->nextBuffer();
-            LOGS(_log, LOG_LVL_DEBUG, _idStr << " Asking for GetResponseData size=" << buffer.size());
+            LOGS(_log, LOG_LVL_DEBUG, _idStr << " AskForResp GetResponseData size=" << buffer.size());
             qr->GetResponseData(&buffer[0], buffer.size());
         }
 
         // Wait for XrdSsi to call ProcessResponseData with the data,
         // which will notify this wait with a call to receivedProcessResponseDataParameters.
         {
+            util::InstanceCount _ic{"2-act-AskForResponseDataCmd"};
```
iagaponenko: Same comment on the name of the local variable which should not start with underscore.

core/modules/qdisp/QueryRequest.cc
```diff
@@ -278,18 +285,45 @@ bool QueryRequest::_importStream(JobQuery::Ptr const& jq) {
...
+    } else if (rating <= proto::ScanInfo::Rating::SLOW) {
+        if (not _largeResult) {
+            _qdispPool->queCmd(cmd, 3);
+        }
+    } else {
+        _qdispPool->queCmd(cmd, 6);
+    }
+
+
+    /* &&&
```
iagaponenko: Why leaving this commented code block here? Does it still have any use?

core/modules/qdisp/QdispPool.h
```diff
@@ -169,7 +169,7 @@ class QdispPool {
 private:
     /// The default priority queue is meant for pool control commands.
     PriorityQueue::Ptr _prQueue = std::make_shared<PriorityQueue>(100, 1, 5); // default (lowest) priority.
-    util::ThreadPool::Ptr _pool{util::ThreadPool::newThreadPool(35, _prQueue)};
+    util::ThreadPool::Ptr _pool{util::ThreadPool::newThreadPool(150, _prQueue)};
```
iagaponenko: Same suggestion as stated earlier - make the numbers configurable.

core/modules/wdb/QueryRunner.cc
```diff
@@ -126,6 +126,18 @@ void QueryRunner::_setDb() {
...
+
+std::mutex totalMemWaitMtx;
+double totalMemWaitTime(0);
+double memWaitCallCount(0);
+std::atomic<uint64_t> lessThan1(0);
+std::atomic<uint64_t> lessThan5(0);
+std::atomic<uint64_t> lessThan10(0);
+std::atomic<uint64_t> lessThan20(0);
+std::atomic<uint64_t> lessThan40(0);
+std::atomic<uint64_t> over40(0);
```
iagaponenko: This set of variables could be turned into a vector of atomics as explained in:
https://stackoverflow.com/questions/13193484/how-to-declare-a-vector-of-atomic-in-c?lq=1

This may add more flexibility to the code and make it significantly smaller (provided that wrapper class around the atomic is put into a separate header sitting in module utility::).

core/modules/wsched/BlendScheduler.h
```diff
@@ -119,13 +119,14 @@ class BlendScheduler : public wsched::SchedulerBase {
...
     int applyAvailableThreads(int tempMax) override { return tempMax;} //< does nothing
 
-    void setFlagReorderScans() { _flagReorderScans = true; }
     int calcAvailableTheads();
 
     bool isScanSnail(SchedulerBase::Ptr const& scan);
     int moveUserQueryToSnail(QueryId qId, SchedulerBase::Ptr const& source);
     int moveUserQuery(QueryId qId, SchedulerBase::Ptr const& source, SchedulerBase::Ptr const& destination);
 
+    void setPrioritizeByInFlight(bool val) { _prioritizeByInFlight = val; }
```
iagaponenko: Is this thread safe?

core/modules/wsched/testSchedulers.cc
```diff
@@ -428,6 +428,10 @@ struct SchedFixture {
                 std::chrono::seconds(_examineAllSleep), 5);
         blend = std::make_shared<wsched::BlendScheduler>("blendSched", queries, maxThreads,
                 group, scanSlow, scanSchedulers);
+        group->setPosition(0);
```
iagaponenko: Why should schedules know about their positions in the list of schedulers? This may result in a convoluted logic because usually elements of a collection should not be aware about any details of their host collections.

core/modules/qdisp/QdispPool.h
```diff
@@ -116,30 +119,43 @@ class QdispPool {
     typedef std::shared_ptr<QdispPool> Ptr;
 
     QdispPool() {
-        _prQueue->addPriQueue(0,1);  // Highest priority queue
-        _prQueue->addPriQueue(1,1);  // High priority queue
-        _prQueue->addPriQueue(2,9); // Normal priority queue
-        _prQueue->addPriQueue(3,3);  // Low priority queue
+        // Numbers are based on 35 threads in the pool. Large results
```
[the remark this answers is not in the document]
jgates108: These should definitely be set in a configuration file, as mentioned in the TODO below. This ticket is already sizable and there are other items that should be in configuration files, so I've added these to DM-10237.

core/modules/qdisp/QueryRequest.cc
```diff
@@ -278,18 +286,31 @@ bool QueryRequest::_importStream(JobQuery::Ptr const& jq) {
...
+        if (_largeResult) {
+            _qdispPool->queCmd(cmd, 6);
         } else {
-            LOGS(_log, LOG_LVL_DEBUG, _jobIdStr << " queueing priority norm");
-            _qdispPool->queCmdNorm(_askForResponseDataCmd);
+            _qdispPool->queCmd(cmd, 3);
+        }
+    } else if (rating <= proto::ScanInfo::Rating::SLOW) {
+        if (not _largeResult) {
+            _qdispPool->queCmd(cmd, 4);
```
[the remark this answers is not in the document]
jgates108: Good catch, fixed.

core/modules/util/Timer.h
```diff
@@ -63,6 +65,47 @@ struct Timer {
...
+
+
+/// This class is useful for getting an idea of how long something usually takes.
+/// It is also fairly easy to locate an abnormally long call in the log by
+/// searching for the first instance of a particular histogram value.
+class TimerHistogram {
+public:
+    struct bucket {
+        bucket(double maxV) : maxVal(maxV) {}
+        double maxVal;
```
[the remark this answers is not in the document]
jgates108: -DBL_MAX or just 0 would be more appropriate (DBL_MIN is the smallest possible positive value). For this struct to work as intended, maxVal really needs to be defined by the user as the value will be added to the next bucket if it is larger than maxVal.  I made it a proper class with _maxVal being private and deleting the default constructor.

core/modules/wdb/QueryRunner.cc
```diff
@@ -145,7 +150,12 @@ bool QueryRunner::runQuery() {
     }
 
     // Wait for memman to finish reserving resources. This can take several seconds.
+    util::Timer memTimer;
+    memTimer.start();
     _task->waitForMemMan();
+    memTimer.stop();
+    auto logMsg = memWaitHisto.addTime(memTimer.getElapsed(), _task->getIdStr());
```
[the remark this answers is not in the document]
jgates108: I think I want the values in the histogram to be accurate even if logging is not on.

core/modules/wsched/BlendScheduler.cc
```diff
@@ -108,27 +113,45 @@ BlendScheduler::~BlendScheduler() {
...
 void BlendScheduler::_sortScanSchedulers() {
-    auto greaterThan = [](SchedulerBase::Ptr const& a, SchedulerBase::Ptr const& b)->bool {
-        // Experiment of sorts, priority depends on number of Tasks in each scheduler.
-        auto aVal = a->getPriority() + a->getUserQueriesInQ();
-        auto bVal = b->getPriority() + b->getUserQueriesInQ();
-        return aVal > bVal;
+    auto lessThan = [this](SchedulerBase::Ptr const& a, SchedulerBase::Ptr const& b) -> bool {
+        // group scheduler is always first
+        if (a == _group) return true;
+        if (b == _group) return false;
```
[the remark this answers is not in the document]
jgates108: _group is the interactive query scheduler. Other scedulers are the ScanSchedulers (fast, medium, slow, and snail).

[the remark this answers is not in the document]
iagaponenko: My general comment (beyond a scope of this particular ticked, more like a "wish list") on the schedulers is that they are presently manually coded. It would be nice to have a generic implementation which would be externally configured. The configuration would define:
- schedulers (identified by some names like "group", "fast", "slow", etc.)
- priorities of the schedulers
- some optional parameters of the (like preemptive-ness, etc.)
- other constrains, like resources allowed to be used by each scheduler (numbers of queries, threads, memory, etc.)

Then we could experiment with different configuration options to get the most optimal performance w/o recoding Qserv each time we wanted to try something out.

core/modules/wsched/BlendScheduler.cc
```diff
@@ -270,37 +306,58 @@ bool BlendScheduler::_ready() {
...
+    double totalTimeHeld = 0.0;
+    bool ready = false;
+    {
+        timeToLock.start();
+        std::unique_lock<std::mutex> lock(util::CommandQueue::_mx);
+        timeToLock.stop();
+        timeHeld.start();
+        if (wait) {
+            // util::CommandQueue::_cv.wait(lock, [this](){return _ready();});
+            while (!_ready()) {
```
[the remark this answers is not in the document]
jgates108: _ready() is supposed to wait until there is something to do. If _ready() is false, there's nothing to do, and no harm in waiting forever.

core/modules/wsched/BlendScheduler.h
jgates108: I made it atomic, but it's really only meant to be called at startup where it may for systems with limited memory.

core/modules/wsched/testSchedulers.cc
jgates108: Position is a form of priority, the position kept within the scheduler is its default position. If the position is changed in the list for temporary conditions, the default position allows for it to be restored easily. I'll change the name for clarity.

core/modules/wsched/BlendScheduler.cc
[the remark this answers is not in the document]
jgates108: Actually, they are all ScanScheduler, except for the GroupScheduler, which has a very different behavior than the ScanScheduler. Most of the ScanScheduler parameters are already set via configuration files. They do have priorities, but priorities are used in a different way than default position (i.e. when position is not good enough). See wconfig/WorkerConfig.cc and xrdsvc::SsiService::SsiService(...).  The coding changes have to do with attempts to improve behavior in corner cases and none of them have really worked that well. The current case with _prioritizeByInFlight is that it does help with scheduler that aren't acquiring resources (memory for large tables) but it does hurt performance of the fast scheduler. I think it has some promise, but it's not useful enough as is and I'm not sure what it really needs.

PR #413, opened 2018-08-20
DM-15433: process antlr4 generated files with sed after build to convert `;;` to `;`

n8pease:


site_scons/site_tools/antlr4.py
```diff
@@ -97,8 +97,10 @@ def generate(env):
         ANTLR4_HSUFFIX='.h',
         ANTLR4_CCSUFFIX='.cpp',
         ANTLR4_TOKENSUFFIX='.tokens',
-        # Antlr command
-        ANTLR4_COM="$ANTLR4 $ANTLR4_FLAGS $SOURCES",
+        # Antlr4 command
+        # The following call to sed is a hack to replace double semi colons (`;;`) with a single semi
+        # colon in files generated by antlr4.
```
andy-slac: Could you also say here why do we have to replace `;;` with `;`?

PR #415, opened 2018-08-31
Fixed a logic error that could cause deadlock.

jgates108:


core/modules/qdisp/QueryRequest.cc
```diff
@@ -354,14 +354,21 @@ XrdSsiRequest::PRD_Xeq QueryRequest::ProcessResponseData(XrdSsiErrInfo const& eI
...
+    bool jobFinished = false;
     {
         std::lock_guard<std::mutex> lock(_finishStatusMutex);
-        if (_finishStatus != ACTIVE || jq == nullptr || jq->isQueryCancelled()) {
-            LOGS(_log, LOG_LVL_INFO, _jobIdStr << "ProcessResponseData job is inactive.");
-            // Something must have killed this job.
-            _errorFinish();
-            return XrdSsiRequest::PRD_Normal;
-        }
+        jobFinished = _finishStatus != ACTIVE || jq == nullptr;
```
andy-slac: I don't actually see `_jobQuery` being reset anywhere in the source, does it make sense to check for nullptr here?

jgates108: See QueryRequest::cleanup(). The move effectively sets it to null.

PR #424, opened 2018-10-22
DM-16166: improve user visible error reporting in qserv with antlr4 parser

n8pease:


core/modules/parser/SelectParser.cc
```diff
@@ -237,6 +237,50 @@ class Antlr2Parser : public AntlrParser {
 };
 
 
+class Antlr4ErrorStrategy : public antlr4::DefaultErrorStrategy {
+public:
+    Antlr4ErrorStrategy(std::string const& statement) : _statement(statement) {}
```
iagaponenko: A few problems with this class contract as per the LSST Style Guide:
- no explicitly declkared default constructor
- no copy constructor
- no assignment operator
- the existing constructor with only parameter must be declared as `explicit`

Here is the minimal interface for the class:
```
class Antlr4ErrorStrategy : public antlr4::DefaultErrorStrategy {
public:
    explicit Antlr4ErrorStrategy(std::string const& statement);
    Antlr4ErrorStrategy() = delete;
    Antlr4ErrorStrategy(Antlr4ErrorStrategy const&) = delete;
    Antlr4ErrorStrategy& operator=(Antlr4ErrorStrategy const&) = delete;
```

This should make the class much safer by 1) avoiding its use in the implicit conversions (hence `explicit`). 2) avoiding accidental construction with uninitialized state (via the default constructor), 3) avoiding potentially dangerous (if you have a state which can't be trivially copied) by the copy ctor or the assignment operator.

See:
https://developer.lsst.io/cpp/style.html#constructors-taking-one-argument-must-be-declared-as-explicit

core/modules/parser/SelectParser.cc
```diff
@@ -237,6 +237,50 @@ class Antlr2Parser : public AntlrParser {
...
+
+private:
+    void recover(antlr4::Parser *recognizer, std::exception_ptr e) override {
```
iagaponenko: Names of the private methods are supposed to begin with underscore (`void _recover()`) ...unless you're overriding a virtual method which isn't under your control (comes from ANTLR4)

core/modules/parser/SelectParser.cc
```diff
@@ -237,6 +237,50 @@ class Antlr2Parser : public AntlrParser {
...
+        LOGS(_log, LOG_LVL_ERROR, __FUNCTION__ <<
+                " antlr4 could not make a parse tree out of the input statement:" << _statement);
+        throw ParseException(std::string("Failed to instantiate query: \"") + _statement +
+                std::string("\""));
+    }
+
+    antlr4::Token* recoverInline(antlr4::Parser *recognizer) override {
```
iagaponenko: Names of the private methods are supposed to begin with underscore ...unless you're overriding a virtual method which isn't under your control (comes from ANTLR4)

core/modules/parser/SelectParser.cc
```diff
@@ -237,6 +237,50 @@ class Antlr2Parser : public AntlrParser {
...
+        LOGS(_log, LOG_LVL_ERROR, __FUNCTION__ <<
+                " antlr4 could not make a parse tree out of the input statement:" << _statement);
+        throw ParseException(std::string("Failed to instantiate query: \"") + _statement +
+                std::string("\""));
+    }
+
+    void sync(antlr4::Parser *recognizer) override {
```
iagaponenko: Names of the private methods are supposed to begin with underscore ...unless you're overriding a virtual method which isn't under your control (comes from ANTLR4)

core/modules/parser/SelectParser.cc
```diff
@@ -237,6 +237,50 @@ class Antlr2Parser : public AntlrParser {
...
+};
+
+
+class NonRecoveringQSMySqlLexer : public QSMySqlLexer {
+public:
+    NonRecoveringQSMySqlLexer(antlr4::CharStream *input, std::string const & statement)
+        : QSMySqlLexer(input), _statement(statement) {}
+
+private:
+    virtual void recover(const antlr4::LexerNoViableAltException &e) {
```
iagaponenko: Names of the private methods are supposed to begin with underscore ...unless you're overriding a virtual method which isn't under your control (comes from ANTLR4)

core/modules/parser/SelectParser.cc
```diff
@@ -237,6 +237,50 @@ class Antlr2Parser : public AntlrParser {
...
+
+    void sync(antlr4::Parser *recognizer) override {
+        // we want this function to be a no-op so we override it.
+    }
+
+    std::string const& _statement;
+};
+
+
+class NonRecoveringQSMySqlLexer : public QSMySqlLexer {
```
iagaponenko: Same comment as earlier on a lack of `explicit` for the main constructor, a lack of the default, copy  constructors and the assignment operator

core/modules/parser/ParseException.h
```diff
@@ -53,6 +53,8 @@ namespace parser {
 /// parse node to be bundled with the exception for greater context.
 class ParseException : public std::runtime_error {
 public:
+    virtual ~ParseException() = default;
```
iagaponenko: `~ParseException() override = default;`

core/modules/parser/QSMySqlListener.cc
```diff
@@ -123,30 +124,54 @@ void QSMySqlListener::exit##NAME(QSMySqlParser::NAME##Context* ctx) {\
...
+    LOGS(_log, LOG_LVL_TRACE, \
+        __FUNCTION__ << " is IGNORED, see warning in enter-function log entry, above."); \
 } \
 
 
-// A fucntion to assert certain condtions during runtime. In the case of a failure an adapter_exectuion_error
-// is raised with a context specific message.
+// A function to assert certain conditions during runtime. In the case of a failure a detailed message is
+// logged, for debugging, and an adapter_exectuion_error is raised with a message for the user.
 #define ASSERT_EXECUTION_CONDITION(CONDITION, MESSAGE, CTX) \
```
iagaponenko: This macro is rather big. Could it be implemented as an _inline_ function?

core/modules/parser/SelectParser.cc
n8pease: yes, it's an override of an antlr4 method :-(

n8pease: (and the same for the other ones below)

PR #427, opened 2018-11-01
DM-16218: verify QSMySQLListener adapters handle or throw grammar possibilities

n8pease:
adds a required function to Adapter called checkContext where subclasses
can/should check the context to be sure that non-context members are
either expected or null (or throw) so that unexpected tokens do not get
silently ignored. Members that return a context subclass typically do
not need to be checked because there will be a child adpater to handle
that context that will get called by the parse tree walker.

### DIFF
--- a/core/modules/parser/QSMySqlListener.cc
+++ b/core/modules/parser/QSMySqlListener.cc
@@ -499,7 +499,15 @@ public:
 
 class UnaryOperatorCBH : public BaseCBH {
 public:
-    virtual void handleUnaryOperator(string const & val) = 0;
+    enum OperatorType {
+        BANG,  // '!'
+        TILDE, // '~'
+        PLUS,  // '+'
+        MINUS, // '-'
+        NOT,   // NOT
+    };
+
+    virtual void handleUnaryOperator(OperatorType operatorType) = 0;
 };
 
 
@@ -545,6 +553,20 @@ class Adapter {
 public:
     Adapter() {}
     virtual ~Adapter() {}
+
+    // checkContext is called after construction and before the adapter is pushed onto the adapter stack.
+    // In this function Adapter subclasses should verify that all the token and terminalNode members of the
+    // context are either handled (as required or optional) or rejected if they are non-null.
+    // Tokens and terminal nodes are members of the context that are _not_ represented by a child context, so
+    // no callbacks will be received for them.
+    // Other contexts that are child contexts will be represented by another adapter. It is verified that
+    // child contexts are handled by verifying that their parent context is a handler for the child.
+    // In the implementations of child context, tokens and terminal nodes in the associated context should
+    // all be mentioned explicitly, those that are handled and required should be named in
+    // assertExecutionCondition calls (because they are required), those that are not handled should be named
+    // in assertNotSupported calls (because they are not supported), and those that are handled but are
+    // optional should be named in a comment that notes that the token or terminal node is optional.
+    virtual void checkContext() const = 0;
 
     // onEnter is called just after the Adapter is pushed onto the context stack
     virtual void onEnter() {}
@@ -618,6 +640,8 @@ public:
         msg << getTypeName(this) << "::" << function;
         msg << " messsage:\"" << message << "\"";
         msg << ", in query:" << getStatementString();
+        msg << ", string tree:" << getStringTree();
+        msg << ", tokens:" << getTokens();
         LOGS(_log, LOG_LVL_ERROR, msg.str());
         throw adapter_execution_error(
                 "Error parsing query, near \"" + getQueryString(ctx) + "\", " + message);
@@ -672,8 +696,17 @@ public:
         _selectStatement = selectStatement;
     }
 
+    void checkContext() const override {
+        // check for required tokens:
+        assertExecutionCondition(__FUNCTION__, _ctx->EOF() != nullptr,
+                "Missing context condition: EOF is null.", _ctx);
+        // optional:
+        // MINUSMINUS (ignored, it indicates a comment)
+    }
+
     void onEnter(QSMySqlParser::RootContext* ctx, QSMySqlListener const * const listener) {
         _ctx = ctx;
+        checkContext();
         qsMySqlListener = listener;
     }
 
@@ -703,6 +736,10 @@ public:
 
     void handleSelectStatement(shared_ptr<query::SelectStmt> const & selectStatement) override {
         _selectStatement = selectStatement;
+    }
+
+    void checkContext() const override {
+        // there is nothing to check
     }
 
     void onExit() override {
@@ -738,6 +775,10 @@ public:
         _groupByClause = groupByClause;
         _havingClause = havingClause;
         _distinct = distinct;
+    }
+
+    void checkContext() const override {
+        // there is nothing to check
     }
 
     void onExit() override {
@@ -797,6 +838,11 @@ public:
         _distinct = distinct;
     }
 
+    void checkContext() const override {
+        // required:
+        assertExecutionCondition(__FUNCTION__, _ctx->SELECT() != nullptr, "Context check failure.", _ctx);
+    }
+
     void onExit() override {
         lockedParent()->handleQuerySpecification(_selectList, _fromList, _whereClause, _orderByClause,
                 _limit, _groupByClause, _havingClause, _distinct);
@@ -845,6 +891,11 @@ public:
 
     void handleSelectExpressionElement(shared_ptr<query::ValueExpr> const & valueExpr) override {
         SelectListFactory::addValueExpr(_selectList, valueExpr);
+    }
+
+    void checkContext() const override {
+        // optional:
+        // star
     }
 
     void onExit() override {
@@ -934,6 +985,19 @@ public:
         _groupByClause->addTerm(query::GroupByTerm(valueExpr, ""));
     }
 
+    void checkContext() const override {
+        // required:
+        assertExecutionCondition(__FUNCTION__, _ctx->FROM() != nullptr, "Context check failure.", _ctx);
+        // optional:
+        // *WHERE();
+        // *GROUP();
+        // *BY();
+        // *HAVING();
+        // not supported:
+        assertNotSupported(__FUNCTION__, _ctx->WITH() == nullptr, "WITH is not supported", _ctx);
+        assertNotSupported(__FUNCTION__, _ctx->ROLLUP() == nullptr, "ROLLUP is not supported", _ctx);
+    }
+
     void onExit() override {
         shared_ptr<query::FromList> fromList = make_shared<query::FromList>(_tableRefList);
         lockedParent()->handleFromClause(fromList, _whereClause, _groupByClause, _havingClause);
@@ -967,6 +1031,10 @@ public:
         _tableRefList->push_back(tableRef);
     }
 
+    void checkContext() const override {
+        // nothing to check.
+    }
+
     void onExit() override {
         lockedParent()->handleTableSources(_tableRefList);
     }
@@ -997,6 +1065,10 @@ public:
 
     void handleNaturalJoin(shared_ptr<query::JoinRef> const & joinRef) override {
         _joinRefs.push_back(joinRef);
+    }
+
+    void checkContext() const override {
+        // nothing to check
     }
 
     void onExit() override {
@@ -1035,6 +1107,13 @@ public:
         _alias = string;
     }
 
+    void checkContext() const override {
+        // optional:
+        // AS();
+        // not supported:
+        assertNotSupported(__FUNCTION__, _ctx->PARTITION() == nullptr, "PARTITION is not supported", _ctx);
+    }
+
     void onExit() override {
         shared_ptr<query::TableRef> tableRef = make_shared<query::TableRef>(_db, _table, _alias);
         lockedParent()->handleAtomTableItem(tableRef);
@@ -1059,6 +1138,10 @@ public:
         lockedParent()->handleTableName(uidlist);
     }
 
+    void checkContext() const override {
+        // nothing to check
+    }
+
     void onExit() override {}
 
     string name() const override { return getTypeName(this); }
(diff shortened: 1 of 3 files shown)


core/modules/parser/QSMySqlListener.cc
```diff
@@ -546,6 +554,14 @@ class Adapter {
     Adapter() {}
     virtual ~Adapter() {}
```
iagaponenko: Perhaps:
`virtual ~Adapter() = default;`

core/modules/parser/QSMySqlListener.cc
```diff
@@ -672,8 +690,17 @@ class RootAdapter :
         _selectStatement = selectStatement;
     }
 
+    void checkContext() const override {
+        // required:
+        assertExecutionCondition(__FUNCTION__, _ctx->EOF() != nullptr,
+                "Missing context condition: EOF is null.", _ctx);
+        // optional:
+        // MINUSMINUS()
```
iagaponenko: Leftover from the testing/prototyping?

core/modules/parser/QSMySqlListener.cc
```diff
@@ -672,8 +690,17 @@ class RootAdapter :
...
+    }
+
     void onEnter(QSMySqlParser::RootContext* ctx, QSMySqlListener const * const listener) {
```
iagaponenko: Why the second `const` before `listener`?

core/modules/parser/QSMySqlListener.cc
```diff
@@ -797,6 +832,12 @@ class QuerySpecificationAdapter :
         _distinct = distinct;
     }
 
+    void checkContext() const override {
+        // required:
```
iagaponenko: What's a meaning of this 'required:'? I've seen it used in other similar methods.

core/modules/parser/QSMySqlListener.cc
```diff
@@ -847,6 +888,11 @@ class SelectElementsAdapter :
         SelectListFactory::addValueExpr(_selectList, valueExpr);
     }
 
+    void checkContext() const override {
+        // optional:
+        // antlr4::Token *star;
```
iagaponenko: Why leave the commented code here? Is it planned to be used?

core/modules/parser/QSMySqlListener.cc
```diff
@@ -934,6 +980,19 @@ class FromClauseAdapter :
         _groupByClause->addTerm(query::GroupByTerm(valueExpr, ""));
     }
 
+    void checkContext() const override {
+        // required:
+        assertExecutionCondition(__FUNCTION__, _ctx->FROM() != nullptr, "Context check failure.", _ctx);
+        // optional:
```
iagaponenko: Multiple comments lines are just confusing. If they're just temporarily disabled then, perhaps this should be explained.

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1035,6 +1102,13 @@ class AtomTableItemAdapter :
         _alias = string;
     }
 
+    void checkContext() const override {
+        // optional:
+        // antlr4::tree::TerminalNode *AS();
```
iagaponenko: Same comment as earlier on the commented code

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1082,6 +1160,11 @@ class FullIdAdapter :
         }
     }
 
+    void checkContext() const override {
+        // optional:
+        // antlr4::tree::TerminalNode *DOT_ID();
```
iagaponenko: Why the score is commented?

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1240,6 +1336,15 @@ class QservFunctionSpecAdapter :
...
     }
 
+    void checkContext() const override {
+        // required:
+        assertExecutionCondition(__FUNCTION__, _ctx->QSERV_AREASPEC_BOX() != nullptr ||
+            _ctx->QSERV_AREASPEC_CIRCLE() != nullptr ||
+            _ctx->QSERV_AREASPEC_ELLIPSE() != nullptr ||
+            _ctx->QSERV_AREASPEC_POLY() != nullptr ||
+            _ctx->QSERV_AREASPEC_HULL() != nullptr, "Context check failure.", _ctx);
+    }
```
iagaponenko: The code formatting could be made a bit better by:
```
assertExecutionCondition(
    __FUNCTION__,
    _ctx->QSERV_AREASPEC_BOX()     != nullptr ||
    _ctx->QSERV_AREASPEC_CIRCLE()  != nullptr ||
    _ctx->QSERV_AREASPEC_ELLIPSE() != nullptr ||
    _ctx->QSERV_AREASPEC_POLY()    != nullptr ||
    _ctx->QSERV_AREASPEC_HULL()    != nullptr,
    "Context check failure.",
    _ctx
);

```

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1432,6 +1545,25 @@ class ComparisonOperatorAdapter :
...
+            return;
+        } else if (_ctx->getText() == "<>") {
+            return;
+        } else if (_ctx->getText() == "!=") {
+            return;
+        } else if (_ctx->getText() == ">=") {
+            return;
+        } else if (_ctx->getText() == "<=") {
+            return;
+        }
```
iagaponenko: My personal approach to this kind of problems is to use the following pattern:
```
bool in(std::string const& what,
        std::vector<std::string>& const collection) {
   return collection.end() != std::find(collection.begin(), collection.end(), what);
}
```
Then you may write your method as:
```
if (in(_ctx->getText(),
       {"=", "!=", "<", "<=", ...})) return;
```
This utility function could be reusable many contexts.
I love functional programing :-)

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1491,6 +1629,13 @@ class OrderByExpressionAdapter :
         _valueExpr = valueExpr;
     }
 
+    void checkContext() const override {
+        // optional:
+        // antlr4::Token *order;
+        // antlr4::tree::TerminalNode *ASC();
+        // antlr4::tree::TerminalNode *DESC();
```
iagaponenko: Why these commented lines are here?

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1537,6 +1682,16 @@ class InnerJoinAdapter :
...
     }
 
+    void checkContext() const override {
+        // required:
+        assertExecutionCondition(__FUNCTION__, _ctx->JOIN() != nullptr, "Context check failure.", _ctx);
+        // optional:
+        // antlr4::tree::TerminalNode *INNER();
+        // antlr4::tree::TerminalNode *CROSS();
+        // antlr4::tree::TerminalNode *ON();
+        // antlr4::tree::TerminalNode *USING();
```
iagaponenko: Why these commented lines are here?

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1621,6 +1776,17 @@ class NaturalJoinAdapter :
...
     }
 
+    void checkContext() const override {
+        // required:
+        assertExecutionCondition(__FUNCTION__, _ctx->NATURAL() != nullptr, "Context check failure.", _ctx);
+        assertExecutionCondition(__FUNCTION__, _ctx->JOIN() != nullptr, "Context check failure.", _ctx);
+        // optional:
+        // antlr4::tree::TerminalNode *LEFT();
+        // antlr4::tree::TerminalNode *RIGHT();
+        // not supported:
```
iagaponenko: Why these commented lines are here?

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1645,7 +1811,10 @@ class SelectSpecAdapter :
 public:
     using AdapterT::AdapterT;
 
-    void onExit() override {
+    void checkContext() const override {
+        // optional:
+        // antlr4::tree::TerminalNode *DISTINCT();
```
iagaponenko: Why these commented lines are here?

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1731,6 +1906,11 @@ class SelectFunctionElementAdapter :
         _functionValueFactor = valueFactor;
     }
 
+    void checkContext() const override {
+        // optional:
+        // antlr4::tree::TerminalNode *AS();
```
iagaponenko: Why these commented lines are here?

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1765,6 +1945,14 @@ class SelectExpressionElementAdapter :
         _valueExpr = valueExpr;
     }
 
+    void checkContext() const override {
+        // optional:
+        // antlr4::tree::TerminalNode *AS();
```
iagaponenko: Why these commented lines are here?

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1867,6 +2095,13 @@ class NullNotnullAdapter :
 public:
     using AdapterT::AdapterT;
 
+    void checkContext() const override {
+        // required:
+        assertExecutionCondition(__FUNCTION__, _ctx->NULL_LITERAL() != nullptr || _ctx->NULL_SPEC_LITERAL() != nullptr, "Context check failure.", _ctx);
+        // optional:
+        // antlr4::tree::TerminalNode *NOT();
```
iagaponenko: Why these commented lines are here?

core/modules/parser/QSMySqlListener.cc
```diff
@@ -1893,6 +2128,11 @@ class SelectColumnElementAdapter :
         _alias = string;
     }
 
+    void checkContext() const override {
+        // optional:
+        // antlr4::tree::TerminalNode *AS();
```
iagaponenko: Why these commented lines are here?

core/modules/parser/QSMySqlListener.cc
```diff
@@ -2122,6 +2395,42 @@ class AggregateWindowedFunctionAdapter :
...
 
+    void checkContext() const override {
+        // optional:
+        // antlr4::tree::TerminalNode *AVG();
+        // antlr4::tree::TerminalNode *MAX();
+        // antlr4::tree::TerminalNode *MIN();
+        // antlr4::tree::TerminalNode *SUM();
+        // antlr4::tree::TerminalNode *COUNT();
+        // antlr4::Token *starArg = nullptr;
+        // FunctionArgContext *functionArg();
```
iagaponenko: Why these commented lines are here?

core/modules/parser/QSMySqlListener.cc
```diff
@@ -2774,8 +3161,26 @@ class UnaryOperatorAdapter :
...
 
 
+    void checkContext() const override {
+        assertNotSupported(__FUNCTION__, _ctx->NOT() == nullptr, "NOT is not supported.", _ctx);
+    }
+
     void onExit() override {
-        lockedParent()->handleUnaryOperator(_ctx->getText());
+        UnaryOperatorCBH::OperatorType operatorType;
+        if (_ctx->getText() == "!") {
```
iagaponenko: See y previous suggestion on the functional programming (function `in()`)

core/modules/qproc/testQueryAnaGeneral.cc
```diff
@@ -1012,7 +1012,7 @@ BOOST_AUTO_TEST_CASE(Case01_1083) {
         "from Source s join RefObjMatch rom using (objectId) "
         "join SimRefObject sro using (refObjectId) where isStar =1 limit 10;";
     // % is not valid for arithmetic in SQL92
-    char const expectedErr[] = "ParseException:Error parsing query, near \"%\"";
+    char const expectedErr[] = "ParseException:Error parsing query, near \"%\", Unhandled operator type:%";
```
iagaponenko: I understand this code was there for year. However, it looks a bit dangerous to depend on a very specific error message.

core/modules/qproc/testQueryAnaGeneral.cc
```diff
@@ -1067,7 +1067,7 @@ BOOST_AUTO_TEST_CASE(Case01_2006) {
     std::string stmt = "SELECT scisql_fluxToAbMag(uFlux_PS) "
         "FROM   Object WHERE  (objectId % 100 ) = 40;";
     // % is not a valid arithmetic operator in SQL92.
-    char const expectedErr[] = "ParseException:Error parsing query, near \"%\"";
+    char const expectedErr[] = "ParseException:Error parsing query, near \"%\", Unhandled operator type:%";
```
iagaponenko: I understand this code was there for year. However, it looks a bit dangerous to depend on a very specific error message.

core/modules/parser/QSMySqlListener.cc
n8pease: const pointer to a const object

core/modules/parser/QSMySqlListener.cc
n8pease: there's a description in the comments where checkContext is declared pure virtual in the Adapter base class. I think it can be improved though, I'll add to it.

core/modules/parser/QSMySqlListener.cc
n8pease: for this and the rest; see the description of checkContext.

core/modules/parser/QSMySqlListener.cc
n8pease: I'm not very much a fan of vertical formatting like that.

core/modules/qproc/testQueryAnaGeneral.cc
n8pease: it's testing for an expected error message; if it changes I want to know.

core/modules/qproc/testQueryAnaGeneral.cc
n8pease: (same)

core/modules/parser/QSMySqlListener.cc
iagaponenko: I'd argue that eliminating the second `const` will not change anything in the application.  The only difference only difference would be that the second `const` disallows modifying the variable within the body of the function. This change won't have any side effects beyond the method.
However, things would look totally different for:
```
void onEnter(...QSMySqlListener const ** const listener) 
```
In this case you do want to have the second `const` to prevent replacing the original pointer's value with a new object assigned within the method.

PR #428, opened 2018-11-06
DM-16407: Like should not be the same as NOT LIKE

n8pease:


core/modules/ccontrol/testCControl.cc
```diff
@@ -615,6 +619,37 @@ BOOST_DATA_TEST_CASE(antlr_compare, QUERIES, query) {
...
+    auto orTerm = std::dynamic_pointer_cast<query::OrTerm>(whereClauseRef.getRootTerm());
+    auto andTerm = std::dynamic_pointer_cast<query::AndTerm>(orTerm->_terms[0]);
+    auto boolFactor = std::dynamic_pointer_cast<query::BoolFactor>(andTerm->_terms[0]);
+    auto likePredicate = std::dynamic_pointer_cast<query::LikePredicate>(boolFactor->_terms[0]);
+    likePredicate->hasNot = true;
+    // now the the LIKE and the NOT LIKE selectStatements should be exactly the same:
+    BOOST_REQUIRE_EQUAL(*selectStatementWithNot, *selectStatementWithoutNot);
+}
+
+
```
iagaponenko: I'm a bit confused here. The comment at the beginning of this test case is mentioning ANTLR2, meanwhile I'm only seeing ANTLR4 when instantiating select statements by `parser::SelectParser::newInstance`. How ANTLR2 is supposed to be tested? Is there any internal switch in the implementation of `parser::SelectParser::newInstance` which would return an older version of the parser?

n8pease: I added more comments to the code, hopefully it's clearer: we can't compare the antlr4-generated IR to the antlr2-generated IR because antlr2 couldn't parse NOT LIKE. so we create the Like statement, and change the NOT var from 0 to 1 (to make it NOT, instead of not-NOT ;-) )

PR #429, opened 2018-11-17
DM-16532: mysql proxy cannot do select on escaped schema/table name

n8pease:


core/modules/parser/SelectParser.cc
```diff
@@ -366,6 +366,13 @@ SelectParser::Ptr SelectParser::newInstance(std::string const& statement, AntlrV
     return std::shared_ptr<SelectParser>(new SelectParser(statement, v));
 }
 
+std::shared_ptr<query::SelectStmt> SelectParser::makeSelectStmt(
+        std::string const& statement, AntlrVersion v) {
+    auto parser = newInstance(statement, v);
+    parser->setup();
```
jgates108: Should the setup call be done in newInstance()?

core/modules/ccontrol/testCControl.cc
```diff
@@ -619,36 +622,82 @@ BOOST_DATA_TEST_CASE(antlr_compare, QUERIES, query) {
...
+        : query(iQuery)
+        , compQuery(iCompQuery)
+        , serializedQuery(iSerializedQuery)
+        , modFunc(iModFunc)
+    {}
+
+    std::string query;
+    std::string compQuery;
+    std::string serializedQuery;
+    std::function<void(query::SelectStmt::Ptr const)> modFunc;
```
jgates108: Some comments would probably make this easier to understand. 'query' - query to test. 'compQuery' - antr2 comparison query. 'serializedQuery' - expected reconstruction.  'modFunc' -  makes the changes to compQuery IR so it should be equivalent to 'query'.

core/modules/parser/QSMySqlListener.cc
```diff
@@ -2056,30 +2056,33 @@ class SimpleIdAdapter :
 
 
 class DottedIdAdapter :
-        public AdapterT<DottedIdCBH, QSMySqlParser::DottedIdContext> {
+        public AdapterT<DottedIdCBH, QSMySqlParser::DottedIdContext>,
```
jgates108: A brief explanation of what this class does or its purpose could be useful.

core/modules/parser/QSMySqlListener.cc
```diff
@@ -2165,6 +2168,10 @@ class UidAdapter :
                    "If value is not set by callback then one of the terminal nodes should be populated.",
                     _ctx);
             _val = _ctx->getText();
+            assertExecutionCondition(__FUNCTION__, (_val.find('`') == 0) && (_val.rfind('`') == _val.size()-1),
+                    "REVERSE QUOTE values should begin and end with a backtick(`).", _ctx);
+            _val.erase(_val.begin());
+            _val.erase(--(_val.end()));
```
jgates108: I'm not familiar with the code. Is there a check to find back ticks in the middle of _val, or does that happen elsewhere in parsing? Is there a test for that?

n8pease: a backtick in the middle wouldn't appear here.

core/modules/parser/SelectParser.cc
n8pease: no, because that makes a new instance of the parser, this makes a SelectStmt, which requires calling setup on the parser. It's a little odd, I know. I don't really like the API of the SelectParser, requiring the call to setup. I haven't wanted to go ahead and refactor it yet, but I may get there at some point. It'll be more likely after we drop the antlr 2 parser entirely.

core/modules/parser/QSMySqlListener.cc
n8pease: There is an explanation of Adapter in the base class (this is an AdapterT, which is an Adapter). There are many specializations of Adapter (one per grammar element) and it would be repetitive to explain each one.

core/modules/ccontrol/testCControl.cc
n8pease: close, but not quite (compQuery will be parsed by antlr4, not antlr2, and then modified by modFunc, as you observed). So, point proven :-) I'll add comments.

PR #431, opened 2018-11-29
Tickets/dm 10425

iagaponenko:


core/modules/replica/tools/qserv-replica-master.cc
```diff
@@ -35,715 +40,322 @@
...
+            _replicationIntervalSec  (60),
+            _workerResponseTimeoutSec(60),
+            _workerEvictTimeoutSec   (3600),
+            _qservSyncTimeoutSec     (60),
+            _numReplicas             (0),
+            _numIter                 (0),
+            _purge          (false),
+            _forceQservSync (false),
+            _permanentDelete(false),
+            _failed         (false),
```
n8pease: would it be better to define these defaults in a cfg or with defines at the top of the file or something?

core/modules/replica/tools/qserv-replica-master.cc
```diff
@@ -35,715 +40,322 @@
...
+            " actions: check - fix-up - replicate - re-balance",
+            _replicationIntervalSec
+        ).option(
+            "worker-response-timeout",
+            "maximum number of seconds to wait before giving up"
+            " on worker probes when checking for workers",
+            _workerResponseTimeoutSec
+        ).option(
+            "worker-evict-timeout",
+            "the maximum number of seconds to allow troubles workers to recover"
```
n8pease: 'troubles' is not the right word here. 'failed', 'slow', 'workers having trouble'?

core/modules/replica/tools/qserv-replica-master.cc
```diff
@@ -35,715 +40,322 @@
...
+            "qserv-sync-force",
+            "the flag which would force Qserv workers to update their list of replicas"
+            " even if some of the chunk replicas were still in use by on-going queries."
+            " This affect replicas to be deleted from the workers during the synchronization"
+            " stages",
+            _forceQservSync
+        ).option(
+            "replicas",
+            "the minimal number of replicas when running the replication phase"
+            " This number if provided and if it's not 0 will override the corresponding value found"
```
n8pease: there could be trouble with the default value changing and this not getting updated. Maybe a `const int kNumDefaultWorkers` _is_ a good idea...

core/modules/replica/tools/qserv-replica-master.cc
```diff
@@ -35,715 +40,322 @@
...
+            _numReplicas
+        ).option(
+            "iter",
+            "the number of iterations (a value of 0 means running indefinitely)",
+            _numIter
+        ).flag(
+            "purge",
+            "The binary flag which if provided would enable the optional purge algorithm in"
+            " the end of each replication cycle in order to eliminate excess replicas which"
+            " might get created by algorithm ran earlier in the cycle.",
```
n8pease: 'an algorithm' or 'algorithms'.
If I may, this may be more clear:
The binary flag which, if provided, enables the 'purge' algorithm at the end of each replication cycle that eliminates excess replicas, which may have been created by algorithms executed earlier in the cycle.

core/modules/replica/tools/qserv-replica-master.cc
```diff
@@ -35,715 +40,322 @@
...
+            _purge
+        ).flag(
+            "permanent-worker-delete",
+            "The flag would trigger the permanent removal of the evicted workers"
+            " from the configuration of the Replication system. Please, use"
+            " this option with caution as it may result in loosing some Configuration"
```
n8pease: loosing->losing
(it's not getting looser, it's getting lost ;-) )
(I make the same mistake all the time)

core/modules/replica/tools/qserv-replica-master.cc
```diff
@@ -35,715 +40,322 @@
...
+            " data for the deleted workers",
```
n8pease: not 'for', I think. 'on' the deleted workers? 'from' the deleted workers?

core/modules/replica/tools/qserv-replica-master-http.cc
```diff
@@ -0,0 +1,380 @@
...
+            _workerResponseTimeoutSec(60),
+            _workerEvictTimeoutSec   (3600),
+            _qservSyncTimeoutSec     (60),
+            _numReplicas             (0),
+            _numIter                 (0),
+            _purge          (false),
+            _forceQservSync (false),
+            _permanentDelete(false),
+            _failed         (false),
+            _log(LOG_GET(logger)) {
```
n8pease: I think I mentioned in another file: it would be an improvement to have the defaults in a cfg area or file.

core/modules/replica/tools/qserv-replica-master-http.cc
```diff
@@ -0,0 +1,380 @@
...
+            " actions: check - fix-up - replicate - re-balance",
+            _replicationIntervalSec
+        ).option(
+            "worker-response-timeout",
+            "maximum number of seconds to wait before giving up"
+            " on worker probes when checking for workers",
+            _workerResponseTimeoutSec
+        ).option(
+            "worker-evict-timeout",
+            "the maximum number of seconds to allow troubles workers to recover"
```
n8pease: troubled

core/modules/replica/tools/qserv-replica-master-http.cc
```diff
@@ -0,0 +1,380 @@
...
+            _numReplicas,
+            _numIter,
+            _purge
+        );
+        _replicationThread->start();
+
+        _healthMonitorThread = HealthMonitorThread::create(
+            _controller,
+            [self] (ControlThread::Ptr const& ptr) {
+                self->_failed = true;
```
n8pease: I went looking for the thread-safety-ness of `_failed`; I see that in use it only transitions from `false` to `true`, which would cause an abort sequence to start, so that's cool. From an interface & sequencing safety perspective, it might be cool if this was a var that could only change state in one direction. 
`class Failer {
public: 
    void doFail() { _failed = true; }
    bool isFailed() const { return _failed; } 
private:
    atomic bool _failed{false};
};

core/modules/replica/tools/qserv-replica-master-http.cc
```diff
@@ -0,0 +1,380 @@
...
+        _deleteWorkerThread->stop();    // it's safe to call this method even if the thread is
+                                        // no longer running.
+
+        _deleteWorkerThread = nullptr;  // the object is no longer needed because it was
+                                        // created for a specific worker.
+
+        // Resume the normal replication sequence unless a catastrophic failure
+        // in the system has been detected
+        
+        if (not _failed) _replicationThread->start();
```
n8pease: I'm not seeing how it's possible to be confident in a `ControlThread` that `start` can't be called while `stop` is pending.

core/modules/replica/ControlThread.h
```diff
@@ -0,0 +1,368 @@
...
+    bool start();
+
+    /**
+     * Stop the thread if it's still running
+     *
+     * @return
+     *   'true' if the thread was already stopped at a time when this
+     *   method was called
+     */
+    bool stop();
```
n8pease: this is effectively `detach`, why not use that name? It would be more clear to those familiar with `std::thread`.

core/modules/replica/ControlThread.h
```diff
@@ -0,0 +1,368 @@
...
+    std::string name() const { return _name; }
+
+    // @return 'true' if the thread is running
+    bool isRunning() const { return _isRunning.load(); }
+
+    /**
+     * Start a subclass-supplied sequence of actions (virtual method 'run')
+     * within a new thread if it's not running. Note that the lifetime of
+     * the current object is guaranteed to be extended for the duration of
+     * the thread.
```
n8pease: what is 'the current object'? Is it the instance of this class? I don't see how the thread guarantees this object will be kept alive. Can you describe it to me please?

core/modules/replica/ControlThread.h
```diff
@@ -0,0 +1,368 @@
...
+    
+        while (numFinishedJobs != jobs.size()) {
+            if (stopRequested()) {
+                for (auto&& job: jobs) {
+                    job->cancel();
+                }
+                info(jobName + ": tracking aborted");
+                throw ControlThreadStopped();
+            }
+            blockPost.wait();
```
n8pease: I think it would be better to use the asio service to manage things that these thread controller classes are doing. Can we spend some time on it (with Fritz, who knows asio better than I do) during the meeting on Wednesday?

core/modules/replica/HealthMonitorThread.cc
```diff
@@ -0,0 +1,233 @@
...
+                controller(),
+                parentJobId,
+                [&numFinishedJobs](ClusterHealthJob::Ptr const& job) {
+                    ++numFinishedJobs;
+                }
+            )
+        );
+        jobs[0]->start();
+
+        track<ClusterHealthJob>("ClusterHealthJob", jobs, numFinishedJobs);
```
n8pease: Based on use of `track` it looks like `ClusterHealthJob` could have a `static std::string name() { return "ClusterHealthJob"; } that gets called inside the track Function? It doesn't make it any less lines of code but may clean up the api a little, if the job name is going to be the same as the class name.

core/modules/replica/HealthMonitorThread.cc
```diff
@@ -0,0 +1,233 @@
...
+
+        std::atomic<size_t> numFinishedJobs{0};
+        std::vector<ClusterHealthJob::Ptr> jobs;
+        jobs.emplace_back(
+            ClusterHealthJob::create(
+                _workerResponseTimeoutSec,
+                controller(),
+                parentJobId,
+                [&numFinishedJobs](ClusterHealthJob::Ptr const& job) {
+                    ++numFinishedJobs;
```
n8pease: it seems awfully brave to me to be sharing a function variable with other threads. I see that this is the run function, so I think I can see it is the run loop of a thread, which maybe makes it better. But still, what if this thread exits, perhaps unexpectedly?
One alternative I can suggest is: in the lambda capture a `weak_ptr` to `this`, and have the body of the function lock the weak ptr and if it succeeds, call a function `incrementFinishedJobs()` (or increment the atomic var directly but I somewhat prefer the accessor function).

core/modules/replica/DatabaseServices.h
```diff
@@ -102,7 +103,7 @@ class DatabaseServices
      *
      * The Job::Option object is explicitly passed as a parameter to avoid
      * making a call back to the job which may create a deadlock because
-     * method saveState() is meant to be called by the jobs' implementations.
+     * method saveState() is meant to be called by the job's implementations.
```
n8pease: it strikes me as a little strange that there are overloads of `saveState` that each have different calling requirements (functions with different API should be named differently? ...maybe they should be named differently, `saveControllerState` and `saveJobState`?)

core/modules/replica/DatabaseServices.h
```diff
@@ -123,7 +124,7 @@ class DatabaseServices
      *
      * The Performance object is explicitly passed as a parameter to avoid
      * making a call back to the request which may create a deadlock because
-     * method saveState() is meant to be called by the requests' implementations.
+     * method saveState() is meant to be called by the request's implementations.
```
n8pease: Does the request have multiple implementations? ("request's implementations")
or a single implementation ("request's implementation") (probably this is the one you want?)
ore are there different requests, with different implementations? "requests' implementations" (apostrophe after 'requests' is possessive of plural)

core/modules/replica/tools/qserv-replica-master.cc
iagaponenko: Yes, I was already considering moving these parameters into the Configuration service (and eventually - into the Configuration database).

core/modules/replica/tools/qserv-replica-master.cc
iagaponenko: 👍

core/modules/replica/tools/qserv-replica-master.cc
iagaponenko: The value of this parameter has nothing to do with the number of workers (as per your suggestion for `kNumDefaultWorkers`). Though, I can see your concern - putting a fixed number 0 may not be a good idea here. I will think about how to rephrase the parameter description to make it less confusing.

core/modules/replica/tools/qserv-replica-master.cc
iagaponenko: 👍

core/modules/replica/tools/qserv-replica-master.cc
iagaponenko: 💯

core/modules/replica/tools/qserv-replica-master.cc
iagaponenko: 👍

core/modules/replica/tools/qserv-replica-master-http.cc
iagaponenko: 💯

core/modules/replica/tools/qserv-replica-master-http.cc
iagaponenko: 👍

core/modules/replica/tools/qserv-replica-master-http.cc
iagaponenko: Very interesting idea! I will definitively consider it.

core/modules/replica/tools/qserv-replica-master-http.cc
iagaponenko: ControlThread's methods *start* and *stop* have a thread-safe implementation. Note the *lock* on a mutex acquired by each method before attempting a state transition. Here is:
```
bool ControlThread::start() {
    util::Lock lock(_mtx, context() + "start");
    if (_isRunning.exchange(true)) {
        return true;
    }
    ..
}


bool ControlThread::stop() {
    util::Lock lock(_mtx, context() + "stop");
    if (not _isRunning) return true;
    _stopRequested = true;
    ....
}
```

core/modules/replica/ControlThread.h
```diff
@@ -0,0 +1,368 @@
...
+class ControlThread
+    :   public std::enable_shared_from_this<ControlThread> {
+
+public:
+
+    /// The pointer type for instances of the class
+    typedef std::shared_ptr<ControlThread> Ptr;
+
+    /// The function type for notifications on the abnormal termination of the threads
+    typedef std::function<void(Ptr)> CallbackType;
```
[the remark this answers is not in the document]
iagaponenko: 👍

core/modules/replica/ControlThread.h
iagaponenko: Well, class `ControlThread` is not just an instance of `std::thread`. This class uses the C++ threads to execute relevant operations. If you look at the implementation of this class then you'll see that method `ControlThread::start()` lunches `std::thread` (and makes it detached), and method `ControlThread::stop()` is attempting to terminate (gracefully) an instance of `std::thread`. So, tis choice of methods makes a sense.

core/modules/replica/ControlThread.h
iagaponenko: This class inherits from:
```
class ControlThread
    :   public std::enable_shared_from_this<ControlThread> {
```
An object of this class is owned by a shared pointer:
```
    typedef std::shared_ptr<ControlThread> Ptr;
```
When class's method `start` is called this method launches `std::thread` and passes a copy of the smart pointer `ControlThread::Ptr` to the object (self). This would guarantee that an instance of `ControlThread` will stay around for a duration of the launched thread (`std::thread`). Therefore it's safe for `std::thread` to call back an instance of `ControlThread`.

core/modules/replica/ControlThread.h
iagaponenko: Surely, we can discuss it! I guess, your question is about the threading oder of the Master Controller application. It's rather trivial:
- there is a (configurable) number of the BOOST ASIO threads which are used for asynchronous operations (completion of the network I/O requests, timers, scheduling delayed activities)
- there 1 dedicated thread launched by `ControlThread::start`  for each of the following subclasses of `ControlThread`: `HttpThread`, 'ReplicationThread`, 'HealthMonitorThread` and `DeleteWorkerThread`.
- and there is obviously the `main` thread

A pool of the BOOST ASIO threads is configured based on a scale of the Replication System's setup, hardware capabilities (the number of cores, memory) of a machine where the Master Controller runs.

The `ControlThread`s threads are managed differently. They need to be independent of the BOOST ASIO's threads. Besides, they're launched/stopped dynamically depending on a status of the Replication System. Note that BOOST ASIO's threads are always there. And those (the later) may be all taken by some on-going processing.

Does it makes things more clear to you? Ot it makes things even more confusing?  :-)

core/modules/replica/HealthMonitorThread.cc
iagaponenko: That'a a good suggestion. It will require some work though. Right now each subclass of class `Job` has its own unique name which is used 1) to identify the job in the log files, and 2) is stored in the database as a key identifying a type of the job. That name is passed to the base class `Job` by the subclass's constructor. Here is an example:
```
ClusterHealthJob::ClusterHealthJob(unsigned int timeoutSec,
                                   Controller::Ptr const& controller,
                                   std::string const& parentJobId,
                                   CallbackType const& onFinish,
                                   Job::Options const& options)
    :   Job(controller,
            parentJobId,
            "CLUSTER_HEALTH",
```
This name is reported by the base class's method:
```
std::string const& Job::name() const { return _name; }
```
Therefore, calling that method by method `track` could be an option. A problem here is that (as you have correctly noticed) I would need to add a static method to class `ClusterHealthJob` to return the very same name `CLUSTER_HEALTH`.

I will consider this possibility.

core/modules/replica/HealthMonitorThread.cc
iagaponenko: Thanks! That's the correct observation (regarding `numFinishedJobs`). I will reinforce the code to pass a shared pointer to `HealthMonitorThread` instead. This will guarantee the lifespan of `numFinishedJobs` for as long as an instance of `ClusterHealthJob` is around.

core/modules/replica/DatabaseServices.h
iagaponenko: That's a reason why C++ has overloading :-)  A contract of a method is defined by its name and a collection (and types) of its parameters.

core/modules/replica/DatabaseServices.h
iagaponenko: Yes, this comment sounds a bit confusing. I would rather remove it altogether than trying to explain too much in this context.

core/modules/replica/tools/qserv-replica-master.cc
iagaponenko: Fixed

core/modules/replica/ControlThread.h
iagaponenko: improved the documentation

core/modules/replica/DatabaseServices.h
iagaponenko: Done

core/modules/replica/HealthMonitorThread.cc
iagaponenko: Done

PR #433, opened 2018-12-13
DM-16432: verify we have unit tests and add if needed for arithmetic operators, logical operators, bit operators, and comparison operators

n8pease:


core/modules/ccontrol/testCControl.cc
```diff
@@ -686,6 +707,246 @@ static const std::vector<Antlr4CompareQueries> ANTLR4_COMPARE_QUERIES = {
...
+              }));
+        },
+        "SELECT objectId FROM Object WHERE (objectID|1)=1"
+    ),
+
+    // tests the >> operator
+    Antlr4CompareQueries(
+        "SELECT objectId from Object where objectID << 10 = 1",
+        "SELECT objectId from Object where objectID = 1",
+        [](query::SelectStmt::Ptr const & selectStatement) {
```
iagaponenko: Perhaps:
```
query::SelectStmt::Ptr const& selectStatement
```

core/modules/ccontrol/testCControl.cc
```diff
@@ -686,6 +707,246 @@ static const std::vector<Antlr4CompareQueries> ANTLR4_COMPARE_QUERIES = {
...
+              }));
+        },
+        "SELECT objectId FROM Object WHERE (objectID<<10)=1"
+    ),
+
+    // tests the << operator
+    Antlr4CompareQueries(
+        "SELECT objectId from Object where objectID >> 10 = 1",
+        "SELECT objectId from Object where objectID = 1",
+        [](query::SelectStmt::Ptr const & selectStatement) {
```
iagaponenko:
```
const& selectStatement
```

core/modules/ccontrol/testCControl.cc
```diff
@@ -686,6 +707,246 @@ static const std::vector<Antlr4CompareQueries> ANTLR4_COMPARE_QUERIES = {
...
+              auto boolFactor = std::dynamic_pointer_cast<query::BoolFactor>(andTerm->_terms[0]);
+              auto compPredicate = std::dynamic_pointer_cast<query::CompPredicate>(boolFactor->_terms[0]);
+              compPredicate->left = std::make_shared<query::ValueExpr>(query::ValueExpr::FactorOpVector({
+                  query::ValueExpr::FactorOp(
+                          query::ValueFactor::newColumnRefFactor(query::ColumnRef::newShared("", "", "objectID")),
+                          query::ValueExpr::BIT_XOR),
+                  query::ValueExpr::FactorOp(query::ValueFactor::newConstFactor("1"))
+              }));
+        },
+        "SELECT objectId FROM Object WHERE (objectID^1)=1"
```
iagaponenko:
```
SELECT objectId FROM Object WHERE (objectID ^ 1) = 1
```

core/modules/ccontrol/testCControl.cc
```diff
@@ -686,6 +707,246 @@ static const std::vector<Antlr4CompareQueries> ANTLR4_COMPARE_QUERIES = {
...
+        "select * from Filter where filterId > 1 AND filterId < 6",
+        [](query::SelectStmt::Ptr const & selectStatement) {
+            // flip the 'not' on the filterId boolFactor to make it 'not'
+            auto whereClauseRef = selectStatement->getWhereClause();
+            auto orTerm = std::dynamic_pointer_cast<query::OrTerm>(whereClauseRef.getRootTerm());
+            auto andTerm = std::dynamic_pointer_cast<query::AndTerm>(orTerm->_terms[0]);
+            auto boolFactor = std::dynamic_pointer_cast<query::BoolFactor>(andTerm->_terms[0]);
+            boolFactor->setHasNot(true);
+        },
+        "SELECT * FROM Filter WHERE NOT filterId>1 AND filterId<6"
```
iagaponenko: This would be easier to read:
```
SELECT * FROM Filter WHERE NOT filterId > 1 AND filterId < 6
```

core/modules/ccontrol/testCControl.cc
```diff
@@ -686,6 +707,246 @@ static const std::vector<Antlr4CompareQueries> ANTLR4_COMPARE_QUERIES = {
...
+        "select * from Filter where (filterId > 1 AND filterId < 6)",
+        [](query::SelectStmt::Ptr const & selectStatement) {
+            // flip the 'not' on the AndTerm to make it 'not'
+            auto whereClauseRef = selectStatement->getWhereClause();
+            auto orTerm = std::dynamic_pointer_cast<query::OrTerm>(whereClauseRef.getRootTerm());
+            auto andTerm = std::dynamic_pointer_cast<query::AndTerm>(orTerm->_terms[0]);
+            auto boolFactor = std::dynamic_pointer_cast<query::BoolFactor>(andTerm->_terms[0]);
+            boolFactor->setHasNot(true);
+        },
+        "SELECT * FROM Filter WHERE NOT(filterId>1 AND filterId<6)"
```
iagaponenko: This would be easier to read:
```
SELECT * FROM Filter WHERE NOT (filterId > 1 AND filterId < 6)
```

core/modules/parser/QSMySqlListener.cc
```diff
@@ -437,6 +464,13 @@ class FunctionArgCBH : public BaseCBH {
 };
 
 
+class NotExpressionCBH : public BaseCBH {
+public:
+    virtual void handleNotExpression(shared_ptr<query::BoolTerm> const & boolTerm,
```
iagaponenko:
```
const& boolTerm
```

core/modules/parser/QSMySqlListener.cc
```diff
@@ -525,13 +545,30 @@ class LogicalOperatorCBH : public BaseCBH {
...
 class MathOperatorCBH : public BaseCBH {
 public:
     enum OperatorType {
         SUBTRACT,
         ADD,
-        DIVIDE,
+        DIVIDE, // `/` operator
         MULTIPLY,
+        DIV,    // `DIV` operator
+        MOD,    // `MOD` operator
```
iagaponenko: Do you really need these trailing comments? I think everything is obvious w/o them.

core/modules/parser/QSMySqlListener.cc
```diff
@@ -3198,13 +3269,48 @@ class LogicalOperatorAdapter :
     }
 
     void onExit() override {
-        if (_ctx->AND() != nullptr) {
+        if (_ctx->AND() != nullptr || _ctx->getText() == "&&") {
```
iagaponenko: Even though, I understand what this is about, a difference between `AND` and `&&` still looks a bit confusing. Any chance the later be automatically converted into an object returned by the former?

core/modules/parser/QSMySqlListener.cc
```diff
@@ -3198,13 +3269,48 @@ class LogicalOperatorAdapter :
...
+            // Qserv IR is not set up to treat AND and && differently, up to and inlcuding that the AndTerm
+            // automatically serializes itself to "AND" (i.e. not to lower case "and" or any other form). If
+            // it becomes important to handle different forms of the lexical AND differently we can add it,
+            // but for now it seems unnecessary.
             lockedParent()->handleLogicalOperator(LogicalOperatorCBH::AND);
-        } else if (_ctx->OR() != nullptr) {
+        } else if (_ctx->OR() != nullptr || _ctx->getText() == "||") {
```
iagaponenko: Same question as for the above asked `AND` and `&&`

core/modules/query/Predicate.h
```diff
@@ -156,9 +163,9 @@ class InPredicate : public Predicate {
...
 public:
-    BetweenPredicate() {}
-    BetweenPredicate(ValueExprPtr iValue, ValueExprPtr iMinValue, ValueExprPtr iMaxValue)
-    : value(iValue), minValue(iMinValue), maxValue(iMaxValue) {}
+    BetweenPredicate() : hasNot(false) {}
+    BetweenPredicate(ValueExprPtr iValue, ValueExprPtr iMinValue, ValueExprPtr iMaxValue, bool iHasNot)
+    : value(iValue), minValue(iMinValue), maxValue(iMaxValue), hasNot(iHasNot) {}
     typedef std::shared_ptr<BetweenPredicate> Ptr;
 
     virtual ~BetweenPredicate() = default;
```
iagaponenko: Perhaps:
```
~BetweenPredicate() override = default;
```

core/modules/query/ValueExpr.cc
```diff
@@ -82,6 +82,14 @@ std::ostream& operator<<(std::ostream& os, const ValueExpr::FactorOp& factorOp)
...
     case ValueExpr::MULTIPLY: os << "MULTIPLY"; break;
     case ValueExpr::DIVIDE: os << "DIVIDE"; break;
+    case ValueExpr::DIV: os << "DIV"; break;
+    case ValueExpr::MODULO: os << "MODULO"; break;
+    case ValueExpr::MOD: os << "MOD"; break;
+    case ValueExpr::BIT_SHIFT_LEFT: os << "BIT_SHIFT_LEFT"; break;
+    case ValueExpr::BIT_SHIFT_RIGHT: os << ">BIT_SHIFT_RIGHT>"; break;
+    case ValueExpr::BIT_AND: os << "BIT_AND"; break;
+    case ValueExpr::BIT_OR: os << "BIT_OR"; break;
+    case ValueExpr::BIT_XOR: os << "BIT_XOR"; break;
```
iagaponenko: To improve the readability of these repeating patters one could align them like tgis:
```
case ValueExpr::DIV:             os << "DIV";               break;
case ValueExpr::MODULO:          os << "MODULO";            break;
case ValueExpr::MOD:             os << "MOD";               break;
case ValueExpr::BIT_SHIFT_LEFT:  os << "BIT_SHIFT_LEFT";    break;
case ValueExpr::BIT_SHIFT_RIGHT: os << ">BIT_SHIFT_RIGHT>"; break;
case ValueExpr::BIT_AND:         os << "BIT_AND";           break;
case ValueExpr::BIT_OR:          os << "BIT_OR";            break;
case ValueExpr::BIT_XOR:         os << "BIT_XOR";           break;
```
ow you can see that `BIT_SHIFT_RIGHT` is represented differently than others with extra symbols at both ends `">BIT_SHIFT_RIGHT>"`. Is this by design?

core/modules/query/ValueExpr.cc
```diff
@@ -296,6 +308,14 @@ void ValueExpr::render::applyToQT(ValueExpr const& ve) {
...
         case ValueExpr::MULTIPLY: _qt.append("*"); break;
         case ValueExpr::DIVIDE: _qt.append("/"); break;
+        case ValueExpr::DIV: _qt.append("DIV"); break;
+        case ValueExpr::MODULO: _qt.append("%"); break;
+        case ValueExpr::MOD: _qt.append("MOD"); break;
+        case ValueExpr::BIT_SHIFT_LEFT: _qt.append("<<"); break;
+        case ValueExpr::BIT_SHIFT_RIGHT: _qt.append(">>"); break;
+        case ValueExpr::BIT_AND: _qt.append("&"); break;
+        case ValueExpr::BIT_OR: _qt.append("|"); break;
+        case ValueExpr::BIT_XOR: _qt.append("^"); break;
```
iagaponenko: I'd recommend the following alignment here:
```
case ValueExpr::MINUS:           _qt.append("-");   break;
case ValueExpr::MULTIPLY:        _qt.append("*");   break;
case ValueExpr::DIVIDE:          _qt.append("/");   break;
case ValueExpr::DIV:             _qt.append("DIV"); break;
case ValueExpr::MODULO:          _qt.append("%");   break;
case ValueExpr::MOD:             _qt.append("MOD"); break;
case ValueExpr::BIT_SHIFT_LEFT:  _qt.append("<<");  break;
case ValueExpr::BIT_SHIFT_RIGHT: _qt.append(">>");  break;
case ValueExpr::BIT_AND:         _qt.append("&");   break;
case ValueExpr::BIT_OR:          _qt.append("|");   break;
case ValueExpr::BIT_XOR:         _qt.append("^");   break;
```

core/modules/query/ValueExpr.h
```diff
@@ -62,8 +62,12 @@ class ValueFactor;
 /// operators. No nesting is allowed yet.
 class ValueExpr {
 public:
-    ValueExpr();
-    enum Op {NONE=200, UNKNOWN, PLUS, MINUS, MULTIPLY, DIVIDE};
+    // in Op: DIVIDE is the `/` operator, "Division; quotient of operands" as specified by MySQL.
+    //        DIV is the `DIV` operator, "Division; integer quotient of operands" as specified by MySQL.
+    //        THE BIT_ values are bitwise operators: BIT_SHIFT_LEFT is <<, BIT_SHIFT_RIGHT is >>,
+    //        BIT_AND is &, BIT_OR is |, BIT_XOR is ^.
+    enum Op {NONE=200, UNKNOWN, PLUS, MINUS, MULTIPLY, DIVIDE, DIV, MOD, MODULO,
```
iagaponenko: Why does it need to start with `200`? Just curious.

core/modules/ccontrol/testCControl.cc
```diff
@@ -686,6 +707,246 @@ static const std::vector<Antlr4CompareQueries> ANTLR4_COMPARE_QUERIES = {
...
+              auto boolFactor = std::dynamic_pointer_cast<query::BoolFactor>(andTerm->_terms[0]);
+              auto compPredicate = std::dynamic_pointer_cast<query::CompPredicate>(boolFactor->_terms[0]);
+              compPredicate->left = std::make_shared<query::ValueExpr>(query::ValueExpr::FactorOpVector({
+                  query::ValueExpr::FactorOp(
+                          query::ValueFactor::newColumnRefFactor(query::ColumnRef::newShared("", "", "objectID")),
+                          query::ValueExpr::BIT_AND),
+                  query::ValueExpr::FactorOp(query::ValueFactor::newConstFactor("1"))
+              }));
+        },
+        "SELECT objectId FROM Object WHERE (objectID&1)=1"
```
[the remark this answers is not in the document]
thecaliforniasons: qserv serializes the IR without spaces, this is the same behavior as before I started working on the antlr4 parser. For now I'm going to leave it as is. If we want to address it we can do it in separate work.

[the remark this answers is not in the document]
n8pease: qserv serializes the intermediate representation without spaces, it's been this way since before I started working on the new parser. If we want to change this behavior it should be done as separate work.

These queries are not executed, this just tests for expected IR representation & serialization. So the column name need not match any of the 'normal' qserv column naming conventions, e.g. `objectId` vs `objectID`

core/modules/ccontrol/testCControl.cc
```diff
@@ -686,6 +707,246 @@ static const std::vector<Antlr4CompareQueries> ANTLR4_COMPARE_QUERIES = {
...
+              auto boolFactor = std::dynamic_pointer_cast<query::BoolFactor>(andTerm->_terms[0]);
+              auto compPredicate = std::dynamic_pointer_cast<query::CompPredicate>(boolFactor->_terms[0]);
+              compPredicate->left = std::make_shared<query::ValueExpr>(query::ValueExpr::FactorOpVector({
+                  query::ValueExpr::FactorOp(
+                          query::ValueFactor::newColumnRefFactor(query::ColumnRef::newShared("", "", "objectID")),
+                          query::ValueExpr::BIT_OR),
+                  query::ValueExpr::FactorOp(query::ValueFactor::newConstFactor("1"))
+              }));
+        },
+        "SELECT objectId FROM Object WHERE (objectID|1)=1"
```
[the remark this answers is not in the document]
n8pease: same as above; this is the expected serialized query.

core/modules/ccontrol/testCControl.cc
```diff
@@ -486,7 +489,24 @@ static const std::vector< std::string > QUERIES = {
     "select shortName from Filter where shortName LIKE 'Z'",
 
     // DM-16532; this verifies that dotted IDs work for table+column ids. quoted dotted IDs are tested in the antlr4 test, below.
-    "SELECT Source.sourceId, Source.objectId From Source WHERE Source.objectId IN (386942193651348) ORDER BY Source.sourceId;"
+    "SELECT Source.sourceId, Source.objectId From Source WHERE Source.objectId IN (386942193651348) ORDER BY Source.sourceId;",
```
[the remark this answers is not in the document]
n8pease: that's a good suggestion. I think this comment got put on the wrong line though? Or there's something I don't understand; I think this would be applied in `BOOST_DATA_TEST_CASE(antlr4_compare, ANTLR4_COMPARE_QUERIES, queryInfo)` around line 961, right?

core/modules/ccontrol/testCControl.cc
n8pease: this is the expected value from serialization.

core/modules/ccontrol/testCControl.cc
n8pease: this is the expected value from serialization.

core/modules/ccontrol/testCControl.cc
n8pease: this is the expected value from serialization.

core/modules/parser/QSMySqlListener.cc
n8pease: it seemed worth being explicit since there is also e.g. `MODULO`, the `%` operator in that enum.

core/modules/parser/QSMySqlListener.cc
n8pease: There is not a way to make "and", "AND", and "&&" return a lexical `AND` token. There are ways to  adjust the grammar so that these 3 strings could have their own grammar element which would then get its own Adapter subclass. The pattern would then be duplicated for `OR`. This seems like overkill. The logical operator types are defined as:
```
logicalOperator
    : AND | '&' '&' | XOR | OR | '|' '|'
```
This encapsulates the logic of the types of `AND` in this one location (and also for the types of `OR`), which seems good enough encapsulation for me. (Note, the `AND` token here is case-insensitive). If the difference between the text "and" and "&&" starts spreading then some more abstraction would be correct, but I'm pretty sure it won't happen.

core/modules/query/ValueExpr.h
n8pease: I don't know. It was that way before I started working on it. It was checked in by Daniel years ago, I can't find an explanation for it. I didn't want to change it without need.

core/modules/ccontrol/testCControl.cc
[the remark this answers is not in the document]
iagaponenko: This s just a general comment/suggestion, not pointing to any specific line of the unit test.

core/modules/ccontrol/testCControl.cc
iagaponenko: 👍

core/modules/parser/QSMySqlListener.cc
iagaponenko: 👍

core/modules/parser/QSMySqlListener.cc
iagaponenko: Got it. Thanks!

core/modules/query/ValueExpr.h
iagaponenko: Makes a sense.

core/modules/ccontrol/testCControl.cc
```diff
@@ -686,6 +707,246 @@ static const std::vector<Antlr4CompareQueries> ANTLR4_COMPARE_QUERIES = {
...
         "SELECT Source.sourceId,Source.objectId FROM Source WHERE Source.objectId IN(386942193651348) ORDER BY Source.sourceId"
     ),
+
+    // tests the null-safe equals operator
+    Antlr4CompareQueries(
+        "SELECT ra_PS FROM Object WHERE objectId<=>417857368235490",
+        "SELECT ra_PS FROM Object WHERE objectId = 417857368235490",
+        [](query::SelectStmt::Ptr const & selectStatement) {
+            // change the equals op to be the null safe equals op
+            auto whereClauseRef = selectStatement->getWhereClause();
```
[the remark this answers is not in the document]
n8pease: it doesn't really work to make `whereClauseRef` const, because it continues getting member objects, until finally an object is in fact changed, so the references/pointers getting there must be non-const.

core/modules/ccontrol/testCControl.cc
```diff
@@ -686,6 +707,246 @@ static const std::vector<Antlr4CompareQueries> ANTLR4_COMPARE_QUERIES = {
...
+            auto boolFactor = std::dynamic_pointer_cast<query::BoolFactor>(andTerm->_terms[0]);
+            auto betweenPredicate = std::dynamic_pointer_cast<query::BetweenPredicate>(boolFactor->_terms[0]);
+            betweenPredicate->hasNot = true;
+        }
+    ),
+
+    // tests the && operator.
+    // The Qserv IR converts && to AND as a result of the IR structure and how it serializes it to string.
+    Antlr4CompareQueries(
+        "select objectId, iRadius_SG, ra_PS, decl_PS from Object where iRadius_SG > .5 && ra_PS < 2 && decl_PS < 3;", // &&
```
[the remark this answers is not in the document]
n8pease: I've found when I come back to these tests later, and when I have to read tests that someone else wrote maybe years earlier, it helps to have come clues about what's going on, what's different, and if that difference was intentional. In this case it may be a little overly verbose, but I'd rather leave more bread crumbs than less.

core/modules/ccontrol/testCControl.cc
```diff
@@ -686,6 +707,246 @@ static const std::vector<Antlr4CompareQueries> ANTLR4_COMPARE_QUERIES = {
...
+        "select objectId, iRadius_SG, ra_PS, decl_PS from Object where iRadius_SG > .5 AND ra_PS < 2 AND decl_PS < 3;", // AND
+        nullptr,
+        "SELECT objectId,iRadius_SG,ra_PS,decl_PS FROM Object WHERE iRadius_SG>.5 AND ra_PS<2 AND decl_PS<3"
+    ),
+
+    // tests the || operator.
+    // The Qserv IR converts || to OR as a result of the IR structure and how it serializes it to string.
+    Antlr4CompareQueries(
+        "select objectId from Object where objectId < 400000000000000 || objectId > 430000000000000 ORDER BY objectId;", // ||
```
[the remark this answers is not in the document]
n8pease: they're arbitrary, except that they do reflect numbers that one may find for objectId in the qserv catalog. I'll leave a comment.

PR #437, opened 2018-12-19
Tickets/dm 16652

jgates108:


core/modules/loader/CentralClient.h
```diff
@@ -68,27 +69,20 @@ class KeyInfoData : public util::Tracker {
...
     ~CentralClient() override = default;
 
     std::string const& getHostName() const { return _hostName; }
     int getUdpPort() const { return _udpPort; }
     int getTcpPort() const { return 0; } ///< No tcp port at this time.
 
-    std::string getWorkerHostName() const { return _workerHostName; }
-    int getWorkerPort() const { return _workerPort; }
+    std::string getDefWorkerHost() const { return _defWorkerHost; }
+    int getDefWorkerPortUdp() const { return _defWorkerPortUdp; }
```
thecaliforniasons: It wasn't obvious to me that 'def' was 'default', but I think that's the case because now I see the comment near the member var. I'd probably prefer to not abbreviate Default, but if you want to abbreviate it I think the methods needed to be commented.

n8pease: It wasn't obvious to me that 'def' was 'default', but I think that's the case because now I see the comment near the member var. I'd probably prefer to not abbreviate Default, but if you want to abbreviate it I think the methods needed to be commented.

core/modules/loader/CentralWorker.h
```diff
@@ -253,14 +254,14 @@ class CentralWorker : public Central {
 
     std::atomic<bool> _shiftAsClientInProgress{false}; ///< True when shifting to or from right neighbor.
 
-    /// Shift if a node has 10% more than it's neighbor TODO config file DM-16652
-    double _thresholdNeighborShift{1.10};
+    /// Shift if a node has % more than it's neighbor (1.1 would be 10% more than neighbor)
+    double _thresholdNeighborShift;
```
n8pease: it appears the 1.1 this refers to has been removed. Maybe you still want to say something like "the percentage threshold is expressed as a decimal, so e.g. 110% of the neighbor would be 1.10"

core/modules/loader/CentralWorker.h
```diff
@@ -253,14 +254,14 @@ class CentralWorker : public Central {
...
 
-    /// Maximum number of keys to shift in one iteration.
-    /// An iteration would be transfer, insert, verify range. TODO config file  DM-16652
+    /// Maximum number of keys to shift in one iteration. 10000 may be reasonable.
```
n8pease: but 20000 would not? is there any more explanation or reason that can go here?

core/modules/loader/WorkerConfig.cc
```diff
@@ -0,0 +1,62 @@
...
+
+// LSST headers
+#include "lsst/log/Log.h"
+
+// Qserv headers
+#include "util/ConfigStore.h"
+#include "util/ConfigStoreError.h"
+
+namespace {
+LOG_LOGGER _log = LOG_GET("lsst.qserv.loader.WorkerConfig");
```
andy-slac: In other file logger name is "lsst.qserv.loader.Config", is WorkerConfig here intentional?

core/modules/loader/WorkerConfig.h
```diff
@@ -0,0 +1,89 @@
...
+namespace lsst {
+namespace qserv {
+namespace loader {
+
+/// A class for reading the configuration file for the worker which consists of
+/// a collection of key-value pairs and provide access functions for those values.
+///
+class WorkerConfig : public ConfigBase {
+public:
+    WorkerConfig(std::string configFileName)
```
andy-slac: const&

core/modules/loader/config/clientBad.cnf
```diff
@@ -0,0 +1,11 @@
...
+#
+[client]
+masterHost = 127.0.0.1
+masterPortUdp = 10042
+defWorkerHost = 127.0.0.1
+defWorkerPortUdp = 10043
+clientortUdp = 10050
+loopSleepTime = 100000
+doListMaxLookups = 99000 
+doListMaxInserts = 99000
```
andy-slac: Is the missing newline at EOF also intentionally malformed?

core/modules/loader/config/masterBad.cnf
```diff
@@ -0,0 +1,8 @@
+# This configuration file is intentionally incorrect for testing.
+#
+[master]
+portdp = 9876
+maxKeysPerWorker = 1050
+threadPoolSize = 10
+# Time to sleep in microseconds.
+loopSleepTime = 100000
```
andy-slac: Missing newline again

core/modules/loader/udpTest.cc
```diff
@@ -207,40 +218,84 @@ int main(int argc, char* argv[]) {
...
+    {
+        bool threw = false;
+        try {
+            ClientConfig workerCfg("core/modules/loader/config/clientBad.cnf");
+        } catch (ConfigErr const& e) {
+            threw = true;
+            LOGS(_log, LOG_LVL_INFO, "ClientConfig clientBad threw " << e.what());
+        }
+        if (not threw) {
+            LOGS(_log, LOG_LVL_ERROR, "ClinetConfig workerBad.cnf should have thrown!!");
```
andy-slac: clientBad.cnf

core/modules/loader/udpTest.cc
```diff
@@ -250,43 +305,66 @@ int main(int argc, char* argv[]) {
...
+        cCentral2A.start();
+    } catch (boost::system::system_error const& e) {
+        LOGS(_log, LOG_LVL_ERROR, "cCentral2A.start() failed e=" << e.what());
+        exit(-1);
+    }
     cCentral2A.run();
 
+    ClientConfig clientCfg3("core/modules/loader/config/client3.cnf");
+    LOGS(_log, LOG_LVL_INFO, "clientCfg3=" << clientCfg3);
+    CentralClient cCentral1B(ioServiceClient1A, ourHost, clientCfg3);
```
andy-slac: ioServiceClient1B?

core/modules/loader/ConfigBase.h
```diff
@@ -0,0 +1,120 @@
...
+private:
+    ConfigElement(std::string const& header, std::string const& key,
+                  bool required, std::string const& default_)
+            : _header(header), _key(key), _required(required), _default(default_) {}
+
+    std::string _header;      ///< name of the header for this element
+    std::string _key;         ///< name of the key under the header
+    std::string _value{""};   ///< value found in config or default
+    bool _required{true};     ///< required to be in config.
+    std::string _default{""}; ///< default value.
```
n8pease: I think the string default ctor sets it to `""`. Is there a benefit to passing `""` here? (and 2 lines above)

core/modules/loader/config/client1.cnf
```diff
@@ -0,0 +1,27 @@
...
+loopSleepTime = 100000
+
+# Maximum number of lookups that can be on the DoList at a time.
+doListMaxLookups = 99000 
+
+# Maximum number of inserts that can be on the DoList at a time.
+doListMaxInserts = 99000
+
+# Client thread pool size
+threadPoolSize = 5
```
n8pease: do you want a newline at the end of these cfg files?

core/modules/loader/CentralWorker.h
```diff
@@ -253,14 +254,14 @@ class CentralWorker : public Central {
...
+    /// An iteration would be transfer, insert, verify range.
     /// Too big a value, and the maps will be paralyzed for a long time during inserts.
     /// Too small and shift operations will take significantly longer.
```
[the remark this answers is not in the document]
jgates108: Between iterations, other inserts and look ups can happen, so it should be somewhat responsive. If there are a million keys to insert in a single iteration, nothing else will be inserted or looked up until it's done with the million keys. Its about breaking the shifts into reasonable sized bites.

PR #441, opened 2019-01-25
Tickets/dm 16655

jgates108:


jgates108:
Insert 1,000,000 rows took 572 seconds with one worker and one client running on a single computer. Travis failed on the qserv integration tests, which is a bit unexpected as I've avoided changing code within qserv itself.

andy-slac:
I tried to restart Travis and it fails in integration test in exactly the same spot with error:
```
2019-01-25 19:04:56,571 - lsst.qserv.tests.benchmark - INFO - Launch 1006_qserv_func_with_equals.1.sql against qserv
2019-01-25 19:04:56,590 - lsst.qserv.admin.commons - CRITICAL - Error code returned by command : mysql --host=master.localdomain --port=4040 --user=qsmaster --batch qservTest_case01_qserv -e SELECT objectId FROM Object WHERE 1 = qserv_areaspec_box(0, 0, 3, 10) ORDER BY objectId 
```
(this is in case01). That `1 = qserv_areaspec_box()` looks like new syntax from Nate. Did you rebase your branch to latest master?

core/modules/loader/appMaster.cc
```diff
@@ -0,0 +1,90 @@
...
+        LOGS(_log, LOG_LVL_ERROR, "cMaster.start() failed e=" << e.what());
+        exit(-1);
+    }
+    cMaster.setMaxKeysPerWorker(100); // delete
+    // Need to start several threads so messages aren't dropped while being processed.
+    cMaster.run();
+    cMaster.run();
+    cMaster.run();
+    cMaster.run();
+    cMaster.run();
```
andy-slac: 5!

jgates108:
With logging only warnings and errors, the time for 1,000,000 inserts drops to 331 seconds.

core/modules/loader/appMaster.cc
```diff
@@ -0,0 +1,90 @@
...
+
+    bool loop = true;
+    while(loop) {
+        sleep(10);
+    }
+    ioService.stop(); // this doesn't seem to work cleanly
```
andy-slac: But it never reaches this point, so does it matter that it does not work cleanly?

core/modules/loader/testLoader.cc
```diff
@@ -193,6 +194,107 @@ BOOST_AUTO_TEST_CASE(LoaderTest) {
...
+        BOOST_CHECK(!(a == b));
+        BOOST_CHECK( (a != b));
+        BOOST_CHECK(!(a < b));
+        BOOST_CHECK( (a > b));
+        BOOST_CHECK(!(a <= b));
+        BOOST_CHECK( (a >= b));
+    }
+
+    {
+        CompositeKey a(34568, "@#WSR$RT%fewsewer");
```
andy-slac: How do you know my password? :cat2:

core/modules/loader/CentralClient.cc
```diff
@@ -64,45 +65,45 @@ void CentralClient::start() {
...
     }
 
     // TODO put in separate thread
-    _handleKeyInfo(inMsg, protoData);
+    _handleKeyLook(inMsg, protoData);
 }
 
 
-void CentralClient::_handleKeyInfo(LoaderMsg const& inMsg, std::unique_ptr<proto::KeyInfo>& protoBuf) {
+void CentralClient::_handleKeyLook(LoaderMsg const& inMsg, std::unique_ptr<proto::KeyInfo>& protoBuf) {
```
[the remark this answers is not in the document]
jgates108: The intention is to put _handleKeyInfo in a separate thread at some point so that it doesn't use one of the boost asio threads. In which case it needs the pointer to keep the memory from being reclaimed. It may be just as well to just create more asio threads and pass a reference.

core/modules/loader/CentralClient.cc
```diff
@@ -198,38 +219,67 @@ void CentralClient::_keyInsertReq(std::string const& key, int chunk, int subchun
...
     protoKeyInsert.set_hops(0);
 
     StringElement strElem;
     protoKeyInsert.SerializeToString(&(strElem.element));
     strElem.appendToData(msgData);
-
-    sendBufferTo(getDefWorkerHost(), getDefWorkerPortUdp(), msgData);
+    try {
+        sendBufferTo(getDefWorkerHost(), getDefWorkerPortUdp(), msgData);
+    } catch (boost::system::system_error e) {
```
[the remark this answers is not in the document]
jgates108: Fixed them

core/modules/loader/ServerUdpBase.cc
```diff
@@ -146,14 +153,11 @@ void ServerUdpBase::_receivePrepare() {
...
-    io_context ioContext;
-    ip::udp::resolver resolver(ioContext);
-    return *resolver.resolve(udp::v4(), hostName, std::to_string(port)).begin();
-    */
+    std::lock_guard<std::mutex> lg(_resolveMtx);
     using namespace boost::asio;
-    ip::udp::endpoint dest(ip::address::from_string(hostName), port);
+    // resolver returns an iterator. This uses the first item only.
+    ip::udp::endpoint dest =
+            *_resolver.resolve(ip::udp::v4(), hostName, std::to_string(port)).begin();
```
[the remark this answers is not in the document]
jgates108: If it fails to resolve anything, it should throw a boost system error. There's a comment in the header file to that effect. I'll add another here.

core/modules/loader/appClientNum.cc
```diff
@@ -0,0 +1,262 @@
...
+    if (numEnd >= numStart) {
+        totalKeyCount = (numEnd - numStart) + 1;
+        for (uint64_t j=numStart; j<=numEnd; ++j) {
+            kList.push_back(clientAdd(cClient, j));
+            // occasionally trim the list
+            if (j%1000 == 0) keyInsertListClean(kList, successCount, failedCount);
+        }
+    } else {
+        totalKeyCount = (numStart - numEnd) + 1;
+        for (uint64_t j=numStart; j>=numEnd; --j) {
```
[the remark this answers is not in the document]
jgates108: Good point.

core/modules/loader/testLoader.cc
jgates108: Your cat is terrible at keeping secrets.

core/modules/loader/CentralClient.h
```diff
@@ -162,14 +164,15 @@ class CentralClient : public Central {
...
+    size_t _doListMaxInserts; ///< Maximum number of concurrent inserts in DoList  DM-16555 &&&
+    int _maxRequestSleepTime{100000}; ///< Time to sleep between checking requests when at max length &&& add config file entry
 
     std::map<std::string, KeyInsertReqOneShot::Ptr> _waitingKeyInsertMap;
-    std::mutex _waitingKeyInsertMtx; ///< protects _waitingKeyInsertMap
+    std::mutex _waitingKeyInsertMtx; ///< protects _waitingKeyInsertMap, _doListMaxInserts
 
-    std::map<std::string, KeyInfoReqOneShot::Ptr> _waitingKeyInfoMap;
-    std::mutex _waitingKeyInfoMtx; ///< protects _waitingKeyInfoMap
+    std::map<std::string, KeyInfoReqOneShot::Ptr> _waitingKeyInfoMap; // &&& change all references of keyInfo to keyLookup &&&, including protobuf keyInfo should only apply to  worker key count and worker key range.
```
iagaponenko: This line looks way to long. I think we have a limit of 105 characters, or something like this. You may also consider putting this comment in front of the member variable and document it using the Doxygen style:
```
/// &&& change all references of keyInfo to keyLookup &&&, including protobuf keyInfo
/// should only apply to  worker key count and worker key range.
std::map<std::string, KeyInfoReqOneShot::Ptr> _waitingKeyInfoMap;
```

core/modules/loader/CentralClient.cc
```diff
@@ -65,45 +65,45 @@ void CentralClient::start() {
 }
 
 
-void CentralClient::handleKeyInfo(LoaderMsg const& inMsg, BufferUdp::Ptr const& data) {
-    LOGS(_log, LOG_LVL_DEBUG, "CentralClient::handleKeyInfo");
+void CentralClient::handleKeyLook(LoaderMsg const& inMsg, BufferUdp::Ptr const& data) {
+    LOGS(_log, LOG_LVL_DEBUG, "CentralClient::handleKeyLook");
 
     StringElement::Ptr sData = std::dynamic_pointer_cast<StringElement>(MsgElement::retrieve(*data));
```
iagaponenko: `auto` would work here better. Besides, variables which are initialized at a point of their declarations and which are never changed later might be `const`, just in case:
```
auto const sData = std::dynamic_pointer_cast<StringElement>(MsgElement::retrieve(*data));
```

core/modules/loader/CentralClient.cc
```diff
@@ -233,30 +233,28 @@ void CentralClient::_keyInsertReq(CompositeKey const& key, int chunk, int subchu
...
-        auto iter = _waitingKeyInfoMap.find(key);
-        while (_waitingKeyInfoMap.size() > _doListMaxLookups
-               && iter == _waitingKeyInfoMap.end()) {
-            size_t sz = _waitingKeyInfoMap.size();
+        auto iter = _waitingKeyLookMap.find(key);
+        while (_waitingKeyLookMap.size() > _doListMaxLookups
+               && iter == _waitingKeyLookMap.end()) {
+            size_t sz = _waitingKeyLookMap.size();
             lck.unlock();
             if (loopCount % 100 == 0) {
```
iagaponenko: This condition may result in non-deterministic behavior as it depends on a value of `_maxRequestSleepTime` which can be changed w/o taking into account this loop . I would rather count the amount of time waited in `usleep(_maxRequestSleepTime)` and print the message, say every N seconds. Then you'd be getting periodic printouts every N seconds regardless of what you choose to put into `_maxRequestSleepTime`.

core/modules/loader/CentralClient.h
```diff
@@ -168,11 +168,11 @@ class CentralClient : public Central {
     size_t _doListMaxInserts{1000}; ///< Maximum number of concurrent inserts in DoList (set by config)
     int _maxRequestSleepTime{100000}; ///< Time to sleep between checking requests when at max length (set by config)
```
iagaponenko: Please, consider this form of the initialization:
```
size_t const _doListMaxInserts = 1000;
int const _maxRequestSleepTime = 100000;
```

core/modules/loader/CentralWorker.cc
```diff
@@ -382,7 +382,7 @@ void CentralWorker::_shift(Direction direction, int keysToShift) {
 
             // TODO This is very similar to code in TcpBaseConnection::_handleShiftToRight and they should be merged.
             int sz = protoKeyList->keypair_size();
-            std::vector<CentralWorker::StringKeyPair> keyList;
+            std::vector<CentralWorker::CompKeyPair> keyList;
```
iagaponenko: I would highly recommend to get rid of that `std::` in the CC-files. This will eliminate a lot of clutter from the code. Just put somewhere on top of the file:
```
using namespace std;
```

core/modules/loader/CentralWorker.cc
```diff
@@ -928,8 +924,6 @@ void CentralWorker::_workerKeyInfoReq(LoaderMsg const& inMsg, std::unique_ptr<pr
         }catch (boost::system::system_error e) {
```
iagaponenko: Space is needed here `}catch`

core/modules/loader/CentralClient.h
```diff
@@ -136,12 +136,12 @@ class CentralClient : public Central {
         CentralClient* central;
     };
 
-    /// Create commands to find a key in the index and get its value.
+    /// Create commands to look up a key in the index and get its value.
     /// It should keep trying this until it works and then drop it from _waitingKeyInfoMap.
-    struct KeyInfoReqOneShot : public DoListItem {
-        using Ptr = std::shared_ptr<KeyInfoReqOneShot>;
+    struct KeyLookReqOneShot : public DoListItem {
+        using Ptr = std::shared_ptr<KeyLookReqOneShot>;
```
iagaponenko: I've noticed that you're inconsistently using these definitions:
```
using Ptr = std::shared_ptr<KeyLookReqOneShot>;
typedef std::shared_ptr<KeyLookReqOneShot> Ptr;
```

core/modules/loader/DoListItem.h
```diff
@@ -104,7 +104,7 @@ class DoListItem : public std::enable_shared_from_this<DoListItem> {
             if (_isOneShotDone()) { return nullptr; }
```
iagaponenko: Curly braces are not needed in the one-line `if` statements

core/modules/loader/LoaderMsg.h
```diff
@@ -60,7 +60,7 @@ class LoaderMsg {
         KEY_INSERT_REQ,        // Insert a new key with info. MSG_RECEIVED + KEY_INFO
         KEY_INSERT_COMPLETE,   // Key has been inserted and logged.
         KEY_INFO_REQ,          // Request info for a single key.
-        KEY_INFO,              // Information about a specific key. (includes file id and row)
+        KEY_LOOK,              // Information about a specific key. (includes file id and row)
```
iagaponenko: Based on the comment to this value `KEY_INFO` was a more appropriate identifier. Could the comment be made consistent with the new name?

core/modules/loader/appClientNum.cc
```diff
@@ -0,0 +1,262 @@
...
+    } catch (boost::system::system_error const& e) {
+        LOGS(_log, LOG_LVL_ERROR, "cWorker.start() failed e=" << e.what());
+        exit(-1);
+    }
+
+    // Need to start several threads so messages aren't dropped while being processed.
+    cClient.run();
+    cClient.run();
+    cClient.run();
+    cClient.run();
```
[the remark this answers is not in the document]
iagaponenko: I guess each creates a new BOOST ASIO thread. This might be better be done explicitly in a loop:
```
int const numClients = 4;
for (int i=0; i<numClients; ++i) {
    cClient.run();
}
```

core/modules/loader/appClientNum.cc
```diff
@@ -151,17 +151,24 @@ int main(int argc, char* argv[]) {
...
+    int totalKeyCount = 0;
+
+    TimeOut::TimePoint insertBegin = TimeOut::Clock::now();
+
     if (numEnd >= numStart) {
+        totalKeyCount = (numEnd - numStart) + 1;
         for (uint64_t j=numStart; j<=numEnd; ++j) {
             kList.push_back(clientAdd(cClient, j));
             // occasionally trim the list
             if (j%1000 == 0) keyInsertListClean(kList, successCount, failedCount);
```
iagaponenko: What's so special about `1000`? Why not `100` instead?
It may make a sense to do the trimming based on something like an interval of time or the total number of requests in a queue, or on the memory, etc.

core/modules/loader/WorkerListItemBase.h
```diff
@@ -30,10 +30,10 @@
 #include <memory>
 #include <mutex>
 
+#include "KeyRange.h"
```
iagaponenko: Misplaced header w/o path.

core/modules/loader/appClientNum.cc
```diff
@@ -0,0 +1,277 @@
...
+ * the GNU General Public License along with this program.  If not,
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+
+
+// Class header
+#include "loader/CentralClient.h"
+#include "loader/ClientConfig.h"
+
+// System headers
```
iagaponenko: I think the convention is to always include the system headers before any other. The only exception is the CC-files of the classes which should include class header as the very first include.

core/modules/loader/appClientNum.cc
```diff
@@ -0,0 +1,277 @@
...
+#include <iostream>
+#include <unistd.h>
+
+// Third-party headers
+#include "boost/lexical_cast.hpp"
+
+
+// qserv headers
```
iagaponenko: No includes - no comment lines!

core/modules/loader/appClientNum.cc
```diff
@@ -0,0 +1,277 @@
...
+using namespace lsst::qserv::loader;
+using  boost::asio::ip::udp;
+
+typedef std::list<KeyInfoData::Ptr> KeyInfoDataList;
+
+/// @return true if the list is empty.
+bool keyInsertListClean(KeyInfoDataList& kList, int& successCount, int& failedCount) {
+    for(auto iter=kList.begin(); iter != kList.end();) {
+        if (*iter == nullptr || (*iter)->isFinished()) {
+            KeyInfoData::Ptr const& kPtr = *iter;
```
iagaponenko: `auto&&` would be sufficient here

core/modules/loader/appClientNum.cc
```diff
@@ -0,0 +1,277 @@
...
+    }
+    return kList.empty();
+}
+
+
+/// Get a repeatable value for the chunk and subchunk numbers. It's arbitrary for
+/// the test as there just needs to be some check that what was written in for
+/// the key is the same as what was read
+int calcChunkFrom(uint64_t j) {
+    return j%10000;
```
iagaponenko: spaces around `%` would make this easier to read: `j % 10000`

core/modules/loader/appClientNum.cc
```diff
@@ -0,0 +1,277 @@
...
+    LOGS(_log, LOG_LVL_INFO, "start=" << numStart << " end=" << numEnd << " cCfg=" << cCfgFile);
+    if (numEnd == 0) {
+        LOGS(_log, LOG_LVL_ERROR, "end cannot equal 0");
+        return 1;
+    }
+
+    std::string ourHost;
+    {
+        char hName[300];
+        if (gethostname(hName, sizeof(hName)) < 0) {
```
iagaponenko: I would rather use BOOST ASIO since you already using it:
```
std::string const name = boost::asio::ip::host_name();
```
It's here:
```
namespace boost {
namespace asio {
namespace ip {

/// Get the current host name.
BOOST_ASIO_DECL std::string host_name();

/// Get the current host name.
BOOST_ASIO_DECL std::string host_name(boost::system::error_code& ec);

} // namespace ip
} // namespace asio
} // namespace boost
```

core/modules/loader/appMaster.cc
```diff
@@ -0,0 +1,84 @@
...
+ * the GNU General Public License along with this program.  If not,
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+
+
+// Class header
+#include "loader/CentralMaster.h"
+
+// System headers
+#include <iostream>
```
iagaponenko: Please, include system headers before any other ones

core/modules/loader/appMaster.cc
```diff
@@ -0,0 +1,84 @@
...
+
+// Third-party headers
```
iagaponenko: No includes - no comments

core/modules/loader/appMaster.cc
```diff
@@ -0,0 +1,84 @@
...
+    std::string mCfgFile("core/modules/loader/config/master.cnf");
+    if (argc > 1) {
+        mCfgFile = argv[1];
+    }
+    LOGS(_log, LOG_LVL_INFO, "masterCfg=" << mCfgFile);
+
+    std::string ourHost;
+    {
+        char hName[300];
+        if (gethostname(hName, sizeof(hName)) < 0) {
```
iagaponenko: Use `boost::asio::ip::host_name()` instead

core/modules/loader/appWorker.cc
```diff
@@ -0,0 +1,86 @@
...
+ * the GNU General Public License along with this program.  If not,
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+
+
+// Class header
+#include "loader/CentralWorker.h"
+
+// System headers
+#include <iostream>
```
iagaponenko: Please, rearrange includes as per the LSST C++ Style guide!

core/modules/loader/appWorker.cc
```diff
@@ -0,0 +1,86 @@
...
+    std::string wCfgFile("core/modules/loader/config/worker1.cnf");
+    if (argc > 1) {
+        wCfgFile = argv[1];
+    }
+    LOGS(_log, LOG_LVL_INFO, "workerCfg=" << wCfgFile);
+
+    std::string ourHost;
+    {
+        char hName[300];
+        if (gethostname(hName, sizeof(hName)) < 0) {
```
iagaponenko: Use `boost::asio::ip::host_name()` instead

core/modules/loader/CentralClient.cc
```diff
@@ -161,10 +164,27 @@ KeyInfoData::Ptr CentralClient::keyInsertReq(std::string const& key, int chunk,
...
         auto iter = _waitingKeyInsertMap.find(key);
+        while (_waitingKeyInsertMap.size() > _doListMaxInserts
+               && iter == _waitingKeyInsertMap.end()) {
+            size_t sz = _waitingKeyInsertMap.size();
+            lck.unlock();
+            if (loopCount % 100 == 0) {
+                LOGS(_log, LOG_LVL_INFO, "keyInsertReq waiting key=" << key <<
+                        "size=" << sz << " loopCount=" << loopCount);
+            }
+            usleep(_maxRequestSleepTime);
```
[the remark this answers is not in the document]
jgates108: Loops like this can burn significant CPU cycles doing nothing useful while waiting for something to finish. A short sleep can significantly reduce wasted cycles and is much simpler than a condition variable solution.

core/modules/loader/CentralClient.cc
```diff
@@ -214,14 +234,38 @@ void CentralClient::_keyInsertReq(std::string const& key, int chunk, int subchun
...
+        auto iter = _waitingKeyInfoMap.find(key);
+        while (_waitingKeyInfoMap.size() > _doListMaxLookups
+               && iter == _waitingKeyInfoMap.end()) {
+            size_t sz = _waitingKeyInfoMap.size();
+            lck.unlock();
+            if (loopCount % 100 == 0) {
+                LOGS(_log, LOG_LVL_INFO, "keyInfoReq waiting key=" << key <<
+                        "size=" << sz << " loopCount=" << loopCount);
+            }
+            usleep(_maxRequestSleepTime);
```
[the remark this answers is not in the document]
jgates108: // Let the CPU do something else while waiting for some requests to finish.

core/modules/loader/Central.h
```diff
@@ -103,19 +101,30 @@ class Central {
...
+                int threadPoolSize, int loopSleepTime,
+                int iOThreads)
                 : ioService(ioService_), _masterAddr(masterHostName, masterPort),
-                  _threadPoolSize(threadPoolSize), _loopSleepTime(loopSleepTime) {
+                  _threadPoolSize(threadPoolSize), _loopSleepTime(loopSleepTime),
+                  _iOThreads(iOThreads){
         _initialize();
     }
 
+    void run();
```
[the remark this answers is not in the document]
iagaponenko: I meant to say - put a comment documenting this method :-)

core/modules/loader/ClientConfig.h
```diff
@@ -86,7 +88,14 @@ class ClientConfig : public ConfigBase {
     /// Maximum number of insert requests allowed in the DoList.
     ConfigElement::Ptr _maxInserts{
         ConfigElement::create(cfgList, header, "maxInserts", ConfigElement::INT, false, "90000")};
-
+    /// When reaching maxInserts or maxLookups, sleep this long before trying to add more,
+    /// in micro seconds. 100000micro = 0.1sec
+    ConfigElement::Ptr _maxRequestSleepTime{
+        ConfigElement::create(cfgList, header,
+                              "maxRequestSleepTime", ConfigElement::INT, false, "100000")};
```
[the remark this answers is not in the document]
jgates108: Eventually, that value gets passed to util::ConfigStore::get() which takes a string for the default argument, and jumping through the hoops to take an integer here only to convert to a string later doesn't seem worth it at this time. 
I really want to avoid changing code in qserv proper until this branch is ready to merge to master.

core/modules/loader/MWorkerList.h
```diff
@@ -30,12 +30,12 @@
 #include <memory>
 #include <mutex>
 
+#include "KeyRange.h"
```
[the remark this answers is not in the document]
iagaponenko: I've ran into the very same problem with NetBeans 😄

core/modules/loader/Central.h
```diff
@@ -143,6 +152,9 @@ class Central {
     std::vector<std::thread> _ioServiceThreads; ///< List of asio io threads created by this
 
     std::thread _checkDoListThread; ///< Thread for running doList checks on DoListItems.
+
+    int _iOThreads{5}; ///< Number of asio IO threads to run, set by config file.
```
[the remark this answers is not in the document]
jgates108: It isn't a conflict as the constructor will overwrite the value.  
I'm not really sure what the correct course of action is here, and I suspect the best method differs depending on the situation.
Moving the default parameter to the constructor defeats the purpose as the default value set here is meant to take hold if the value is not set by a constructor.
I'm a pretty big believer in setting values when variables are declared. In classes, that means in many cases, you only have to set the default value once in the header. If that value changes, you don't have to hunt it down in all the constructors.

core/modules/loader/Central.h
```diff
@@ -143,6 +152,9 @@ class Central {
...
+    int _runningIOThreads{0}; ///< Number of asio IO threads started.
```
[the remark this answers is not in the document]
jgates108: Being explicit here tells the reader that the default value should be zero and that variable initialization was not overlooked.

PR #442, opened 2019-01-28
Fix for czar message table insert fail

fritzm:
Small proposed fix for the czar side of this issue

### DIFF
--- a/core/modules/czar/MessageTable.cc
+++ b/core/modules/czar/MessageTable.cc
@@ -42,16 +42,17 @@ namespace {
 
 LOG_LOGGER _log = LOG_GET("lsst.qserv.czar.MessageTable");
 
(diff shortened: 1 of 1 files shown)


core/modules/czar/MessageTable.cc
```diff
@@ -42,16 +42,17 @@ namespace {
 
 LOG_LOGGER _log = LOG_GET("lsst.qserv.czar.MessageTable");
 
+#define MAX_MESSAGE_LEN "1024"
```
ktlim: Maybe a comment to indicate that this must be a string rather than an integer?

fritzm:
@iagaponenko but it is a change to the (per-query, transient) message table schema, not to qmeta?

iagaponenko:
If this is a temporary table then it's not needed.

PR #445, opened 2019-01-30
DM-17092: look into 'group by question'

n8pease:
the bug was that an extra comma when the IR was reserialzied to a string, so `GROUP BY a, b` became `GROUP BY a,,b`. 

no need to mannualy add commas, the inserter does it automatically.

also extends the antlr4 expected-value-construction unit test to support
the GroupByClause and a deque in addition to vectors in the pusher
function.

### DIFF
--- a/core/modules/ccontrol/testAntlr4GeneratedIR.cc
+++ b/core/modules/ccontrol/testAntlr4GeneratedIR.cc
@@ -81,9 +81,9 @@ BOOST_AUTO_TEST_SUITE(Suite)
  * @param container The container to push objects into.
  * @param first The object to push into the container.
  */
-template <template <typename... Args> class Container, typename... Types, typename T, typename... TArgs>
-void pusher(Container<Types...>& container, T first) {
-    container.push_back(first);
+template <typename Container, typename T>
+void pusher(Container& container, T&& first) {
+    container.push_back(std::forward<T>(first));
 }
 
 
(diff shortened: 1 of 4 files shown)


core/modules/ccontrol/testAntlr4GeneratedIR.cc
```diff
@@ -65,60 +66,52 @@ using namespace lsst::qserv;
...
+ *               and the allocator type)
+ * @tparam T The type of the object to push into the container.
+ * @tparam TArgs The type of the remaining objects to push onto the vector (this should match T)
+ * @param container The container to push objects into.
+ * @param first The object to push into the container.
  */
-template <typename T, typename U>
-void pusher(vector<T>& vec, U v) {
-    vec.push_back(v);
+template <template <typename... Args> class Container, typename... Types, typename T, typename... TArgs>
```
iagaponenko: I'm not sure I understand why such complex template specification is needed in this context.

core/modules/ccontrol/testAntlr4GeneratedIR.cc
```diff
@@ -65,60 +66,52 @@ using namespace lsst::qserv;
...
+ * @tparam TArgs The type of the remaining objects to push onto the vector (this should match T)
+ * @param container The container to push objects into.
+ * @param first The object to push into the container.
+ * @param args The rest of the objects to push into the container.
  */
-template <typename T, typename...Args>
-void pusher(vector<T>& vec, T first, Args... args) {
-    vec.push_back(first);
-    pusher(vec, args...);
+template <template <typename... Args> class Container, typename... Types, typename T, typename... TArgs>
```
iagaponenko: Same comment regarding the complexity of the template specification

core/modules/ccontrol/testAntlr4GeneratedIR.cc
```diff
@@ -65,60 +66,52 @@ using namespace lsst::qserv;
...
+void pusher(Container<Types...>& container, T first, TArgs... args) {
```
iagaponenko: I think that would be easier to understand:
```
template <class CONTAINER, class D, class T, typename... Targs>
void pusher(CONTAINER<D>& cont, T first, Targs... args) {
   ...
}
```
This will make the template specification easier to understand, and it would also allow a flexibility in the `T -> D` conversion.

If the only reason for such complex template is in supporting multi-parametric containers, such as `map<K,V>` then I would rather go with the template specialization by adding extra set of methods for this set of problems. For example:
```
template <class CONTAINER, class K, class V,  class T, typename... Targs>
void pusher(CONTAINER<K,V>& cont, T first, Targs... args) {
   ...
}
```
Since these functions are rather small this won't blow your code. But it would add some clarity to it.

PR #447, opened 2019-02-05
DM-17659: Use != rather than "is not" to compare to an int

timj:


timj:
Yes, the new flake8 now recognizes `var is not constant` as a problem (code F632).

andy-slac:
Ah, I should have checked the ticket first :sorry:

PR #449, opened 2019-02-19
Tickets/dm 16033

jgates108:


core/modules/qmeta/QMetaMysql.cc
```diff
@@ -817,6 +817,129 @@ QMetaMysql::_checkDb() {
...
+    // Try to create the temporary table if it is not there.
+    QMetaTransaction trans(_conn);
+    sql::SqlErrorObject errObj;
+    sql::SqlResults results;
+    std::string query = "CREATE TABLE IF NOT EXISTS QStatsTmp (queryId bigint(20), "
+                        "totalChunks int, completedChunks int, "
+                        "queryBegin timestamp DEFAULT 0, "
+                        "lastUpdate timestamp DEFAULT 0, "
+                        "PRIMARY KEY (queryId)) "
+                        "ENGINE = MEMORY;";
```
andy-slac: No need to add semicolon to the statements (here and all other places).

core/modules/qmeta/QMetaMysql.cc
```diff
@@ -817,6 +817,129 @@ QMetaMysql::_checkDb() {
                                value);
     }
 
+    createQueryStatsTmpTable();
```
andy-slac: What happens if exception is thrown by this method, entire QMeta becomes unusable? Maybe add try/catch around it and handle this condition. Or rather move all of that stuff to a separate thing independent of QMeta.

core/modules/qmeta/QMetaMysql.cc
```diff
@@ -817,6 +817,129 @@ QMetaMysql::_checkDb() {
...
+
+
+bool QMetaMysql::queryStatsTmpRegister(QueryId queryId, int totalChunks) {
+    std::lock_guard<std::mutex> sync(_dbMutex);
+    QMetaTransaction trans(_conn);
+    sql::SqlErrorObject errObj;
+    sql::SqlResults results;
+    std::string query = "INSERT INTO QStatsTmp (queryId, totalChunks, completedChunks, queryBegin, lastUpdate) "
+                        "VALUES ( ";
+    query += boost::lexical_cast<std::string>(queryId) + ", ";
```
andy-slac: I'd probably do `std::to_string` which is a bit shorter and more standard.

core/modules/qmeta/QMetaMysql.cc
```diff
@@ -817,6 +817,129 @@ QMetaMysql::_checkDb() {
...
+    QMetaTransaction trans(_conn);
+    sql::SqlErrorObject errObj;
+    sql::SqlResults results;
+    std::string query = "UPDATE QStatsTmp SET completedChunks = ";
+    query += boost::lexical_cast<std::string>(completedChunks);
+    query += ", lastUpdate = NOW() WHERE queryId = ";
+    query += boost::lexical_cast<std::string>(queryId) + ";";
+
+    LOGS(_log, LOG_LVL_DEBUG, "Executing query: " << query);
+    if (not _conn.runQuery(query, results, errObj)) {
```
andy-slac: You don't need `results` argument with UPDATE query.

core/modules/qmeta/QMetaMysql.cc
```diff
@@ -817,6 +817,129 @@ QMetaMysql::_checkDb() {
...
+bool QMetaMysql::queryStatsTmpRemove(QueryId queryId) {
+    std::lock_guard<std::mutex> sync(_dbMutex);
+    QMetaTransaction trans(_conn);
+    sql::SqlErrorObject errObj;
+    sql::SqlResults results;
+    std::string query = "DELETE FROM QStatsTmp WHERE queryId =";
+    query += boost::lexical_cast<std::string>(queryId) + ";";
+
+    LOGS(_log, LOG_LVL_DEBUG, "Executing query: " << query);
+    if (not _conn.runQuery(query, results, errObj)) {
```
andy-slac: `results` can be dropped here too

core/modules/qmeta/QMetaMysql.cc
andy-slac: Just a note for the future - if schema of this table ever changes you'll also need to DROP old one.

core/modules/qdisp/Executive.h
```diff
@@ -157,6 +162,7 @@ class Executive : public std::enable_shared_from_this<Executive> {
     XrdSsiService* _xrdSsiService; ///< RPC interface
     JobMap _jobMap; ///< Contains information about all jobs.
     JobMap _incompleteJobs; ///< Map of incomplete jobs.
+    std::atomic<int> _totalJobs{1}; ///< How many jobs are used in this query.
```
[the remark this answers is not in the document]
jgates108: It's set to 1 so that there isn't a 0 out of 0 chunks completed possibility being sent to the user when the worker queries are being generated. The value is overwritten by the correct value when that value is know.

core/modules/qmeta/QMeta.h
```diff
@@ -272,6 +273,37 @@ class QMeta {
...
+     * @return QStats object containing query completion information.
+     * @throw QueryIdError, SqlError
+     */
+    virtual QStats queryStatsTmpGet(QueryId queryId) = 0;
+
+    /**
+     * @ brief Remove row for completion status when the query is done.
+     * @return true if successful.
+     */
+    virtual bool queryStatsTmpRemove(QueryId queryId) = 0;
```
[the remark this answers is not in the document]
jgates108: I had thought this information was going to be put in the central database in in-memory tables?

core/modules/qdisp/Executive.h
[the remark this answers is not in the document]
andy-slac: Can you add a comment saying that it should be initialized with non-zero value?

core/modules/qmeta/QMeta.h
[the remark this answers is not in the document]
andy-slac: My impression was that we wanted to avoid pounding central server with the updates that no one reads (or reads very rarely). Still, if it's going to be a central service I think some extra flexibility may be beneficial, we could move it to some specialized in memory-service like memcached or whatnot.

core/modules/qmeta/QMetaMysql.cc
jgates108: If (probably when) a schema change becomes needed, DROP'ing the old table should probably be done with a script. All of the data is temporary in nature, so there's no need to worry about keeping it if qserv is stopped.

jgates108:
The code still is in qmeta, but is now outside of the QMeta class. It should be very easy to move the table to another database if needed. The data in the table is metadata for a query and I think belongs with the other QMeta data and should only be moved if it does cause a performance problem with multiple czars.

core/modules/qmeta/QStatus.h
```diff
@@ -0,0 +1,87 @@
...
+#include "qmeta/QStats.h"
+#include "qmeta/types.h"
+
+
+namespace lsst {
+namespace qserv {
+namespace qmeta {
+
+
+class QStatus {
```
iagaponenko: Some documentation explaining a role of the class would be helpful.

core/modules/qmeta/QStatus.h
```diff
@@ -0,0 +1,87 @@
...
+public:
+	typedef std::shared_ptr<QStatus> Ptr;
+    /**
+     *  Type for representing the list of tables, first item in pair is
+     *  database name, second is table name.
+     */
+    typedef std::vector<std::pair<std::string, std::string> > TableNames;
```
iagaponenko: Since C++11 there is no need to have spaces in template declarations. I mean this: `> >`

core/modules/qmeta/QStatus.cc
```diff
@@ -0,0 +1,55 @@
...
+
+LOG_LOGGER _log = LOG_GET("lsst.qserv.qmeta.QStatus");
+
+}
+
+namespace lsst {
+namespace qserv {
+namespace qmeta {
+
+
```
iagaponenko: This file is empty. Why it's here?

core/modules/qmeta/QStatusMysql.h
```diff
@@ -0,0 +1,73 @@
...
+    typedef std::shared_ptr<QStatusMysql> Ptr;
+
+    QStatusMysql(mysql::MySqlConfig const& mysqlConf);
+
+    QStatusMysql(QStatusMysql const&) = delete;
+    QStatusMysql& operator=(QStatusMysql const&) = delete;
+
+    virtual ~QStatusMysql() = default;
+
+    /// @see QMeta::createQueryStatsTable()
```
iagaponenko: I think the comment is misleading. Should it be referring to the base class `QStatus::` instead?

core/modules/qmeta/QStatusMysql.h
```diff
@@ -0,0 +1,73 @@
...
+    void createQueryStatsTmpTable() override;
+
+    /// @see QMeta::queryStatsTmpRegister(QueryId queryId, int totalChunks)
```
iagaponenko: Same comment for this an other `virtual` methods of this class. I think the comment is misleading. Should it be referring to the base class `QStatus::` instead?

core/modules/qmeta/QStatusMysql.h
```diff
@@ -0,0 +1,73 @@
...
+class QStatusMysql : public QStatus {
+public:
+    typedef std::shared_ptr<QStatusMysql> Ptr;
+
+    QStatusMysql(mysql::MySqlConfig const& mysqlConf);
+
+    QStatusMysql(QStatusMysql const&) = delete;
+    QStatusMysql& operator=(QStatusMysql const&) = delete;
+
+    virtual ~QStatusMysql() = default;
```
iagaponenko: Since this is a subclass then the destructor needs to be defined as:
```
~QStatusMysql() override = default;
```
Or:
```
~QStatusMysql() final = default;
```

core/modules/qmeta/QStatusMysql.h
```diff
@@ -0,0 +1,73 @@
+/*
+ * LSST Data Management System
+ * Copyright 2015 AURA/LSST.
```
iagaponenko: 2019

core/modules/qmeta/QStatusMysql.cc
```diff
@@ -0,0 +1,189 @@
...
+                        "UNIX_TIMESTAMP(queryBegin), UNIX_TIMESTAMP(lastUpdate) "
+                        "FROM QStatsTmp WHERE queryId= ";
+    query += std::to_string(queryId);
+    LOGS(_log, LOG_LVL_DEBUG, "Executing query: " << query);
+    if (not _conn.runQuery(query, results, errObj)) {
+        LOGS(_log, LOG_LVL_ERROR, "SQL query failed: " << query);
+        throw SqlError(ERR_LOC, errObj);
+    }
+
+    sql::SqlResults::iterator rowIter = results.begin();
```
iagaponenko: Why not to use `auto` here?

core/modules/qmeta/QStatusMysql.cc
```diff
@@ -0,0 +1,189 @@
...
+    if (rowIter == results.end()) {
+        // no records found
+        throw QueryIdError(ERR_LOC, queryId);
+    }
+
+    // make sure that iterator does not move until we are done with row
+    sql::SqlResults::value_type const& row = *rowIter;
```
iagaponenko: Why not to use `auto` or `auto const&` here?

core/modules/qmeta/QStats.h
```diff
@@ -0,0 +1,56 @@
...
+        : queryId(queryId_), totalChunks(totalChunks_), completedChunks(completedChunks_),
+          begin(begin_), lastUpdate(lastUpdate_) {}
+    // Default operators and destructor should be fine.
+
+
+    QueryId queryId{0}; ///< Query Id
+    int totalChunks{0}; ///< Total number of chunks to be searched by the query.
+    int completedChunks{0}; ///< Number of chunks that have been searched.
+    std::time_t begin{0}; ///< Time the query was started.
+    std::time_t lastUpdate{0}; ///< Last time this row was updated.
```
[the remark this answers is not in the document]
jgates108: When initializing member variable values in the class definition, other methods don't parse.

core/modules/qmeta/QStatusMysql.cc
```diff
@@ -0,0 +1,189 @@
...
+}
+
+
+void QStatusMysql::queryStatsTmpRemove(QueryId queryId) {
+    std::lock_guard<std::mutex> sync(_dbMutex);
+    QMetaTransaction trans(_conn);
+    sql::SqlErrorObject errObj;
+    //sql::SqlResults results; &&&
+    std::string query = "DELETE FROM QStatsTmp WHERE queryId =";
+    query += std::to_string(queryId);
```
iagaponenko: These two lines could be a single line:
```
auto const query = "DELETE FROM QStatsTmp WHERE queryId =" + std::to_string(queryId);
```
The compiler would automatically make an object of type `std::string`

core/modules/qmeta/QStatusMysql.cc
```diff
@@ -0,0 +1,189 @@
...
+
+void QStatusMysql::queryStatsTmpChunkUpdate(QueryId queryId, int completedChunks) {
+    std::lock_guard<std::mutex> sync(_dbMutex);
+    QMetaTransaction trans(_conn);
+    sql::SqlErrorObject errObj;
+    //sql::SqlResults results; &&&
+    std::string query = "UPDATE QStatsTmp SET completedChunks = ";
+    query += std::to_string(completedChunks);
+    query += ", lastUpdate = NOW() WHERE queryId = ";
+    query += std::to_string(queryId);
```
iagaponenko: These 4 lines could be make one by:
```
auto const query =
    "UPDATE QStatsTmp SET completedChunks = " + std::to_string(completedChunks) +
    ", lastUpdate = NOW() WHERE queryId = " + std::to_string(queryId);
```

core/modules/qmeta/QStatusMysql.cc
```diff
@@ -0,0 +1,189 @@
...
+
+void QStatusMysql::queryStatsTmpRegister(QueryId queryId, int totalChunks) {
+    std::lock_guard<std::mutex> sync(_dbMutex);
+    QMetaTransaction trans(_conn);
+    sql::SqlErrorObject errObj;
+    // sql::SqlResults results; &&&
+    std::string query = "INSERT INTO QStatsTmp (queryId, totalChunks, completedChunks, queryBegin, lastUpdate) "
+                        "VALUES ( ";
+    query += std::to_string(queryId) + ", ";
+    query += std::to_string(totalChunks) + ", ";
```
iagaponenko: CC files could be made more readable by eliminating `std::` using `using namespace std;` at the beginning of the file.

iagaponenko: Then the query generation could be rewritten in one line:
```
auto const query =
    "INSERT INTO QStatsTmp (queryId, totalChunks, completedChunks, queryBegin, lastUpdate) "
    "VALUES ("+ to_string(queryId) + "," + to_string(totalChunks) + ", " + "0, NOW(), NOW())";
```

core/modules/qmeta/QStatusMysql.cc
```diff
@@ -0,0 +1,189 @@
...
+// Qserv headers
+#include "qmeta/Exceptions.h"
+#include "qmeta/QMetaTransaction.h"
+#include "sql/SqlResults.h"
+
+
+namespace {
+
+// Current version of QStatus schema.
+char const VERSION_STR[] = "0";
```
iagaponenko: I don't see any uses for this variable in the rest of the code. Besides, we already have a schema versioning mechanism implemented in table  `qservMeta.QMetadata`:
```
MariaDB [qservMeta]> SELECT * FROM qservMeta.QMetadata;
+---------+-------+
| metakey | value |
+---------+-------+
| version | 1     |
+---------+-------+
```

core/modules/qmeta/QStatus.h
jgates108: Actually, no real need for that typedef.

core/modules/qmeta/QStats.h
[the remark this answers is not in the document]
iagaponenko: That's not quite true. Here is what we're supposed to do:
```
    int totalChunks = 0;
    int completedChunks = 0;
    std::time_t begin = 0;
    std::time_t lastUpdate = 0;
...
```
Besides, there is no need to explicitly initialize numeric variables with zeroes due to "zero initialization" guaranteed by all C++11 compliant compilers. See:
https://en.cppreference.com/w/cpp/language/zero_initialization

core/modules/qmeta/QStatus.h
andy-slac: Maybe rename it to QQueryStatus, QStatus sounds very non-specific?

PR #451, opened 2019-03-08
Tickets/dm 16788

iagaponenko:


core/modules/replica/ClusterHealthJob.cc
```diff
@@ -147,13 +147,43 @@ ClusterHealth const& ClusterHealthJob::clusterHealth() const {
...
 
+
+std::list<std::pair<std::string,std::string>> ClusterHealthJob::persistentLogData() const {
+
+    std::list<std::pair<std::string,std::string>> result;
+
+    auto&& health = clusterHealth();
+
+    for (auto&& entry: health.qserv()) {
+
```
n8pease: is this an extra (incorrect) line inside the for block? You do it again below, so I think it's intentional. I would not put it there (IMO), but does our coding standard say anything about it?

core/modules/replica/ConfigApp.h
```diff
@@ -256,210 +257,8 @@ class ConfigApp: public Application {
...
-        std::string const key         = "WORKER_FS_BUF_SIZE_BYTES";
-        std::string const description = "Buffer size for file and network operations at worker's file server.";
-        size_t            value;
-
-        void save(Configuration::Ptr const& config) {
-            if (value != 0) config->setWorkerFsBufferSizeBytes(value);
-        }
-    } _workerFsBufferSizeBytes;
+    /// General parameters
+    ConfigurationGeneralParams _general;
```
n8pease: nice; seems like a good improvement.

core/modules/replica/ConfigurationTypes.cc
```diff
@@ -0,0 +1,79 @@
...
+    result.push_back(::paramToJson(databaseHost,                config));
+    result.push_back(::paramToJson(databasePort,                config));
+    result.push_back(::paramToJson(databaseUser,                config));
+    result.push_back(::paramToJson(databasePassword,            config));
+    result.push_back(::paramToJson(databaseName,                config));
+    result.push_back(::paramToJson(databaseServicesPoolSize,    config));
+    result.push_back(::paramToJson(workerTechnology,            config));
+    result.push_back(::paramToJson(workerNumProcessingThreads,  config));
+    result.push_back(::paramToJson(fsNumProcessingThreads,      config));
+    result.push_back(::paramToJson(workerFsBufferSizeBytes,     config));
```
n8pease: does result offer any kind of `reserve` functionality like a `vector` (since it seems to borrow from the `vector` api...). And or static ctor `{...}` could be very cool.

core/modules/replica/ConfigurationTypes.h
```diff
@@ -0,0 +1,418 @@
...
+        }
+        unsigned int get(Configuration::Ptr const& config) const { return config->jobHeartbeatTimeoutSec(); }
+        std::string  str(Configuration::Ptr const& config) const { return std::to_string(get(config)); }
+
+    } jobHeartbeatTimeoutSec;
+
+    struct {
+
+        std::string const key         = "QSERV_AUTO_NOTIFY";
+        std::string const description = "Automatically notify Qserv on changes in replica disposition (0 disables this feature).";
```
n8pease: nitpick: this line is too long

core/modules/replica/ConfigurationTypes.h
```diff
@@ -0,0 +1,418 @@
...
+ */
+struct ConfigurationGeneralParams {
+
+    struct {
+
+        std::string const key         = "NET_BUF_SIZE_BYTES";
+        std::string const description = "The default buffer size for network communications.";
+        size_t            value;
+
+        bool const updatable = true;
```
n8pease: It's not clear to me (yet, anyway) how `updatable` is used. It seems that these are meant to be simple, lightweight, helper classes, but I wonder if some documentation is appropriate. Maybe you could explain the pattern a little more in the comment around line 45? If you gave a summary of the expected parameters & functions of each struct are and how it is to be used I think it would be helpful.

core/modules/replica/CreateReplicaJob.h
```diff
@@ -145,7 +145,12 @@ class CreateReplicaJob
      *
      * @see Job::extendedPersistentState()
      */
-    std::list<std::pair<std::string,std::string>> extendedPersistentState() const override;
+    std::list<std::pair<std::string,std::string>> extendedPersistentState() const final;
```
n8pease: I'm pretty sure `final` does not imply `override` (it does seem to imply `virtual`, but that's not exactly the same...). Do you really mean to remove `override`?

core/modules/replica/CreateReplicaJob.h
```diff
@@ -145,7 +145,12 @@ class CreateReplicaJob
...
+
+    /**
+     * @see Job::persistentLogData()
+     */
+    std::list<std::pair<std::string,std::string>> persistentLogData() const final;
```
n8pease: does this also want `override`? (since you reference a base class method in the documentation)

core/modules/replica/DatabaseServices.h
```diff
@@ -331,7 +558,156 @@ class DatabaseServices
...
+     * @param maxEntries
+     *   (optional) the maximum number of events to be reported. The default
+     *   values of 0 doesn't impose any limits.
+     *
+     * @return
+     *   collection of events found within the specified time interval
+     */
+    virtual std::list<ControllerEvent> readControllerEvents(std::string const& controllerId,
+                                                            uint64_t fromTimeStamp=0,
+                                                            uint64_t toTimeStamp=std::numeric_limits<uint64_t>::max(),
```
n8pease: nit: this line is too long (it makes the line wrapping weird in code review (I wish GitHub wouldn't wrap lines, and scrolled instead))

core/modules/replica/DatabaseServicesMySQL.cc
```diff
@@ -1,6 +1,6 @@
 /*
  * LSST Data Management System
- * Copyright 2017 LSST Corporation.
+ * Copyright 2019 LSST Corporation.
```
n8pease: are you correcting the date or updating it? ...if updating, I thought we were supposed to give a range of years, e.g. 2017-2019

core/modules/replica/CreateReplicaJob.h
n8pease: (I see other places you use `final` without override, I'm not going to mark it all up, just know that the question applies generally 😎 )

core/modules/replica/HttpProcessor.cc
```diff
@@ -168,217 +430,835 @@ void HttpProcessor::_getReplicationLevel(qhttp::Request::Ptr req,
...
+                    unsigned int const level = entry.first;
+                    size_t const numChunks   = entry.second;
+                    double const percent     = 100. * numChunks / numOnlineQservChunks;
+                    databaseJson["levels"][level]["qserv"]["online"]["num_chunks"] = numChunks;
+                    databaseJson["levels"][level]["qserv"]["online"]["percent"   ] = percent;
+                }
+                for (auto&& entry: allQservLevels) {
+                    unsigned int const level = entry.first;
+                    size_t const numChunks   = entry.second;
+                    double const percent     = 100. * numChunks / numAllQservChunks;
```
n8pease: we're sure there's no case (some kind of error condition or something? where `numAllQservChunks` can be 0?

core/modules/replica/HttpProcessor.cc
```diff
@@ -168,217 +430,835 @@ void HttpProcessor::_getReplicationLevel(qhttp::Request::Ptr req,
...
+    _debug(__func__);
     
-    // Update the cache
-    _replicationLevelReport = resultJson.dump();
-    _replicationLevelReportTimeMs = PerformanceUtils::now();
+    try {
+
+        // Extract optional parameters of the query
+
+        uint64_t const fromTimeStamp = ::getQueryParam(req->query, "from");
```
n8pease: I think the `::` tells the compiler to look the function up in the global namespace, is that right? As far as I can tell these functions are declared in this file in the anonymous namespace. Do you need the `::`? I'd like to understand this better.

core/modules/replica/HttpProcessor.h
```diff
@@ -131,119 +134,233 @@ class HttpProcessor
...
     /**
-     * Process "GET" requests for a list of resources of the given type
+     * Process a request which return the status of the replicas.
      *
      * @param req   request received from a client
      * @param resp  response to be sent back
      */
-    void _testList(qhttp::Request::Ptr req,
-                   qhttp::Response::Ptr resp);
+    void _getReplicationLevel(qhttp::Request::Ptr req,
```
n8pease: is "level" the same as "status", or would it be more clear name the function `_getReplicationStatus`?

core/modules/replica/RebalanceJob.h
```diff
@@ -83,9 +83,9 @@ struct RebalanceJobResult {
 
     // Parameters of the planner
 
-    size_t totalWorkers    {0};     // not counting workers which failed to report chunks
-    size_t totalGoodChunks {0};     // good chunks reported by the precursor job
-    size_t avgChunks       {0};     // per worker average
+    size_t totalWorkers    = 0;     /// not counting workers which failed to report chunks
+    size_t totalGoodChunks = 0;     /// good chunks reported by the precursor job
+    size_t avgChunks       = 0;     /// per worker average
```
n8pease: For my education; why change from {} to =?

core/modules/replica/ReplicateJob.cc
```diff
@@ -1,4 +1,4 @@
-/*
+    /*
  * LSST Data Management System
  * Copyright 2017 LSST Corporation.
```
n8pease: 2017-2019?

core/modules/replica/ReplicateJob.h
```diff
@@ -109,7 +109,7 @@ class ReplicateJob
     ReplicateJob(ReplicateJob const&) = delete;
     ReplicateJob& operator=(ReplicateJob const&) = delete;
 
-    ~ReplicateJob() = default;
+    ~ReplicateJob() final = default;
```
n8pease: I'm confused; the coliru compiler (g++ -std=c++11 -O2 -Wall -pedantic -pthread main.cpp && ./a.out) says it's an error to declare a member function `final` if it's not declared `virtual`. I reread the [cppreference page about 'final'](https://en.cppreference.com/w/cpp/language/final), it says
> When used in a virtual function declaration or definition, final ensures that the function is virtual and specifies that it may not be overridden by derived classes. The program is ill-formed (a compile-time error is generated) otherwise.
which I originally took to mean it makes the function virtual, but after reading it closer I think it means it _requires_ the function to be declared virtual by the programmer. But then, I wonder why our compiler is not complaining?

Also, I found one thread suggesting that since the effect of declaring the destructor `final` makes the class final then the more explicit thing to do is to declare the class itself final instead (and don't declare the dtor final). Did you consider this already, do you have any thoughts about it?

iagaponenko: Goof suggestion indeed to declare the whole class a `final`:
```
class Subclass final : public Base {
   ...
```

core/modules/replica/ClusterHealthJob.cc
iagaponenko: I don't think we have any specific coding standard for empty lines within the bodies of methods.

core/modules/replica/ConfigurationTypes.cc
iagaponenko: AFAIK, the `son` objects don't have method `reserve`. At least, I don't see this in the documentation:
https://github.com/nlohmann/json

core/modules/replica/testReplicaInfo.cc
```diff
@@ -0,0 +1,91 @@
...
+
+BOOST_AUTO_TEST_SUITE(Suite)
+
+BOOST_AUTO_TEST_CASE(ReplicaInfoTest) {
+
+    LOGS_INFO("ReplicaInfo test begins");
+
+    QservReplicaCollection one;
+    one.emplace_back(1U, "db1", 0);
+    one.emplace_back(1U, "db2", 0);   // only here
```
n8pease: "only here" not not very meaningful. Would it be worth expanding on this comment?

core/modules/replica/ConfigurationTypes.h
iagaponenko: These attributes are used by the CLI and Web UI tools to determine if the Configuration parameter is allowed to be updated before even attempting it ... failing. For the Web UI these parameters will be "grayed out" in the UI. In the CLI tools (`qserv-replica-config`) they won't be allowed to be modified.
I could put comments explaining these structures in the header file. This was a good suggestion.

n8pease: I think it would be helpful to have those comments.

core/modules/replica/ReplicateJob.h
n8pease: Goof(y) suggestion indeed 😆

core/modules/replica/CreateReplicaJob.h
iagaponenko: According to the C++11 documentation:
```
In both cases, virt-specifier-seq, if used, is either override or final, or final override or override final
```


See: https://en.cppreference.com/w/cpp/language/override

iagaponenko: Also see these examples which contradict to you theory:
https://en.cppreference.com/w/cpp/language/final

n8pease: per our discussion; the functions are marked virtual in the base class. It makes sense to me now why this works.

core/modules/replica/ConfigurationTypes.h
iagaponenko: Fixed

core/modules/replica/ConfigurationTypes.h
iagaponenko: Comments added.

core/modules/replica/DatabaseServicesMySQL.cc
iagaponenko: Good point. Fixed

core/modules/replica/DatabaseServices.h
iagaponenko: Fixed

core/modules/replica/HttpProcessor.cc
iagaponenko: Good catch! Reinforced

core/modules/replica/HttpProcessor.cc
iagaponenko: That's an interesting problem. Formally, symbols defined in the unnamed namespaces appear to a code as if they're global (`extern` is the real Evil in C++. AFAIK XRootD is the only code which still uses them in `qserv`), even though (technically) they aren't the same. That's why the compiler is happy with what I'm doing above. I'm also using prefix `::` to indicate that the symbol isn't defined in a scope of a class (or any of its base classes).

core/modules/replica/HttpProcessor.h
iagaponenko: The method returns a JSON object presenting numbers of chunks at various replication levels. I tried to avoid using 'status' in this context because this term is already heavily use in related contexts, including:
* a status of a replication request
* a status of a Job
* a status of the Replication system
Though, I would be happy to rename the method into its plural form `_getReplicationLevels`. This in turn would have another counter-argument - the REST APIs are typically modeled around objects (nouns), like:
```
/service/object`
/service/object/:id`
``` 
And HTTP methods (`GET`, `PUT`,'POST` and `DELETE`) used for operations over objects. This particular method is a handler of `GET` requests to the following REST URI  `/replication/v1/level`.

core/modules/replica/RebalanceJob.h
iagaponenko: According to the following section of the LSST C++ Style Guide:
"... Non-aggregate PODs should be constructed using an assignment-like syntax (note that this does not actually invoke the assignment operator):
```
double a = 4;           // good
double a{4};            // discouraged to maintain consistency
double a = {4};         // discouraged to maintain consistency
```
See:
https://developer.lsst.io/cpp/style.html#a-constructor-calls-should-use-c-98-syntax-when-possible

iagaponenko: That vague reason "...discouraged to maintain consistency..." is referring to consistency with Python 😆

core/modules/replica/ReplicateJob.cc
iagaponenko: Formally, yes, you're right! In reality we're long overdue to replace these repeated copyrights in all headers an CC files with one file per folder `Copyright` where all that legal stuff would be placed. This would reduce the amount of noise in the files. Someone should probably swallow the pain and migrate our code to that.

iagaponenko: Having said that, I would probably leave this as is for now and wait before the "Copyright migration"

core/modules/replica/testReplicaInfo.cc
iagaponenko: Fixed. For the record, these comments are referring to those elements of the collections which are only present on one collection and not in an opposite collection. These elements are used for testing an algorithm which finds differences between collections.

PR #452, opened 2019-03-14
Tickets/dm 18401

iagaponenko:


core/modules/replica/AddReplicaQservMgtRequest.h
```diff
@@ -131,6 +122,32 @@ class AddReplicaQservMgtRequest
...
+     *
+     * @param chunk
+     *   the chunk number
+     *
+     * @param databases
+     *   the names of databases
+     *
+     * @param onFinish
+     *   callback function to be called upon request completion
+     */
```
jgates108: Is this recommended somewhere? I preferred the previous comment block to this one as this one uses way more space than it needs. Pretty much all of the parameters are self explanatory, " serviceProvider reference to a provider of services" is pointless as documentation since the documentation says exactly the same thing as the code. Only the onFinish comment is somewhat useful. The first version only ate up 9 lines of header file instead of 18, which bothers me in that it takes that much longer to scan through the header files to find the important things.

fritzm: :100:  This kind of comment is USELESS DISTRACTING GARBAGE and FILLS ME WITH RAGE.  Write your code for PEOPLE to read.

fritzm: I will also point out the raging internal consistency between being bothered by license headers but not by this kind of faffery.  How can you sleep at night?

iagaponenko: I don't care much about "space utilization", it's free. Putting parameter descriptions (which is permitted by Oxygen) onto a separate line fixed a need to align descriptions with was causing me troubles in the previous version of the comments.
Regarding self-explanatory, you may be right here. However, I need to document all parameters, not just one.

core/modules/replica/ApplicationTypes.h
```diff
@@ -1,6 +1,5 @@
 /*
  * LSST Data Management System
- * Copyright 2017 LSST Corporation.
  *
```
jgates108: I think you should check with KT (or someone familiar with the copyright rules) to make sure this meets up with LSST  copyright requirements in all cases.

core/modules/replica/AddReplicaQservMgtRequest.h
fritzm: Many linefeeds gave their lives...

iagaponenko: To @fritzm , I always have a habit to extensively comment my code, even if that might seem an overkill to others. I do this for many years ... and for myself. Later, when I'm looking at a code I usually first look at its structure, then I  always read comments, and then (if needed) read what that code does. This helps me to remember details after I forget it (which is a case for all of us).
As far as "faffery" is concerned, adding comments to the code is really cheep compared to the time & efforts spent to understand which code to write how to write it.

fritzm: I have no problem with comments.  I have issue with "content-free" comments, and formatting that makes it harder for me to have more code on the screen at once.

fritzm: I want to read the _code_.  I want comments when the code is tricky or non-obvious (hopefully not that common!)

iagaponenko: Regular expressions inside an IDE do the job. For example (in Netbeans), this pattern:
```
@param (.+)\s-\s(.+)$
```
And the replacement:
```
@param $1\n     *   $2
```

fritzm: I glad to hear this won't be hard for you to fix! :-)

core/modules/replica/ApplicationTypes.h
iagaponenko: KT has already posted notes on tis subject last Monday. So, the copyright section has been removed from files as per the following guidelines:
https://developer.lsst.io/stack/license-and-copyright.html

core/modules/replica/AddReplicaQservMgtRequest.h
fritzm: (Don't take me too seriously.  This objection is not "blocking" for review.  I just respect you less.)

iagaponenko: I'm glad that I still have some credit of respect 😄 
To be serious, when it comes to the code I tend to be consistent in a way I write it which means: 1) always explaining each class/struct, 2) commenting non-trivial methods, 3) when commenting a method then I comment all parameters o the method, 4) commenting non-trivial members of a class/struct, 5) commenting statements or groups of statements in method bodies when it's not clear what they do (I just do not make an assumption that people should waste there tim trying to figure what that code is meant to do).
I do value the "self-explanatory" approach to naming identifiers. However this has its own limits.

iagaponenko: In the end, tis effort was to make everything look consistent (I had a few different styles in the previous version of the code, which was bothering me).

fritzm: There is something to be said for consistency...  I've taken more than enough time/space now to soapbox my opinion.  Peace, friend!

core/modules/replica/FindAllRequest.h
```diff
@@ -128,56 +166,41 @@ class FindAllRequest
...
      * Callback handler for the asynchronous operation
      *
      * @param ec - error code to be checked
      */
-    void awaken(boost::system::error_code const& ec);
+    void _awaken(boost::system::error_code const& ec);
 
     /**
      * Send the serialized content of the buffer to a worker
      *
```
jgates108: I don't think these blank line really help readability. The '@' makes a pretty distinctive visual break by itself.

core/modules/replica/StopRequestBase.h
```diff
@@ -107,45 +110,43 @@ class StopRequestBase
...
+     * Initiate request-specific operation with the persistent state
+     * service to store replica status.
+     *
+     * This method must be implemented by subclasses.
+     */
+    virtual void saveReplicaInfo() = 0;
+
+    /**
+     * @see Request::savePersistentState()
+     */
```
jgates108: I think "/// @see Request::savePersistentState()" would work just as well and takes up less space. This goes for most places you have @see.

core/modules/replica/VerifyJob.h
```diff
@@ -248,23 +238,44 @@ class VerifyJob
...
+    void _onRequestFinish(FindRequest::Ptr const& request);
 
     /**
      * Find the next replicas to be inspected.
      *
-     * @param lock        - the lock must be acquired by a caller of the method
-     * @param replicas    - a collection of replicas returned from the database
-     * @param numReplicas - a desired number of replicas to be pulled from the database
+     * @param lock
+     *   the lock must be acquired by a caller of the method
```
jgates108: For all of these lock comments, the really important thing is the name of the underlying mutex. It's probably worth putting the name of the mutex in the variable name or something as passing a Lock with the wrong mutex could make debugging more difficult.

core/modules/replica/tools/qserv-replica-job-verify.cc
```diff
@@ -31,14 +30,15 @@
 // Qserv headers
 #include "replica/VerifyApp.h"
 
+using namespace std;
```
n8pease: I wonder if this should be alphabetized, and go below `using namespace lsst...`?

core/modules/replica/testConfiguration.cc
```diff
@@ -268,77 +267,77 @@ BOOST_AUTO_TEST_CASE(ConfigurationTest) {
...
-        std::vector<std::string> databases3 = config->databases("test");
-        std::sort(databases3.begin(), databases3.end());
+        vector<string> databases3 = config->databases("test");
+        sort(databases3.begin(), databases3.end());
         BOOST_CHECK(databases3.size() == 2);
-        BOOST_CHECK(databases3 == std::vector<std::string>({"db4","db5"}));
+        BOOST_CHECK(databases3 == vector<string>({"db4","db5"}));
     
-        for (auto&& name: std::vector<std::string>({"db1","db2","db3","db4","db5"})) {
+        for (auto&& name: vector<string>({"db1","db2","db3","db4","db5"})) {
```
n8pease: I'd prefer spaces after commas, I don't know if we have a coding standard rule about it tho.

core/modules/replica/WorkerServerConnection.h
```diff
@@ -183,22 +195,25 @@ class WorkerServerConnection
...
+     *
+     * @param bytes_transferred
+     *   the number of bytes sent (if successful)
      */
-    void sent(boost::system::error_code const& ec,
-              size_t bytes_transferred);
+    void _sent(boost::system::error_code const& ec,
+               size_t bytes_transferred);
 
 private:
```
n8pease: this `private` is redundant.

core/modules/replica/testConfiguration.cc
iagaponenko: (this is also a note to myself for the future reviews) I think the review should be based on adopted conventions (a way most of the code is written in package `qserv`), common standards (including LSST Style Guide, etc.), and manager's requirements... rather than on personal preferences of reviewers (unless the reviewer is the manager) 😸

Seriously speaking, for the former (the standards) I've found such requirement in the Style Guide:
```
Commas should be followed by a white space.
```
Details in:
https://developer.lsst.io/cpp/style.html#white-space
So, I will fix the code accordingly.

core/modules/replica/WorkerServerConnection.h
iagaponenko: Fixed

core/modules/replica/tools/qserv-replica-job-verify.cc
iagaponenko: Good question. Though, I've found nothing on this subject in the Style Guide:
https://developer.lsst.io/cpp/style.html#appendix-on-using-using

core/modules/replica/VerifyJob.h
iagaponenko: Good suggestion! Thank you!

core/modules/replica/AddReplicaQservMgtRequest.h
iagaponenko: @jgates108 

Regarding 1) earlier (beginning of this year, or end of the last year) I was already criticized by one of reviewers that I didn't document all parameters of a method (I missed one or two). So, now I'm hearing a conflicting requirement. I would rather stick to the current model of documenting all of them.

Regarding 2), your are making a good point on a possibility of changing a type of chunk identifiers. Coincidently I already have this in my TODO list. Leaving `@param chunk` w/o any description is fine with me. Though, I would rather revisit each method (at lest factory methods or constructors) where the chunk identifier are passed and expand parameter descriptions with an intended purpose of passing those numbers. For example: "an identifier of a chunk to be deleted from a worker" where it's not clear from a context why the chunk number is needed here.

Regarding 3) the Replication System was actually designed as an extendable framework which can be modified by anyone of us (or not just by us - by the people who'll maintain this software in many years from now). Having some in-code documentation (I hope) will help. It also helps to myself in the future should I need to modify this code. My memory is fading quicker than I want it to 😄 

Anyway, thank you very much for the good comments and suggestions!

core/modules/replica/ChunkLocker.cc
```diff
@@ -149,23 +157,24 @@ bool ChunkLocker::releaseImpl(util::Lock const& mLock,
...
     // Get rid of chunks owned by the specified owner, and also collect
     // those (removed) chunks into a vector to be returned to a caller.
 
     OwnerToChunks owner2chunks;
-    lockedImpl(mLock,
-               owner,
-               owner2chunks);
+    _lockedImpl(mLock,
+                owner,
+                owner2chunks);
```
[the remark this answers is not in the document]
iagaponenko: It makes it easier to see all parameters of a call and the order they're passed into the method. I do realize this increases the number of lines. But I made this tradeoff in favor of more readable code.

core/modules/replica/ChunkLocker.h
```diff
@@ -119,47 +125,60 @@ class ChunkLocker {
...
-     * @param owner - reference to a string which will be initialized with
-     *                an identifier of an owner which had a claim on the chunk
-     *                at a time of the method call
+     * @param chunk
+     *   chunk to be released
+     *
+     * @param owner
+     *   reference to a string which will be initialized with
+     *   an identifier of an owner which had a claim on the chunk
+     *   at a time of the method call
```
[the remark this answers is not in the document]
iagaponenko: Yes, the string will be modified by the method upon a successful completion of the method. That's why it's not `const`. Since you're asking this question then my description didn't make it clear. I will fix it.

core/modules/replica/ConfigurationMySQL.cc
```diff
@@ -1014,12 +1015,12 @@ void ConfigurationMySQL::loadConfigurationImpl(util::Lock const& lock,
...
-void ConfigurationMySQL::_setImp(std::string const& category,
-                                 std::string const& param,
+void ConfigurationMySQL::_setImp(string const& category,
+                                 string const& param,
                                  SetValueExprFunc const& setValueExprFunc,
-                                 std::function<void()> const& onSuccess) {
+                                 function<void()> const& onSuccess) {
 
-    std::string const context_ = context() + std::string(__func__);
+    string const context_ = context() + string(__func__);
```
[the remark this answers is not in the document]
iagaponenko: Absolutely agree with this statement! Will fix it.

core/modules/replica/ConfigurationMySQL.cc
```diff
@@ -30,57 +29,61 @@
...
     row.get("value", value);
     return true;
 }
 
+
 template <typename T>
 void readMandatoryParameter(database::mysql::Row& row,
-                            std::string const&    name,
+                            string const&    name,
                             T&                    value) {
```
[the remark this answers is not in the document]
iagaponenko: Fixed

core/modules/replica/EchoRequest.h
```diff
@@ -124,56 +160,42 @@ class EchoRequest
...
-
-    /**
-     * @see Request::savePersistentState()
-     */
-    void savePersistentState(util::Lock const& lock) final;
-
-    /**
-     * @see Request::extendedPersistentState()
+     * @param success
+     *   flag indicating if the response succeeded
```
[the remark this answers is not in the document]
iagaponenko: Good suggestion! Fixed.

core/modules/replica/StopRequestBase.h
iagaponenko: Good point! Fixed

core/modules/replica/FindAllRequest.h
iagaponenko: noted

core/modules/replica/ConfigurationStore.cc
```diff
@@ -45,24 +46,25 @@ using namespace lsst::qserv;
  *
  * The function may throw the following exceptions:
  *
- *   std::bad_lexical_cast
+ *   bad_lexical_cast
```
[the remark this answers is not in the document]
iagaponenko: Fixed

core/modules/replica/ConfigurationStore.cc
```diff
@@ -249,47 +262,47 @@ DatabaseFamilyInfo ConfigurationStore::addDatabaseFamily(DatabaseFamilyInfo cons
     util::Lock lock(_mtx, context() + __func__);
     
     if (info.name.empty()) {
-        throw std::invalid_argument(context() + std::string(__func__) + "  the family name can't be empty");
+        throw invalid_argument(context() + string(__func__) + "  the family name can't be empty");
```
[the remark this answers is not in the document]
iagaponenko: Good point! Fixed

core/modules/replica/Controller.cc
```diff
@@ -746,15 +765,16 @@ ServiceSuspendRequest::Ptr Controller::suspendWorkerService(
...
 ServiceResumeRequest::Ptr Controller::resumeWorkerService(
-                                    std::string const& workerName,
+                                    string const& workerName,
                                     ServiceResumeRequest::CallbackType const& onFinish,
-                                    std::string const& jobId,
+                                    string const& jobId,
                                     unsigned int requestExpirationIvalSec) {
 
-    LOGS(_log, LOG_LVL_DEBUG, context() << "resumeWorkerService  workerName: " << workerName);
+    LOGS(_log, LOG_LVL_DEBUG, _context() << string(__func__) << "  workerName: " << workerName);
```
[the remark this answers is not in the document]
iagaponenko: Fixed!

core/modules/replica/ConfigurationStore.cc
```diff
@@ -499,20 +512,20 @@ void ConfigurationStore::loadConfiguration(util::ConfigStore const& configStore)
...
-    for (std::string const& name: workers) {
+    for (string const& name: workers) {
 
-        std::string const section = "worker:" + name;
+        string const section = "worker:" + name;
         if (_workerInfo.count(name)) {
-            throw std::range_error(
-                    "ConfigurationStore::" + std::string(__func__) + "  duplicate worker entry: '" +
+            throw range_error(
+                    "ConfigurationStore::" + string(__func__) + "  duplicate worker entry: '" +
```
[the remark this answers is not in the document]
iagaponenko: Agreed. Refactored.

core/modules/replica/Controller.cc
```diff
@@ -765,15 +785,16 @@ ServiceResumeRequest::Ptr Controller::resumeWorkerService(
...
 ServiceStatusRequest::Ptr Controller::statusOfWorkerService(
-                                    std::string const& workerName,
+                                    string const& workerName,
                                     ServiceStatusRequest::CallbackType const& onFinish,
-                                    std::string const& jobId,
+                                    string const& jobId,
                                     unsigned int requestExpirationIvalSec) {
 
-    LOGS(_log, LOG_LVL_DEBUG, context() << "statusOfWorkerService  workerName: " << workerName);
+    LOGS(_log, LOG_LVL_DEBUG, _context() << string(__func__) << "  workerName: " << workerName);
```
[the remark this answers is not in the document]
iagaponenko: Fixed

core/modules/replica/ControllerApp.h
```diff
@@ -68,14 +67,16 @@ class ControllerApp: public Application {
...
+private:
 
     /**
-     * @see Application::runImpl()
+     * @see ControllerApp::create()
      */
-    int runImpl() final;
+    ControllerApp(int argc, char* argv[]);
 
 private:
```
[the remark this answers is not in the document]
iagaponenko: Fixed everywhere

core/modules/replica/CreateReplicaJob.cc
```diff
@@ -199,29 +205,28 @@ void CreateReplicaJob::startImpl(util::Lock const& lock) {
...
             chunk(),
             destinationWorker(),
             databaseFamily());
 
-    } catch (std::invalid_argument const& ex) {
+    } catch (invalid_argument const& ex) {
 
-        LOGS(_log, LOG_LVL_ERROR, context() << "startImpl  "
-             << "** misconfigured application ** "
+        LOGS(_log, LOG_LVL_ERROR, context() << string(__func__) << "  ** MISCONFIGURED ** "
```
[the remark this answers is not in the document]
iagaponenko: Fixed everywhere

core/modules/replica/DatabaseMySQL.cc
```diff
@@ -49,7 +50,7 @@ namespace replica {
 namespace database {
 namespace mysql {
 
-std::atomic<size_t> Connection::_nextId{0};
+atomic<size_t> Connection::_nextId{0};
```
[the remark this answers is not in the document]
iagaponenko: Unfortunately, this won't work for `atomic` types due to:
```
core/modules/replica/DatabaseMySQL.cc:53:38: error: use of deleted function 'std::atomic<long unsigned int>::atomic(const std::atomic<long unsigned int>&)'
 atomic<size_t> Connection::_nextId = 0;
```

core/modules/replica/ConfigurationStore.cc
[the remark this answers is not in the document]
andy-slac: Should it be `boost::bad_lexical_cast`?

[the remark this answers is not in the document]
iagaponenko: Good catch! Fixed.

PR #453, opened 2019-03-26
Tickets/dm 17812

iagaponenko:


core/modules/replica/SqlApp.cc
```diff
@@ -0,0 +1,171 @@
...
+
+    parser().flag(
+        "all-workers",
+        "The flag for selecting all workers regardless of their status (DISABLED or READ-ONLY).",
+        _allWorkers);
+
+    parser().option(
+        "max-rows",
+        "The maximum number of rows to be pulled from result set at workers"
+        " whene processing queries. NOTE: This parameter nothing to do with"
```
n8pease: when

n8pease: this parameter *has* nothing to do

core/modules/replica/SqlApp.cc
```diff
@@ -0,0 +1,171 @@
...
+        [&finished] (SqlJob::Ptr const& job) {
+            finished = true;
+        }
+    );
+    job->start();
+
+    util::BlockPost blockPost(1000,2000);
+    while (not finished) {
+        blockPost.wait();
+    }
```
n8pease: spin lock :-( 
there's not a better way?

core/modules/replica/SqlApp.cc
```diff
@@ -0,0 +1,171 @@
...
+    // Analyze and display results for each worker
+
+    auto const& resultData = job->getResultData();
+    for (auto&& itr : resultData.resultSets) {
+        auto&& worker = itr.first;
+        auto&& resultSet = itr.second;
+
+        bool const succeeded = resultData.workers.at(worker);
+        if (not succeeded) {
+            cout << "worker: " << worker << "  error: " << resultSet.error << endl;
```
n8pease: personally I'd put a comma or other delimiter after worker, ie `", error: "`, ymmv.

core/modules/replica/SqlJob.cc
```diff
@@ -0,0 +1,260 @@
...
+namespace replica {
+
+string SqlJob::typeName() { return "SqlJob"; }
+
+
+Job::Options const& SqlJob::defaultOptions() {
+    static Job::Options const options{
+        2,      /* priority */
+        false,  /* exclusive */
+        true    /* exclusive */
```
n8pease: exclusive is false and true? 🙃 
maybe you can elaborate on these comments?

n8pease: FWIW I *do* appreciate the comments on opaque variables. (what does true or false mean in these contexts? who knows? 😆).

core/modules/replica/SqlJob.h
```diff
@@ -0,0 +1,223 @@
...
+     * @param allWorkers
+     *   engage all known workers regardless of their status. If the flag
+     *   is set to 'false' then only 'ENABLED' workers which are not in
+     *   the 'READ-ONLY' state will be involved into the operation.
+     *
+     * @param controller
+     *   is needed launching requests and accessing the Configuration
+     *
+     * @param parentJobId
+     *   (optional) identifier of a parent job
```
n8pease: what's the opt-out value?

core/modules/replica/SqlJob.h
```diff
@@ -0,0 +1,223 @@
...
+     *
+     * @param onFinish
+     *   a callback function to be called upon a completion of the job
+     *
+     * @param options
+     *   defines the job priority, etc.
```
n8pease: looks like this one is optional too. The other is not maybe?

core/modules/replica/DatabaseMySQL.cc
```diff
@@ -423,13 +423,13 @@ void Connection::exportField(ProtocolResponseSqlField* ptr,
...
-    ptr->set_db(        field.db,        field.db_length);
-    ptr->set_catalog(   field.catalog,   field.catalog_length);
-    ptr->set_def(       field.def,       field.def_length);
+    ptr->set_name(      field.name/*,      field.name_length*/);
+    ptr->set_org_name(  field.org_name/*,  field.org_name_length*/);
+    ptr->set_table(     field.table/*,     field.table_length*/);
+    ptr->set_org_table( field.org_table/*, field.org_table_length*/);
+    ptr->set_db(        field.db/*,        field.db_length*/);
+    ptr->set_catalog(   field.catalog/*,   field.catalog_length*/);
+    ptr->set_def(       field.def/*,       field.def_length*/);
```
n8pease: it looks like this block was changed in the previous commit, and then changed back here. You could just `fixup` this commit into the previous one and not have this change. Or: why is there commented code here, can it go away?

core/modules/replica/ControllerApp.cc
```diff
@@ -50,6 +55,98 @@ string const description =
...
+        for (size_t columnIdx = 0; columnIdx < numColumns; ++columnIdx) {
+            auto tableColumnPtr = make_shared<vector<string>>();
+            tableColumnPtr->reserve(numRows);
+            tableDataByColumns.push_back(tableColumnPtr);
+        }
+        for (auto const& row :  result.rows) {                
+            for (size_t columnIdx = 0; columnIdx < numColumns; ++columnIdx) {
+
+                auto const& field  = result.fields[columnIdx];
+                auto const& cell   = row   .cells [columnIdx];
```
n8pease: I know you like to align things vertically but IMO the space in `row   .cells` goes too far. I'd probably accept
    auto const& field  = result.fields[columnIdx];
    auto const& cell   =     row.cells [columnIdx];

core/modules/replica/ControllerApp.cc
```diff
@@ -50,6 +55,98 @@ string const description =
...
+
+                auto const& tableColumnPtr = tableDataByColumns[columnIdx];
+                if (row.nulls[columnIdx]) {
+                    tableColumnPtr->push_back("NULL");
```
n8pease: `const` tho. (line 87). I hate that about const shared_ptr.

core/modules/replica/ControllerApp.cc
```diff
@@ -50,6 +55,98 @@ string const description =
...
+                } else {
+                    if (sqlShowType) {
+                        string typeName ="UNKNOWN MYSQL TYPE:" + to_string(field.type);
+                        switch (field.type) {
```
n8pease: no default?

core/modules/replica/sql/replication_config_ncsa.sql
```diff
@@ -47,36 +47,36 @@ INSERT INTO `config` VALUES ('worker', 'db_password',                'CHANGEME')
...
-INSERT INTO `config_worker` VALUES ('db25', 1, 0, 'lsst-qserv-db25',  NULL, 'lsst-qserv-db25',  NULL, NULL, 'lsst-qserv-db25', NULL, NULL, NULL);
-INSERT INTO `config_worker` VALUES ('db26', 1, 0, 'lsst-qserv-db26',  NULL, 'lsst-qserv-db26',  NULL, NULL, 'lsst-qserv-db26', NULL, NULL, NULL);
-INSERT INTO `config_worker` VALUES ('db27', 1, 0, 'lsst-qserv-db27',  NULL, 'lsst-qserv-db27',  NULL, NULL, 'lsst-qserv-db27', NULL, NULL, NULL);
-INSERT INTO `config_worker` VALUES ('db28', 1, 0, 'lsst-qserv-db28',  NULL, 'lsst-qserv-db28',  NULL, NULL, 'lsst-qserv-db28', NULL, NULL, NULL);
-INSERT INTO `config_worker` VALUES ('db29', 1, 0, 'lsst-qserv-db29',  NULL, 'lsst-qserv-db29',  NULL, NULL, 'lsst-qserv-db29', NULL, NULL, NULL);
-INSERT INTO `config_worker` VALUES ('db30', 1, 0, 'lsst-qserv-db30',  NULL, 'lsst-qserv-db30',  NULL, NULL, 'lsst-qserv-db30', NULL, NULL, NULL);
+INSERT INTO `config_worker` VALUES ('db01', 1, 0, 'lsst-qserv-db01',  NULL, 'lsst-qserv-db01',  NULL, NULL, '127.0.0.1', NULL, NULL, NULL);
+INSERT INTO `config_worker` VALUES ('db02', 1, 0, 'lsst-qserv-db02',  NULL, 'lsst-qserv-db02',  NULL, NULL, '127.0.0.1', NULL, NULL, NULL);
```
n8pease: just making sure you want to change to localhost. maybe you want to define a constant for that? Probably not necessary...

core/modules/replica/SqlApp.cc
iagaponenko: Fixed

core/modules/replica/SqlJob.cc
iagaponenko: Fixed the comment. In general, I agree with your observation. In this specific case the function creates an object of the following type:
```
# file: replica/Job.h
class Job {

    /// The job options container
    struct Options {

        /// the priority level
        int  priority;

        /// the flag indicating of this job can't be run simultaneously along with other jobs.
        bool exclusive;

        /// the flag indicating if the job is allowed to be interrupted by other jobs.
        bool preemptable;
    };
    ...
```
Right now, the only meaningful parameter here is the `priority`, which affects an order in which requests are processed by the worker services. Others were reserved for a (non-existing yet) job scheduler. And I'm not sure if I want to have the one in the Replication system. That scheduler is like Buttler or SuperTask - something which would be a universal solution for everything 😄

core/modules/replica/SqlJob.h
iagaponenko: It's supposed to be `std::string()`. I fixed this in the contract of the method.

core/modules/replica/DatabaseMySQL.cc
iagaponenko: I'm sorry, the commit structure is not perfect! But I'm working on it (making my commits as perfect as yours). The commented blocks were removed later.

core/modules/replica/ControllerApp.cc
iagaponenko: Good suggestion! Honestly., I'm still undecided if I should keep aligning my code like I do it now, instead of just "streaming" it. I've doing it both. Apart from obvious benefits of the alignment, it also causes causes troubles when adding more lines. Then I may need to realign the rest of this "table".

core/modules/replica/ControllerApp.cc
iagaponenko: Agree. To avoid this confusion I should probably do:
```
auto& tableColumnPtr = tableDataByColumns[columnIdx];
```

core/modules/replica/ControllerApp.cc
iagaponenko: This was fixed in another commit.

core/modules/replica/Common.h
```diff
@@ -114,7 +114,7 @@ struct ReplicationRequestParams {
     unsigned int chunk;
     std::string  sourceWorker;
 
-    ReplicationRequestParams();
+    ReplicationRequestParams() = default;
```
n8pease: The default ctor used to initialize member variables, are you sure it's safe to not be doing that? Maybe you want to init them in the body of the class if you're going to `default` the ctor? (Check the other classes you changed too, I won't add noisy comments on all of them)

core/modules/replica/ControllerApp.cc
iagaponenko: Though, this was fixed in a later commit.

core/modules/replica/Controller.h
```diff
@@ -378,6 +378,60 @@ class Controller : public std::enable_shared_from_this<Controller> {
...
+     *
+     * @param user
+     *   the name of a database account for connecting to the database service
+     *
+     * @param password
+     *   a database for connecting to the database service
+     *
+     * @param maxRows
+     *   (optional) limit for the maximum number of rows to be returned with the request.
+     *   Laving the default value of the parameter to 0 will result in not imposing any
```
n8pease: Leaving?

core/modules/replica/Controller.h
```diff
@@ -378,6 +378,60 @@ class Controller : public std::enable_shared_from_this<Controller> {
...
+     *   explicit restrictions on a size of the result set. NOte that other, resource-defined
```
n8pease: NOte

core/modules/replica/sql/replication_config_ncsa.sql
iagaponenko: This is within an SQL DML file. And this value is forced by the DDL file `replica/sql/replication.sql` where the filed `db_host` is defined as `NOT NULL `:
```
CREATE TABLE IF NOT EXISTS `config_worker` (

  `name`         VARCHAR(255)       NOT NULL ,      -- the name of the worker

  `is_enabled`   BOOLEAN            NOT NULL ,      -- is enabled for replication
  `is_read_only` BOOLEAN            NOT NULL ,      -- a subclass of 'is_enabled' which restricts use of
                                                    -- the worker for reading replicas. No new replicas can't be
                                                    -- placed onto this class of workers.

  `svc_host`     VARCHAR(255)       NOT NULL ,      -- the host name on which the worker server runs
  `svc_port`     SMALLINT UNSIGNED  DEFAULT NULL ,  -- override for the global default

  `fs_host`      VARCHAR(255)       NOT NULL ,      -- the host name on which the built-in FileServer runs
  `fs_port`      SMALLINT UNSIGNED  DEFAULT NULL ,  -- override for the global default

  `data_dir`     VARCHAR(255)       DEFAULT NULL ,  -- a file system path to the databases

  -- Connection parameters for the Qserv worker database

  `db_host`      VARCHAR(255)       NOT NULL ,      -- the host name on which the worker database server runs
  `db_port`      SMALLINT UNSIGNED  DEFAULT NULL ,  -- override for the global default
  `db_user`      VARCHAR(255)       DEFAULT NULL ,  -- override for the global default
  `db_password`  VARCHAR(255)       DEFAULT NULL ,  -- override for the global default

  PRIMARY KEY (`name`) ,

  UNIQUE  KEY (`svc_host`, `svc_port`) ,
  UNIQUE  KEY (`fs_host`,  `fs_port`) ,
  UNIQUE  KEY (`db_host`,  `db_port`)
)
ENGINE = InnoDB;
```
The default value (the 'constant') which might address this could be defined either via a common default here:
```
...
INSERT INTO `config` VALUES ('worker', 'svc_port',                   '25000');
INSERT INTO `config` VALUES ('worker', 'fs_port',                    '25001');
INSERT INTO `config` VALUES ('worker', 'data_dir',                   '/qserv/data/mysql');
INSERT INTO `config` VALUES ('worker', 'db_port',                    '3306');
INSERT INTO `config` VALUES ('worker', 'db_user',                    'root');
...
```
Or, perhaps in the above mentioned table schema:
```
  `db_host`      VARCHAR(255)       DEFAULT '127.0.0.1. ,      -- the host name on which the worker database server runs
```

iagaponenko: Though, I tried to avoid this kind of defaults in the general DDL schema files.

core/modules/replica/Common.h
iagaponenko: This is safe because what I was doing earlier in the removed constructor was doing what the C++11 are supposed to do by the Standard:
- initializing strings with empty strings
- zero value initialization of the numeric fields
These are the members:
```
    int          priority;
    std::string  database;
    unsigned int chunk;
    std::string  sourceWorker;
```

core/modules/replica/SqlRequest.cc
```diff
@@ -0,0 +1,334 @@
...
+    _responseData.set(message);
+
+    // Extract target request type-specific parameters from the response
+    if (message.has_request()) {
+        _targetRequestParams = SqlRequestParams(message.request());
+    }
+    switch (message.status()) {
+
+        case ProtocolStatus::SUCCESS:
+
```
n8pease: the blank line here is inconsistent with other cases in the switch.

core/modules/replica/Common.h
iagaponenko: The default version of the c-tor would effectively do:
```
    int          priority = 0;
    std::string  databaser = std::string();
    unsigned int chunk = 0U;
    std::string  sourceWorker = std::string();
```

core/modules/replica/SqlRequest.cc
iagaponenko: 👍

core/modules/replica/SqlJob.h
iagaponenko: Yes, it is. Fixed.

core/modules/replica/SqlResultSet.h
```diff
@@ -0,0 +1,122 @@
...
+
+    /// The row type of a result set. The number of elements in each row
+    /// must match the number of fields.
+    struct Row {
+
+        /// Values at the cells
+        std::vector<std::string> cells;
+
+        /// Flags indicating if the corresponding values of the cells represent SQL NULL
+        std::vector<bool> nulls;
```
n8pease: beware! `vector<bool>` is not your standard vector. Scott Meyers has a good writeup on this in one of his "Effective C++" books. You might/should consider a different container type.

core/modules/replica/SqlApp.cc
iagaponenko: Potentially, I could use `cv`. But this may not worth the effort here. The `BlockPost` allows me to print periodic 'heartbeats' while waiting for the completion of the job.

core/modules/replica/WorkerSqlRequest.cc
```diff
@@ -48,56 +52,61 @@ WorkerSqlRequest::Ptr WorkerSqlRequest::create(ServiceProvider::Ptr const& servi
...
-    // Return the performance of the target request
+    // Update the performance of the target request before returning it
+    _response.set_allocated_target_performance(performance().info());
 
-    response.set_allocated_target_performance(performance().info());
-
-    // TODO: implement extracting and setting a result set
-    //response.set_data(data());
+    // Note, we shall not invalidate the local state of the cached response
+    // object because, by a design of the Replication Framework,  the state
```
n8pease: IMO I'd prefer a single space after the comma than have the end of the word `state` line up with the end of the word `response` 🙃

core/modules/replica/SqlResultSet.h
iagaponenko: Him. That's interesting. I heard something on this subject long time ago, but I didn't pay attention to tis problem. And it looks like old issues are still relevant on a context of C++11 (or later versions) as per:
https://en.cppreference.com/w/cpp/container/vector_bool

I will redesign tis code to get rid of `bool`

iagaponenko: Thank you!

core/modules/replica/EchoRequest.cc
```diff
@@ -83,7 +83,7 @@ EchoRequest::EchoRequest(ServiceProvider::Ptr const& serviceProvider,
                          shared_ptr<Messenger> const& messenger)
     :   RequestMessenger(serviceProvider,
                          io_service,
-                         "REPLICA_ECHO",
+                         "ECHO",
```
n8pease: didn't you make this TEST_ECHO in WorkerEchoRequest? Do you want to be consistent?

core/modules/replica/WorkerProcessor.cc
```diff
@@ -690,47 +690,23 @@ void WorkerProcessor::_setServiceResponseInfo(
...
+    if (nullptr != dynamic_pointer_cast<WorkerReplicationRequest>(request)) {
+        info->set_queued_type(ProtocolQueuedRequestType::REPLICA_CREATE);
+    } else if (nullptr != dynamic_pointer_cast<WorkerDeleteRequest>(request)) {
+        info->set_queued_type(ProtocolQueuedRequestType::REPLICA_DELETE);
+    } else if (nullptr != dynamic_pointer_cast<WorkerFindRequest>(request)) {
+        info->set_queued_type(ProtocolQueuedRequestType::REPLICA_FIND);
+    } else if (nullptr != dynamic_pointer_cast<WorkerFindAllRequest>(request)) {
+        info->set_queued_type(ProtocolQueuedRequestType::REPLICA_FIND_ALL);
+    } else if (nullptr != dynamic_pointer_cast<WorkerEchoRequest>(request)) {
+        info->set_queued_type(ProtocolQueuedRequestType::TEST_ECHO);
```
n8pease: all this dynamic_casting is not great I think. It would be good if there was a way to leverage the polymorphism to make a single call to get the proper ProtocolQueuedRequestType.

core/modules/replica/Performance.h
```diff
@@ -22,31 +22,24 @@
...
-
 // This header declarations
 
 namespace lsst {
 namespace qserv {
 namespace replica {
 
+// Forward declarations
+
+class ProtocolPerformance;
```
n8pease: the way the forward declarations were is consistent with forward declarations in qserv.

core/modules/replica/WorkerProcessor.cc
```diff
@@ -46,34 +46,33 @@ namespace {
...
 bool ifDuplicateRequest(PROTOCOL_RESPONSE_TYPE& response,
-                        replica::WorkerRequest::Ptr const& p,
+                        WorkerRequest::Ptr const& p,
                         PROTOCOL_REQUEST_TYPE const& request) {
 
     bool isDuplicate = false;
 
-    if (replica::WorkerReplicationRequest::Ptr ptr =
-        dynamic_pointer_cast<replica::WorkerReplicationRequest>(p)) {
+    if (WorkerReplicationRequest::Ptr ptr =
```
n8pease: `auto` would work here and then you wouldn't have to line wrap which would be nice. 
implicit pointer conversion to bool is [disallowed](https://developer.lsst.io/cpp/style.html#implicit-test-for-0-should-not-be-used-other-than-for-boolean-variables)

core/modules/replica/testConfiguration.cc
```diff
@@ -89,6 +92,10 @@ BOOST_AUTO_TEST_CASE(ConfigurationTest) {
         {"worker:worker-A.fs_host",      "host-A"},
         {"worker:worker-A.fs_port",      "52001"},
         {"worker:worker-A.data_dir",     "/data/A"},
+        {"worker:worker-A.db_host",      "host-A"},
+        {"worker:worker-A.db_port",      "53306"},
+        {"worker:worker-A.db_user",      "qsmaster"},
+        {"worker:worker-A.db_password",  "CHANGEME"},
```
n8pease: you're sure you want this on GitHub?

core/modules/replica/testConfiguration.cc
```diff
@@ -97,6 +104,10 @@ BOOST_AUTO_TEST_CASE(ConfigurationTest) {
         {"worker:worker-B.fs_host",      "host-B"},
      // {"worker:worker-B.fs_port",      "52002"},      // assuming default
         {"worker:worker-B.data_dir",     "/data/B"},
+        {"worker:worker-B.db_host",      "host-B"},
+     // {"worker:worker-B.db_port",      "3306"},       // assuming default
+     // {"worker:worker-B.db_user",      "root"},       // assuming default
+     // {"worker:worker-B.db_password",  ""},           // assuming default
```
n8pease: should the commented out code get removed?

core/modules/replica/ControllerApp.cc
n8pease: one of my objections to aligning is it can make for noisy code reviews when you have to realign a block.

core/modules/replica/ControllerApp.cc
n8pease: It could have been `fixup`ed back into this commit.   ...I chose to review by commit because it seemed there was a lot of namespace changing and it was becoming hard to read the whole lot of changes at the same time.

core/modules/replica/ControllerApp.cc
n8pease: `fixup`?

core/modules/replica/Common.h
n8pease: oh, that's a change in C++11 I wasn't aware of. cool.

core/modules/replica/SqlResultSet.h
n8pease: I think it's one of those things where they can't fix it without possibly breaking existing code so they don't.

n8pease:
I'm done with the review. I've left notes but there's nothing glaring, I release Igor on his own recognizance to do the Right Thing.

core/modules/replica/EchoRequest.cc
iagaponenko: You're right!

core/modules/replica/WorkerProcessor.cc
iagaponenko: You've hit right at the spot! The dynamic casting is a price to pay for my attempt to keep a hierarchy of transient classes (subclasses of `WorkerRequest`) to be completely independent of the types in the Protobuf definition file `replica/protocol.proto`. The only reason to have the ugly "staircase" of the `if` statements is to deduce a specific variant of this `enum` from th protocol definition file:
```
enum ProtocolQueuedRequestType {

    REPLICA_CREATE   = 0;    // replicate a chunk
    REPLICA_DELETE   = 1;    // delete an existing replica of a chunk
    REPLICA_FIND     = 2;    // find out if a replica is present, and report its state
    REPLICA_FIND_ALL = 3;    // find all replicas and report their states
    TEST_ECHO        = 4;    // test the worker-side framework
    SQL              = 5;    // execute an SQL statement against the worker's database
}
```
I could (in theory) introduce a method:
```
ProtocolQueuedRequestType WorkerRequest::requestType() const;
```
But this will add the direct dependency of the Worker request classes on the types generated by the Protobuf compiler.
Keeping transient classes separate from the protocol definitions has certain advantages...as well as problems. And, frankly, I'm still undecided which approach is the best. For now, I'd suggest keeping this code "as is". Otherwise, this code review will result in a significant amount of work. Let's address this issue in a separate refactoring ticket.

core/modules/replica/WorkerSqlRequest.cc
iagaponenko: This confusing comment was eliminated in a later commit

core/modules/replica/Performance.h
iagaponenko: If someone told me what is the "qserv way" of making forward declarations then I would be happy to follow the lead 😸 When I'm looking at the headers in Qserv modules with:
```
egrep -B 5 -A 5 '[Ff]orward' core/modules/*/*.h
```
Then I'm seeing a spectrum of patterns, including (empty lines inside namespaces and no indentations):
```
core/modules/wbase/MsgProcessor.h:// Forward declarations
core/modules/wbase/MsgProcessor.h-namespace lsst {
core/modules/wbase/MsgProcessor.h-namespace qserv {
core/modules/wbase/MsgProcessor.h-namespace wbase {
core/modules/wbase/MsgProcessor.h-
core/modules/wbase/MsgProcessor.h-class Task;
core/modules/wbase/MsgProcessor.h-class WorkerCommand;
core/modules/wbase/MsgProcessor.h-
core/modules/wbase/MsgProcessor.h:}}} // End of forward declarations
core/modules/wbase/MsgProcessor.h-
```
And this:
```
core/modules/xrdsvc/SsiService.h:// Forward declarations
core/modules/xrdsvc/SsiService.h-class XrdSsiLogger;
core/modules/xrdsvc/SsiService.h-
core/modules/xrdsvc/SsiService.h-namespace lsst {
core/modules/xrdsvc/SsiService.h-namespace qserv {
core/modules/xrdsvc/SsiService.h-namespace wcontrol {
core/modules/xrdsvc/SsiService.h-  class Foreman;
core/modules/xrdsvc/SsiService.h-}
core/modules/xrdsvc/SsiService.h-namespace wpublish {
core/modules/xrdsvc/SsiService.h-  class ChunkInventory;
core/modules/xrdsvc/SsiService.h:}}} // End of forward declarations
```
core/modules/ccontrol/UserQuery.h:// Forward decl
core/modules/ccontrol/UserQuery.h-namespace lsst {
core/modules/ccontrol/UserQuery.h-namespace qserv {
core/modules/ccontrol/UserQuery.h-namespace qdisp {
core/modules/ccontrol/UserQuery.h-class MessageStore;
core/modules/ccontrol/UserQuery.h-}}}
```
And this:
```
core/modules/parser/SelectParser.h:// Forward declarations
core/modules/parser/SelectParser.h-namespace lsst {
core/modules/parser/SelectParser.h-namespace qserv {
core/modules/parser/SelectParser.h-namespace parser {
core/modules/parser/SelectParser.h-    class AntlrParser; // Internally-defined in SelectParser.cc
core/modules/parser/SelectParser.h-}
core/modules/parser/SelectParser.h-namespace query {
core/modules/parser/SelectParser.h-    class SelectStmt;
core/modules/parser/SelectParser.h:}}} // End of forward declarations
```

Comment (iagaponenko) [reply core/modules/replica/Performance.h]: And, unfortunately, there s noting on this subject in the Style Guide:
https://developer.lsst.io/cpp/style.html

File: core/modules/replica/WorkerProcessor.cc (hunk 21 above)
Comment (iagaponenko) [reply core/modules/replica/WorkerProcessor.cc]: God points. Fixed.

File: core/modules/replica/testConfiguration.cc (hunk 22 above)
Comment (iagaponenko) [reply core/modules/replica/testConfiguration.cc]: That `CHANGEME` suggests this is some mock "password", especially in the in-memory unit test 😸 The production environment should not be using this type of passwords.

File: core/modules/replica/testConfiguration.cc (hunk 23 above)
Comment (iagaponenko) [reply core/modules/replica/testConfiguration.cc]: The comments here are on purpose to illustrate what to expect in the unit test made downstream. I'm sorry, if these comments look confusing!

File: core/modules/replica/ControllerApp.cc (hunk 8 above)
Comment (iagaponenko) [reply core/modules/replica/ControllerApp.cc]: Yes, that's my concern too.

File: core/modules/replica/ControllerApp.cc (hunk 10 above)
Comment (iagaponenko) [reply core/modules/replica/ControllerApp.cc]: The code was refactored and moved into a more appropriate class/method (where it must belong):
```
class SqlResultType {
    ...
    /// Structure Field stores a content captured from MYSQL_FIELD
    struct Field {
        ...
        int32_t     type;       /// Field type (see MySQL headers for enum enum_field_types)
        ...
        /// @return string representatio of the type
        std::string type2string() const;
    };
}
```

File: core/modules/replica/Common.h (hunk 12 above)
Comment (iagaponenko) [reply core/modules/replica/Common.h]: Neither was I 😸 Andy-s once mentioned it to me when reviewing my code. Here is what the Standard says:
https://en.cppreference.com/w/cpp/language/zero_initialization

File: core/modules/replica/ControllerApp.cc (hunk 9 above)
Comment (iagaponenko) [reply core/modules/replica/ControllerApp.cc]: Sorry about that! I'm working on improving my `git` commit habits.

File: core/modules/replica/WorkerProcessor.cc (hunk 19 above)
Comment (andy-slac) [reply core/modules/replica/WorkerProcessor.cc]: I do not think we'll have more than one protocol supported (at a time) so direct dependency on protocol classes should be OK. Otherwise you'd need double dispatch to support multiple protocols.

File: core/modules/replica/WorkerSqlRequest.cc [hunk 24]
```diff
@@ -0,0 +1,217 @@
...
+    util::Lock lock(_mtx, context(__func__));
+
+    // Update the performance of the target request before returning it
+    response.set_allocated_target_performance(performance().info());
+
+    // Carry over the result of the query only after the request
+    // has finished (or failed).
+    switch (status()) {
+        case STATUS_SUCCEEDED:
+        case STATUS_FAILED:
```
andy-slac: DO you need `fall-through` comment here to suppress warnings?

core/modules/replica/WorkerSqlRequest.cc
```diff
@@ -0,0 +1,217 @@
...
+                workerInfo.dbHost,
+                workerInfo.dbPort,
+                user(),
+                password(),
+                ""));
+
+        auto self = shared_from_base<WorkerSqlRequest>();
+        conn->execute([self](decltype(conn) const& conn_) {
+            conn_->begin();
+            conn_->execute(self->query());
```
andy-slac: It could make sense building query outside transaction if it has chance to throw.

core/modules/replica/WorkerSqlRequest.h
```diff
@@ -0,0 +1,181 @@
...
+namespace lsst {
+namespace qserv {
+namespace replica {
+
+/**
+ * Class WorkerSqlRequest executes queries against the worker database
+ * and return results sets (if any) back to a caller.
+ *
+ * @note
+ *   queries passed into this operation must be well formed.
```
andy-slac: What happens if they are not?

core/modules/replica/WorkerSqlRequest.h
```diff
@@ -0,0 +1,181 @@
...
+     *   indicates the importance of the request
+     *
+     * @param query
+     *   the query to be executed
+     *
+     * @param user
+     *   the name of a database account for connecting to the database service
+     *
+     * @param password
+     *   a database for connecting to the database service
```
andy-slac: password

core/modules/replica/protocol.proto
```diff
@@ -465,45 +492,151 @@ message ReplicationResponseEcho {
...
+                                        //   MYSQL_TYPE_NEWDECIMAL
+                                        //   MYSQL_TYPE_ENUM
+                                        //   MYSQL_TYPE_SET
+                                        //   MYSQL_TYPE_TINY_BLOB
+                                        //   MYSQL_TYPE_MEDIUM_BLOB
+                                        //   MYSQL_TYPE_LONG_BLOB
+                                        //   MYSQL_TYPE_BLOB
+                                        //   MYSQL_TYPE_VAR_STRING
+                                        //   MYSQL_TYPE_STRING
+                                        //   MYSQL_TYPE_GEOMETRY
```
andy-slac: I don't think this list is good to keep here, better to provide a link to online doc, or the file where you define the enum. Replicating this list in many places is not a purpose or `replica` module. 😉

core/modules/replica/protocol.proto
```diff
@@ -465,45 +492,151 @@ message ReplicationResponseEcho {
...
+}
+
+/// The message representing data of a row from a result set
+message ProtocolResponseSqlRow {
+
+    /// An array of the raw (non-null terminated) byte strings, where each string
+    /// represents a cell.
+    /// Note that the number of elements in the array must match
+    /// the number of fields.
```
andy-slac: Is there anything to be said here about encoding of the bytes, is it guaranteed to be utf-8 for example?

core/modules/replica/Performance.h
n8pease: Sorry, my point was the forward declarations are consistently in their own block, not in the namespacing scope of the class being declared.

iagaponenko: Got it! I will refactor my code to follow this convention.

admin/tools/docker/replication/container/tools/ncsa/Dockerfile
```diff
@@ -5,6 +5,7 @@ USER 0
 
 RUN mkdir -p /qserv/bin
 RUN mkdir -p /qserv/lib
+RUN mkdir -p /qserv/work
 RUN chown -R 1001:1001 /qserv
```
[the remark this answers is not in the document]
iagaponenko: Actually, this is a good point! I'll improve the Docker file the way it's suggest in the article.

admin/tools/docker/replication/deployment/ncsa/run.sh
```diff
@@ -52,12 +52,13 @@ docker run \
...
     -e "TOOL=${TOOL}" \
     -e "PARAMETERS=${PARAMETERS}" \
     -e "LSST_LOG_CONFIG=${LSST_LOG_CONFIG}" \
     -e "CONFIG=${CONFIG}" \
     -e "WORK_DIR=${WORK_DIR}" \
     --name "${TOOL}" \
     "${REPLICATION_IMAGE_TAG}" \
-    bash -c 'cd ${WORK_DIR}; /qserv/bin/${TOOL} ${PARAMETERS} --config=${CONFIG}'
+    /qserv/bin/${TOOL} "$@" --config=${CONFIG}
+#    bash -c 'cd ${WORK_DIR}; /qserv/bin/${TOOL} ${PARAMETERS} --config=${CONFIG}'
```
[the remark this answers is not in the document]
iagaponenko: Done

core/modules/replica/ChunksApp.cc
```diff
@@ -41,14 +41,16 @@ using namespace std;
 
 namespace {
 
+using namespace lsst::qserv::replica;
```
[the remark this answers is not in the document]
iagaponenko: I'm not sure if this is the right thing to do as per the de-jure LSST Style Guide or Qserv de-factor one. Note that this file has the following structure:
```
namespace {
    using namespace lsst::qserv::replica;
    void foo() { ... }
}
namespace last {
namespace qserv {
namespace replica {
void ChunkApp::bar() {
  ::foo();
}}}
```

core/modules/replica/SqlApp.cc
iagaponenko: to sure if I'm really "suffering" 😄 However, using `cv` would probably make this code more complex. Besides, this (spill-lock) pattern is mostly (if not to say - exclusively) used in the smile command-line applications. Class `SqlApp` is one of those. All classes ending with `App` are the command-line tools. I need s simple (nothin fancy) mechanism for waiting for a completion of an asynchronous operation.

core/modules/replica/Common.h
iagaponenko: This is actually a very interesting (an important) topic. I've just revisited the documentation on the C++11 (or later) and, I think, I'm convinced that for non-static members of classes, th default constructor will always perform zero-value initialization. See case 3 of following document:
https://en.cppreference.com/w/cpp/language/default_initialization
Which says:
```
Default initialization is performed in three situations:
...
3) when a base class or a non-static data member is not mentioned in a constructor initializer list and that constructor is called.
```
The second document explains what the default constructor does (how it initializes non-static members of the class):
https://en.cppreference.com/w/cpp/language/default_constructor
According to that document:
```
ClassName() = default;
```
is the same as:
```
ClassName() { }
```
And, for the missing initializations, zero value initialization would kick-in.

Obviously, I'm not a C++ lawyer 😸 So, I may be missing something here.

core/modules/replica/WorkerProcessor.cc
iagaponenko: Even though I may agree with (both of) you, let's leave this topic for now. I will keep it in my _TO DO_ list. To tell the truth, refactoring the worker-side Replication Framework would require an effort comparable with the one of the current ticket.

core/modules/replica/Common.h
iagaponenko: For now I'm putting the explicit initializers (as suggested by @andy-slac ) back. Just in case.

admin/tools/docker/replication/deployment/ncsa/pull.sh
```diff
@@ -33,11 +33,11 @@ set -e
...
-ssh -n $HOST docker pull $DB_IMAGE_TAG
+ssh -n $HOST docker pull $DB_IMAGE_TAG &
 
 # Replication System's tools
 
 for n in $MASTER $WORKERS; do
     HOST="qserv-${n}"
-    ssh -n $HOST docker pull $REPLICATION_IMAGE_TAG
+    ssh -n $HOST docker pull $REPLICATION_IMAGE_TAG &
 done
```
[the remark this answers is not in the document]
iagaponenko: Good catch! Thank you!

admin/tools/docker/replication/deployment/ncsa/stop.sh
```diff
@@ -32,15 +32,15 @@ set -e
...
     HOST="qserv-${WORKER}"
-    ssh -n $HOST 'echo "["'$WORKER'"] worker agent: "$(docker stop '$WORKER_CONTAINER_NAME')" "$(docker rm '$WORKER_CONTAINER_NAME')'
+    ssh -n $HOST 'echo "["'$WORKER'"] worker agent: "$(docker stop '$WORKER_CONTAINER_NAME')" "$(docker rm '$WORKER_CONTAINER_NAME')' &
 done
 
 if [ -n "${DB_SERVICE}" ]; then
     HOST="qserv-${MASTER}"
-    ssh -n $HOST 'echo "["'$MASTER'"] database service: "$(docker stop '$DB_CONTAINER_NAME')" "$(docker rm '$DB_CONTAINER_NAME')'
+    ssh -n $HOST 'echo "["'$MASTER'"] database service: "$(docker stop '$DB_CONTAINER_NAME')" "$(docker rm '$DB_CONTAINER_NAME')' &
 fi
```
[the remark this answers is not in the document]
iagaponenko: Fixed as well.

core/modules/replica/WorkerSqlRequest.cc
iagaponenko: This is kust a trivial non-throwing accessor method. The query is supposed to be prepared.

core/modules/replica/WorkerSqlRequest.h
iagaponenko: The query execution will fail and the database exception `replica::database::mysql::Error` with an explanation of a problem with the syntax of the query would be thrown. The will be caught by:
```
try {
    ...
} catch(database::mysql::Error const& ex) {
    ..
}````
within a method executing the query:
```
bool WorkerSqlRequest::execute() {
```
And be reported back to the calling `Controller` in `Protobuf` structure:
```
message ProtocolResponseSql {
    ...
    /// MySQL-specific error
    optional string error = 5 [default = ""];
    ...
}
```
I'm going to extend the `Doxygen` documentation for the class to make it clear.

File: core/modules/replica/protocol.proto (hunk 28 above)
Comment (iagaponenko) [reply core/modules/replica/protocol.proto]: Good point! Here is the link:
https://dev.mysql.com/doc/refman/8.0/en/c-api-data-structures.html
I will put it into the protocol definition file.

File: core/modules/replica/WorkerSqlRequest.h (hunk 27 above)
Comment (iagaponenko) [reply core/modules/replica/WorkerSqlRequest.h]: Fixed

File: core/modules/replica/protocol.proto (hunk 29 above)
Comment (iagaponenko) [reply core/modules/replica/protocol.proto]: The type `bytes` which is used here is a variable length array of the 8-bit bytes (similar to `char` in C++), and it's populated from `std::string` (same as `std::basic_string<char>` or `char*` + `length`). This is a raw result set pulled via the MySQL C-level API after executing a query. It's up to a recipient of the Protobuf message how to interpret the byte stream.

File: core/modules/replica/Common.cc [hunk 35]
```diff
@@ -41,96 +38,102 @@ namespace replica {
...
+        case ProtocolStatusExt::FILE_RENAME:   return ExtendedCompletionStatus::EXT_STATUS_FILE_RENAME;
+        case ProtocolStatusExt::FILE_EXISTS:   return ExtendedCompletionStatus::EXT_STATUS_FILE_EXISTS;
+        case ProtocolStatusExt::SPACE_REQ:     return ExtendedCompletionStatus::EXT_STATUS_SPACE_REQ;
+        case ProtocolStatusExt::NO_FOLDER:     return ExtendedCompletionStatus::EXT_STATUS_NO_FOLDER;
+        case ProtocolStatusExt::NO_FILE:       return ExtendedCompletionStatus::EXT_STATUS_NO_FILE;
+        case ProtocolStatusExt::NO_ACCESS:     return ExtendedCompletionStatus::EXT_STATUS_NO_ACCESS;
+        case ProtocolStatusExt::NO_SPACE:      return ExtendedCompletionStatus::EXT_STATUS_NO_SPACE;
+        case ProtocolStatusExt::FILE_MTIME:    return ExtendedCompletionStatus::EXT_STATUS_FILE_MTIME;
+        case ProtocolStatusExt::MYSQL_ERROR:   return ExtendedCompletionStatus::EXT_STATUS_MYSQL_ERROR;
+        case ProtocolStatusExt::LARGE_RESULT:  return ExtendedCompletionStatus::EXT_STATUS_LARGE_RESULT;
```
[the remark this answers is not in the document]
iagaponenko: I do understand the concern. There are some obstacles to that though. First, I can't use use C++ types in the `protobuf` schema definition (which has its own language). Secondly, I'm hesitant (so far) to go other way around - proliferating `protobuf`-generated types in the rest of the Replication Framework (at both its ends - the Controller an the workers ones). This is a code architecture "moment", not just an "issues" with this particular fragment.

core/modules/replica/ConfigApp.cc
```diff
@@ -165,6 +165,26 @@ ConfigApp::ConfigApp(int argc, char* argv[])
...
+
+    updateWorkerCmd.option(
+        "worker-db-user",
+        "The name of a user account for the worker's database service.",
+        _workerInfo.dbUser);
+
+    updateWorkerCmd.option(
+        "worker-db-password",
+        "A password for the worker's database service.",
+        _workerInfo.dbPassword);
```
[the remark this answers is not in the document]
iagaponenko: Good suggestion! I will think about it. Or, better, put it into my `TO DO` list of improvements.

core/modules/replica/Configuration.cc
```diff
@@ -188,6 +194,10 @@ uint16_t     const Configuration::defaultWorkerSvcPort                (50000);
 string       const Configuration::defaultWorkerFsHost                 ("localhost");
 uint16_t     const Configuration::defaultWorkerFsPort                 (50001);
 string       const Configuration::defaultDataDir                      ("{worker}");
+string       const Configuration::defaultWorkerDbHost                 ("localhost");
+uint16_t     const Configuration::defaultWorkerDbPort                 (3306);
+string       const Configuration::defaultWorkerDbUser                 (FileUtils::getEffectiveUser());
+string       const Configuration::defaultWorkerDbPassword             ("");
```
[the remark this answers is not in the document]
iagaponenko: Good point! Normally, I do this with the assignment operator. This is a 'legacy' code which is left from a period of the development when I wasn't doing this right. Fixed.

core/modules/replica/Configuration.cc
```diff
@@ -465,6 +475,10 @@ string Configuration::asString() const {
     ss << context() << "defaultWorkerFsHost:                  " << defaultWorkerFsHost << "\n";
     ss << context() << "defaultWorkerFsPort:                  " << defaultWorkerFsPort << "\n";
     ss << context() << "defaultDataDir:                       " << defaultDataDir << "\n";
+    ss << context() << "defaultWorkerDbHost:                  " << defaultWorkerDbHost << "\n";
+    ss << context() << "defaultWorkerDbPort:                  " << defaultWorkerDbPort << "\n";
+    ss << context() << "defaultWorkerDbUser:                  " << defaultWorkerDbUser << "\n";
+    ss << context() << "defaultWorkerDbPassword:              " << "*****" << "\n";
```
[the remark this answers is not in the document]
iagaponenko: Just for the sake of consistency :)

core/modules/replica/ConfigurationFile.cc
```diff
@@ -109,6 +109,15 @@ string ConfigurationFile::dump2init(Configuration::Ptr const& config) {
...
         << "fs_buf_size_bytes          = " << to_string(config->workerFsBufferSizeBytes())    << "\n"
+        << "svc_host                   = " <<           defaultWorkerSvcHost                  << "\n"
+        << "svc_port                   = " << to_string(defaultWorkerSvcPort)                 << "\n"
+        << "fs_host                    = " <<           defaultWorkerFsHost                   << "\n"
+        << "fs_port                    = " << to_string(defaultWorkerFsPort)                  << "\n"
+        << "data_dir                   = " <<           defaultDataDir                        << "\n"
+        << "db_host                    = " <<           defaultWorkerDbHost                   << "\n"
+        << "db_port                    = " << to_string(defaultWorkerDbPort)                  << "\n"
+        << "db_user                    = " <<           defaultWorkerDbUser                   << "\n"
+        << "db_password                = " <<           defaultWorkerDbPassword               << "\n"
```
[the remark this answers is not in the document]
iagaponenko: Good point! Improved.

core/modules/replica/ConfigurationMySQL.cc
```diff
@@ -132,7 +141,11 @@ string ConfigurationMySQL::dump2init(Configuration::Ptr const& config) {
             <<         info.svcPort << ","
             << "'" <<  info.fsHost  << "',"
             <<         info.fsPort  << ","
-            << "'" <<  info.dataDir << "'"
+            << "'" <<  info.dataDir << "',"
+            << "'" <<  info.dbHost  << "',"
+            <<         info.dbPort  << ","
+            << "'" <<  info.dbUser  << "',"
+            << "'" <<  info.dbPassword << "'"
```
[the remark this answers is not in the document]
iagaponenko: Agreed. Removed support for the worker database passwords everywhere.

core/modules/replica/ConfigurationStore.cc
```diff
@@ -262,6 +262,70 @@ WorkerInfo ConfigurationStore::setWorkerDataDir(string const& name,
 }
 
 
+WorkerInfo ConfigurationStore::setWorkerDbHost(std::string const& name,
+                                               std::string const& host) {
+    LOGS(_log, LOG_LVL_DEBUG, context(__func__) << "  name=" << name << " host=" << host);
+    util::Lock lock(_mtx, context(__func__));
+
+    auto&& itr = _workerInfo.find(name);
```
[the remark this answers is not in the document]
iagaponenko: Good point. Fixed

core/modules/replica/ConfigurationFile.cc
```diff
@@ -122,6 +131,10 @@ string ConfigurationFile::dump2init(Configuration::Ptr const& config) {
             << "fs_host      = " <<           info.fsHost                  << "\n"
             << "fs_port      = " << to_string(info.fsPort)                 << "\n"
             << "data_dir     = " <<           info.dataDir                 << "\n"
+            << "db_host      = " <<           info.dbHost                  << "\n"
+            << "db_port      = " << to_string(info.dbPort)                 << "\n"
```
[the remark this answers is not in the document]
iagaponenko: Removed...along with the table-like formatting

core/modules/replica/ConfigApp.cc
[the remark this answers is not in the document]
iagaponenko: Actually, I decided to eliminate worker database passwords from being stored in the Configuration (the code above is meant to upload a password into the MySQL-based Configuration).

core/modules/replica/ConfigurationStore.cc
```diff
@@ -262,6 +262,70 @@ WorkerInfo ConfigurationStore::setWorkerDataDir(string const& name,
...
+WorkerInfo ConfigurationStore::setWorkerDbPort(std::string const& name,
+                                               uint16_t port) {
+    LOGS(_log, LOG_LVL_DEBUG, context(__func__) << "  name=" << name << " port=" << port);
+    util::Lock lock(_mtx, context(__func__));
+
+    auto&& itr = _workerInfo.find(name);
+    if (_workerInfo.end() == itr) {
+        throw invalid_argument(
+                ::classMethodContext(__func__) + "  no such worker: " + name);
+    }
```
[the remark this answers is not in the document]
iagaponenko: Since I don't like using C/C++ macros here then I may go for the following code pattern instead (which would be using a member function):
```
assertIfTrue(_workerInfo.end() == ptr, __func__, "  no such worker: " + name);
```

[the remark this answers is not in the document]
iagaponenko: Refactored. It turns out I already have a base class's method which encapsulates this pattern:  `Configuration::safeFindWorker(...)`. This particular class was not refactored to use that method.

core/modules/replica/Controller.h
```diff
@@ -378,6 +378,59 @@ class Controller : public std::enable_shared_from_this<Controller> {
...
+    SqlRequestPtr sql(std::string const& workerName,
+                      std::string const& query,
+                      std::string const& user,
+                      std::string const& password,
+                      uint64_t maxRows,
+                      SqlRequestCallbackType const& onFinish=nullptr,
+                      int  priority=0,
+                      bool keepTracking=true,
+                      std::string const& jobId="",
+                      unsigned int requestExpirationIvalSec=0);
```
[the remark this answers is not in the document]
iagaponenko: This will require a significant overhaul of the code. I will put this into my _TO DO_ list for now.

core/modules/replica/ControllerApp.cc
```diff
@@ -44,29 +45,81 @@ using namespace std;
...
 
-/// Report result of the operation
+void printAsTable(SqlResultSet const& result,
+                  size_t const pageSize) {
+
+    cout << "\n"
+         << "error:     " << result.error << "\n"
+         << "hasResult: " << (result.hasResult ? "yes" : "no") << "\n"
+         << "fields:    " << to_string(result.fields.size()) << "\n"
+         << "rows:      " << to_string(result.rows.size()) << "\n"
```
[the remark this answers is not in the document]
iagaponenko: Fixed

core/modules/replica/ControllerApp.cc
```diff
@@ -245,6 +304,42 @@ ControllerApp::ControllerApp(int argc, char* argv[])
...
+
+    sqlCmd.required(
+        "user",
+        "The name of a user for establishing a connection with the worker's database.",
+        _sqlUser);
+
+    sqlCmd.required(
+        "password",
+        "A password which is used along with the user name for establishing a connection"
+        " with the worker's database.",
```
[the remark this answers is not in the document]
iagaponenko: 👍

core/modules/replica/ControllerApp.cc
```diff
@@ -44,29 +45,81 @@ using namespace std;
...
+                 pageSize);
+}
+
+
+void printRequest(StopSqlRequest::Ptr const& request, size_t const pageSize) {
+    cout << request->context() << "\n"
+         << "  performance: " << request->performance() << endl;
+    printAsTable(request->responseData(),
+                 pageSize);
+}
```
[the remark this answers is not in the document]
iagaponenko: Refactored

core/modules/replica/ControllerApp.cc
```diff
@@ -245,6 +304,42 @@ ControllerApp::ControllerApp(int argc, char* argv[])
 
     /// Request-specific parameters, options, flags
 
+    auto& sqlCmd = parser().command("SQL");
+
+    echoCmd.description(
```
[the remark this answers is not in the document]
iagaponenko: Good catch!

core/modules/replica/ControllerApp.cc
```diff
@@ -561,14 +693,22 @@ int ControllerApp::runImpl() {
...
 
-    util::BlockPost blockPost(0, 5000);     // for random delays (milliseconds) between iterations
+    util::BlockPost blockPost(10, 20);  // for random delays (in milliseconds) between iterations
 
+    size_t const printIvalMs   = 5000;
+    size_t       currentIvalMs = 0;
     while (not finished) {
-        cout << "HEARTBEAT: " << blockPost.wait() << " msec" << endl;;
+        currentIvalMs += blockPost.wait();
+        if (currentIvalMs >= printIvalMs) {
```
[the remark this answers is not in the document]
iagaponenko: Good point!

core/modules/replica/DatabaseMySQLRow.cc
```diff
@@ -175,9 +178,31 @@ Row::Cell const& Row::getDataCell(string const& columnName) const {
...
 
+
+void Row::exportRow(ProtocolResponseSqlRow* ptr) const {
+
+    string const context = "Row::" + string(__func__) + "  ";
+    if (not isValid()) {
+        throw logic_error(context + "the object is not valid");
+    }
+    if (nullptr == ptr) {
+        invalid_argument(context + "null pointer passed as a parameter");
```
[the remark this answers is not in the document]
iagaponenko: Good catch

core/modules/replica/DatabaseMySQLRow.h
```diff
@@ -191,6 +200,20 @@ class Row {
...
+     * Fill a Protobuf object representing a row.
+     * 
+     * @note
+     *   The previous content of the 
+     *
+     * @param ptr
+     *   a valid pointer to the Protobuf object to be populated.
+     * 
+     * @param std::invalid_argument
+     *   if the input pointer is 0
```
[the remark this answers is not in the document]
iagaponenko: Then I will probably end up with an undefined behavior when attempting to dereference the pointer. A problem is that I'm getting this pointer from the Protobuf API. This is a mutable object which I'm supposed to fill in. Th safety check inside the method is to ensure I didn't mess up when calling the method.

core/modules/replica/HttpProcessor.cc
```diff
@@ -310,6 +312,79 @@ bool saveConfigParameter(T& struct_,
...
 }
 
+
+/**
+ * Helper class HttpRequestBody parses a body of an HTTP request
+ * which has the following header:
+ * 
+ *   Content-Type: application/json
+ * 
+ * Exceptions are thrown
```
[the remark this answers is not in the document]
iagaponenko: Expanded the class documentation

core/modules/replica/HttpProcessor.cc
```diff
@@ -310,6 +312,79 @@ bool saveConfigParameter(T& struct_,
...
+ *   Content-Type: application/json
+ * 
+ * Exceptions are thrown 
+ */
+class HttpRequestBody {
+
+public:
+
+    /// Key-value pairs
+    map<string,string> kv;
```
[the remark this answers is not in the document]
iagaponenko: It's easier that way. This is a helper class which is meant to parse the body of requests (the _POST_ method). An the only product of this class is that `kv` map.

core/modules/replica/HttpProcessor.cc
```diff
@@ -310,6 +312,79 @@ bool saveConfigParameter(T& struct_,
...
+     * - the required HTTP header is not found i nthe request
+     * - the body doesn't have a valid JSON string (unless the body is empty)
+     * 
+     * @param req
+     *   the request to be parsed
+     */
+    explicit HttpRequestBody(qhttp::Request::Ptr req) {
+
+        string const contentType = req->header["Content-Type"];
+        string const requiredContentType = "application/json";
```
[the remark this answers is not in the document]
iagaponenko: It's easier to compare C++ strings than C-style strings. IN general I avoid using any C strings unless I have to.

core/modules/replica/HttpProcessor.cc
```diff
@@ -385,7 +460,9 @@ void HttpProcessor::_initialize() {
         {"DELETE", "/replication/v1/config/database/:name", bind(&HttpProcessor::_deleteDatabaseConfig, self, _1, _2)},
         {"POST",   "/replication/v1/config/database",       bind(&HttpProcessor::_addDatabaseConfig,    self, _1, _2)},
         {"DELETE", "/replication/v1/config/table/:name",    bind(&HttpProcessor::_deleteTableConfig,    self, _1, _2)},
-        {"POST",   "/replication/v1/config/table",          bind(&HttpProcessor::_addTableConfig,       self, _1, _2)}
+        {"POST",   "/replication/v1/config/table",          bind(&HttpProcessor::_addTableConfig,       self, _1, _2)},
+
+        {"POST",   "/replication/v1/sql/query", bind(&HttpProcessor::_sqlQuery, self, _1, _2)}
```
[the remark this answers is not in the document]
iagaponenko: "...no alignment - no cry..." (after Bob Marley's "No Woman no cry"). Got rid of the alignments and empty lines which "reset the alignment" 😺

core/modules/replica/HttpProcessor.cc
```diff
@@ -1226,6 +1303,61 @@ void HttpProcessor::_addTableConfig(qhttp::Request::Ptr req,
...
+        auto const password =        body.required("password");
+        auto const maxRows  = stoull(body.optional("max_rows", "0"));
+
+        _debug(string(__func__) + " worker="   +           worker);
+        _debug(string(__func__) + " query="    +           query);
+        _debug(string(__func__) + " user="     +           user);
+        _debug(string(__func__) + " password=" +           password);
+        _debug(string(__func__) + " maxRows="  + to_string(maxRows));
+
+        atomic<bool> finished{false};
```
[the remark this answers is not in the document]
iagaponenko: OK

core/modules/replica/ChunksApp.cc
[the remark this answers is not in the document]
andy-slac: Declarations inside anonymous namespace are visible in a global namespace too. Placing `using` inside anonymous namespace may give someone the idea that it only applies to that namespace which is not true. Move it outside to avoid this confusion.

core/modules/replica/Common.cc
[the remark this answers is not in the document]
andy-slac: May be easier to just avoid defining protocol enums completely, pass that attribute as integer and covert back to enum. That extra type safety is not worth this extra code that you need to maintain.

core/modules/replica/Common.h
andy-slac: You are right about point 3 above, but it says that default initialization applies in that case, not zero initialization. Default initialization for basic types without constructor means no initialization. In the same document few lines below:

  nothing is done: the objects with automatic storage duration (and their subobjects) are initialized to indeterminate values.

core/modules/replica/DatabaseMySQLRow.h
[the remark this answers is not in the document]
andy-slac: You should definitely check for nullptr before dereferencing, but my concern is mostly with API, if a method should never receive a nullptr then (in my opinion) its argument should be a reference. What you explain is that you made API dependent on a separate external concern (how you obtain that object) which should not be even known to this method.

core/modules/replica/HttpProcessor.cc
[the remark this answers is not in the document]
andy-slac: I'm not sure what is easier here. I can't see `kv` used anywhere outside the class and you have two methods which provide reasonable access to the data in `kv`.

core/modules/replica/HttpProcessor.cc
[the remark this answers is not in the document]
andy-slac: It is not terribly difficult to compare std::string with C string too.

core/modules/replica/WorkerSqlRequest.cc
andy-slac: How do I know it here? What if `query()` implementation changes?

core/modules/replica/protocol.proto
andy-slac: So what you say is that encoding is not known? I like this new API even less than I did before 🙁

core/modules/replica/HttpProcessor.cc
```diff
@@ -1226,6 +1303,61 @@ void HttpProcessor::_addTableConfig(qhttp::Request::Ptr req,
...
+            password,
+            maxRows,
+            [&finished] (SqlRequest::Ptr const& request) {
+                finished = true;
+            }
+        );
+        util::BlockPost blockPost(10, 20);  // for random delays (in milliseconds) between iterations
+        while (not finished) {
+            blockPost.wait();
+        }
```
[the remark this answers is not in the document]
iagaponenko: Making the async requests and jobs classes synchronous would be messy. Though, In theory I could introduce wrapper methods which would launch operations on the original async `*Job` or `*Request` classes  with a custom callback, and then wait for completion of the job/request using `cv` or that BlockPost (in case if I need heartbeats). I will put this side onto my _TO DO_ list.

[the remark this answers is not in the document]
iagaponenko: I've put the code improvement proposal still under development) here:
https://confluence.lsstcorp.org/pages/viewpage.action?pageId=102471961

core/modules/replica/HttpProcessor.h
```diff
@@ -158,259 +158,152 @@ class HttpProcessor : public std::enable_shared_from_this<HttpProcessor> {
...
-                        qhttp::Response::Ptr resp);
+                         qhttp::Response::Ptr resp);
+
+    /**
+     * Process a request for executing a query against a worker database.
+     * A result set of the query will be returned for those query types which
+     * have the one upon a successful completion of a request.
+     */
+    void _sqlQuery(qhttp::Request::Ptr req,
+                   qhttp::Response::Ptr resp);
```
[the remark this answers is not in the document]
iagaponenko: Migrated to `const&`

core/modules/replica/Performance.h
```diff
@@ -172,7 +165,7 @@ class WorkerPerformance {
      * OWNERSHIP TRANSFER NOTE: this method allocates a new object and
      * returns a pointer along with its ownership.
      */
-    proto::ReplicationPerformance* info() const;
+    ProtocolPerformance* info() const;
```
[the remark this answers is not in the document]
iagaponenko: Nice pattern. I'm happy to trade some of my Doxygen-style comments for this pattern 😄

core/modules/replica/ReplicaInfo.h
```diff
@@ -187,7 +181,7 @@ class ReplicaInfo {
      * @param info
      *   Protobuf object
      */
-    explicit ReplicaInfo(proto::ReplicationReplicaInfo const* info);
+    explicit ReplicaInfo(ProtocolReplicaInfo const* info);
```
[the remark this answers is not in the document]
iagaponenko: I'm getting this object from the Protobuf API as a plain-vanilla pointer as I'm iterating over a collection of replicas received from a worker at the Controller-side of the fence.

core/modules/replica/RequestTypesFwd.h
```diff
@@ -92,18 +103,21 @@ using StopDeleteRequest      = StopRequest<StopDeleteRequestPolicy>;
...
 typedef std::shared_ptr<StopFindAllRequest>     StopFindAllRequestPtr;
 typedef std::shared_ptr<StopEchoRequest>        StopEchoRequestPtr;
+typedef std::shared_ptr<StopSqlRequest>         StopSqlRequestPtr;
 
 typedef std::function<void(StopReplicationRequestPtr)> StopReplicationRequestCallbackType;
 typedef std::function<void(StopDeleteRequestPtr)>      StopDeleteRequestCallbackType;
 typedef std::function<void(StopFindRequestPtr)>        StopFindRequestCallbackType;
 typedef std::function<void(StopFindAllRequestPtr)>     StopFindAllRequestCallbackType;
 typedef std::function<void(StopEchoRequestPtr)>        StopEchoRequestCallbackType;
+typedef std::function<void(StopSqlRequestPtr)>         StopSqlRequestCallbackType;
```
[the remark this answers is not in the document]
iagaponenko: According to the C++11/14 Standard, they are both the same
```
2 A typedef-name can also be introduced by an alias-declaration. The identifier following the using keyword
becomes a typedef-name and the optional attribute-specifier-seq following the identifier appertains to that
typedef-name. It has the same semantics as if it were introduced by the typedef specifier. In particular, it
does not define a new type and it shall not appear in the type-id. [
```
See section 7.1.3 at:
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4296.pdf

So, why bother?

core/modules/replica/SqlApp.h
```diff
@@ -0,0 +1,105 @@
...
+    /// sets at workers
+    /// @note
+    ///   This parameter nothing to do with the SQL's 'LIMIT <num-rows>'.
+    ///   It serves as an additional fail safe mechanism preventing
+    ///   protocol buffers from being overloaded by huge result sets
+    ///   which might be accidentally initiated by users.
+    uint64_t _maxRows = 10000;
+
+    /// Permanently delete a worker from the Configuration
+    bool _allWorkers = false;
```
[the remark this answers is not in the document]
iagaponenko: 👍

core/modules/replica/SqlApp.h
```diff
@@ -0,0 +1,105 @@
...
+    /// The query to be executed
+    std::string _query;
+
+    /// The maximum number of rows to be pulled from result
+    /// sets at workers
+    /// @note
+    ///   This parameter nothing to do with the SQL's 'LIMIT <num-rows>'.
+    ///   It serves as an additional fail safe mechanism preventing
+    ///   protocol buffers from being overloaded by huge result sets
+    ///   which might be accidentally initiated by users.
```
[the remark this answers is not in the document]
iagaponenko: Removed

core/modules/replica/SqlRequest.cc
```diff
@@ -0,0 +1,342 @@
...
+    // upon a completion of the request within method _send() - the only
+    // client of _analyze(). So, we should take care of proper locking and watch
+    // for possible state transition which might occur while the async I/O was
+    // still in a progress.
+
+    // IMPORTANT: the final state is required to be tested twice. The first time
+    // it's done in order to avoid deadlock on the "in-flight" callbacks reporting
+    // their completion while the request termination is in a progress. And the second
+    // test is made after acquiring the lock to recheck the state in case if it
+    // has transitioned while acquiring the lock.
```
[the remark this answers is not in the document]
iagaponenko: State transitions in the Controller-side Request classes may occur in the following cases:
# a completion of a response from a worker
# an expiration of a request (there is an asynchronous timer started when the request starts)
# e explicit cancellation of a request by a client who might get lost his/her interest in the request
A problem is that these events may happen in any thread. Each even will progress the state of a request in one direction - towards its state `Request::FINISHED`. And, what is really important here, is that in either of those cases a notification callback would be initiated (indirectly) by calling method `Request::finish`. It's is a *REQUIREMENT* that the callback was called once and only once, regardless of a completion status of the request. What you've seen provides a guarantee for that. The first check prevents deadlocks in case if more then one completion condition (among those above mentioned) is being executed by different threads.

core/modules/replica/WorkerSqlRequest.cc
iagaponenko: Ot if the `case` is empty. The complier complains for non-trivial scenarios only, such as:
```
switch(val) {
    case 123:
        a = 123;
    case 345:
        break;

File: core/modules/replica/WorkerProcessor.cc [hunk 65]
```diff
@@ -431,8 +433,51 @@ void WorkerProcessor::enqueueForEcho(string const& id,
...
+                           ProtocolStatusExt::INVALID_PARAM);
+    }
+}
+
+
+void WorkerProcessor::enqueueForSql(std::string const& id,
+                                    ProtocolRequestSql const& request,
+                                    ProtocolResponseSql& response) {
+
+    LOGS(_log, LOG_LVL_DEBUG, _context() << __func__
```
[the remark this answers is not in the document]
iagaponenko: Yes, I do have. I started migrating to a more optimal way of generating those contexts not so long time ago. Some classes have already been migrated. This one hasn't. I'm going to do this (for tis class) now. Others will be eventually migrated as well.

core/modules/replica/ChunksApp.cc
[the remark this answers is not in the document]
iagaponenko: I didn't know that. Thanks! Every time I'm asking you to review my PR I'm learning something new.

core/modules/replica/protocol.proto
iagaponenko: One way to detect it is to make the following query:
```
SHOW VARIABLES LIKE '%collation%'
```
Which could return something like:
```
 Variable_name        | Value            
----------------------+-------------------
 collation_connection | latin1_swedish_ci
 collation_database   | latin1_swedish_ci
 collation_server     | latin1_swedish_ci
```
I have also extended the API to return a result obtained for the current connection from:
https://dev.mysql.com/doc/refman/5.7/en/mysql-character-set-name.html
I hope this solves the immediate problem. If that's not enough then the protocol and the API could be expanded to pass (back an forth) the desired character set:
https://dev.mysql.com/doc/refman/5.7/en/mysql-get-character-set-info.html
https://dev.mysql.com/doc/refman/5.7/en/mysql-set-character-set.html
Though, I think this is going to be an overkill for  problem I'm trying to solve in a context of this development. Character sets other than `latin1` are rarely used in the English-speaking world or scientific applications.

I should also mention that nothing in the MySQL connection API implemented in Qserv (no the Replication system) is sampling the encoding. Perhaps, I'm missing something?

core/modules/replica/Common.cc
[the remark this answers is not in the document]
iagaponenko: For the record, this concern has been registered in my list of proposed/discussed "Improvements" at:
https://confluence.lsstcorp.org/display/DM/Allow+protobuf+dependencies+in+the+worker-side+replication+requests
It will be evaluated later based on development priorities in the scope of the project.

PR #457, opened 2019-03-31
Tickets/dm 18743

iagaponenko:


core/modules/replica/ChunksApp.h
```diff
@@ -86,8 +86,9 @@ class ChunksApp : public Application {
     /// The flag (if set) for pulling chunk disposition from Qserv workers for the combined analysis
     bool _pullQservReplicas = false;
 
-    /// The maximum timeout for the replica disposition requests
-    unsigned int _timeoutSec = 60;
+    /// The maximum timeout for the completion of requests sent to
+    /// the Replication System's and Qserv workers
+    unsigned int _timeoutSec = 0;
```
andy-slac: `0` means no timeout?

core/modules/replica/ControllerApp.cc
```diff
@@ -381,318 +375,245 @@ ControllerApp::ControllerApp(int argc, char* argv[])
...
+        ptr = controller->replicate(
             _workerName,
             _sourceWorkerName,
             _databaseName,
             _chunkNumber,
-            [&finished] (ReplicationRequest::Ptr const& request) {
-                ::printRequest<ReplicationRequest>(request);
-                finished = true;
+            [] (ReplicationRequest::Ptr const& ptr_) {
+                ::printRequest<ReplicationRequest>(ptr_);
```
andy-slac: I think `request` was easier to read than `ptr_`

core/modules/replica/Job.cc
```diff
@@ -207,6 +213,19 @@ void Job::start() {
...
+void Job::wait() {
+ 
+    LOGS(_log, LOG_LVL_DEBUG, context() << __func__);
+
+    if (state() == State::FINISHED) return;
+
+    unique_lock<mutex> lock(_onFinishMtx);
+    _onFinishCv.wait(lock, [this] {
+        return state() == State::FINISHED;
+    });
```
andy-slac: I think this has a chance to block forever. I'll approve PR if you explain how it can happen (and fix it) 😉 

Also condition variables are messy in general, maybe `std::future` would be easier to use for this kind of pattern?

iagaponenko: I've added `unique_lock<mutex> lock(_onFinishMtx);` before modifying `_state`. Now, both update and acquisition of the state are mutually exclusive for the purpose of implementing the proposed synchronization. Otherwise, there was a chance of missing the `notify_all` signal.
Regarding `std::future`, I find it messy and counter-intuitive.

core/modules/replica/ControllerApp.cc
iagaponenko: The previous version of the code was saying "request" way too much.

core/modules/replica/ChunksApp.h
iagaponenko: It means that a default value of the timeout would be pulled from the Configuration. I've updated the comment to make it clear.

core/modules/replica/Job.cc
andy-slac: Fix looks fine.

And you need to work in you intuition, `std::future` is supposed to be more intuitive than condition variables :) I think your use case can be implemented trivially with something like this:

    class Job {

        std::promise<void> _finishPromise;
        std::future<void> _finishFuture;

        Job() : _finishFuture(_finishPromise.get_future()) {}

        void finish() {
            // Unblock threads (if any) waiting on the synchronization...
            _finishPromise.set_value();
        }

        void wait() { _finishFuture.wait(); }

    };

PR #458, opened 2019-04-02
Fix centos7-llvm build

fritzm:
Clang was flagging on private embbeded class replica::Controller::RequestWrapper;
made this embedded class public instead.


n8pease:
looks good.

PR #460, opened 2019-04-02
Tickets/dm 18844

fritzm:
Cleanups for various warnings generated by clang

### DIFF
--- a/core/modules/qana/DuplSelectExprPlugin.h
+++ b/core/modules/qana/DuplSelectExprPlugin.h
@@ -81,7 +81,7 @@ public:
      */
     static std::string const ERR_MSG;
 
-    virtual ~DuplSelectExprPlugin() {}
+    ~DuplSelectExprPlugin() override {}
 
     /**
      * Prevent execution of queries which have duplicated select fields names.
@@ -89,7 +89,7 @@ public:
      * @throw AnalysisError if duplicated select fields name are detected
      * @see QueryPlugin::applyLogical()
      */
-    virtual void applyLogical(query::SelectStmt& stmt, query::QueryContext&);
+    void applyLogical(query::SelectStmt& stmt, query::QueryContext&) override;
 
     /**
     * Return the name of the plugin class for logging.
--- a/core/modules/qana/TableInfo.h
+++ b/core/modules/qana/TableInfo.h
@@ -187,28 +187,28 @@ struct DirTableInfo : TableInfo {
     DirTableInfo(std::string const& db, std::string const& t, double o) :
         TableInfo(db, t, DIRECTOR), partitioningId(0), overlap(o) {}
 
-    virtual std::vector<ColumnRefConstPtr> const makeColumnRefs(
-        std::string const& tableAlias) const;
+    std::vector<ColumnRefConstPtr> const makeColumnRefs(
+        std::string const& tableAlias) const override;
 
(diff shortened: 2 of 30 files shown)


core/modules/replica/ReplicationTask.h
```diff
@@ -105,10 +105,6 @@ class ReplicationTask : public Task {
     /// on the Qserv synchronization requests.
     unsigned int _qservSyncTimeoutSec;
 
-    /// The number of seconds to wait in the end of each iteration loop before
-    /// to begin the new one.
-    unsigned int _replicationIntervalSec;
```
iagaponenko: 👍

core/modules/replica/HttpProcessor.h
```diff
@@ -248,8 +248,6 @@ class HttpProcessor : public std::enable_shared_from_this<HttpProcessor> {
     // the reference counters to the pointed objects.
 
     HealthMonitorTask::Ptr const& _healthMonitorTask;
-    ReplicationTask::Ptr   const& _replicationTask;
-    DeleteWorkerTask::Ptr  const& _deleteWorkerTask;
```
iagaponenko: This is good for now. Though, I will need to put both back later when I will actually need these references. An idea behind having these references here was to allow REST API to suspend/resume those tasks if needed.

core/modules/replica/StatusRequest.h
```diff
@@ -58,7 +58,8 @@ namespace replica {
 //   Customizations for specific request types require dedicated policies
 // ========================================================================
 
-struct StatusReplicationRequestPolicy {
+class StatusReplicationRequestPolicy {
```
iagaponenko: The change is fine. I'm just curious about a reason for this change.

fritzm: clang / llvm warns on inconsistent use of {{class}} vs {{struct}}, so we should pick one in each case.

And easier, I think, to just start using {{class}} everywhere, even for POD?  That way you don't have to think about rules of what is POD and what is not, have people interpret differently and argue, remember to change/update code when appropriate, etc.

core/modules/qana/TableInfo.h
```diff
@@ -187,28 +187,28 @@ struct DirTableInfo : TableInfo {
     DirTableInfo(std::string const& db, std::string const& t, double o) :
         TableInfo(db, t, DIRECTOR), partitioningId(0), overlap(o) {}
 
-    virtual std::vector<ColumnRefConstPtr> const makeColumnRefs(
-        std::string const& tableAlias) const;
+    std::vector<ColumnRefConstPtr> const makeColumnRefs(
+        std::string const& tableAlias) const override;
 
-    virtual bool isEqPredAdmissible(TableInfo const& t,
+    bool isEqPredAdmissible(TableInfo const& t,
```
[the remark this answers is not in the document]
fritzm: thx; will fix...

core/modules/qana/DuplSelectExprPlugin.h
```diff
@@ -81,15 +81,15 @@ class DuplSelectExprPlugin : public QueryPlugin {
      */
     static std::string const ERR_MSG;
 
-    virtual ~DuplSelectExprPlugin() {}
+    ~DuplSelectExprPlugin() override {}
```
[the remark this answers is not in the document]
fritzm: Perhaps, consistently, throughout, on a future changelist.

PR #462, opened 2019-04-12
DM-19259: invalid column in select statement crashes qserv

n8pease:


core/modules/ccontrol/UserQuery.h
```diff
@@ -105,6 +105,14 @@ class UserQuery {
...
     /// @return True if query is async query
     virtual bool isAsync() const { return false; }
+
+    /**
+     * @brief set up the merge table (stores results from workers)
+     *
+     * @param errMsg if creating the merge table fails this will contain an error message.
+     * @return bool true if success, false if there was an error.
+     */
+    virtual bool setupMerger(std::string& errMsg) { return true; }
```
andy-slac: This feels like anti-pattern. Can't you just throw an exception when it fails?

core/modules/ccontrol/UserQuerySelect.cc
```diff
@@ -389,9 +386,14 @@ void UserQuerySelect::_setupMerger() {
     _infileMerger = std::make_shared<rproc::InfileMerger>(*_infileMergerConfig);
 
     auto&& preFlightStmt = _qSession->getPreFlightStmt();
-    if (preFlightStmt == nullptr || not _infileMerger->makeResultsTableForQuery(*preFlightStmt)) {
-        throw UserQueryError(getQueryIdString() + " Could not create results table for query (no worker queries).");
+    if (preFlightStmt == nullptr) {
+        errMsg = " Could not create results table for query (no worker queries).";
```
andy-slac: Do you need that space at the beginning of message?

core/modules/rproc/InfileMerger.h
```diff
@@ -191,7 +191,7 @@ class InfileMerger {
     bool scrubResults(int jobId, int attempt);
     int makeJobIdAttempt(int jobId, int attemptCount);
 
-    bool makeResultsTableForQuery(query::SelectStmt const& stmt);
+    bool makeResultsTableForQuery(query::SelectStmt const& stmt, std::string& errMsg);
```
andy-slac: This should probably throw too if there is an error.

PR #464, opened 2019-04-26
tickets/DM-19234

jgates108:


core/modules/xrdsvc/SsiRequest.cc
```diff
@@ -385,22 +387,33 @@ bool SsiRequest::replyFile(int fd, long long fSize) {
...
-       }
+        _stream = std::make_shared<ChannelStream>();
+        if (SetResponse(_stream.get()) != XrdSsiResponder::Status::wasPosted) {
+            LOGS(_log, LOG_LVL_WARN, "SetResponse stream failed, calling Recycle for sBuf");
+            // SetResponse return value indicates XrdSsi wont call Recycle().
+            sBuf->Recycle();
+            return false;
+        }
     } else if (_stream->closed()) {
+        // XrdSsi isn't going to call Recycle if we wind up here.
```
abh3: This doesn't seem right. The stream gets closed it the last buffer has been added to the vector of buffers. If so, why is the code trying to add another one? That would indicate a serious logic error.

core/modules/wdb/QueryRunner.cc
```diff
@@ -314,24 +315,31 @@ void QueryRunner::_transmit(bool last, uint rowCount, size_t tSize) {
...
+    } else {
+        LOGS(_log, LOG_LVL_DEBUG, "_transmit cancelled");
+    }
+    _largeResult = true; // Transmits after the first are considered large results.
+}
+
+
+void QueryRunner::_sendBuf(xrdsvc::StreamBuffer::Ptr& streamBuf,
+                           util::TimerHistogram& histo, std::string const& note) {
+    bool sent = _task->sendChannel->sendStream(streamBuf, false);
```
iagaponenko: Why `last` is always set to `false`? Was this a bug fix? In the previous version of the code a value of `last` was passed into the method from:
```
void QueryRunner::_transmit(bool last, ...)
```

core/modules/wdb/QueryRunner.cc
```diff
@@ -314,24 +315,31 @@ void QueryRunner::_transmit(bool last, uint rowCount, size_t tSize) {
...
+    if (!sent) {
+        LOGS(_log, LOG_LVL_ERROR, _task->getIdStr() << " Failed to transmit " << note << "!");
+        _cancelled = true;
+        // Normally, sendStream calls Recycle(), but if this path is taken, it may not have been called.
+        // Calling Recycle() twice should be harmless.
```
iagaponenko: What's said in the comment above is a true "black magic" to me 😸

core/modules/wdb/QueryRunner.h
```diff
@@ -46,10 +46,21 @@
...
 namespace qserv {
+
 namespace proto {
 class ProtoHeader;
 class Result;
-}}}
+}
+
+namespace util {
+class TimerHistogram;
```
iagaponenko: Forward declarations need to be indented.

core/modules/wdb/QueryRunner.h
```diff
@@ -46,10 +46,21 @@
...
+}
+
+namespace xrdsvc {
+class StreamBuffer;
```
iagaponenko: Forward declarations need to be indented.

core/modules/xrdsvc/SsiRequest.h
```diff
@@ -57,7 +57,8 @@ namespace lsst {
 namespace qserv {
 namespace xrdsvc {
 
-class ChannelStream; // Forward declaration
+class ChannelStream;
+class StreamBuffer;
```
iagaponenko: Forward declarations need to be indented.

abh3:
Hi Igor,

This is a classic case of a producer-consumer problem when both sides run 
at different speeds. The producer is the worker and it creates buffers to 
be consumed by the czar. It may run faster or slower than the czar. When 
the speeeds match (or teh czar is faster) all the buffers are consumed and 
recycled. When the worker runs faster (which is more than likely) then 
buffers get queued. So, when the consumer decides to not consume any more 
then the worker has to dispose of the queued buffers. No buffer ever gets 
recycled twice it's just the matter of who has to recycle buffers. A 
cnsumed buffer is recycled by the consumer and an unconsumed buffer is 
recycled by the producer. Pretty much how it works in a restaurant, eh?

Andy

iagaponenko:
@abh3 your explanation makes me feel even more confused:

"...This is a classic case of a producer-consumer problem when both sides run
at different speeds. The producer is the worker and it creates buffers to
be consumed by the czar...."

We both know that "czar" and "workers" live in completely different memory spaces, so they can not share the same buffers, neither they can "consume" buffers. They may produce and consume data. And each side of this "food servicing chain" maintains its own sets of buffers to store data (which are either received or being received, from an opposite party, or before/while sending the data to an opposite party). Therefore, the buffer management relationships are happening between Qserv czar's code and the XRootD/SSI runtime on the czar's side, and (independently) between worker's code and the XRootD/SSI runtime in the worker process. And the "mediator" (XRootD/SSI) machinery moves data between these two sets of buffers (on the opposite sides of the "food chain").

My second problem is that I'm just confused by the terminology. In my view, all operation with buffers are around the "ownership transfer" (who owns the given buffer at each moment). Hence, if "Recycle" is just another way of saying that my thread is "Releasing Ownership" or "Taking Ownership" (over given buffer) then it's fine with me. If this is something else then I'm begging even more confused.

abh3:
On Fri, 26 Apr 2019, Igor Gaponenko wrote:

> @abh3 your explanation makes me feel even more confused:
>
> "...This is a classic case of a producer-consumer problem when both sides run
> at different speeds. The producer is the worker and it creates buffers to
> be consumed by the czar...."
>
> We both know that "czar" and "workers" live in completely different 
>memory spaces, so they can not share the same buffers, neither they can 
>"consume" buffers. They may produce and consume data. And each side of 
>this "food servicing chain" maintains its own sets of buffers to store 
>data (which are either received or being received, from an opposite 
>party, or before/while sending the data to an opposite party). Therefore, 
>the buffer management relationships are happening between Qserv czar's 
>code and the XRootD/SSI runtime on the czar's side, and (independently) 
>between worker's code and the XRootD/SSI runtime in the worker process. 
>And the "mediator" (XRootD/SSI) machinery moves data between these two 
>sets of buffers (on the opposite sides of the "food chain").
----
I think you may have taken my statement a bit too literally. What you say 
is technically correct. However, it doesn't materially affect the 
relationship. The mediator doesn't do anything more than what it is 
asked to do. In this case, it's the czar that's asking so, in effect, the 
mediator is the consumer working on behalf of the czar. I don't see where 
the confusion arrises. I suppose it's all a matter of perspective.

> My second problem is that I'm just confused by the terminology. In my 
view, all operation with buffers are around the "ownership transfer" (who 
owns the given buffer at each moment). Hence, if "Recycle" is just 
another way of saying that my thread is "Releasing Ownership" or "Taking 
Ownership" (over given buffer) then it's fine with me. If this is 
something else then I'm begging even more confused.
----
That's certainly one valid way of looking at it. I simply used a different 
analogy. That was to avoid the missing piece above which begs the question 
of how do I get ownership. Clearly, Recycle releases ownership but there 
is no concept here of how ownership is obtained. The consumer-producer 
analogy adresses both sides without explicitly saying how it's done. I 
hope this doesn't confuse things even more. I was simply trying to 
establish the ownership flow.

core/modules/wdb/QueryRunner.h
jgates108: Where is this stated? They aren't indented anywhere else (aside from replica) and "The contents of namespaces are not indented." (6-21b).

core/modules/xrdsvc/SsiRequest.cc
jgates108: I can add an error message to that affect, or throw. Either way, if this path is taken, Recycle should still be called as finding the source of a schedule wedge is difficult.

core/modules/wdb/QueryRunner.cc
jgates108: Thanks, fixed now.

core/modules/wdb/QueryRunner.h
iagaponenko: CC: @n8pease @andy-slac @fritzm 

I was "beaten to death" by members of our team when I expressed a similar objection in one of my tickets's review 😸 Claims I heard included:
"...please, do what the rest of Qserv does..."
"...it's a common practice do indent forward declaration..."
In the end, I just gave up and did what they wanted from me. So, I'm just 'forwarding" my pain to you 😸 Feel free to ignore my suggestion though.
And, speaking of which code does it one way or another, the `replica` module is a big contributor, but not the only one. I've counted a number of headers where indents (the first case are used) versus the ones where they're aren't:
```
% egrep '   class .+;' core/modules/*/*.h | awk -F: '{print $1}' | sort -u | wc -l
91
% egrep '^class .+;' core/modules/*/*.h | awk -F: '{print $1}' | sort -u | wc -l
51
```
An this is the frequency (the number of headers) in each module where the indents are used in the forward declarations:
```
 1 ccontrol
 1 global
 1 mysql
 2 parser
 7 qana
 1 qdisp
 3 qproc
30 query
31 replica
 1 rproc
 1 sql
 1 util
 3 wbase
 1 wcontrol
 3 wdb
 2 wpublish
 2 wsched
```
And there is a list of counters for modules where no spacing is present in front of the forward declarations:
```
7 ccontrol
2 css
1 czar
2 memman
3 mysql
1 parser
1 proto
2 qana
6 qdisp
1 qhttp
2 qproc
2 query
3 replica
1 util
4 wbase
2 wdb
7 wpublish
1 wsched
3 xrdsvc
```

iagaponenko: Not only my heart is with you, but the LSST C++ Style guide itself says:
"...The contents of namespaces are not indented..."

See:
https://developer.lsst.io/cpp/style.html#b-nested-namespaces-should-be-aligned-left-with-each-level-of-nesting-on-a-new-line

I wish I had read that before adding indents in my forward declarations 😸

andy-slac: Igor, your `grep` may be finding class declarations inside classes, which should be indented.

iagaponenko: That's certainly true. I have inspected the headers, and found 12 "false-positives". Still, the majority of forward declarations are indented.

I'm not arguing that the code had to be fixed (one way or another). I think this is one of the topics we should never touch upon on the code reviews. I myself was quite frustrated when I was requested to add those indents because "...the rest of Qserv code does it...". I think we (always) have a bigger "fish to fry" 😸 

Here is my proposal of *DONT'S* for the code reviews (this list could be extended):
- spacing, indents (except the standard ones for code blocks)
- alignments 
- empty lines
- the line length (exceeding 100 or whatever number of characters) should  be fine as long as it's needed
- code documentation (it has to be up to a developer how or if to document their code, using Doxygen, etc.).

These are the ones which are causing most disagreements and confusions. If people feel we really need to unify the "look & feel" of our code then one of those automated code formatting tools would help.

fritzm: ^so basically, the complete list of usual Igor sins ;-)  I'd schedule us some group therapy during PCW, but I guess that's what drinks by the pool are for 🍺

fritzm: More seriously, after taking a closer look, @jgates108 and @iagaponenko  I think the "prevailing style" could be described as:

- For "short" namespaces (in particular, those near the head of a file containing a few forward declarations), **DO** indent, because indenting those improves readability.
- For "long" namespaces (in particular, those spanning the entirety or a class declaration or method definitions, "set and forget" for majority of a file) **DO NOT** indent, because indenting those does not improve readability.

What do you think?

jgates108: Honestly, indent sometimes sounds awful to me. Difference between short and long is? You have a short one and a long one together?
I think Forward declarations are not at all interesting. They just need to be there. I'd rather not have any special rules for them.

jgates108: To be clear, I have nothing against Igor (or anyone) going through and marking style issues. If there is an existing style rule, I'll follow it.  If there isn't, I can ignore him :)

If there is not an existing rule, I don't want to make more unless REALLY necessary. If Igor really wants to indent his forward declarations, it doesn't really bother me.

fritzm: For clarity only: the difference between "long" and "short" to me is whether you parse it visually as single chunk or not.  Anything that involves scrolling around certainly does not qualify.

However, one person's readable is another person's fugly (c.f. the recent discussion about Doxygen comments).  We've certainly bigger fish, so I am offering an opinion but no argument.

iagaponenko: As per the classic: "Indent or not indent - that is THE question!" 😸 

Seriously speaking, my only problem in this context is that all IDEs I've used so far always try to indent within namespaces whenever I'm adding a new class, function, variable into a namespace. Otherwise, either way works for me just fine. I have no philosophical reservations for this other than practical ones.

andy-slac: And let's not start IDE holy war ⚔️

PR #472, opened 2019-07-09
Tickets/dm 20208

jgates108:


iagaponenko:
Note that Travis CI is failing on this PR.

jgates108:
Travis is failing integration tests because it doesn't have the changes I made to qserv_testData.

core/modules/css/DbInterfaceMySql.cc
```diff
@@ -0,0 +1,126 @@
...
+    if (not results.extractFirstColumn(emptyChunks, errObj)) {
+        LOGS(_log, LOG_LVL_ERROR, funcName << "Failed to extract empty chunks from query result");
+        throw CssError(ERR_LOC, errObj);
+    }
+
+    trans.commit();
+
+    // Create the set
+    std::set<int> emptyChunkSet;
+    for (auto j:emptyChunks) {
```
andy-slac: `auto const&` may be more efficient.

core/modules/css/EmptyChunks.h
```diff
@@ -46,37 +46,52 @@ namespace lsst {
...
 namespace css {
 
+class DbInterfaceMySql;
+
 /// High-level empty-chunk-tracking class. Tracks empty chunks
 /// per-database. In the future, we will likely migrate to a
 /// per-partitioning-group scheme, at which point, we will re-think
 /// the db-based dispatch as well (user tables in the partitioning
 /// group may be extremely sparse).
 class EmptyChunks {
```
andy-slac: Or maybe even `EmptyChunks` class itself can be an abstract interface with different implementations. We should untangle this mess, I do not particularly like EmptyChunk interface too 🙂 Something for the future ticket, no need to clean all mess in one sweep.

core/modules/css/EmptyChunks.h
```diff
@@ -46,37 +46,52 @@ namespace lsst {
...
                 std::string const& fallbackFile="emptyChunks.txt")
-        : _path(path), _fallbackFile(fallbackFile) {}
+        : _dbI(dbI), _path(path), _fallbackFile(fallbackFile) {}
+
+    EmptyChunks() = delete;
+    EmptyChunks(EmptyChunks const&) = delete;
+    EmptyChunks(EmptyChunks&&) = delete;
+    EmptyChunks& operator=(EmptyChunks const&) = delete;
+
+    virtual ~EmptyChunks() = default;
```
andy-slac: Is `virtual` needed? Or are you already preparing for what I just said above? 😸

core/modules/css/KvInterfaceImplMySql.cc
```diff
@@ -114,66 +114,23 @@ namespace lsst {
...
 
-// this is copied more or less directly from QMetaTransaction. Refactor to shared location?
-// proper place is to port it into db module
-// https://github.com/LSST/db
-// right now it's python only.
-// create a ticket to make generalized code to be used between c++ and python
-class KvTransaction
-{
+
+class KvTransaction : public sql::SqlTransactionStd {
```
andy-slac: This definitely collides with `KvTransaction` in other file. It can explode, be careful. One Definition Rule can do horrible things if violated.

core/modules/qmeta/QMeta.cc
```diff
@@ -15,6 +15,8 @@
 #include "util/ConfigStore.h"
 #include "util/ConfigStoreError.h"
 
+using namespace std;
+
```
andy-slac: I'd probably revert most changes to `qmeta`, I think your ticket is only supposed to touch `css`. Some of these look like style-only changes which don't need to be on this ticket.

core/modules/qmeta/QMetaMysql.h
```diff
@@ -38,7 +38,7 @@ namespace qmeta {
 
 /// @addtogroup qmeta
 
-/**
+/**)))
```
andy-slac: Is this a "Cyclops' big smile" ascii art? 👁  😃

core/modules/qproc/QuerySession.cc
```diff
@@ -230,7 +231,15 @@ QuerySession::getDbStriping() {
 std::shared_ptr<IntSet const>
 QuerySession::getEmptyChunks() {
     // FIXME: do we need to catch an exception here?
-    return _css->getEmptyChunks().getEmpty(_context->dominantDb);
+    if (_css != nullptr) {
+        LOGS(_log, LOG_LVL_DEBUG, "QuerySession::getEmptyChunks " << _context->dominantDb);
+        std::shared_ptr<IntSet const> result = _css->getEmptyChunks()->getEmpty(_context->dominantDb);
+        return result;
+    } else {
+        LOGS(_log, LOG_LVL_WARN, "QuerySession::getEmptyChunks no _qmeta");
```
andy-slac: _qmeta -> _css

core/modules/qproc/QuerySession.cc
```diff
@@ -64,6 +64,7 @@
 #include "qana/ScanTablePlugin.h"
 #include "qana/TablePlugin.h"
 #include "qana/WherePlugin.h"
+#include "qmeta/QMeta.h"
```
andy-slac: Probably not needed anymore

core/modules/qproc/QuerySession.h
```diff
@@ -58,6 +58,9 @@ namespace qserv {
 namespace css {
     class StripingParams;
 }
+namespace qmeta {
+    class QMeta;
+}
```
andy-slac: And this too?

core/modules/sql/SqlTransactionStd.h
```diff
@@ -0,0 +1,76 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the LSST License Statement and
+ * the GNU General Public License along with this program.  If not,
+ * see <https://www.lsstcorp.org/LegalNotices/>.
+ */
+#ifndef LSST_QSERV_SQL_SQLTRANSACTIONNORMAL_H
+#define LSST_QSERV_SQL_SQLTRANSACTIONNORMAL_H
```
andy-slac: `LSST_QSERV_SQL_SQLTRANSACTIONSTD_H`

core/modules/sql/SqlTransactionStd.h
```diff
@@ -0,0 +1,76 @@
+/*
+ * LSST Data Management System
+ * Copyright 2015 AURA/LSST.
```
andy-slac: Use different preamble or fix year? (and in .cc file too)

core/modules/sql/SqlTransactionStd.h
```diff
@@ -0,0 +1,76 @@
...
+
+// Qserv headers
+#include "sql/SqlTransaction.h"
+#include "util/Issue.h"
+
+namespace lsst {
+namespace qserv {
+namespace sql {
+
+/// A standard way to use SqlTransaction.
```
andy-slac: I would probably rephrased this and renamed the class. The class implements an RAII idiom for transaction management, it may be "standard" but most people are probably used more to the names like "Guard" or "Scope".

core/modules/sql/SqlTransactionStd.cc
```diff
@@ -0,0 +1,82 @@
...
+namespace lsst {
+namespace qserv {
+namespace sql {
+
+// Constructors
+SqlTransactionStd::SqlTransactionStd(sql::SqlConnection& conn)
+    : errObj(), trans(conn, errObj) {
+    if (errObj.isSet()) {
+        LOGS(_log, LOG_LVL_ERROR, "Constructor failed (" << errObj.errNo() << ") " << errObj.errMsg());
+        throwException(ERR_LOC, "Constructor failed");
```
andy-slac: Calling virtual methods from constructor is asking for trouble ⛔️ Just throw what you want to throw.

core/modules/css/DbInterfaceMySql.h
```diff
@@ -0,0 +1,68 @@
+// -*- LSST-C++ -*-
+/*
+ * LSST Data Management System
+ * Copyright 2019 LSST/AURA.
```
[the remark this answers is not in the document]
iagaponenko: The problem affects nearly all files in the Qserv package. So, this has to be the package-wide refactoring.

core/modules/css/DbInterfaceMySql.h
```diff
@@ -0,0 +1,68 @@
...
+#include "mysql/MySqlConfig.h"
+#include "sql/SqlConnection.h"
+
+namespace lsst {
+namespace qserv {
+namespace css {
+
+
+/// This class is used to connect to tabular data in a mysql database.
+class DbInterfaceMySql {
```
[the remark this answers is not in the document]
iagaponenko: Ditto. I would also suggest the in-memory implementation (from `std::list`, etc.) for the base interface to facilitate unit testing.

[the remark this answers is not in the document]
jgates108: This is not really meant to be the empty chunk interface. Everything else in CSS is key-value, this is meant to be an interface for things better represented by database tables, hence DbInterfaceMySql. I've been holding off making an interface class as it complicates the code I don't have immediate need of it.

core/modules/css/DbInterfaceMySql.h
```diff
@@ -0,0 +1,68 @@
...
+public:
+
+    /// @param mysqlConf: Configuration object for mysql connection
+    DbInterfaceMySql(mysql::MySqlConfig const& mysqlConf);
```
iagaponenko: This may need to be declared as `explicit` to avoid being used in the implicit conversions.

core/modules/css/DbInterfaceMySql.h
```diff
@@ -0,0 +1,68 @@
...
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+
+#ifndef LSST_QSERV_CSS_DBINTERFACE_H
+#define LSST_QSERV_CSS_DBINTERFACE_H
+
+// System headers
+#include <mutex>
+#include <set>
+
```
iagaponenko: Technically one would also have to add this:
```
#include <string>
```

core/modules/css/DbInterfaceMySql.h
```diff
@@ -0,0 +1,68 @@
...
+    DbInterfaceMySql(DbInterfaceMySql const&) = delete;
+    DbInterfaceMySql& operator=(DbInterfaceMySql const&) = delete;
+
+    // Destructor
+    ~DbInterfaceMySql() = default;
+
+    /// @return a set of empty chunks for database 'dbName'
+    //  @throws QMetaError if the table cannot be opened.
+    std::set<int> getEmptyChunks(std::string const& dbName);
```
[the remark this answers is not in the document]
jgates108: Interesting point. vector's are certainly more cache friendly but usually there are going to be several searches into the set to see if the chunk is in it, which should be faster with a set than a vector.

core/modules/css/DbInterfaceMySql.cc
```diff
@@ -0,0 +1,126 @@
...
+namespace lsst {
+namespace qserv {
+namespace css {
+
+
+class KvTransaction : public sql::SqlTransactionStd
+{
+public:
+    // Constructors
+    KvTransaction(sql::SqlConnection& conn)
```
iagaponenko: This may need to be as well declared as `explicit` to avoid being used in the implicit type conversion. See: https://en.cppreference.com/w/cpp/language/explicit

core/modules/css/EmptyChunks.h
jgates108: Empty chunks is just supposed to be a cache of empty chunk information read from elsewhere.

core/modules/css/DbInterfaceMySql.cc
```diff
@@ -0,0 +1,126 @@
...
+};
+
+
+DbInterfaceMySql::DbInterfaceMySql(mysql::MySqlConfig const& mysqlConf)
+: _conn(mysqlConf) {
+}
+
+
+std::set<int> DbInterfaceMySql::getEmptyChunks(std::string const& dbName) {
+    std::string funcName("getEmptyChunksFromDb");
```
iagaponenko: An alternative to this is the predefined `__func__`. With that one would not need to chase down all occurrences of `getEmptyChunks*` and rename them into the new name in case of code refactoring.

core/modules/css/DbInterfaceMySql.h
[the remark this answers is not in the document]
andy-slac: Hmm, without being able to predict what other data do we need in CSS in tabular format it's hard to tell how useful this tabular data interface is going to be. It may well be that empty chunk list is the only thing that we need.

core/modules/css/DbInterfaceMySql.cc
```diff
@@ -0,0 +1,126 @@
...
+    std::lock_guard<std::mutex> sync(_dbMutex);
+
+    KvTransaction trans(_conn);
+
+    /// This must match the name of the table generated by
+    /// dataloader.py DataLoader _makeEmptyChunksTable()
+    std::string emptyChunkTbl = dbName + "_EmptyChunks";
```
iagaponenko: All non-mutable variable which are initialized once are better to be declared with `const`

core/modules/css/DbInterfaceMySql.h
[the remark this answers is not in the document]
andy-slac: Binary search in a sorted vector is theoretically the same as search in binary tree, but due to locality it may as well be that vector search is faster.

core/modules/css/EmptyChunks.h
andy-slac: What I do not like about it is that it knows about all possible ways in which empty chunk list can be stored, i.e. it has too many dependencies. In other words its not just a cache, it's also code that populates cache.

core/modules/css/DbInterfaceMySql.cc
```diff
@@ -0,0 +1,126 @@
...
+
+    // run query
+    sql::SqlErrorObject errObj;
+    sql::SqlResults results;
+    std::string const query = "SELECT * FROM " + emptyChunkTbl;
```
iagaponenko: I understand that the current implementation of the method is built on assumption that the empty chunk list tables have a single column (`chunkId`, etc.) and that is the very first column in the schema. Although it may seem to be sufficient to `SELECT * ...` now this makes code more susceptible to the schema migrations. Hence, to protect this code from the migrations one may consider a small reinforcement like `SELECT chunkId ...`.

core/modules/css/DbInterfaceMySql.cc
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    trans.commit();
+
+    // Create the set
+    std::set<int> emptyChunkSet;
+    for (auto j:emptyChunks) {
+        try {
+            emptyChunkSet.insert(std::stoi(j));
+        } catch (std::invalid_argument const& e) {
```
iagaponenko: Strictly speaking, you may also need to intercept the `out-of-band` exceptions when a value is not falling into the allowed range for type `int`. Then you would need to catch `std::out_of_range` as well.

core/modules/css/EmptyChunks.cc
```diff
@@ -35,71 +35,50 @@
 #include "lsst/log/Log.h"
 
 // Qserv headers
+#include "css/DbInterfaceMySql.h"
+#include "css/CssError.h"
```
iagaponenko: Headers would need to be swapped to comply with the adopted header sorting rules.

core/modules/sql/SqlTransactionStd.cc
jgates108: The things calling this are catching different specific exceptions. Throwing something else will cause them to break. I'll need to think about this.

core/modules/css/EmptyChunks.h
```diff
@@ -46,37 +46,52 @@ namespace lsst {
...
     // Convenience types
     typedef std::shared_ptr<IntSet> IntSetPtr;
     typedef std::shared_ptr<IntSet const> IntSetConstPtr;
 
+    /// @return all the empty chunks for database 'db'.
+    IntSet _populate(std::string const& db);
+
+
     typedef std::map<std::string, IntSetPtr> IntSetMap;
+    std::shared_ptr<DbInterfaceMySql> _dbI; ///< allow access to empty chunks table.
```
iagaponenko: This is better to be declared as `const`

core/modules/sql/SqlTransactionStd.cc
andy-slac: Well, the point is that overriding `throwException` does not work if you call it from constructor, in that respect things are already broken so you need to fix it somehow.

core/modules/qproc/QuerySession.cc
```diff
@@ -230,7 +231,15 @@ QuerySession::getDbStriping() {
 std::shared_ptr<IntSet const>
 QuerySession::getEmptyChunks() {
     // FIXME: do we need to catch an exception here?
-    return _css->getEmptyChunks().getEmpty(_context->dominantDb);
+    if (_css != nullptr) {
```
iagaponenko: Is it even possible to not have CSS initialized in this context?

core/modules/css/EmptyChunks.h
iagaponenko: I agree with Andy. In theory a better design would be to turn `EmptyChunk` into an interface (an abstract class) and to feed a shared pointer to the interface into the normal constructor of `CssAccess` as it used to be in the previous version of the code. This would also eliminate extra parameters passed into `CssAccess` which are presently forwarded to the multi-faced `ChunkList` (like database connector, file path, etc.). Then implement `EmptyChunk` in different ways, for specific sources of the empty chunks. In theory one can see the factory method to be added to class `EmptyChunk`, something along these lines:
```
class EmptyChunk: std::enable_shared_from_this<EmptyChunk>{
public:
    static std::shared_ptr< EmptyChunk> load(<db-connector>, <file-name>, etc...);

   /// The current interface goes below...
   ...
};
```

core/modules/sql/SqlTransactionStd.cc
jgates108: More of a need to think about how to fix it as it needs to be fixed.

PR #474, opened 2019-07-18
DM-20693: Fix aliasing for long SELECT list items

n8pease:


core/modules/ccontrol/UserQuerySelect.h
```diff
@@ -178,7 +178,7 @@ class UserQuerySelect : public UserQuery {
     std::string _queryIdStr{QueryIdHelper::makeIdStr(0, true)};
     bool _killed{false};
     std::mutex _killMutex;
-    std::string _errorExtra;    ///< Additional error information
+    mutable std::string _errorExtra;    ///< Additional error information
```
n8pease: so the error can be set from a const method

andy-slac: We should really switch to more reasonable error reporting that does not need this internal state variable.

core/modules/ccontrol/UserQuerySelect.cc
```diff
@@ -214,11 +214,13 @@ std::string UserQuerySelect::getResultQuery() const {
...
     for (auto const& valueExpr : valueExprList) {
         if (valueExpr->isStar()) {
-            std::string errMsg;
             auto useSelectList = std::make_shared<query::SelectList>();
             useSelectList->addValueExpr(valueExpr);
             query::SelectStmt starStmt(useSelectList, _qSession->getStmt().getFromList().clone());
-            auto schema = _infileMerger->getSchemaForQueryResults(starStmt, errMsg);
+            sql::Schema schema;
+            if (not _infileMerger->getSchemaForQueryResults(starStmt, schema)) {
+                _errorExtra = "Could not get schema for query results.";
```
andy-slac: This message may be too cryptic if this is all that we pass to client, maybe some extra context would be useful if we want to make it easier for users and ourselves to debug this rare situation later.

andy-slac: Also, don't you want to stop right away if `getSchemaForQueryResults()` failed? Is it defined what state `schema` is going to have in that case?

core/modules/qana/TablePlugin.cc
```diff
@@ -267,14 +275,28 @@ TablePlugin::applyPhysical(QueryPlugin::Plan& p,
 }
 
 
+std::string TablePlugin::_getNextValueExprAlias() {
+    return "valueExprAlias_" + std::to_string(_nextValueExprAlias++);
```
andy-slac: I'm not very familiar with plugins, is this code thread safe (or is it called from other protected code)?

core/modules/qana/TablePlugin.cc
```diff
@@ -267,14 +275,28 @@ TablePlugin::applyPhysical(QueryPlugin::Plan& p,
...
         return;
     }
     if (not tableRef->hasAlias()) {
-        tableRef->setAlias(tableRef->hasDb() ?
-                           tableRef->getDb() + "." + tableRef->getTable() :
-                           tableRef->getTable());
+        std::string alias = tableRef->hasDb() ?
+                tableRef->getDb() + "." + tableRef->getTable() :
+                tableRef->getTable();
+        if (alias.size() >MYSQL_FIELD_MAX_LEN) {
```
andy-slac: Add space after `>`

core/modules/qana/TablePlugin.h
```diff
@@ -48,7 +48,7 @@ class TablePlugin : public QueryPlugin {
     // Types
     typedef std::shared_ptr<TablePlugin> Ptr;
 
-    TablePlugin() {}
+    TablePlugin() : _nextValueExprAlias(0), _nextTableRefAlias(0) {}
```
andy-slac: You could also write initializer in member declaration, saves some typing.

core/modules/qana/testTablePlugin.cc
```diff
@@ -59,8 +59,11 @@ using lsst::qserv::query::TestFactory;
 
 
 struct TestFixture {
-    TestFixture(void) : schemaCfg(sql::SqlConfig(sql::SqlConfig::MockDbTableColumns(
-            {{"Somedb", {{"Object", {"objectId", "ra_PS", "decl_PS", "rFlux_PS_Sigma"}}}}}))) {
+    TestFixture(void)
+            : schemaCfg(sql::SqlConfig(sql::SqlConfig::MockDbTableColumns(
+            {{"Somedb", {{"Object", {"objectId", "ra_PS", "decl_PS", "rFlux_PS_Sigma"}}}},
+             {"Very_Long_Database_Name_So_That_It_And_The", {{"Table_are_65_char_long", {"ColumnName"}}}},
```
andy-slac: Hard for me to verify that, I'm going to trust you blindly 🙈

core/modules/rproc/InfileMerger.cc
```diff
@@ -350,51 +350,42 @@ int InfileMerger::makeJobIdAttempt(int jobId, int attemptCount) {
...
-        return sql::Schema();
+        LOGS(_log, LOG_LVL_ERROR, "Failed to extract schema from result: " << errObj.errMsg());
+        _error = InfileMergerError(util::ErrorCode::INTERNAL, "Failed to extract schema from result: " + errObj.errMsg());
+        return false;
     }
-    auto schema = results.makeSchema(errObj);
+    schema = results.makeSchema(errObj);
     LOGS(_log, LOG_LVL_DEBUG, "InfileMerger extracted schema: " << schema);
-    return schema;
+    return true;
```
andy-slac: Next logical step would be to make it throw an exception (and return schema), that would make reviewers even happier. 😸

n8pease: I'm all with you, but this is where you already dinged me once for throwing where the rest of the class was return-resulting. I agree, it wants a wholesale refactor, but not today 🙃

core/modules/qana/TablePlugin.cc
n8pease: it's single threaded at this point.

PR #475, opened 2019-07-24
DM-17295: Improve qserv_areaspec hinting

n8pease:


core/modules/qana/QservRestrictorPlugin.cc
```diff
@@ -64,27 +64,354 @@
...
+    std::shared_ptr<const query::FuncExpr> scisqlFunc;
+    if (nullptr == topLevelAnd) return nullptr;
+    for (auto const& boolTerm : topLevelAnd->_terms) {
+        auto boolFactor = std::dynamic_pointer_cast<const query::BoolFactor>(boolTerm);
+        if (nullptr == boolFactor) continue;
+        for (auto boolFactorTerm : boolFactor->_terms) {
+            auto compPredicate = std::dynamic_pointer_cast<const query::CompPredicate>(boolFactorTerm);
+            if (nullptr == compPredicate) continue;
+            if (compPredicate->op != query::CompPredicate::EQUALS_OP) continue;
+            for (auto const& valueExpr: {compPredicate->left, compPredicate->right})
```
iagaponenko: Please, put `{` in the end of the line and indent subsequent `if`. Otherwise the code looks confusing (the first time one reads it 😸 ), and the subsequent if looks like just an ordinary standalone statement.

core/modules/qana/QservRestrictorPlugin.cc
```diff
@@ -64,27 +64,354 @@
...
+bool isScisqlAreaFunc(query::ValueExpr const& valueExpr) {
+    if (not valueExpr.isFunction()) return false;
+    auto&& funcExpr = valueExpr.getFunction();
+    if (funcExpr->getName().find("scisql_s2Pt") != 0) return false;
+    return true;
+}
+
+
+/**
+ * @brief If there is exactly one scisql area restrictor in the top level AND of the where clause,
```
iagaponenko: I'm not sure the implementation of the method finds "exactly one" or just the very first restrictor(?)  Is it guaranteed that the the top level ("root") term has zero or 1 "bool factor" term?

core/modules/qana/QservRestrictorPlugin.cc
```diff
@@ -64,27 +64,354 @@
...
+        for (auto boolFactorTerm : boolFactor->_terms) {
+            auto compPredicate = std::dynamic_pointer_cast<const query::CompPredicate>(boolFactorTerm);
+            if (nullptr == compPredicate) continue;
+            if (compPredicate->op != query::CompPredicate::EQUALS_OP) continue;
+            for (auto const& valueExpr: {compPredicate->left, compPredicate->right})
+            if (isScisqlAreaFunc(*valueExpr)) {
+                if (scisqlFunc != nullptr) {
+                    return nullptr;
+                }
+                scisqlFunc = valueExpr->getFunction();
```
iagaponenko: So, if there is more than one such term then the last one would be used?

core/modules/qana/QservRestrictorPlugin.cc
```diff
@@ -64,27 +64,354 @@
...
+                           query::QueryContext& context) {
+    if (restrictors.empty()) return;
+
+    auto const& tableList = fromList.getTableRefList();
+    RestrictoryEntryList chunkedTables;
+    getTable gt(*context.css, chunkedTables);
+    std::for_each(tableList.begin(), tableList.end(), gt);
+    // chunkedTables is now populated with a RestrictorEntry for each table in the FROM list that is chunked.
+    if (chunkedTables.empty()) {
+        throw qana::AnalysisError("Spatial restrictor without partitioned table.");
```
iagaponenko: I thought that any spatial restrictors will be simply ignored for non-partitioned (non-chunked) tables as per its, `2` of the proposal in:
https://jira.lsstcorp.org/browse/DM-17295

core/modules/qana/QservRestrictorPlugin.cc
```diff
@@ -146,7 +457,8 @@ query::QsRestrictor::Ptr newRestrictor(
 {
     // Extract the literals, bailing out if we see a non-literal
     bool isValid = true;
-    std::for_each(values.begin(), values.end(), validateLiteral(isValid));
+    std::for_each(values.begin(), values.end(),
+        [&](query::ValueExprPtr p) { isValid = isValid && p != nullptr && not p->copyAsLiteral().empty(); });
```
iagaponenko: Better to restrict it to:
```
   [&isValid](uery::ValueExprPtr p) { ...
```

core/modules/query/Constraint.h
```diff
@@ -45,6 +45,12 @@ namespace query {
...
 class Constraint {
 public:
+    Constraint() = default;
+
+    Constraint(std::string const& name_, std::vector<std::string> const& params_)
+            : name(name_), params(params_) {
+    }
+
     std::string name;
     std::vector<std::string> params;
```
iagaponenko: If `name` and `params` are only set by the constructor(s) then they may be declared as `const`

core/modules/query/QueryContext.cc
```diff
@@ -142,6 +142,18 @@ QueryContext::getValueExprMatch(std::shared_ptr<query::ValueExpr> const& valExpr
 }
 
 
+void QueryContext::addRestrictors(RestrList const& restrictors_) {
+    if (restrictors_.empty()) {
+        return;
+    }
+    if (nullptr == restrictors) {
```
iagaponenko: This test is really confusing, primarily because the type `RestrList` of the input parameter has no indication that it's the shared pointer type. Could be renamed into `RestrListPtr` perhaps?

core/modules/query/ValueExpr.cc
```diff
@@ -308,7 +308,28 @@ std::shared_ptr<ValueFactor> ValueExpr::getFactor() {
 }
 
 
-/// @return true if holding a single ValueFactor
+std::string ValueExpr::getConstVal() const {
+    if (_factorOps.size()) {
```
iagaponenko: Explicitly compare with 0?

core/modules/query/ValueExpr.cc
```diff
@@ -321,11 +342,13 @@ bool ValueExpr::isColumnRef() const {
...
-        ValueFactor const& factor = *_factorOps.front().factor;
-        if (factor.getType() == ValueFactor::FUNCTION) {
-            return true;
-        }
+    return getFunction() != nullptr;
+}
+
+
+bool ValueExpr::isConstVal() const {
+    if (_factorOps.size() == 1 && _factorOps[0].factor->isConstVal()) {
```
iagaponenko: Could be all done in one line:
```
return _factorOps.size() == 1 && _factorOps[0].factor->isConstVal();
```

core/modules/query/WhereClause.h
```diff
@@ -83,6 +83,18 @@ class WhereClause {
...
     }
+
+    /**
+     * @brief Determine if there are any QsRestrictors in the WHERE clause.
+     *
+     * If there are restrictors, this guarantees that getRestrs() will return a valid pointer to a vector
+     * that has one or more objects.
+     *
+     * @return true if there are restrictors.
+     * @return false if there are not restrictors.
```
iagaponenko: That's redundant 😸

core/modules/qana/QservRestrictorPlugin.cc
```diff
@@ -517,51 +538,49 @@ QservRestrictorPlugin::applyLogical(query::SelectStmt& stmt,
...
-    getTable gt(*context.css, entries);
-    std::for_each(tList.begin(), tList.end(), gt);
 
+    // If there's no where clause then there's no need to do any work here.
     if (!stmt.hasWhereClause()) { return; }
 
+    // Get a list of the chunked tables (as a list of RestrictorEntry).
+    query::TableRefList& tList = stmt.getFromList().getTableRefList();
+    RestrictoryEntryList chunkedTables;
+    getTable gt(*context.css, chunkedTables);
```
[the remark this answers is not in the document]
n8pease: Yeah, since it was moved I didn't get into cleaning that stuff up at all. I can go ahead with caps there tho.

core/modules/qana/QservRestrictorPlugin.h
```diff
@@ -80,7 +68,7 @@ class QservRestrictorPlugin : public QueryPlugin {
     std::string name() const override { return "QservRestrictorPlugin"; }
 
 private:
-    std::shared_ptr<query::BoolTerm> _makeCondition(std::shared_ptr<query::QsRestrictor> const restr,
+    std::shared_ptr<query::BoolTerm> _makeCondition(std::shared_ptr<const query::QsRestrictor> const restr,
```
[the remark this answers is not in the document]
n8pease: I'll do even one better and make it a const ref to a QsRestrictor, and deref the pointer on the way in (it's only used temporarily)

core/modules/qana/QservRestrictorPlugin.cc
```diff
@@ -64,27 +64,354 @@
...
+        if (alias.empty()) {
+            // For now, only accept aliased tablerefs (should have
+            // been done earlier)
+            throw qana::AnalysisBug("Unexpected unaliased table reference");
+        }
+        std::vector<std::string> pCols = partParam.partitionCols();
+        RestrictorEntry se(alias,
+                           StringPair(pCols[0], pCols[1]),
+                           pCols[2]);
+        _chunkedTables.push_back(se);
```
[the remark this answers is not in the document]
n8pease: no

core/modules/qana/QservRestrictorPlugin.cc
```diff
@@ -64,27 +64,354 @@
...
+        query::JoinRefPtrVector& joinRefs = t.getJoins();
+        typedef query::JoinRefPtrVector::iterator Iter;
+        for (Iter i=joinRefs.begin(), e=joinRefs.end(); i != e; ++i) {
```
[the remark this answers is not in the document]
n8pease: cargo cult. I'll clean it up tho.

core/modules/qana/QservRestrictorPlugin.cc
```diff
@@ -64,27 +64,354 @@
...
+    Restriction(query::QsRestrictor const& r)
+        : _name(r._name) {
+        _setGenerator(r);
+    }
+
+    query::BoolFactor::Ptr generate(RestrictorEntry const& e) {
+        return (*_generator)(e);
+    }
+
+    class Generator {
```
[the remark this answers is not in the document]
n8pease: good call, no.

core/modules/qana/QservRestrictorPlugin.cc
```diff
@@ -64,27 +64,354 @@
...
+            const int use_string = AreaGenerator::USE_STRING;
+            _generator = std::make_shared<AreaGenerator>("s2PtInCPoly",
+                                                         use_string,
+                                                         r._params
+                                                         );
+        } else {
+            throw qana::AnalysisBug("Unmatched restriction spec: " + _name);
+        }
+    }
+    std::string _name;
```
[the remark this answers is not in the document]
n8pease: oh, excellent.

core/modules/qana/QservRestrictorPlugin.cc
```diff
@@ -64,27 +64,354 @@
...
+ *
+ * This is a helper function for handleScisqlRestrictors.
+ *
+ * @param valueExpr The ValueExpr to check.
+ * @return true if the ValueExpr is a function and starts with `scisql_s2PtIn` else false.
+ */
+bool isScisqlAreaFunc(query::ValueExpr const& valueExpr) {
+    if (not valueExpr.isFunction()) return false;
+    auto&& funcExpr = valueExpr.getFunction();
+    if (funcExpr->getName().find("scisql_s2Pt") != 0) return false;
```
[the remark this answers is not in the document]
n8pease: but it calls `find` on the returned string from `getName`, and `find` returns the location where the beginning of the search string was found. This is like `boost::starts_with` but no boost.

core/modules/qana/QservRestrictorPlugin.cc
```diff
@@ -64,27 +64,354 @@
...
+    void operator()(query::TableRef::Ptr t) {
+        // FIXME: Modify so we can use TableRef::apply()
+        if (!t) {
+            throw qana::AnalysisBug("NULL TableRefN::Ptr");
+        }
+        (*this)(*t);
+    }
+    virtual void operator()(query::TableRef& t) {
+        std::string const& db = t.getDb();
+        std::string const& table = t.getTable();
```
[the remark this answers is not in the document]
n8pease: I've been thinking bout this lately; and I want to be sure I'm getting a `const&`, which if I understand `auto` correctly, if the function returns a non-const reference then the variable will also be non-const. Is that right? I'd rather be const correct...

core/modules/qana/QservRestrictorPlugin.cc
n8pease: oh man that was a typo not having braces for the for. I'm just lucky I got away with it because there was exctly one 'if' statement in the loop 🤣

core/modules/qana/QservRestrictorPlugin.cc
n8pease: it's exactly one; see ```if (scisqlFunc != nullptr) {
                        return nullptr;```

core/modules/qana/QservRestrictorPlugin.cc
n8pease: I'm not sure if you're referring to line 309 or line 311:
WRT 309: the first match it finds sets `scisqlFunction` to a non-null value, and the second match will go into line 309 and returns `nullptr`.
WRT 311, `ValueExpr::getFunction` returns a valid pointer only if the list of `ValueFactor` in the `ValueExpr` contains exactly one item.

core/modules/qana/QservRestrictorPlugin.cc
n8pease: My takeaway was if there is a `qserv_areaspec_` restrictor, then use 'old' behavior (including throwing if there is not a chunked table. Otherwise, do the new thing.

n8pease: (this code handles the 'old' case)

core/modules/qana/QservRestrictorPlugin.cc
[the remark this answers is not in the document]
n8pease: ...which is to say, I'm looking for exactly the beginning (position 0) as opposed to 'anything bu the end'.

core/modules/qana/QservRestrictorPlugin.cc
n8pease: Thanks! I'm still getting the hang of lambdas...

core/modules/query/QueryContext.cc
n8pease: I'm disinclined to rename a type to `...Ptr` this without a broader effort (at least in the Czar code); there's a lot of places where containers are owned by a shared_ptr and _don't_ follow that convention, and to mix & match, at least not without an explicit intention of going one way or the other (and I've been admonished for trying to introduce Hungarian Notation before...) seems like it's introducing more confusion than clarity. OTOH, there are 2 variables here: `restrictors` and `restrictors_`, the latter has that underscore which we often use to mean "input variable" (grumble). I'll rename the input var to `newRestrictors` or something, to make it more clear.

core/modules/qana/QservRestrictorPlugin.cc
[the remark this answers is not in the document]
iagaponenko: 👍

core/modules/qana/QservRestrictorPlugin.cc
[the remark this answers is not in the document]
iagaponenko: You may be right, unless you want to turn the table reference into `const` in the operator declaration:
```
    virtual void operator()(query::TableRef const& t) {
```

PR #476, opened 2019-07-25
DM-19337: qserv can't parse functions with an alias in the select statement

n8pease:


core/modules/parser/QSMySqlListener.cc
```diff
@@ -1999,6 +1999,10 @@ class SelectExpressionElementAdapter :
         _valueExpr = valueExpr;
     }
 
+    virtual void handleUid(string const & string) {
```
iagaponenko: I guess the name of the parameter should be more descriptive, something like `uid`. That `string` is too generic. Also, remove extra space in `const &`.

iagaponenko: And the second note or question - is this method overriding the one from the base class/interface `handleUid`? If so then it should be declared as:
```
void handleUid(string const& uid) override {
```

PR #479, opened 2019-08-05
Fixed a bug in the Replication System's monitoring probe witin Qserv …

iagaponenko:
…workers.


jgates108:
It looks good. I'm concerned about the Travis CI test hanging (unresponsive for 10m0s) but I don't think this change is responsible for that.

fritzm:
I'm concerned about the Travis hang, too.  Are we sure we didn't introduce a deadlock when taking that mutex?

PR #486, opened 2019-09-05
DM-21043: Qserv log diet: use named context for query ID

andy-slac:
Switch to using MDC for some context information instead of adding explicit info to each message.

### DIFF
--- a/core/modules/czar/Czar.cc
+++ b/core/modules/czar/Czar.cc
@@ -157,7 +157,6 @@ Czar::submitQuery(std::string const& query,
         std::lock_guard<std::mutex> lock(_mutex);
         uq = _uqFactory->newUserQuery(query, defaultDb, getQdispPool(), userQueryId, msgTableName, resultDb);
     }
-    auto queryIdStr = uq->getQueryIdString();
 
     // Add logging context with query ID
     QSERV_LOGCONTEXT_QUERY(uq->getQueryId());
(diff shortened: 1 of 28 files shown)


core/modules/czar/Czar.cc
```diff
@@ -134,6 +136,9 @@ Czar::submitQuery(std::string const& query,
     std::string const msgTableName = "message_" + userQueryId;
     std::string const lockName = resultDb + "." + msgTableName;
 
+    // Add logging context with user query ID
+    LOG_MDC_SCOPE("TID", userQueryId);
```
ktlim: Curious: How is TID different from QID?

andy-slac:
@jgates108, travis is failing, this is because it does not know how to pull new `log` version.

core/modules/czar/Czar.cc
andy-slac: We just had 1-hour discussion yesterday about how we want to get rid of TID 🙂 TID is a sort of temporary ID which is assigned to a request before the request is registered in QMeta and gets it actual QID. Some requests never go to QMeta, so for those we only have TID. The whole things is a mess and we'd like to get rid of TID, but it's not quite trivial.

andy-slac: Just to add - TID is a sort of randomly generated but not unique and not persistent, QID is unique and persistent.

PR #492, opened 2019-09-30
Improve the API and the implementation of the MySQL of the Replicatio…

iagaponenko:
…n & Ingest

system. Specifically:
- add exceptions for handling errors with queries involving MySQL partitions
- add ConnectionPool class to allow persistent connections to the MySQL services
- improve the RAII ConnectionHandler class to allow using the above mentioned
  class ConnectionPool in addition to ordinary objects of class Connection.
- do minor refactoring (and - where it makes a sense - the reduction) in
  the Doxygen style documentation.
- fix a bug in the SQL code generation

### DIFF
--- a/core/modules/replica/DatabaseMySQL.cc
+++ b/core/modules/replica/DatabaseMySQL.cc
@@ -237,6 +237,15 @@ void Connection::_processLastError(string const& context,
         case ER_DUP_ENTRY:
             throw DuplicateKeyError(msg);
 
+        case  ER_NO_SUCH_TABLE:
+            throw NoSuchTable(msg);
+
+        case ER_PARTITION_MGMT_ON_NONPARTITIONED:
+            throw NotPartitionedTable(msg);
+
+        case ER_UNKNOWN_PARTITION:
+            throw NoSuchPartition(msg);
+
         case ER_ABORTING_CONNECTION:
         case ER_NEW_ABORTING_CONNECTION:
 
@@ -572,6 +581,10 @@ void Connection::_connectOnce() {
         throw Error(context + "mysql_init failed");
     }
 
+    // Only allow TCP/IP, no UNIX sockets for now
+    enum mysql_protocol_type prot_type= MYSQL_PROTOCOL_TCP;
+    mysql_optionsv(_mysql, MYSQL_OPT_PROTOCOL,(void*)&prot_type);
+
     // Make a connection attempt
 
     if (nullptr == mysql_real_connect(
@@ -612,7 +625,21 @@ void Connection::_connectOnce() {
     // Set session attributes
 
     vector<string> queries;
-    queries.push_back("SET SESSION SQL_MODE='ANSI'");
+
+    // The change is related to the following bug in MariaDB before 10.2.8:
+    //    https://jira.mariadb.org/browse/MDEV-16792
+    //
+    // This bug causes problems with partitioned tables if partitions are created
+    // with SQL_MODE='ANSI'. This mode sets 'ANSI_QUOTES` (double quotes for identifiers).
+    // Details are in:
+    //   https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_ansi_quotes
+    //
+    // The problem is seen in clients which aren't setting this mode. Hence a workaround
+    // is to either keep this mode here or to set that mode in all clients (Qserv).
+    // The current solution is to disable this:
+    //
+    //   queries.push_back("SET SESSION SQL_MODE='ANSI'");
+
     queries.push_back("SET SESSION AUTOCOMMIT=0");
     //
     // TODO: Reconsider this because it won't work in the modern versions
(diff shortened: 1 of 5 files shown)


core/modules/replica/DatabaseMySQL.cc
```diff
@@ -572,6 +581,10 @@ void Connection::_connectOnce() {
         throw Error(context + "mysql_init failed");
     }
 
+    // Only allow TCP/IP, no UNIX sockets for now
+    enum mysql_protocol_type prot_type= MYSQL_PROTOCOL_TCP;
```
andy-slac: Space before `=`

core/modules/replica/DatabaseMySQL.cc
```diff
@@ -664,4 +691,110 @@ void Connection::_assertTransaction(bool inTransaction) const {
...
+
+    // Open a new connection and return it right away if the limit hasn't
+    // been reached yet.
+    //
+    // TODO: the factory method called below may put a calling thread
+    // in the blocking state while the (database) service becomes available.
+    // This will prevent operations with the pool by other threads. Investigate
+    // a non-blocking algorithm.
+
+    if (_availableConnections.size() + _usedConnections.size() < _maxConnections) {
```
andy-slac: Can you reuse a connection from `_availableConnections` if that is non-empty?

core/modules/replica/DatabaseMySQL.cc
```diff
@@ -664,4 +691,110 @@ void Connection::_assertTransaction(bool inTransaction) const {
...
+        auto conn = Connection::open(_params);
+        _usedConnections.push_back(conn);
+        return conn;
+    }
+
+    // Otherwise grab an existing one (which may require to wait before
+    // it would become available).
+
+    auto const self = shared_from_this();
```
andy-slac: Not sure I understand this - if you are in this method then there should be no danger that reference count drops to zero, you can pass `this` to lambda below instead?

core/modules/replica/DatabaseMySQL.cc
```diff
@@ -664,4 +691,110 @@ void Connection::_assertTransaction(bool inTransaction) const {
...
+    string const context = "ConnectionPool::" + string(__func__) + "  ";
+
+    LOGS(_log, LOG_LVL_DEBUG, context);
+
+    unique_lock<mutex> lock(_mtx);
+
+    // Move it between queues.
+
+    size_t numRemoved = 0;
+    _usedConnections.remove_if(
```
andy-slac: `remove_if` is C++20 method, not sure we are allowed to use it.

core/modules/replica/DatabaseMySQLExceptions.h
```diff
@@ -55,14 +55,8 @@ namespace mysql {
...
-
-    /**
-     * @param what
-     *   reason for the exception
-     */
-    explicit Error(std::string const& what)
-        :   std::runtime_error(what) {
-    }
+    /// @param what  reason for the exception
+    explicit Error(std::string const& what) : std::runtime_error(what) {}
```
andy-slac: You can avoid writing this by using inherited constructor: `using std::runtime_error::runtime_error`.

core/modules/replica/DatabaseMySQLExceptions.h
```diff
@@ -162,14 +136,55 @@ class Reconnected : public Error {
...
-     */
-    explicit DuplicateKeyError(std::string const& what)
-        :   Error(what) {
-    }
+
+/**
+ * Instances of this exception class are thrown on queries involving non-existing
+ * tables.
+ */
+class NoSuchTable : public Error {
```
andy-slac: I believe that our standard convention is to add `Error` to all exception class names (e.g. `NoSuchTableError`).

core/modules/replica/DatabaseMySQLTypes.h
```diff
@@ -124,11 +124,13 @@ struct ConnectionParams {
 
     /**
      * @return
-     *   a string representation of all (but the password) parameters.
+     *   a string representation of all (but the password unless requested) parameters.
      *   The result will be formatted similarly to the one expected by
      *   the non-default constructor of the class.
+     *
+     * @oparam showPassword  if 'false' then hash a password in the result
```
andy-slac: `*o*param`?

`@param` usually goes above `@return`

core/modules/replica/DatabaseMySQL.cc
iagaponenko: That's certainly true for the algorithm `std::remove_if`. However, what I'm using here is a method of class `std::list::remove_if` which is available in the Standard library before `C++20`. See:
https://en.cppreference.com/w/cpp/container/list/remove
Unlike the newer version that pre-version-20 method won't return any result. And I don't expect it here.

core/modules/replica/DatabaseMySQL.cc
iagaponenko: 👍

core/modules/replica/DatabaseMySQL.cc
iagaponenko: Great suggestion! Thanks! I will implement it.

core/modules/replica/DatabaseMySQL.cc
andy-slac: OK, my fault, I looked at the wrong page.

core/modules/replica/DatabaseMySQL.cc
iagaponenko: Good point. I have reviewed tis class, and I don't see any reason why it should be based on the `std::enable_shared_from_this` paradigm because it doesn't initiate any asynchronous operations. I'm going to simplify the class design.

core/modules/replica/DatabaseMySQLExceptions.h
iagaponenko: I wasn't aware about this trick! Thanks!

iagaponenko: Okay. I have tried this, and it's not going to work since this is the `C++17` feature as per:
https://en.cppreference.com/w/cpp/language/using_declaration
Though, I like the idea. And I will keep it in mind.

iagaponenko: Or, to me more precise - most probably it would work with out current "almost-17" compiler which includes many features of `c++17` as "the experimental support", but we're not supposed to use this feature for now.

core/modules/replica/DatabaseMySQLExceptions.h
iagaponenko: If that's the case then I'd differ this to some later refactoring as I have many other cases within the same `replica` module where I'm not following this convention. At the point I will try to minimize changes in the current branch (PR request).

core/modules/replica/DatabaseMySQLTypes.h
iagaponenko: 👍

core/modules/replica/DatabaseMySQLExceptions.h
andy-slac: This is actually C++11 feature, should be safe to use.

PR #494, opened 2019-10-06
Tickets/dm 21524

iagaponenko:


core/modules/replica/Controller.h
```diff
@@ -27,22 +27,174 @@
...
+    class FindAllRequest;
+
+    typedef std::shared_ptr<FindRequest>    FindRequestPtr;
+    typedef std::shared_ptr<FindAllRequest> FindAllRequestPtr;
+
+    class EchoRequest;
+
+    typedef std::shared_ptr<EchoRequest> EchoRequestPtr;
+
+    class IndexRequest;
```
n8pease: nit: I don't understand why these comein groups of ones and twos.

core/modules/replica/Controller.h
```diff
@@ -27,22 +27,174 @@
...
+    typedef std::shared_ptr<SqlRemoveTablePartitionsRequest> SqlRemoveTablePartitionsRequestPtr;
+    typedef std::shared_ptr<SqlDeleteTablePartitionRequest>  SqlDeleteTablePartitionRequestPtr;
+
+    class StopReplicationRequestPolicy;
+    class StopDeleteRequestPolicy;
+    class StopFindRequestPolicy;
+    class StopFindAllRequestPolicy;
+    class StopEchoRequestPolicy;
+    class StopIndexRequestPolicy;
+    class StopSqlRequestPolicy;
```
n8pease: I guess this is the group of Stop Policies, and so forth.

core/modules/replica/Controller.h
```diff
@@ -103,204 +256,304 @@ class Controller : public std::enable_shared_from_this<Controller> {
...
+            std::string const& table,
+            uint32_t transactionId,
+            std::function<void(SqlDeleteTablePartitionRequestPtr)> const& onFinish=nullptr,
+            int priority=0,
+            bool keepTracking=true,
+            std::string const& jobId="",
+            unsigned int requestExpirationIvalSec=0);
+
+    template <class REQUEST>
+    typename REQUEST::Ptr stopById(
```
n8pease: is it worth documenting any of these?

core/modules/replica/Controller.h
```diff
@@ -103,204 +256,304 @@ class Controller : public std::enable_shared_from_this<Controller> {
...
+            std::string const& workerName,
+            std::string const& targetRequestId,
+            typename REQUEST::CallbackType const& onFinish=nullptr,
+            int priority=0,
+            bool keepTracking=true,
+            std::string const& jobId="",
+            unsigned int requestExpirationIvalSec=0) {
+
+        _debug(__func__, "targetRequestId: " + targetRequestId);
```
n8pease: doc?

core/modules/replica/Controller.cc
```diff
@@ -232,681 +89,537 @@ string Controller::_context(string const& func) const {
...
+        string const& password,
+        uint64_t maxRows,
+        SqlQueryRequest::CallbackType const& onFinish,
+        int priority,
+        bool keepTracking,
+        string const& jobId,
+        unsigned int requestExpirationIvalSec) {
 
-    util::Lock lock(_mtx, _context(__func__));
+    LOGS(_log, LOG_LVL_DEBUG, _context(__func__));
```
n8pease: does this want to go in as DEBUG? Or can it be TRACE?

n8pease: and so forth for other log statements, I don't make any more noise about it.

core/modules/replica/ControllerApp.h
```diff
@@ -111,6 +112,24 @@ class ControllerApp : public Application {
...
+    /// The name of a table 
+    std::string _sqlTable;
+
+    /// The name of a MySQL engine for tables to be created
+    std::string _sqlEngine;
+
+    /// The name of a file where to read table schema from
+    std::string _sqlSchemaFile;
+
+    /// The name of the PK for the MySQL partitioned tables
```
n8pease: what's a PK?

core/modules/replica/ControllerApp.cc
```diff
@@ -95,6 +98,42 @@ void printRequest(Request::Ptr const& request,
 }
 
 
+void printRequest(IndexRequest::Ptr const& request, string const& fileName) {
+    auto&& responseData = request->responseData();
+    cout << request->context() << "\n"
+         << "responseData.error: " << responseData.error << "\n"
+         << "performance:  " << request->performance() << endl;
+
+    if (fileName.empty()) cout << responseData.data;
```
n8pease: is it ok to mix single-line if with multi-line, braced, else? (I wouldn't think so...)

core/modules/replica/ControllerApp.cc
```diff
@@ -370,6 +577,10 @@ ControllerApp::ControllerApp(int argc, char* argv[])
     parser().command("SERVICE_DRAIN").description(
         "Drain all requests by stopping cancelling all ongoing requests"
         " and emptying all queues.");
+
+    parser().command("SERVICE_RECONFIG").description(
+        "Reload worker's Configuration. Requests known to a worker won't be affected"
+        " by the operation.");
```
n8pease: parser setup is almost 400 lines long. Maybe it could be factored out of the ctor for readability?

core/modules/replica/ControllerApp.cc
```diff
@@ -95,6 +98,42 @@ void printRequest(Request::Ptr const& request,
...
+    }
+}
+
+
+template <>
+void printRequest<StatusIndexRequest>(StatusIndexRequest::Ptr const& request) {
+    auto&& responseData = request->responseData();
+    cout << request->context() << "\n"
+         << "responseData.error: " << responseData.error << "\n"
+         << "performance:  " << request->performance() << endl;
```
n8pease: Having lots of variations of a templated printRequest function feels inside-out to me, why not add a virtual function (maybe pure?) to the base class & implement the specific variations of the function in the subclasses?

core/modules/replica/ControllerApp.cc
```diff
@@ -434,14 +645,145 @@ int ControllerApp::runImpl() {
...
+            not _doNotTrackRequest);
+    } else if ("SQL_DELETE_DATABASE" == _request) {
+        ptr = controller->sqlDeleteDb(
+            _workerName,
+            _sqlDatabase,
+            [&] (SqlDeleteDbRequest::Ptr const& ptr_) {
+                ::printRequest(ptr_,
+                               ptr_->responseData(),
+                               ptr_->performance(),
+                               _sqlPageSize);
```
n8pease: all these duplicate printRequest are visually noisy, and you're passing a pointer, and then member data extracted from the pointer. Why not just pass the pointer and use it in the called function. Or as mentioned above, why not make `printRequest` (or data extraction for `printRequest` a member function of the printed class (e.g. in this case `SqlDeleteDbRequest`)?

core/modules/replica/Request.cc
```diff
@@ -150,6 +151,13 @@ string const& Request::remoteId() const {
 }
 
 
+unsigned int Request::nextTimeIvalMsec() {
+    auto result = _currentTimeIvalMsec;
+    _currentTimeIvalMsec = min(2 * _currentTimeIvalMsec, 1000 * timerIvalSec());
```
n8pease: some explanation about this calculation would be good IMO.

core/modules/replica/Request.h
```diff
@@ -312,6 +312,32 @@ class Request : public std::enable_shared_from_this<Request>  {
     /// @return suggested interval (seconds) between retries in communications with workers
     unsigned int timerIvalSec() const { return _timerIvalSec; }
 
+    /**
+     * This method allows requests to implements an adaptive tracking algorithm
```
n8pease: implements->implement

core/modules/replica/SqlRequest.h
```diff
@@ -46,57 +48,173 @@ namespace qserv {
...
+private:
+
+    /**
+     * Start the timer before attempting the previously failed
+     * or successful (if a status check is needed) step.
+     *
+     * @param lock
+     *   a lock on Request::_mtx must be acquired before calling this method
+     */
+    void _wait(util::Lock const& lock);
```
n8pease: `_wait` is a funny name for a method that starts a timer...

core/modules/replica/SqlRequest.h
```diff
@@ -46,57 +48,173 @@ namespace qserv {
...
+    /**
+     * Send the serialized content of the buffer to a worker
+     *
+     * @param lock
+     *   a lock on Request::_mtx must be acquired before calling this method
+     */
+    void _send(util::Lock const& lock);
+
+    /**
+     * Process the completion of the requested operation
```
n8pease: process the completion, or analyze the completion?

core/modules/replica/SqlRequest.h
```diff
@@ -46,57 +48,173 @@ namespace qserv {
...
+    /// The results reported by a worker service
+    SqlResultSet _responseData;
+};
+
+
+/**
+ * Class SqlQueryRequest represents Controller-side requests for initiating
+ * arbitrary database queries at a remote worker nodes.
+ */
+class SqlQueryRequest : public SqlBaseRequest  {
```
n8pease: when should this class get its own file?

n8pease: (and so on with the other classes in this file...)

core/modules/replica/SqlRequest.cc
```diff
@@ -292,39 +251,652 @@ void SqlRequest::_analyze(bool success,
...
+        string const& worker,
+        std::string const& database,
+        CallbackType const& onFinish,
+        int priority,
+        bool keepTracking,
+        shared_ptr<Messenger> const& messenger)
+    :   SqlBaseRequest(serviceProvider,
+                       io_service,
+                       worker,
+                       0 /* maxRows */,
```
n8pease: nit: IMO `// maxRows` is a little better, safer (b/c `/* ... */` doesn't nest, and because there's nothing else on this line `//` is preferable.

n8pease: (and so on for the other cases in this file)

core/modules/replica/WorkerRequestFactory.cc
```diff
@@ -52,8 +56,11 @@ namespace replica {
 ///////////////////// WorkerRequestFactoryBase ////////////////////
 ///////////////////////////////////////////////////////////////////
 
-WorkerRequestFactoryBase::WorkerRequestFactoryBase(ServiceProvider::Ptr const& serviceProvider)
-    :   _serviceProvider(serviceProvider) {
+WorkerRequestFactoryBase::WorkerRequestFactoryBase(
+                            ServiceProvider::Ptr const& serviceProvider,
+                            database::mysql::ConnectionPool::Ptr const& connectionPool)
```
n8pease: I guess this is over-indented.

core/modules/replica/protocol.proto
```diff
@@ -168,25 +170,71 @@ message ProtocolRequestEcho {
...
+        DROP_DATABASE = 2;
+        ENABLE_DATABASE = 3;    // in Qserv
+        DISABLE_DATABASE = 4;   // in Qserv
+        GRANT_ACCESS = 5;
+        CREATE_TABLE = 6;
+        DROP_TABLE = 7;
+        REMOVE_TABLE_PARTITIONING = 8;
+        DROP_TABLE_PARTITION = 9;
+    }
+    required Type type = 2;
```
n8pease: You are not supposed to reuse field numbers.

n8pease:
I'm done reviewing. I think we should not reuse field numbers in protobuf files. Otherwise my comments are FWIW, and you can make changes as you see fit.

fritzm:
Affirmative; you may *not* reuse protobuf field IDs within the .proto file.  This must be corrected and re-reviewed before merging to master.

On Oct 8, 2019, at 11:08 AM, Nate Pease <notifications@github.com<mailto:notifications@github.com>> wrote:


I'm done reviewing. I think we should not reuse field numbers in protobuf files. Otherwise my comments are FWIW, and you can make changes as you see fit.

fritzm:
Also, please rebase and force-push this PR when addressing review feedback; that way we can use Travis to validate your build before merge.

core/modules/replica/Controller.h
iagaponenko: This is just due to historical reasons. I've rearranged type definitions by regrouping them.

core/modules/replica/Controller.h
iagaponenko: Correct! These forward declarations are used as template parameters of:
```
    template <typename POLICY> class StopRequest;
```

core/modules/replica/Controller.h
iagaponenko: Well, after receiving cute a bit of criticism for the Doxygen-style comments :smh I started culling the comments in favor of the "self-documented" code. In this particular case the method is used to stop a request of a specific type. Class `Controller` does 3 things:
* starts new requests
* stops requests by their identifiers
* reports a status of the ongoing requests
These three roles of the class are explained in the comment section above class definition. And I tough, the rest should be more-or-less obvious.

core/modules/replica/Controller.cc
iagaponenko: Thanks! I agree with you observation! I will migrate the log level.

core/modules/replica/ControllerApp.h
iagaponenko: The PRIMARY KEY of a table. I will correct the comment

core/modules/replica/ControllerApp.cc
iagaponenko: 👍

core/modules/replica/ControllerApp.cc
iagaponenko: This is an interesting problem. I'm using templates functions to extend classes. So, in theory, that `printRequest` functions could be virtual methods of the classes whose state is bing inspected by the current code. There were two reasons why I didn't do this when I was facing a similar dilemma. These were:
* I think the function templates produce less boiler-plate code compared with the virtual methods simply because most request classes would have similar implementations
* and the current class seems to be the only client of these functions. If I had more than one client then moving the functions elsewhere would be a good alternative.
Having said that, I'm still open to further refactoring of this code. I may not just do this today. My "rule of thumb" is that when I'm starting making improvements to some code and when I see something which makes it difficult to understand or modify the code then I go for a refactoring. Some sort of "lazy improvement" :smh

core/modules/replica/ControllerApp.cc
iagaponenko: I will consider code refactoring here!

iagaponenko: ...actually, having though more on this issue, I'v decided to go for the refactoring now. The proposal has been registered in my _TODO_ list:
https://confluence.lsstcorp.org/display/DM/Refactor+results+reporting+in+the+implementation+of+class+ControllerApp

core/modules/replica/Request.cc
iagaponenko: It's in the header file in the Doxygen documentation for the method. I'm not sure it's safe to repeat it here in case if I'd gave to change the algorithm. Then I would make nods in 3 places instead of 2.

core/modules/replica/Request.h
iagaponenko: 👍

core/modules/replica/SqlRequest.h
iagaponenko: Well, that's where my fantasy is falling short. I'm going to rename the method into `_waitAsync` so that the name matched what it really does - initiate the timer and register a callback for the completion of the interval.

core/modules/replica/SqlRequest.h
```diff
@@ -46,57 +48,173 @@ namespace qserv {
...
+     */
+    void _wait(util::Lock const& lock);
+
+    /**
+     * Callback handler for the asynchronous operation
+     *
+     * @param ec
+     *   error code to be checked
+     */
+    void _awaken(boost::system::error_code const& ec);
```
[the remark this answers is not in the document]
iagaponenko: I think `_handleAsync` is too generic because virtually all operations performed in a context of the request classes are asynchronous by the very nature of the Replication Framework (based on `BOOST ASIO`). Also, after I rename method `_wait` into `_waitAsync` the meaning of method `_awaken` becomes a bit more apparent. I'm also going to reject it more clear in the method documentation.

core/modules/replica/SqlRequest.h
iagaponenko: Good point! Fixed.

core/modules/replica/SqlRequest.h
iagaponenko: Agreed. I will refactor this huge file into many (one file per class).

core/modules/replica/protocol.proto
iagaponenko: Fixed

PR #497, opened 2019-10-18
DM-21343: query with ORDER BY and SELECT * fails

n8pease:


core/modules/qana/TablePlugin.cc
```diff
@@ -215,7 +215,11 @@ TablePlugin::applyLogical(query::SelectStmt& stmt,
 
     if (stmt.hasOrderBy()) {
         matchTableRefs(context, stmt.getOrderBy(), false);
-        matchValueExprs(context, stmt.getOrderBy(), true);
+        // The columns in the ORDER BY statement do not have to have a match in the select list, yet.
+        // This allows a query like `SELECT * ... ORDER BY col`, where col exists in the table (and so this
+        // is a valid query) but is not yet explicitly named. The existance of the ORDER BY columns will be
```
andy-slac: https://en.wiktionary.org/wiki/existance

core/modules/query/ValueExpr.h
```diff
@@ -91,8 +91,28 @@ class ValueExpr {
...
+     * The state of the member variable `aliasIsUserDefined` will not be changed.
+     */
     void setAlias(std::string const& alias);
+
+    /**
+     * @brief Set the Alias of the ValueExpr, and indicate if the alias is user defined or not.
+     *
+     * The state of the member variable `aliasIsUserDefined` will be changed.
+     */
+    void setAlias(std::string const& alias, bool aliasIsUserDefined);
```
andy-slac: I'm not sure how these two overloads of `setAlias` work w.r.t. `aliasIsUserDefined`. I think it's clear what happens if you call second overload - the state becomes either "user-defined" or "not-user-defined". But if you call first overload the state is "no-one-gives-a-damn"? Is that state useful? It's likely just my confusion as I do not know much about this code and I generally dislike modifiable state. 🙂

core/modules/query/ValueExpr.h
```diff
@@ -233,10 +261,12 @@ class ValueExpr {
     // determine if this object is the same as or a less complete description of the passed in object.
     bool isSubsetOf(ValueExpr const& valueExpr) const;
 
+    void setToAliasOnly();
```
andy-slac: Comment would be useful here.

core/modules/qana/TablePlugin.cc
n8pease: brutal! 😆

n8pease: (oh, as noted above)

fritzm:
Closing this pending further rework (see comments in Jira issue).

PR #498, opened 2019-10-24
Tickets/dm 21803

iagaponenko:


core/modules/replica/HttpModule.h
```diff
@@ -0,0 +1,140 @@
...
+#include "qhttp/Server.h"
+#include "replica/EventLogger.h"
+
+// This header declarations
+namespace lsst {
+namespace qserv {
+namespace replica {
+
+/**
+ * Class HttpModule is a base class for requests processing modules an HTTP server
```
n8pease: requests->request

core/modules/replica/HttpModule.h
```diff
@@ -0,0 +1,140 @@
...
+     * processors. These error conditions will be reported to as errors to callers.
+     * 
+     * @param req  the HTTP request
+     * @param resp  the HTTP response channel
+     * @param subModuleName  this optional parameter allows modules to have
+     * multiple sub-modules. A value of this parameter will be forwarded to
+     * the subclass-specific implementation of the pure virtual method
+     * HttpModule::executeImpl().
+     */
+    void execute(qhttp::Request::Ptr const& req,
```
n8pease: should this be marked either `virtual` or `final`? I guess I can't tell if this is supposed to be overridden or if it calls a `do` method that subclasses are supposed to override. Maybe I'll find out later... :-)

n8pease: Ok I looked it up and you can't do that. Too bad! The theory is that since it wasn't marked `virtual` then you're supposed to know that it's intentional that you should not override it. I think that places too much faith on the future readers to trust the past authors of intentional omission, but I guess that's the way it goes. Maybe if it's clearly explained that this calls function that subclasses should override?

core/modules/replica/HttpModule.h
iagaponenko: Correct. Fixed.

core/modules/replica/HttpModule.h
iagaponenko: Okay, I have rewritten the documentation of the method. Hopefully, this will help.

core/modules/replica/HttpWorkerStatusModule.h
```diff
@@ -0,0 +1,84 @@
...
+                           unsigned int workerResponseTimeoutSec,
+                           HealthMonitorTask::Ptr const& healthMonitorTask);
+
+    // Input parameters
+
+    /// A reference to the smart pointer is used to avoid increasing
+    /// the reference counter to the pointed object and to avoid the circular
+    /// dependency which would prevent object destruction.
+    /// TODO: consider the weak pointer?
+    HealthMonitorTask::Ptr const& _healthMonitorTask;
```
n8pease: weak pointer is the proper solution for circular dependencies. Do not keep a reference to a shared pointer, you're setting up a trap for future developers. IMO this must be changed.

iagaponenko: Agreed.

core/modules/replica/HttpRequestsModule.cc
```diff
@@ -0,0 +1,125 @@
...
+                   taskName,
+                   workerResponseTimeoutSec) {
+}
+
+
+void HttpRequestsModule::executeImpl(qhttp::Request::Ptr const& req,
+                                     qhttp::Response::Ptr const& resp,
+                                     string const& subModuleName) {
+
+    if      (subModuleName.empty()) _requests(req, resp);
```
n8pease: mixing with and without brackets in the if/else if/else clause is suspect, IMO.

core/modules/replica/HttpReplicationLevelsModule.h
```diff
@@ -0,0 +1,94 @@
...
+                                unsigned int workerResponseTimeoutSec,
+                                HealthMonitorTask::Ptr const& healthMonitorTask);
+
+    // Input parameters
+
+    /// A reference to the smart pointer is used to avoid increasing
+    /// the reference counter to the pointed object and to avoid the circular
+    /// dependency which would prevent object destruction.
+    /// TODO: consider the weak pointer?
+    HealthMonitorTask::Ptr const& _healthMonitorTask;
```
n8pease: must use weak_ptr

iagaponenko: Agreed.

core/modules/replica/HttpRequestsModule.cc
iagaponenko: The code is really trivial here. It's meant to dispatch incoming requests to the corresponding sub-modules. The way it's written is meant to make it more readable. I can still fix it to be in full compliance with the Style Guide.

iagaponenko: For the record. Fixed in this and other modules.

core/modules/replica/HttpReplicationLevelsModule.cc
```diff
@@ -0,0 +1,267 @@
...
+                double const percent = numAllReplicationSystemChunks == 0
+                        ? 0 : 100. * numOrphanReplicationSystemChunks / numAllReplicationSystemChunks;
+                databaseJson["levels"][0]["replication"]["online"]["num_chunks"] = numOrphanReplicationSystemChunks;
+                databaseJson["levels"][0]["replication"]["online"]["percent"   ] = percent;
+            }
+            result["families"][family]["databases"][database] = databaseJson;
+        }
+    }
+
+    // Update the cache before sending th eresponse
```
n8pease: th->the

core/modules/replica/HttpQservMonitorModule.cc
```diff
@@ -0,0 +1,477 @@
...
+            )
+        )
+    );
+
+    // Extract descriptions of those queries from qservMeta
+
+    json result;
+
+    if (not qids.empty()) {
+        h.conn->execute([&](decltype(h.conn) conn) {
```
n8pease: this lambda is too long, IMHO

core/modules/replica/HttpQservMonitorModule.cc
```diff
@@ -0,0 +1,477 @@
...
+            conn->begin();
+            conn->execute(
+                "SELECT * FROM " + conn->sqlId("QInfo") + "  WHERE " + conn->sqlIn("queryId", qids)
+            );
```
andy-slac: I think it's better to use QMeta API for qmeta access, not hand-written queries.

core/modules/replica/HttpIngestModule.h
```diff
@@ -0,0 +1,300 @@
...
+     *   TRANSACTIONS              for many transactions (possible selected by various criteria)
+     *   SELECT-TRANSACTION-BY-ID  for a single transaction
+     *   BEGIN-TRANSACTION         for starting a new transaction
+     *   END-TRANSACTION           for finishing/aborting a transaction
+     *   ADD-DATABASE              for adding a new database for the data ingest
+     *   PUBLISH-DATABASE          for publishing a database when data ingest is over
+     *   DELETE-DATABASE           for deleting an unpublished (partially ingest) database
+     *   ADD-TABLE                 for adding a new table for the data ingest
+     *   ADD-CHUNK                 for registering (or requesting a status of) of a new chunk
+     *   BUILD-CHUNK-LIST          for building (or rebuilding) an "empty chunk list"
```
n8pease: FWIW I'm not _totally_ buying this string-based submodule dispatch thing. 

I think an enum might be better?

core/modules/replica/HttpIngestModule.h
```diff
@@ -0,0 +1,300 @@
...
+                         qhttp::Response::Ptr const& resp);
+
+    /**
+     * Register a database table for an ingest
+     */
+    void _addTable(qhttp::Request::Ptr const& req,
+                   qhttp::Response::Ptr const& resp);
+
+    /**
+     * Register (if it's not register yet) a chunk for ingest.
```
n8pease: register->registered

core/modules/replica/HttpIngestModule.cc
```diff
@@ -0,0 +1,1465 @@
...
+    if (schema.is_null()) {
+        sendError(resp, __func__, "table schema is empty");
+        return;
+    }
+    if (not schema.is_array()) {
+        sendError(resp, __func__, "table schema is not defined as an array");
+        return;
+    }
+
+    list<pair<string,string>> columns;
```
n8pease: consider a struct or class, this is pretty deep nesting.

core/modules/replica/HttpIngestModule.cc
```diff
@@ -0,0 +1,1465 @@
...
+            cssAccess->createTable(
+                databaseInfo.name,
+                table,
+                databaseInfo.schema4css(table),
+                partParams,
+                scanParams
+            );
+        }
+    }
+    for (auto const& table: databaseInfo.partitionedTables) {
```
n8pease: I would have expected a space between `table` and the colon; `...table : databaseInfo...` but 🤷‍♂

core/modules/replica/HttpConfigurationModule.h
```diff
@@ -0,0 +1,165 @@
...
+                     qhttp::Response::Ptr const& resp,
+                     std::string const& subModuleName) final;
+
+private:
+
+    HttpConfigurationModule(Controller::Ptr const& controller,
+                            std::string const& taskName,
+                            unsigned int workerResponseTimeoutSec);
+
+    void _get(qhttp::Request::Ptr const& req,
```
n8pease: needs doc

core/modules/replica/HttpCatalogsModule.cc
```diff
@@ -0,0 +1,198 @@
...
+    vector<ReplicaInfo> replicas;
+    if (not dummyReport) databaseServices->findDatabaseReplicas(replicas, database);
+
+    json result;
+    result["chunks"]["unique"] = chunks.size();
+    result["chunks"]["with_replicas"] = replicas.size();
+
+    map<string, map<string, size_t>> stats;
+    for (auto&& table: config->databaseInfo(database).partitionedTables) {
+        stats[table]["data_unique_in_chunks_data"] = 0;
```
n8pease: it seems inefficient to not cache a ref to `stats[table]` and operate on that.

core/modules/replica/HttpCatalogsModule.cc
```diff
@@ -0,0 +1,198 @@
...
+                else           stats[table]["data_with_replica_in_chunks_data"]   += size;
+            }
+            if (isIndex) {
+                if (isOverlap) stats[table]["data_with_replica_in_overlaps_index"] += size;
+                else           stats[table]["data_with_replica_in_chunks_index"]   += size;
+            }            
+        }
+    }
+    for (auto&& table: config->databaseInfo(database).partitionedTables) {
+        result["tables"][table]["is_partitioned"] = 1;
```
n8pease: can you cache a reference to a json object at `result["tables"][table]`?

core/modules/replica/HttpCatalogsModule.cc
```diff
@@ -0,0 +1,198 @@
...
+    for (auto&& table: config->databaseInfo(database).regularTables) {
+
+        // TODO: implement this when the Replication system will support regular tables
+
+        size_t data_unique_data  = 0;
+        size_t data_unique_index = 0;
+        size_t data_with_replicas_data  = 0;
+        size_t data_with_replicas_index = 0;
+
+        result["tables"][table]["is_partitioned"] = 0;
```
n8pease: same re. ref

n8pease:
I'm done reviewing. IMO the shared_ptr& member vars must be changed to weak_ptr. Otherwise, comments are left for you to do with what you will 🙃

core/modules/replica/HttpQservMonitorModule.cc
iagaponenko: Refactored

core/modules/replica/HttpQservMonitorModule.cc
iagaponenko: No way 😹 I don't want to break my beautifully crafter Replication system by relying on an API which won't even do reconnect if the MySQL service goes down. I may come back to this idea after proper improvements will be done to that API. I'm 100% sure this will bite us really hard after migrating Qserv to Kubernetes.

core/modules/replica/HttpIngestModule.h
iagaponenko: I understand. When I was designing this "mini-framework" I went through a similar process by weighting "pros" and "const" of `enum` vs `name`. A problem with `enum` is that I would have to make then sub-class-specific. For example, if I had some module:
```
HttpMyModule : public HttpModule {
public:
    typedef td::shared<HttpMyModule> Ptr;
    enum SubModule {
        GetSomeSomethingDone,
        PullInfo,
        ..
    };
protected:
    void execureImpl(SubModule subModule, ... ) final;
    ..
};
```
Then I would have to store a pointer to that module in a header of class `HttpProcessor` by its final class, something along these lines:
```
class HttProcessor {
...
private:
    HttpMyModule::Ptr const _myModule;
};
```
Right now I'm not exposing final classes of modules in that header:
```
    HttpModule::Ptr const _myModule;
};
```
Exposing the final classes through the header would violate one of the fundamental principle of the OOP which is known as the "open-closed" principle:
https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle

core/modules/replica/HttpIngestModule.cc
iagaponenko: Well, this is an area which already caught my attention on a number of occasions when I was reviewing your code. And I was always wondering if the way you were doing it was right ... or wrong. So Now I wanted to finally settle the case 😹 and did a little research on this subject. First I looked at the LSST Style Guide for C++ hoping to find an example. Nothing specific for the C++11's `range for` found in there. Then I look at the documentation for this in here:
https://en.cppreference.com/w/cpp/language/range-for
And it turns out you win here!
I'm tempted to modify this file right away. Unfortunately, things get complicated by a fact that I have hundreds of places in my code which I would have to fix. And it would be silly to modify just this file w/o touching others. So, let's leave this till to a refactoring to be performed later. At the mean time (so that I wouldn't forget about tis) I have registered this case at my *TODO* list. Please, look at the very bottom of the document:
https://confluence.lsstcorp.org/display/DM/Completed%2C+on-going+and+proposed+improvements

core/modules/replica/HttpCatalogsModule.cc
iagaponenko: Good point! Fixed.

core/modules/replica/HttpConfigurationModule.h
iagaponenko: Added. Thank you for noticing this!

core/modules/replica/HttpCatalogsModule.cc
iagaponenko: Good point! Refactored.
Though, to be honest, caching references makes me feel a bit uneasy because what I'm dealing with is not a Standard Library's container. This is a class from the `nlohmann` JSON library which may not behave the way you might expect it. It may reallocate entries on each insert/modifications. I have already seen a few cases of the very strange (counter-intuitive) behavior in tis library. So, if my refactoring will break the code then all blame on you 😹

core/modules/replica/HttpCatalogsModule.cc
iagaponenko: Refactored as well.

core/modules/replica/HttpIngestModule.cc
iagaponenko: From my point of view, this statement is BOTH incorrect and correct. It's correct because I should have used a struct (or a class) here. And it's incorrect in is part which is related to the motivation behind replacing `<pair<string,string>` with a specialized type. The "deep nesting" is not a problem here at all. A reason for the refactoring has a different nature. The expressive type power of C++11 (on which , I guess, we both agree) allows and encourages using abstractions for everything. This makes the code more robust and more readable (less documentation is needed). The ability of expressing "light-weight abstractions" in the code is one of the most fundamental design ideas behind C++. You may read on this directly from the source (the language author's Web site):
http://www.stroustrup.com/ETAPS-corrected-draft.pdf

Compare, for example two ways of expressing column definitions:
```
pair<string,string> coldef;
// I know the first is the name of a column
coldef.first = "Column A";
// And I know that the second is its type
coldef.second = "VARCHAR(255)";
```
A problem with that class `pair` is way to generic, so I would have to tell everyone what is `first` and what is `second` in all contexts I would be using this definitions.
However if I have a structure which would look (for example) like this:
```
struct ColDef {
    std::string name;
    std::string type;
};
```
Then it becomes perfectly clear "who is who" and "what is what".
Having said that, I agree to refactor the code.

core/modules/replica/HttpQservMonitorModule.cc
andy-slac: QMeta uses sql/MysqlConnection which does try to re-connect is connection is lost.

_beautifully crafter_? I'm not going to comment on this but you know what I can say 😉

PR #500, opened 2019-11-05
Tickets/dm 19628

iagaponenko:


core/modules/replica/FileIngestApp.cc
```diff
@@ -0,0 +1,291 @@
...
+            throw invalid_argument(
+                    context
+                    + "Failed to parse the content of file " + _fileListName + ":" + to_string(lineNum)
+                    + ", expected fields: <worker-host> <worker-port> <transaction-id> <table> <type> <infile>.");
+        }
+
+        FileIngestSpec file;
+        file.workerHost = match[1].str();
+
+        auto const workerPort = stoul(match[2].str());
```
n8pease: I guess you're sure these will not throw, based on the regex. But the regex is a little obscure to read, and could be changed by some unwitting future developer (probably named Nate). It might be safer to try/catch these for safety?  Up to you tho.

core/modules/replica/FileIngestApp.cc
```diff
@@ -0,0 +1,291 @@
...
+    }
+    
+    // For partitioned tables analyze file name and extract a chunk number and
+    // the 'overlap' attribute
+    unsigned int chunk = 0;
+    bool isOverlap = false;
+
+    if (file.tableType == "P") {
+        string const filename = fs::absolute(path).filename().string();
+        regex const re("^chunk_([0-9]+)(_overlap)?\\.txt$", regex::extended);
```
n8pease: Some unit testing would make me more comfortable about things in general (including maintainability and poor future developers who must do it), and especially about knowing these regex strings working the way they're supposed to.

core/modules/replica/IngestClient.h
```diff
@@ -0,0 +1,228 @@
...
+     * @param chunk the number of a chunk. The parameter is ignored for
+     *   non-partitioned tables.
+     * @param isOverlap a flag indicating if this is the chunk 'overlap' table.
+     *   The parameter is ignored for non-partitioned tables.
+     * @param inputFilePath the path (relative or absolute) name for a file
+     *   whose content will be transferred to to the remote service.
+     * @param columnSeparator a character which separates columns within each row
+     * @throws IngestClientError for any problem occurred when establishing
+     *   a connection or during the initial handshake with the server
+     */
```
n8pease: nice comments.

core/modules/replica/IngestServerConnection.cc
```diff
@@ -0,0 +1,520 @@
...
+                           to_string(_transactionId) + "-%%%%" + fileExt;
+    fs::path const baseFileName = fs::unique_path(pattern, errCode);
+    if (errCode.value() != 0) {
+        _failed("failed to allocate the name for a temporary file, error: " + errCode.message());
+        return;
+    }
+
+    fs::path const filePath = fs::path(_workerInfo.loaderTmpDir) / baseFileName;
+    _fileName = filePath.string();
+    _file.open(_fileName, ofstream::out);
```
jgates108: This may be good thing to make a separate function and put it in util. For fs::unique_path to avoid using an existing file, wouldn't it need the full path in the pattern? Also, since the file is created after the name is generated, is there a tiny chance that another process (or thread) used the name to open file in that gap? It may be worth making multiple attempts if the file can't be opened due to a file already existing.

core/modules/replica/IngestClient.cc
```diff
@@ -0,0 +1,310 @@
...
+
+    // Connect to the server synchronously using error codes to process errors
+    // instead of exceptions.
+    boost::asio::ip::tcp::resolver resolver(_io_service);
+    boost::asio::ip::tcp::resolver::iterator iter =
+        resolver.resolve(
+            boost::asio::ip::tcp::resolver::query(
+                _workerHost,
+                to_string(_workerPort)),
+        ec
```
n8pease: seems like `ec` should be tabbed over one more time?

core/modules/replica/IngestClient.cc
```diff
@@ -0,0 +1,310 @@
...
+                
+void IngestClient::_closeConnection() {
+
+    LOGS(_log, LOG_LVL_DEBUG, _context(__func__));
+
+    // Always attempt to shutdown and close the socket. This code deliberately
+    // ignores any abnormal conditions should they happen during the operation.
+    boost::system::error_code ec;
+    _socket.shutdown(boost::asio::ip::tcp::socket::shutdown_both, ec);
+    _socket.close(ec);
```
n8pease: I guess it's ok to shutdown and close a socket multiple times? (the docs were not much help on this)

core/modules/replica/IngestServerConnection.cc
```diff
@@ -0,0 +1,520 @@
...
+        }
+        _isPartitioned = _databaseInfo.partitionedTables.end() != find(
+            _databaseInfo.partitionedTables.begin(),
+            _databaseInfo.partitionedTables.end(),
+            _table);
+        if (not _isPartitioned) {
+            if (_databaseInfo.regularTables.end() != find(
+                _databaseInfo.regularTables.begin(),
+                _databaseInfo.regularTables.end(),
+                _table)) {
```
n8pease: is a wrapped `if` supposed to be indented 2x? Otherwise it blends in with the body of the `if`, which can be hard to read.

core/modules/replica/IngestServerConnection.cc
n8pease: and then you could unit test the tmp file code in util! 🙃

core/modules/replica/FileIngestApp.cc
```diff
@@ -0,0 +1,291 @@
...
+        if (not regex_search(filename, match, re) or match.size() != 3) {
+            throw invalid_argument(
+                    context + "allowed file names for contributions into partitioned tables:"
+                    " 'chunk_<chunk>.txt', 'chunk_<chunk>_overlap.txt'");
+        }
+        chunk = stoul(match[1].str());
+        isOverlap = not match[2].str().empty();
+
+    } else if (file.tableType == "R") {
+        ;
```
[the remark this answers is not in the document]
n8pease: oh, I'd assumed that there was nothing to do for this case, and was the same as saying 
```
} else if (file.tableType != "R") {
    throw... // as on line 244
}
```
I agree some comment here would be good.

core/modules/replica/IngestServer.cc
```diff
@@ -0,0 +1,128 @@
...
+            new thread(boost::bind(&boost::asio::io_service::run,
+                       &_io_service))
+        );
+        threads[i] = ptr;
+    }
+
+    // Wait for all threads in the pool to exit.
+    for (size_t i = 0; i < threads.size(); ++i) {
+        threads[i]->join();
+    }
```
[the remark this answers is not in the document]
n8pease: I agree that would be more modern.
I wonder if `auto&` would be more correct tho. The `&&` case allows an object assignment, right? IE `thrd` would own the object, not be a reference to it, right? Is there any way in this case where that is reasonable?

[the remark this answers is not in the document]
jgates108: auto&& will always return a reference and it should work even if the vector contains a simple type like int, which I think is an issue for auto&.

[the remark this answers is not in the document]
andy-slac: I think I remembered why `auto&&` was better than `auto&` but it was long time ago. In this case there should not be any difference, and I believe both forms are equivalent for simple types. I think the difference matters when iterator returns some non-trivial type as a "reference" type, e.g. for `vector<bool>` specialization (but no one should be using that class).

core/modules/replica/IngestClient.cc
jgates108: I think it is harmless. It might be useful to log failures that aren't related to the socket already being closed.

core/modules/replica/protocol.proto
```diff
@@ -759,3 +759,55 @@ message ProtocolFileResponse {
...
+    enum ColumnSeparator {
+        COMMA = 0;
+        TAB   = 1;
+    }
+    required ColumnSeparator column_separator = 5;
+}
+
+/// A message which is meant for sending rows to a server after receiving
+/// status READY_TO_READ_DATA. This message is sent once after each READY_TO_READ_DATA.
+/// When the client has no date then it should send the last message with
```
andy-slac: date -> data?

core/modules/replica/protocol.proto
```diff
@@ -759,3 +759,55 @@ message ProtocolFileResponse {
...
+/// zero number of rows. And the server should respond with FINISHED. Then both
```
andy-slac: Hmm, from the code that I saw it looks like "last" flag can be set to true even if number of rows is non-zero?

core/modules/replica/protocol.proto
```diff
@@ -759,3 +759,55 @@ message ProtocolFileResponse {
...
+}
+
+/// The response message sent by a server on each client's request
+message ProtocolIngestResponse {
+
+    enum Status {
+        ILLEGAL_PARAMETERS = 0;     /// incorrect parameters received from a client
+        READY_TO_READ_DATA = 1;     /// server is waiting for data
+        FINISHED           = 2;     /// a confirmation that the file content accepted
+        FAILED             = 3;     /// to report problems
```
andy-slac: Why is not `ILLEGAL_PARAMETERS` the same as `FAILED` with `error="illegal params"`?

core/modules/replica/IngestServer.h
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * This header declares class IngestServer which is used as a worker-side
+ * end point for ingesting catalog data into the Qserv worker's MySQL database.
+ */
+
+// System headers
+#include <memory>
+
+// Third party headers
+#include <boost/asio.hpp>
```
fritzm: angle brackets system headers only

core/modules/replica/IngestServer.h
```diff
@@ -0,0 +1,130 @@
...
+
+    // Input parameters
+
+    ServiceProvider::Ptr const _serviceProvider;
+    std::string          const _workerName;
+
+    /// Cached worker descriptor obtained from the configuration
+    WorkerInfo const _workerInfo;
+
+    boost::asio::io_service        _io_service;
```
fritzm: For your consideration: I like a pattern where the asio::io_service is passed in at service construction time.  This lets a server implementor (implementing a server that may provide more than one type of service) decide on the appropriate threading model, control which services share and asio::io_service instance and which have their own, gives better control over service lifetime vs. server lifetime, etc.

core/modules/replica/IngestServer.cc
```diff
@@ -0,0 +1,128 @@
...
+    // failures.
+    _acceptor.set_option(boost::asio::socket_base::reuse_address(true));
+}
+
+
+void IngestServer::run() {
+
+    // We shall do so before running the io_service. Otherwise it will
+    // immediately finish as soon as it will discover that there are
+    // outstanding operations.
```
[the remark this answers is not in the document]
fritzm: How about just:
```// Queue some work for the io_service, so it doesn't immediately tail out when started.```

File: core/modules/replica/IngestServer.cc [hunk 16]
```diff
@@ -0,0 +1,128 @@
...
+
+    IngestServerConnection::Ptr connection =
+        IngestServerConnection::create(
+            _serviceProvider,
+            _workerName,
+            _io_service);
+
+    _acceptor.async_accept(
+        connection->socket(),
+        boost::bind(
```
fritzm: PLEASE favor C++ lambda over boost::bind!

core/modules/replica/IngestServerConnection.h
```diff
@@ -0,0 +1,292 @@
...
+#ifndef LSST_QSERV_REPLICA_INGESTSERVERCONNECTION_H
+#define LSST_QSERV_REPLICA_INGESTSERVERCONNECTION_H
+
+// System headers
+#include <cstdio>
+#include <fstream>
+#include <memory>
+
+// Third party headers
+#include <boost/asio.hpp>
```
fritzm: angle brackets for system headers only...

core/modules/replica/IngestServerConnection.cc
```diff
@@ -0,0 +1,520 @@
...
+
+// System headers
+#include <algorithm>
+#include <cerrno>
+#include <cstring>
+#include <ctime>
+#include <stdexcept>
+
+// Third party headers
+#include <boost/bind.hpp>
```
fritzm: you know what I'm gonna say...

core/modules/replica/IngestServerConnection.cc
```diff
@@ -0,0 +1,520 @@
...
+
+    const size_t bytes = sizeof(uint32_t);
+
+    _bufferPtr->resize(bytes);
+
+    boost::asio::async_read(
+        _socket,
+        boost::asio::buffer(_bufferPtr->data(), bytes),
+        boost::asio::transfer_at_least(bytes),
+        boost::bind(
```
fritzm: lambda, not boost::bind PLEASE!

core/modules/replica/IngestServerConnection.cc
```diff
@@ -0,0 +1,520 @@
...
+
+
+void IngestServerConnection::_sendResponse() {
+
+    LOGS(_log, LOG_LVL_DEBUG, context << __func__);
+
+    boost::asio::async_write(
+        _socket,
+        boost::asio::buffer(_bufferPtr->data(), _bufferPtr->size()),
+        boost::bind(
```
fritzm: lambda please

core/modules/replica/IngestServerConnection.cc
```diff
@@ -0,0 +1,520 @@
...
+
+    const size_t bytes = sizeof(uint32_t);
+
+    _bufferPtr->resize(bytes);
+
+    boost::asio::async_read(
+        _socket,
+        boost::asio::buffer(_bufferPtr->data(), bytes),
+        boost::asio::transfer_at_least(bytes),
+        boost::bind(
```
fritzm: lambda please

core/modules/replica/IngestClient.h
```diff
@@ -0,0 +1,228 @@
...
+namespace qserv {
+namespace replica {
+
+/**
+ * Class IngestClientError represents exceptions thrown by IngestClient
+ * on errors.
+ */
+class IngestClientError : public std::runtime_error {
+public:
+    IngestClientError(std::string const& msg)
```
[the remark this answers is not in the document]
iagaponenko: Unfortunately we have to wait till C++17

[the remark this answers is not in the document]
andy-slac: Inheriting constructors are in C++11.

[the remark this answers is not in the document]
iagaponenko: Apparently, you're right! I will refactor the code. Thank you!

core/modules/replica/IngestClient.cc
```diff
@@ -0,0 +1,310 @@
...
+
+        // Read up to the maximum number of lines requested by the server
+        // into a data message
+        ProtocolIngestData data;
+        string row;
+        size_t numRows = 0;
+        while (numRows < _numRowsPerSend) {
+            if (getline(file, row)) {
+                data.add_rows(row);
+                ++numRows;
```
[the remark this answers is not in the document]
iagaponenko: Good question! I have explored Protobuf implementation to see how expensive is `rows_size()`. It turns out it's a chain of 3 `inline` calls to fetch a data member of some template container class. So, I'm going to follow your suggestion and replace my counter with  calls to that method.

core/modules/replica/IngestServer.h
iagaponenko: Noted. I will consider this.

core/modules/replica/protocol.proto
iagaponenko: Makes a sense

core/modules/replica/protocol.proto
iagaponenko: Fixed

core/modules/replica/IngestServerConnection.cc
```diff
@@ -0,0 +1,520 @@
...
+
+            vector<string> const tablesToBeCreated = {
+                sqlTable,
+                sqlFullOverlapTable,
+                sqlDummyChunkTable,
+                sqlOverlapDummyChunkTable
+            };
+            for (auto&& table: tablesToBeCreated) {
+                statements.push_back(
+                    "CREATE TABLE IF NOT EXISTS " + table + " LIKE " + sqlProtoTable
```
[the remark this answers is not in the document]
iagaponenko: Yes, it does.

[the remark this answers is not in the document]
andy-slac: My first reaction is that it should not.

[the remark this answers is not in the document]
iagaponenko: Why?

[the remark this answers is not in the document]
iagaponenko: All other (chunked) tables at workers have this column. The column could (after the ingest is done) be eliminated if needed. Though, it's better to keep them for the bookkeeping and Q&A purposes to investigate problems with ingested data later.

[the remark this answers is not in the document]
iagaponenko: This column is no different from `chunkId` and `subChunkId` which are present in both prototype and chunked tables.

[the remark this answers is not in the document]
andy-slac: If someone else want to use that schema for any other purpose they'd have to know that there are columns in that schema which are specific to your implementation of partitioning mechanism (and that mechanism could change too). It would probably be better to have an original schema without any implementation-specific details and extend schema when needed.

core/modules/replica/IngestServerConnection.cc
```diff
@@ -0,0 +1,520 @@
...
+    _table           = request.table();
+    _chunk           = request.chunk();
+    _isOverlap       = request.is_overlap();
+    _columnSeparator = request.column_separator() == ProtocolIngestHandshakeRequest::COMMA ? ',' : '\t';
+
+    // Check if a context of the request is valid
+    
+    try {
+        auto transactionInfo = _serviceProvider->databaseServices()->transaction(_transactionId);
+        if (transactionInfo.state != "STARTED") {
```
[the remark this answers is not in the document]
iagaponenko: The transaction state is presently defined as a string in the database schema. I may migrate it to `enum` later on.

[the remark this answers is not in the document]
andy-slac: I don't care much how it is defined in database, I'm thinking about C++ side. Even if it a string you could define constants for those strings to avoid typos.

PR #502, opened 2019-11-19
DM-22234: Remove python 2 support

timj:


site_scons/convertVocab.py
```diff
@@ -29,10 +28,8 @@
 import sys
 
 from string import digits
-try:
-    from string import ascii_letters
-except ImportError:
-    from string import letters as ascii_letters
+from string import ascii_letters
```
andy-slac: Should probably combine two imports into one line.

timj:
@andy-slac given that it passed travis, does that mean I can merge? Or is there some other process that needs to be followed for Qserv?

andy-slac:
OK to merge if Travis builds successfully. It does not probably exercise most of the python code that changed but if it builds it means that nothing important is broken. The rest is our worries.

PR #507, opened 2019-12-06
DM-22506: quote identifiers when rendering SQL

n8pease:
never quote the internal representation strings
always quote identifiers when rendering SQL to send to mysql

### DIFF
--- a/core/modules/query/QueryTemplate.h
+++ b/core/modules/query/QueryTemplate.h
@@ -127,6 +127,12 @@ public:
 
     QueryTemplate(SetAliasMode aliasMode) : _aliasMode(aliasMode) {}
 
+    // set if the output should quote identifiers or not
+    void quoteIdentifiers(bool quoteIdentifiers) { _quoteIdentifiers = quoteIdentifiers; };
+
+    // get if the output should quote identifiers or not
+    bool quoteIdentifiers() const { return _quoteIdentifiers; }
+
     void append(std::string const& s);
     void append(ColumnRef const& cr);
     void append(Entry::Ptr const& e);
@@ -201,6 +207,7 @@ public:
 private:
     EntryPtrVector _entries;
     SetAliasMode _aliasMode{USE_ALIAS};
+    bool _quoteIdentifiers{true}; // if true, identifiers will be quoted.
     bool _useColumnOnly{false}; // if true, ColumnRef won't print db or table, only column name.
 };
 
--- a/core/modules/query/QueryTemplate.h
+++ b/core/modules/query/QueryTemplate.h
@@ -127,8 +127,19 @@ public:
 
     QueryTemplate(SetAliasMode aliasMode) : _aliasMode(aliasMode) {}
 
+    /**
+     * @brief Get a version of the identifier that has any needed quoting applied.
+     *
+     * Curently this means putting backtick quotes around the identifier if quoteIdentifiers is set to true,
+     * but may include other formatting as needed in the future.
+     *
+     * @param identifier a reference to a string to be formatted.
+     * @return std::string& a reference to the passed-in string so the function may be used inline.
+     */
+    std::string formatIdentifier(std::string const& identifier) const;
+
     // set if the output should quote identifiers or not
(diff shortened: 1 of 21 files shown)


core/modules/query/TableRef.cc
```diff
@@ -205,9 +205,14 @@ std::string TableRef::sqlFragment() const {
 
 void TableRef::putTemplate(QueryTemplate& qt) const {
     auto aliasMode = qt.getTableAliasMode();
+    bool quoteIdentifiers = qt.quoteIdentifiers();
     if (QueryTemplate::USE == aliasMode) {
         if (hasAlias()) {
-            qt.append("`" + _alias + "`");
+            if (quoteIdentifiers) {
+                qt.append("`" + _alias + "`");
```
iagaponenko: Since this construct is seen in multiple places:
```
"`" + _alias + "`"
```
Then you may replace it as an inline function:
```
string sqlQuotedId(string const& id) const {
    return "`" + id + "`";
}
```
And use it where needed. This approach has the following benefits:
* it clearly expresses the semantic of the operation (unlike just concatenating strings)
* it prevents possible mistakes in the code
* it allows replacing MySQL-specific back ticks with the SQL standard double quote symbol `"`

core/modules/query/TableRef.cc
```diff
@@ -205,9 +205,14 @@ std::string TableRef::sqlFragment() const {
 
 void TableRef::putTemplate(QueryTemplate& qt) const {
     auto aliasMode = qt.getTableAliasMode();
+    bool quoteIdentifiers = qt.quoteIdentifiers();
```
iagaponenko: For read-only variables which are cached it's usually safer to make them `const` like in:
```
bool const quoteIdentifiers = ...
```
This helps code readers and compilers (optimization).

core/modules/query/QueryTemplate.h
```diff
@@ -127,6 +127,12 @@ class QueryTemplate {
 
     QueryTemplate(SetAliasMode aliasMode) : _aliasMode(aliasMode) {}
 
+    // set if the output should quote identifiers or not
+    void quoteIdentifiers(bool quoteIdentifiers) { _quoteIdentifiers = quoteIdentifiers; };
```
iagaponenko: This way of overloading methods looks a bit confusing. I would either name this particular overload as (my favorite option):
```
void setQuteIdentifiers(bool quoteIdentifiers) { ... }
```
or let it to return a value:
```
bool quoteIdentifiers(bool quoteIdentifiers) { ... }
```

core/modules/query/TableRef.cc
```diff
@@ -225,7 +230,11 @@ void TableRef::putTemplate(QueryTemplate& qt) const {
     if (QueryTemplate::DEFINE == aliasMode) {
         if (hasAlias()) {
             qt.append("AS");
-            qt.append("`" + _alias + "`");
+            if (quoteIdentifiers) {
+                qt.append("`" + _alias + "`");
+            } else {
+                qt.append(_alias);
+            }
```
iagaponenko: If you accepted my suggestion to encapsulate the quoting into an inline method `sqlQuotedId` then these 5 lines could be replaced with just 1:
```
qt.apend(quoteIdentifiers ? sqlQuotedId(_alias) : _alias);
```
I'm not insisting on this. Just a suggestion.

core/modules/query/QueryTemplate.h
```diff
@@ -201,6 +207,7 @@ class QueryTemplate {
 private:
     EntryPtrVector _entries;
     SetAliasMode _aliasMode{USE_ALIAS};
+    bool _quoteIdentifiers{true}; // if true, identifiers will be quoted.
```
iagaponenko: This isn't a request, just a suggestion to use the following form of the in-header initialization:
```
bool _quoteIdentifiers = true;    ///< if true, the identifiers will be quoted.
```

core/modules/query/TableRef.cc
n8pease: thanks for noticing this repeated pattern. You made me realize the whole quoting thing could go in QueryTemplate, and I can add a function called `QueryTemplate::appendIdentifier(string)` that does the Right Thing. I'll upload changes soon.

PR #508, opened 2019-12-13
DM-17656: Experimental distributed secondary index

jgates108:


core/modules/loader/CentralWorker.h
```diff
@@ -0,0 +1,276 @@
...
+
+    /// Open the UDP and TCP ports and start monitoring. This can throw boost::system::system_error.
+    void startService() override;
+    void startMonitoring() override;
+
+    ~CentralWorker() override;
+
+    int getTcpPort() const override { return _tcpPort; }
+
+    uint32_t getOurId() const {
```
iagaponenko: Why this method is defined as `inline`?

core/modules/loader/CentralWorker.cc
```diff
@@ -0,0 +1,1146 @@
+// -*- LSST-C++ -*-
```
iagaponenko: This file is huge. I happen to review it many months ago. And it's not clear from tis PR what were the  changes to it in a context of this development.

core/modules/loader/CentralWorker.cc
```diff
@@ -0,0 +1,1146 @@
...
+ * the GNU General Public License along with this program.  If not,
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+
+
+// Class header
+#include "loader/CentralWorker.h"
+
+// system headers
+#include <boost/asio.hpp>
```
iagaponenko: The third party header. And it must be included in the double quotes.

core/modules/loader/ClientServer.h
```diff
@@ -0,0 +1,81 @@
...
+ */
+#ifndef LSST_QSERV_LOADER_CLIENTSERVER_H
+#define LSST_QSERV_LOADER_CLIENTSERVER_H
+
+// system headers
+#include <cstdlib>
+#include <iostream>
+
+// third party headers
+#include <boost/asio.hpp>
```
iagaponenko: The non-system headers should not be included with the angle brackets.

core/modules/loader/WorkerServer.h
```diff
@@ -0,0 +1,70 @@
...
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ *
+ */
+#ifndef LSST_QSERV_LOADER_WORKERSERVER_H
+#define LSST_QSERV_LOADER_WORKERSERVER_H
+
+// system headers
+#include <cstdlib>
+#include <iostream>
+#include <boost/bind.hpp>
```
iagaponenko: These two are the third party headers, and they also need to be included within the doube quotes.

iagaponenko (approved):
I have made the first sweep through classes "appClientNum, appClientMaster, appWorker, BufferUdp, CentralFollower, CentralClient, CentralWorker, ClientServer, WorkerServer, WWorkerList" and files in folder "admin/tools/docker/index". I have left comments where I could understand the code. My main problem was that this is not the usual "diff" ticket. It appears as a completely new development. So, it makes it difficult to judge how the claimed goal of the ticket has been achieved with this code.
My second big complain with the code is the naming convention behind many classes and some methods. The naming convention is way to abstract/generic. It makes it difficult to understand roles and behaviors of classes and methods.
Ideally I would like to have an in-person conversation to better understand the overall design of this application.
I approve it, conditionally, provided that my comments and suggestions will be considered.

admin/tools/docker/index/container/dev/clientNum/Dockerfile
```diff
@@ -0,0 +1,13 @@
+#
```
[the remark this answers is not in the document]
jgates108: It made the yaml  scripts easier to write and get working (which was not easy) and an actual instance of the system would not have the client apps, as they are only for testing. It would make sense to eventually trim down the containers so they only contain what they really need.
Currently, the containers all share the same qserv/indexbase:dev. The differences for the individual containers really comes down the entrypoint designations and bash scripts, which is small. There could be significant differences between master and worker containers in the future.

admin/tools/docker/index/container/dev/clientNum/appClientNum.bash
```diff
@@ -0,0 +1,23 @@
...
+setup -r . -t qserv-dev
+
+export LSST_LOG_CONFIG=/home/qserv/dev/qserv/admin/templates/configuration/etc/log4cxx.index.properties
+
+echo appClientNum $1 $2 $3
+
+/home/qserv/dev/qserv/build/loader/appClientNum $1 $2 /home/qserv/dev/qserv/core/modules/loader/config/$3 
+
+child=$!
+wait "$child"
```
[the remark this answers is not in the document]
jgates108: I believe Kubernetes or microk8s wasn't able to kill these containers until all this was added. They'd get stuck terminating the pods.

core/modules/loader/BufferUdp.cc
```diff
@@ -0,0 +1,202 @@
...
+    dump(os, hexDump, charDump);
+    return os.str();
+}
+
+
+std::ostream& BufferUdp::dump(std::ostream &os, bool hexDump, bool charDump) const {
+    os << "maxLength=" << _length;
+    os << " buffer=" << (void*)_buffer;
+    os << " wCurLen=" << getAvailableWriteLength();
+    os << " wCursor=" << (void*)_wCursor;
```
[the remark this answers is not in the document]
jgates108: It makes it easier to tell if a raw pointer is within the buffer.

core/modules/loader/BufferUdp.h
```diff
@@ -0,0 +1,174 @@
...
+    ~BufferUdp() { if (_ourBuffer) { delete[] _buffer; } }
+
+    /// Resets the cursors in the buffer so it is effectively empty.
+    void reset() { _setupBuffer(); }
+
+    /// Return true only if this object owns the buffer.
+    bool releaseOwnership();
+
+    /// Make this object is responsible for deleting _buffer.
+    void makeOwnerOfBuffer() { _ourBuffer = true; }
```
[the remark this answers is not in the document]
jgates108: I agree, but aspects of asio communication make it very difficult to avoid them. I've tried to wrap the buffers up as much as possible. For the sake of efficiency (as these things can be very large), I've left in the ability to just hand the class a loaded buffer to avoid copying.

fritzm:
Closing this PR, which seems unlikely to merge.  We will retain the branch, in case we wish to draw from this work revisting distributed director indices in the future.

PR #509, opened 2019-12-17
Tickets/dm 22428

iagaponenko:


core/modules/replica/WorkerIndexRequest.cc
```diff
@@ -52,30 +52,44 @@ namespace lsst {
...
+        ProtocolRequestIndex const& request) {
+    return WorkerIndexRequest::Ptr(new WorkerIndexRequest(serviceProvider,
+        connectionPool,
+        worker,
+        id,
+        priority,
+        onExpired,
+        requestExpirationIvalSec,
+        request
+    ));
```
jgates108: There are a lot of these (half dozen arguments passed from create to constructor to constructor again). Would it help to have a data classes for WorkerRequest and all of its children? Then you could define WorkerRequestData and WorkerIndexRequestData and use them as arguments to the create functions and constructor functions. Reduced numbers of arguments passed and reduced duplicate comments.
``` 
struct WorkerRequestData {
    //  copy constructor
    std::string const worker;
    std::string const type;
    std::string const id;
    int         const priority;
    ExpirationCallbackType onExpired;
};

class WorkerRequest : public std::enable_shared_from_this<WorkerRequest> {
public:
    WorkerReques(WorkerRequestData const& workerRequestData) : _data(workerRequestData) { ... }
 ...
private:
   WorkerRequestData _data;
...
};
```

core/modules/replica/DisposeRequest.cc
```diff
@@ -0,0 +1,184 @@
...
+
+void DisposeRequest::_analyze(bool success,
+                              ProtocolResponseDispose const& message) {
+    LOGS(_log, LOG_LVL_DEBUG, context() << __func__ << "  success=" << (success ? "true" : "false"));
+
+    if (state() == State::FINISHED) return;
+
+    util::Lock lock(_mtx, context() + __func__);
+
+    if (state() == State::FINISHED) return;
```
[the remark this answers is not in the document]
iagaponenko: That's a very interesting idea! I will investigate it.

core/modules/replica/DisposeRequest.h
```diff
@@ -0,0 +1,182 @@
...
+    std::vector<Status> ids;
+};
+
+
+/**
+ * Class DisposeRequest represents Controller-side requests for "garbage
+ * collecting requests at workers.
+ *
+ * @note Requests of this type don't have any persistent states. 
+ */
```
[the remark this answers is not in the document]
iagaponenko: For the record (as we have discussed this over phone today), only a few types of requests (those which take time and resources to execute) are put into the worker queues. For those there are worker-side request classes (deriving from `WorkerRequest` as a remote counterpart for the Controller-side `Request` class). There is a set of other request types which are executed instantaneously by the worker services.Hence there is nothing to be disposed for those requests.

core/modules/replica/WorkerIndexRequest.cc
iagaponenko: That's an interesting suggestion! I will investigate it

n8pease: My experience with such "data" classes is not good; they can introduce ownership confusion and when refactoring you may have to modify things in unexpected locations. I think the problem John sites is a symptom of too-much-boilerplate, but IMHO the proposed fix is an anti pattern.

core/modules/replica/WorkerRequestFactory.h
```diff
@@ -83,124 +80,78 @@ class WorkerRequestFactoryBase {
...
-    *
-    * @see class WorkerDeleteRequest
-    *
-    * @return
-    *   a pointer to the newly created object
-    */
+            WorkerRequest::ExpirationCallbackType const& onExpired,
+            unsigned int requestExpirationIvalSec,
+            ProtocolRequestReplicate const& request) const = 0;
+
```
n8pease: WRT boilerplate, I think you could write a single function templated on `std::shared_ptr<WorkerReplicationRequest>` and `ProtocolRequestReplicate` and get rid of all these `createReplication` variants.

core/modules/replica/WorkerProcessor.cc
```diff
@@ -299,9 +298,10 @@ void WorkerProcessor::enqueueForDeletion(string const& id,
         auto const ptr = _requestFactory.createDeleteRequest(
             _worker,
             id,
-            request.priority(),
-            request.database(),
-            request.chunk()
+            priority,
+            bind(&WorkerProcessor::dispose, shared_from_this(), _1),
```
n8pease: would a lambda, even one that just calls `dispose` be more clear here? (I bet you could pass shared_from_this in as an argument?

iagaponenko: A reason why I chose to avoid a lambda function here would be because the code would look like:
```
auto const self = shared_from_this();
auto const ptr = _requestFactory.createDeleteRequest(
             _worker,
             id,
            request.priority(),
            request.database(),
            request.chunk()
            priority,
            [self] (string const& id) {
                self->dispose(id);
            },
```
I think `bind` is a more concise way of doing it.

core/modules/replica/WorkerRequestFactory.h
iagaponenko: Unfortunately, this won't work. I will explain you why in person.

core/modules/replica/WorkerProcessor.cc
fritzm: I find the latter one more readable.  We already have to understand how to read lambdas in this modern world; I would like to be able to dispense with the cognitive load of bind and reclaim those brain cells...

I would argue that putting the bind on one line in such a simple case would also be acceptable.

core/modules/replica/DisposeRequest.cc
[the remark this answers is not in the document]
iagaponenko: Actually, there is probably a simple way - just remove the first (before the lock) check for `state() == FINISHED`. I have already registered this as one of the code refactoring efforts in my TODO list: https://confluence.lsstcorp.org/display/DM/More+robust+implementation+of+the+asynchronous+handlers
I will do it as a dedicated package wide effort rather then fixing one handlers in a context of this PR.

core/modules/replica/WorkerIndexRequest.cc
iagaponenko: Okay, since I have to opposite opinions on this subject then I choose do nothing 😸

core/modules/replica/WorkerProcessor.cc
```diff
@@ -805,6 +796,63 @@ void WorkerProcessor::_setServiceResponseInfo(
...
+                ptr->dispose();
+                found = true;
+                return true;
+            }
+            return false;
+        });
+        if (not found) {
+            // Has it already finished?
+            queue = "finished";
+            _finishedRequests.remove_if([&found,id](auto&& ptr) {
```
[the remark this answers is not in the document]
iagaponenko: Good point! I will refactor the code.

PR #510, opened 2019-12-19
Avoid fatal error when encoutering a zero length file.

abh3:


iagaponenko:
I have reviewed proposed changes, and I approve them.

PR #511, opened 2019-12-20
Tickets/dm 22656

iagaponenko:


core/modules/replica/ConfigurationMySQL.h
```diff
@@ -98,174 +89,240 @@ class ConfigurationMySQL : public ConfigurationBase {
...
 
-    void setRequestBufferSizeBytes(size_t val) final {
+    void setRequestBufferSizeBytes(size_t val,
+                                   bool updatePersistentState) final {
         _set(_requestBufferSizeBytes,
              "common",
              "request_buf_size_bytes",
-             val);
+             val,
+             false,
```
n8pease: a comment expressing the use/intent of the bool var might be useful. and so on below.

iagaponenko: Good point! I will fix it.

PR #515, opened 2020-01-24
Tickets/dm 23087

iagaponenko:


core/modules/replica/HttpIngestModule.cc
```diff
@@ -86,6 +86,43 @@ string leastLoadedWorker(DatabaseServices::Ptr const& databaseServices,
...
+string leastLoadedWorker(map<string,size_t>& worker2replicasCache,
+                         DatabaseServices::Ptr const& databaseServices,
+                         COLLECTION_OF_WORKERS const& workers) {
+    string worker;
+    string const noSpecificDatabase;
+    bool   const allDatabases = true;
+    size_t numReplicas = numeric_limits<size_t>::max();
+    for (auto&& candidateWorker: workers) {
+        if (worker2replicasCache.count(candidateWorker) == 0) {
+            worker2replicasCache[candidateWorker] =
```
n8pease: IIRC Scott Meyers discourages use of the [] operator with maps. In this case I think you'd be better off using something like `emplace` and keep and use the returned iterator instead of performing so many map lookups.

n8pease: you could also skip the call to `count`; `emplace` (or `insert`) will _not_ create a new key if one already exists and in either case will return an iterator to the item at the key, which you can then carry on using.
...It looks like you may want to avoid calling `databaseServices->numWorkerReplicas` for keys that already exist in the map tho, in which case I'd call `find` and then if the returned `itr==end()` then perform the insert, and use the returned itr for the rest of the calls...

iagaponenko: Exactly! The goal of this optimization was to avoid making expensive (and unnecessary) database queries on each invocation of the function. As it was discovered in a large scale setup of the `lsp-int` at NCSA, it takes about `0.1 sec` to fetch the worker-specific info from MySQL at a presence of ~1 million replicas (across all catalogs). This resulted in the poor performance of the chunk allocation mechanism during the ingests:
```
0.1 sec * 30 workers * 150,000 chunks = 3 sec/chunk * 150,000 chunks = 450,000 seconds = many hours
```
The new implementation does it all within less than 9 minutes.

As far as your suggestion to move to the "iterator-based" implementation then I have a conflicting feeling about it. I must admit, I had the very same thought (as you) before I implemented the function. Then I ruled iterators out because the map is rather tiny - it has as many elements as many workers are in a Qserv setup, which is on a scale of 100. On that scale any lookup would take a few microseconds at most, which would be 10,000 - 100,000 faster compared with the database query. So, the performance of the lookup operations is not an issue here. It's obvious that iterators won't make any measurable speedup here. What really discouraged me initially from using iterators with maps (unless I have a good case for this) is an added complexity to the code. The iterator-based implementations becomes harder to read and comprehend.

Having said that, I will still look if I will find a good way of expression the algorithm's intent with iterators.

core/modules/replica/HttpIngestModule.cc
```diff
@@ -86,6 +86,43 @@ string leastLoadedWorker(DatabaseServices::Ptr const& databaseServices,
...
+template<typename COLLECTION_OF_WORKERS>
+string leastLoadedWorker(map<string,size_t>& worker2replicasCache,
+                         DatabaseServices::Ptr const& databaseServices,
+                         COLLECTION_OF_WORKERS const& workers) {
+    string worker;
+    string const noSpecificDatabase;
+    bool   const allDatabases = true;
+    size_t numReplicas = numeric_limits<size_t>::max();
+    for (auto&& candidateWorker: workers) {
+        if (worker2replicasCache.count(candidateWorker) == 0) {
```
n8pease: `count` is a funny choice for a map, I would probably have used `find` and compared to the map `::end`

iagaponenko: Yes, indeed 😸

core/modules/replica/HttpIngestModule.cc
n8pease: I don't know how frequently this is called or for how many `workers`, but `::insert` in this case would be slightly more efficient.

core/modules/replica/HttpRequestBody.h
```diff
@@ -98,7 +98,33 @@ class HttpRequestBody {
...
+                    " is missing in the request body");
+        }
+        if (not itr->is_array()) {
+            throw std::invalid_argument(
+                    "HttpRequestBody::" + std::string(__func__) + "<T> a value of the required parameter " + name +
+                    " is not an array");
+        }
+        std::vector<T> coll;
+        for (size_t i = 0, size = itr->size(); i < size; ++i) {
+            T val = (*itr)[i];
```
n8pease: is `val` a copy here? Is that ever expensive? Maybe you want a reference? Maybe you want to emplace_back in one line instead? 🤷‍♂

iagaponenko: Unfortunately, it needs to be expressed exactly like that. This is a conversion operation from the fancy `nlohmann::json` object.  Otherwise it will get an object of class `nlohmann::json` instead of proposed value of the template type `T`. Further details can be found in the documentation: https://github.com/nlohmann/json

core/modules/replica/HttpRequestQuery.cc
```diff
@@ -145,4 +145,15 @@ bool HttpRequestQuery::optionalBool(string const& param,
...
 }
 
+
+double HttpRequestQuery::requiredDouble(string const& param) const {
+    auto&& itr = _query.find(param);
+    if (itr == _query.end()) {
+        throw invalid_argument(
+                string(__func__) + " parameter '" + param + "' is missing");
+    }
+    return stod(itr->second);
```
n8pease: you're sure this is never going to throw? Or is it handled upstream?

iagaponenko: This function is expected to throw `std::invalid_argument` for miss-formed (or out-of-range) inputs. Exceptions of this kind are caught and properly reported (via the LSST Logger) and sent back to REST API clients as `json` object by the module execution driver:
```
// File: core/modules/replica/HttpModule.cc
void HttpModule::execute(qhttp::Request::Ptr const& req,
                         qhttp::Response::Ptr const& resp,
                         string const& subModuleName) {
    try {
        executeImpl(req, resp, subModuleName);
    } catch (invalid_argument const& ex) {
        sendError(resp, __func__, "invalid parameters of the request, ex: " + string(ex.what()));
    } catch (exception const& ex) {
        sendError(resp, __func__, "operation failed due to: " + string(ex.what()));
    }
}
```

iagaponenko: This is the mini-REST-API framework which I implemented to reduce boilerplate code when processing the REST requests by the Master Replication Controller.

core/modules/replica/HttpIngestModule.cc
iagaponenko: You have convinced me 👍 I have reimplemented this function in terms of the iterators. Though, after testing the new new code I see no difference in its performance as compared with the `[]`-based implementation. As I already mentioned in some previous comment, taking away expensive MySQL queries against a table of millions of replicas (`0.1s` per query) was the main goal of this ticket. This optimization alone has allowed to speed up the chunk placement code by a factor of x100. Replacing `[]` with iterators made no measurable improvement to the performance. Though, it probably made a sense from the code stye point of view in order to make everything ore "C++-ish".

PR #517, opened 2020-02-06
Changed to notify_one instead of notify_all for queCmd.

jgates108:


core/modules/mysql/MySqlConfig.h
```diff
@@ -60,16 +60,14 @@ class MySqlConfig {
...
      *
      * @throws std::runtime_error: if checkValid is true and some parameters are invalid
      */
     MySqlConfig(std::string const& username, std::string const& password,
                 std::string const& hostname,
                 unsigned int const port,
                 std::string const& socket,
-                std::string const& db = "",
-                size_t maxresultsize = 0);
+                std::string const& db = "");
```
iagaponenko, line 70: I think the C+ Style guide allows no spaces around the assignment operator here.

core/modules/qdisp/QdispPool.cc
```diff
@@ -100,21 +102,22 @@ void PriorityQueue::queCmd(PriorityCommand::Ptr const& cmd, int priority) {
...
     while (true) {
         _changed = false;
-        LOGS(_log, LOG_LVL_DEBUG, "priQueGet " << *this);
-        { /// &&& delete block
-            ++localLogLimiter;
-            if (localLogLimiter%25 == 0) {
-                LOGS(_log, LOG_LVL_WARN, "priQueGet " << *this); // &&&
-            }
+        ++localLogLimiter;
+        if (localLogLimiter%50 == 0) {
```
iagaponenko, line 114: What is this "magic" number 50?
Also, it would be easier to read this:
```
        if (localLogLimiter % 50 == 0) {
```
It would be also nice if this logic for separating what's reported at which log level is so important here.

core/modules/qdisp/QdispPool.cc
```diff
@@ -201,17 +204,46 @@ std::string PriorityQueue::statsStr() {
...
+       << " maxPriority=" << largestPriority << " maxRunSizes=" << util::prettyCharList(maxRunSizes)
+       << " minRunningSizes=" << util::prettyCharList(minRunningSizes);
+    if (poolSize < 1 || largestPriority < 0 || maxRunSizes.size() < static_cast<size_t>(largestPriority) + 1) {
+        LOGS(_log, LOG_LVL_ERROR, "QdispPool invalid paramater " << os.str());
+        throw std::invalid_argument(os.str());
+    }
+
+    LOGS(_log, LOG_LVL_INFO, "QdispPool creating " << os.str());
+    _prQueue = std::make_shared<PriorityQueue>(100, 1, 1); // default (lowest) priority.
+    for (unsigned int pri = 0; pri <= static_cast<unsigned int>(largestPriority); ++pri) {
```
iagaponenko, line 220: Is there a protection that this loop won't go "wild" and create thousands or millions of priority queues? Someone may just configure Qserv to set the highest priority to 10000000 (which is quit elegal) in the config file and bad things will happen here!.

core/modules/qdisp/QdispPool.cc
```diff
@@ -201,17 +204,46 @@ std::string PriorityQueue::statsStr() {
...
+        size_t minRun = 0;
+        if (minRunningSizes.size() > pri) {
+            minRun = minRunningSizes[pri];
+        }
+        size_t maxRun = 1;
+        if (maxRunSizes.size() > pri) {
+            maxRun = maxRunSizes[pri];
+        }
```
iagaponenko, line 228: The last 8 lines could be turned into 2 lines:
```
size_t const minRun = minRunningSizes.size() > pri ? minRunningSizes[pri] : 1;
size_t const maxRun = maxRunSizes.size() > pri ? maxRunSizes[pri] : 1;
```

core/modules/qdisp/QdispPool.cc
```diff
@@ -201,17 +204,46 @@ std::string PriorityQueue::statsStr() {
...
         // tend to be slow to give up their threads, thus can't be allowed
         // to eat up the pool. Bandwidth also makes running many of the
         // slow queries at the same time a burden on the system.
         // TODO: Set up thread pool size and queues in configuration. DM-10237
-        _prQueue = std::make_shared<PriorityQueue>(100, 1, 5); // default (lowest) priority.
+        _prQueue = std::make_shared<PriorityQueue>(100, 1, 1); // default (lowest) priority.
         _pool = util::ThreadPool::newThreadPool(2400, _prQueue);
         _prQueue->addPriQueue(0, 1, 90);  // Highest priority - interactive queries
-        _prQueue->addPriQueue(1, 1, 1000);  // Outgoing shared scan queries.
+        _prQueue->addPriQueue(1, 1, 1000); // Outgoing shared scan queries.
```
iagaponenko, line 246: This is the second time I'm complaining about using the hard-wired numbers in initializing parameters of `QdispPool`. I did so when I was reviewing another ticket a few months ago. It would be =the right thing to move these values into `czar`'s configuration file.

core/modules/qdisp/QdispPool.h
```diff
@@ -135,6 +135,25 @@ class QdispPool {
 public:
     typedef std::shared_ptr<QdispPool> Ptr;
 
+    /// poolSize - total number of threads in the pool
+    /// largestPriority - highest priority is 0, lowest possible priority is
+    ///            100 and is reserved for default priority. largestPriority=4 would
```
iagaponenko, line 140: Where this "magic" limit of 100 is enforced? And why it's not loaded from a configuration file?

core/modules/qdisp/QdispPool.h
```diff
@@ -135,6 +135,25 @@ class QdispPool {
...
+    ///             ex. 5, 10, 10, 3, 3 would apply to the priorities above as
+    ///              priority 0 can have up to 5 concurrent running commands
+    ///              priorities 1 and 2 can have up to 10
+    ///              priorities 3 and 4 can have up to 3
+    /// minRunningSizes - Each entry represents the minimum number of threads
+    ///             to be running (defaults to 0). Non-zero values can keep
+    ///             lower priorities from being completely stared and/or
+    ///             reduce deadlocks from high priorities depending on lower
+    ///             priorities.
+    QdispPool(int poolSize, int largestPriority, std::vector<int> const& maxRunSizes,
```
iagaponenko, line 155: Why all of these parameters are not `unsigned`? My understanding none of them can be less than 0.

core/modules/rproc/InfileMerger.cc
```diff
@@ -102,82 +102,48 @@ const char JOB_ID_BASE_NAME[] = "jobId";
...
+        LOGS(_log, LOG_LVL_INFO, "Engine is MYISAM, serial");
+        if (!_setupConnectionMyIsam()) {
+            throw InfileMergerError(util::ErrorCode::MYSQLCONNECT, "InfileMerger mysql connect failure.");
+        }
+    } else {
+        if (_dbEngine == INNODB)  {
+            LOGS(_log, LOG_LVL_INFO, "Engine is INNODB, parallel, semaMgrConn=" << _semaMgrConn);
+            // TODO: Size calculation does not work correctly for INNODB
+            //       increase max allowed size to reduce false positives
+            //       until a reasonably accurate method is implemented.
```
iagaponenko, line 133: Yes, that would be the right ting to do - count bytes as they're arriving from workers rather than measuring the size of the result table.

core/modules/rproc/InfileMerger.cc
```diff
@@ -193,18 +159,43 @@ InfileMerger::InfileMerger(InfileMergerConfig const& c,
...
+std::string InfileMerger::engineToStr(InfileMerger::DbEngine engine) {
+    switch (engine) {
+        case MYISAM:
+            return "MYISAM";
+        case INNODB:
+            return "INNODB";
+        case MEMORY:
+            return "MEMORY";
+        default:
+            return "UNKNOWN";
```
iagaponenko, line 189: I would recommend throwing an exception here. Besides, the code would have to be rewritten like:
```
switch (engine) {
        case MYISAM: return "MYISAM";
        case INNODB: return "INNODB";
        case MEMORY: return "MEMORY";
}
// For others 
throw ....;
```
In this case the complier will also warn you if you missed explicitly mentioning enum values in the `switch`.

core/modules/rproc/InfileMerger.cc
```diff
@@ -193,18 +159,43 @@ InfileMerger::InfileMerger(InfileMergerConfig const& c,
...
+    for (auto&& c:engineName) { eName += toupper(c); }
+    if (eName == "INNODB") {
+        _dbEngine = INNODB;
+    } else if (eName == "MEMORY") {
+        _dbEngine = MEMORY;
+    } else if (eName == "MYISAM") {
+        _dbEngine = MYISAM;
+    } else {
+        LOGS(_log, LOG_LVL_ERROR, "unknown dbEngine=" << engineName << " using default MYISAM");
+        _dbEngine = MYISAM;
```
iagaponenko, line 175: An exception needs to be thrown here to tell users that Qserv is misconfigured, or there was a problem reading a configuration ile.

core/modules/rproc/InfileMerger.cc
```diff
@@ -261,19 +241,18 @@ bool InfileMerger::merge(std::shared_ptr<proto::WorkerResponse> response) {
     if (response->result.row_size() == 0) {
         return true;
     }
-    _sizeCheckRowCount += response->result.row_size();
-
 
-    SemaLock semaLock(semaMgrLocal); // &&&
+    std::unique_ptr<util::SemaLock> semaLock;
+    if (_dbEngine != MYISAM) {
```
iagaponenko, line 240: Based on subsequent comment that "// needed for INNODB and MYISAM", did you mean to compare with `MEMORY` instead?```
    if (_dbEngine == MEMORY) {
        // needed for INNODB and MYISAM
        ...
```

File: core/modules/rproc/InfileMerger.cc [hunk 12]
```diff
@@ -284,18 +263,23 @@ bool InfileMerger::merge(std::shared_ptr<proto::WorkerResponse> response) {
 
 
     auto start = std::chrono::system_clock::now();
-    // &&& ret = _applyMysql(infileStatement);
-    ret = _applyMysqlNew(infileStatement, infileMgr, virtFile);
+    if (_dbEngine == MYISAM) {
+        ret = _applyMysqlMyIsam(infileStatement);
+    } else { // Used for both INNODB and MEMORY
+        ret = _applyMysqlInnoDb(infileStatement);
+    }
```
iagaponenko, line 264: I think using `case` here would be future-proof, in case if more engine options would be used. For others throw an exception:
```
switch (_dbEngine) {
    case MYISAM:
        ret = _applyMysqlMyIsam(infileStatement);
        break;
    case INNODB:
    case MEMORY:
        ret = _applyMysqlInnoDb(infileStatement);
        break;
    default:
        throw ...
}```

File: core/modules/rproc/InfileMerger.cc [hunk 13]
```diff
@@ -319,34 +303,38 @@ bool InfileMerger::merge(std::shared_ptr<proto::WorkerResponse> response) {
...
 }
 
-/* &&&INFILE
-bool InfileMerger::_applyMysql(std::string const& query) {
-    std::lock_guard<std::mutex> lock(_mysqlMutex);
-    if (!_mysqlConn.connected()) {
+
+bool InfileMerger::_applyMysqlMyIsam(std::string const& query) {
+    std::unique_lock<std::mutex> lock(_mysqlMutex);
+    for (int j=0;!_mysqlConn.connected(); ++j) {
```
iagaponenko, line 307: Space is needed at:
```
    for (int j=0; !_mysqlConn.connected(); ++j) {
```

core/modules/rproc/InfileMerger.cc
```diff
@@ -319,34 +303,38 @@ bool InfileMerger::merge(std::shared_ptr<proto::WorkerResponse> response) {
...
         // should have connected during construction
         // Try reconnecting--maybe we timed out.
-        if (!_setupConnection()) {
-            LOGS(_log, LOG_LVL_ERROR, "InfileMerger::_applyMysql _setupConnection() failed!!!");
-            return false; // Reconnection failed. This is an error.
+        if (!_setupConnectionMyIsam()) {
+            if (j < 10) {
+                lock.unlock();
+                sleep(1);
```
iagaponenko, line 313: Why all these magic numbers here? Application-level connection timeouts should be configurable.

core/modules/rproc/InfileMerger.cc
```diff
@@ -356,16 +344,18 @@ bool InfileMerger::_applyMysqlNew(std::string const& query, mysql::LocalInfile::
...
-        //&&& std::lock_guard<std::mutex> aLG(localAttachMtx);
-        LOGS(_log, LOG_LVL_WARN, "&&&setupConnectionNew start " << virtFile);
-        // &&&INFILE _infileMgr.attach(mySConn.getMySql());
-        infileMgr.attach(mySConn.getMySql());
-        LOGS(_log, LOG_LVL_WARN, "&&&setupConnectionNew end " << virtFile << std::endl);
-        return true;
+bool InfileMerger::_setupConnectionInnoDb(mysql::MySqlConnection& mySConn) {
+    // Make 10 attempts to open the connection. They can fail when the
+    // system is busy.
+    for (int j=0; j < 10; ++j) {
```
iagaponenko, line 345: The limit should be configurable, not hard-wired.

core/modules/rproc/InfileMerger.cc
```diff
@@ -502,9 +492,13 @@ bool InfileMerger::makeResultsTableForQuery(query::SelectStmt const& stmt) {
...
-    //createStmt += " ENGINE=MyISAM"; // &&&
-    //createStmt += " ENGINE=MEMORY"; // &&&
-    createStmt += " ENGINE=InnoDB"; // &&&
+    if (_dbEngine == INNODB) {
+        createStmt += " ENGINE=InnoDB";
+    } else if (_dbEngine == MEMORY) {
+        createStmt += " ENGINE=MEMORY";
+    } else { // MYISAM
+        createStmt += " ENGINE=MyISAM";
+    }
```
iagaponenko, line 495: This has to be `switch(_dbEngine) { ...` and an exception `throw` for the rest (unknown table type).

core/modules/util/SemaMgr.h
```diff
@@ -0,0 +1,110 @@
...
+
+    void _release() {
+        --_totalCount;
+        --_usedCount;
+        _tCv.notify_one();
+    }
+
+    std::atomic<int> _totalCount{0};
+    std::atomic<int> _usedCount{0};
+    std::atomic<int> _max;
```
iagaponenko, line 84: Why these are not `unsigned`?

core/modules/util/StringToVector.h
```diff
@@ -0,0 +1,51 @@
...
+// Qserv headers
+#include "util/Command.h"
+
+namespace lsst {
+namespace qserv {
+namespace util {
+
+/// @return a vector of strings resulting from splitting 'str' into separate strings
+/// using 'separator' as the delimiter.
+std::vector<std::string> splitString(std::string const& str, std::string const& separator);
```
iagaponenko, line 41: I would argue that this and other method was put into a utility class rather than having them as stand-alone functions in the namespace.

core/modules/util/StringToVector.cc
```diff
@@ -0,0 +1,84 @@
...
+
+LOG_LOGGER _log = LOG_GET("lsst.qserv.util.StringToVector");
+
+}
+
+namespace lsst {
+namespace qserv {
+namespace util {
+
+std::vector<std::string> splitString(std::string const& original, std::string const& separator) {
```
iagaponenko, line 43: This implementation is too complex. There are much simple and straightforward alternatives like:
```
#include <string>
#include <sstream>
#include <algorithm>
#include <iterator>
 
template <class Container>
Container split2(const std::string& str, char delim=':') {
    std::stringstream ss(str);
    std::string token;
    Container cont;
    while (std::getline(ss, token, delim)) {
        cont.push_back(token);
    }
    return cont;
}
```
Besides, you may also do type conversions into a desired numeric type right within this loop. So that you could return a collection of numeric values right here. And you won't need two separate functions `splitString` and `getIntVectFromStr`.
You may look for other ideas here: http://www.martinbroadhurst.com/how-to-split-a-string-in-c.html

core/modules/wdb/QueryRunner.cc
```diff
@@ -254,6 +261,21 @@ bool QueryRunner::_fillRows(MYSQL_RES* result, int numFields, uint& rowCount, si
...
+            // This query has been answered by the database and the
+            // scheduler for this worker should stop waiting for it.
+            // leavePool() will tell the scheduler this task is finished
+            // and create a new thread in the pool to replace this one.
+            auto pet = _task->getAndNullPoolEventThread();
+            _removedFromThreadPool = true;
+            if (pet != nullptr) {
+                pet->leavePool();
+            } else {
+                LOGS(_log, LOG_LVL_WARN, "Result PoolEventThread was null. Probably already moved.");
```
iagaponenko, line 275: Is this a bug or a feature of the new code? If the former then the process need to be terminated right away. Otherwise these warnings will be buried in the long files.

core/modules/wcontrol/TransmitMgr.cc
```diff
@@ -0,0 +1,97 @@
...
+void TransmitMgr::_take(bool interactive, bool alreadyTransmitting) {
+    ++_totalCount;
+    if (not interactive || _transmitCount >= _maxTransmits || _alreadyTransCount >= _maxAlreadyTran) {
+        // If not alreadyTransmitting, it needs to wait until the number of already transmitting
+        // jobs drops below _maxAlreadyTran before it can start transmitting.
+        if (alreadyTransmitting) {
+            unique_lock<mutex> uLock(_mtx);
+            ++_alreadyTransCount;
+            LOGS(_log, LOG_LVL_DEBUG, " ++_alreadyTransCount=" << _alreadyTransCount);
+            _tCv.wait(uLock, [this](){ return _transmitCount < _maxTransmits; });
```
iagaponenko, line 49: I think the language allows removing empty brackets if no arguments are defined for a lambda function:
```
            _tCv.wait(uLock, [this] { return _transmitCount < _maxTransmits; });
```

core/modules/wcontrol/TransmitMgr.cc
```diff
@@ -0,0 +1,97 @@
...
+        } else {
+            unique_lock<mutex> uLock(_mtx);
+            _tCv.wait(uLock, [this](){
+                return (_transmitCount < _maxTransmits) && (_alreadyTransCount < _maxAlreadyTran);
+            });
+        }
+    }
+    // Incrementing outside the mutex may result in occasionally having more than
+    // _maxTransmits happening at a time.
+    ++_transmitCount;
```
iagaponenko, line 59: The implementation of the method is a bit convoluted. Why not to acquire the mutex at the very start of the method to ensure you won't need to make this type of comments?
```
void TransmitMgr::_take(bool interactive, bool alreadyTransmitting) {
    unique_lock<mutex> uLock(_mtx);
    ...
    ++_transmitCount;
}

File: core/modules/wcontrol/TransmitMgr.cc [hunk 23]
```diff
@@ -0,0 +1,97 @@
...
+void TransmitMgr::_release(bool interactive, bool alreadyTransmitting) {
+    --_totalCount;
+    --_transmitCount;
+    if (not interactive && alreadyTransmitting) {
+        --_alreadyTransCount;
+        LOGS(_log, LOG_LVL_DEBUG, "--_alreadyTransCount=" << _alreadyTransCount);
+        _tCv.notify_one();
+    }
+    _tCv.notify_one();
+}
```
iagaponenko, line 72: Wy two `_tCv.notify_one();`? This is confusing.
I would ALSO argue that this method had to acquire a lock on the mutex before _consistently_ modifying its state even if it's made of atomics:
```
void TransmitMgr::_release(bool interactive, bool alreadyTransmitting) {
    unique_lock<mutex> uLock(_mtx);
    ...
}
```

core/modules/wcontrol/SqlConnMgr.h
```diff
@@ -0,0 +1,108 @@
...
+
+private:
+    void _take(bool scanQuery);
+    void _release();
+
+    std::atomic<int> _totalCount{0};
+    std::atomic<int> _sqlConnCount{0};
+    int _maxSqlConnections;
+    int _maxScanSqlConnections;
+    std::mutex _mtx;
```
iagaponenko, line 83: All counters are better to be declared as `unsigned int`.

core/modules/wcontrol/SqlConnMgr.cc
```diff
@@ -0,0 +1,82 @@
...
+    dump(os);
+    return os.str();
+}
+
+
+ostream& operator<<(ostream &os, SqlConnMgr const& mgr) {
+    return mgr.dump(os);
+}
+
+
```
iagaponenko, line 81: Extra line

core/modules/wcontrol/SqlConnMgr.cc
```diff
@@ -0,0 +1,82 @@
...
+    ++_sqlConnCount;
+}
+
+void SqlConnMgr::_release() {
+    --_sqlConnCount;
+    --_totalCount;
+    _tCv.notify_one();
+}
+
+
```
iagaponenko, line 59: Extra line.

core/modules/wcontrol/SqlConnMgr.cc
```diff
@@ -0,0 +1,82 @@
...
+    _tCv.wait(uLock, [this, scanQuery](){
+        bool ok = _sqlConnCount < _maxScanSqlConnections;
+        if (not scanQuery) {
+            ok = _sqlConnCount < _maxSqlConnections;
+        }
+        return ok;
+    });
+    ++_sqlConnCount;
+}
+
```
iagaponenko, line 52: Missing empty line. There should be 2 lines between methods in the implementation files.

core/modules/wcontrol/SqlConnMgr.cc
```diff
@@ -0,0 +1,82 @@
...
+LOG_LOGGER _log = LOG_GET("lsst.qserv.wcontrol.SqlConnMgr");
+}
+
+using namespace std;
+
+namespace lsst {
+namespace qserv {
+namespace wcontrol {
+
+
```
iagaponenko, line 39: No need for an extra line here.

core/modules/wcontrol/Foreman.h
```diff
@@ -127,6 +131,12 @@ class Foreman
 
     mysql::MySqlConfig const        _mySqlConfig;
     wpublish::QueriesAndChunks::Ptr _queries;
+
+    /// For limiting the number of MySQL connections used for tasks.
+    std::shared_ptr<wcontrol::SqlConnMgr> _sqlConnMgr;
+
+    /// For limiting the number of transmits back to czars
+    std::shared_ptr<wcontrol::TransmitMgr> _transmitMgr;
 };
```
iagaponenko, line 140: These are better to be declared as `const`:
```
    std::shared_ptr<wcontrol::SqlConnMgr> const _sqlConnMgr;
    std::shared_ptr<wcontrol::TransmitMgr> const _transmitMgr;
```

admin/templates/configuration/etc/xrdssi.cnf
```diff
@@ -28,9 +28,8 @@ location = {{QSERV_DATA_DIR}}/mysql
 [scheduler]
 
 # Thread pool size
-# thread_pool_size = 20
 #thread_pool_size = 12
-thread_pool_size = 100
+thread_pool_size = 21
```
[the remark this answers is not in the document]
jgates108: I changed it to 20, but finding the best value requires experimenting. More threads could be better, but contention can become issues pretty easily.

admin/templates/configuration/etc/xrdssi.cnf
```diff
@@ -75,3 +72,16 @@ maxactivechunks_med = 50
 
 # Maximum number of Tasks that can take too long before moving a query to the snail scan.
 # maxtasksbootedperuserquery = 5
+
+[sqlconnections]
+# Maximum number of SQL connections for tasks
+maxsqlconn = 801
```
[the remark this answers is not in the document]
jgates108: No reason, but I do want to leave plenty of room for other things to connect to the database. If qserv has several failures to connect, queries will start failing, which could lead to a cascading failure.

core/modules/ccontrol/UserQueryResources.cc
```diff
@@ -26,12 +26,17 @@
...
+#include "util/SemaMgr.h"
+
 namespace lsst {
 namespace qserv {
 namespace ccontrol {
 
 
-UserQuerySharedResources::UserQuerySharedResources(std::shared_ptr<css::CssAccess> css_,
+UserQuerySharedResources::UserQuerySharedResources(czar::CzarConfig const& czarConfig_,
+                                                   std::shared_ptr<css::CssAccess> css_,
```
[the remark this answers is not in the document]
jgates108: I didn't notice that const& was missing in the parameter list for this and several others.

admin/templates/configuration/etc/xrdssi.cnf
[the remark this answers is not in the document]
iagaponenko: Multi-parametric optimization is a hard problem due to various indirect dependencies between parameters - changing a number of threads may require changing buffer sizes, a size of the MySQL connection pool, etc. Besides, a lot of things may depend on a hardware where Qserv is installed, a size of Qserv itself, as well as on the sizes and complexity of the catalogs. So, what's set for one unique combination of conditions may not work for others.

It would be nice to have just a few parameters which would have a predicted behavior ... or let Qserv to dynamically tune these parameters depending on the observed performance, latencies, resource utilization. This is just a "wish list".

core/modules/ccontrol/UserQueryResources.h
```diff
@@ -81,7 +88,7 @@ class UserQuerySharedResources {
 
     UserQuerySharedResources(UserQuerySharedResources const& rhs) = default;
     UserQuerySharedResources& operator=(UserQuerySharedResources const& rhs) = delete;
-
+    czar::CzarConfig const& czarConfig;
```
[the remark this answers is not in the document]
jgates108: czarConfig is a member of Czar and has the same lifetime as the czar itself. If it has gone missing, something is seriously wrong. A shared pointer cannot be used to pass it since Czar::czarConfig wasn't created that way.

core/modules/czar/CzarConfig.h
```diff
@@ -152,6 +152,27 @@ class CzarConfig {
...
+    /// Getters for QdispPool configuration
+    /// @return the number of threads to create for the pool.
+    int getQdispPoolSize() const { return _qdispPoolSize; }
+    /// @return the maximum priority for a queue. The number of queues
+    ///    equals this value +1.
+    int getQdispMaxPriority() const { return _qdispMaxPriority; }
+    /// @return a string with substrings separated by ':' like "2:45:32:9"
+    ///      The values indicate the maximum number of commands for each
+    ///      priority that can be running concurrently.
+    std::string getQdispVectRunSizes() const { return _qdispVectRunSizes; }
```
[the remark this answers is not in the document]
jgates108: There is not a fixed number of these, so trying to use a several different parameters gets pretty ugly with many of them not being used. I think this is better than that. A json style configuration would be a better solution, but that's a major change.

core/modules/czar/CzarConfig.h
```diff
@@ -170,6 +196,12 @@ class CzarConfig {
     int const _xrootdCBThreadsMax;
     int const _xrootdCBThreadsInit;
     int const _qMetaSecsBetweenChunkCompletionUpdates;
+
+    // Parameters for QdispPool configuration
+    int const _qdispPoolSize;
+    int const _qdispMaxPriority;
```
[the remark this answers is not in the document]
jgates108: ConfigStore doesn't have a method for unsigned int It just has getInt.  I don't trust it to behave appropriately if someone puts in -1.

core/modules/qdisp/QdispPool.cc
jgates108: I was hoping the name would be self explanatory. It gives a good indication of the czar's work load but logging it every time is a waste of disk space, unless you're logging everything anyway. Comment added.

core/modules/qdisp/QdispPool.cc
jgates108: I added upper limits for poolSize and largestPriority.

core/modules/qdisp/QdispPool.cc
jgates108: This is hanging code and is now removed.

core/modules/qdisp/QdispPool.h
jgates108: It's been changed to a member of the class and is used limit the value of largestPriority (set from configuration file). If someone needs more than 100 priorities, it's probably worth altering this class for that kind of use.

core/modules/rproc/InfileMerger.cc
jgates108: Typo, should be   // needed for INNODB and MEMORY. 

MyISAM opens one connection when the user query starts and reuses for every merge. INNODB and MEMORY have parallel merges and make new connections as needed and need SemaLock to keep from opening too many connections to MySQL.

core/modules/util/SemaMgr.h
jgates108: The extended positive range for unsigned isn't needed.
If there's a race condition and one of them goes negative, it's an instant disaster for an unsigned int.
I expect other classes with similar behavior may need to be derived from this and may require negative values.

core/modules/util/StringToVector.cc
jgates108: I tried this out and it has some bad behaviors. It fails unit tests that the other one passes. The one I'm really concerned about is this one. getline comes back with 0 tokens instead of 1. 
```
    {
        auto vect = util::splitString<std::vector<std::string>>("qsa4$3", ':');
        BOOST_CHECK(vect.size() == 1);
        BOOST_CHECK(vect[0] == "qsa4$3");
    }
```

core/modules/wcontrol/SqlConnMgr.h
jgates108: unsigned int has no real benefit here. The extended range isn't needed and if anything winds up going below zero with unsigned int, the logic breaks.

core/modules/wcontrol/TransmitMgr.cc
jgates108: The weirdness of not having locks is probably not worth the performance improvement.
There are 2 notify_one because I was trying to avoid a notify_all. If both alreadyTransCount and transmitCount are decremented, any waiting thread can be woken up. If only transmitCount is decremented, there's a good chance that thread woken really needed both variables decremented and nothing would advance. It really needs a notify all when only one is decremented.

core/modules/wcontrol/TransmitMgr.h
```diff
@@ -0,0 +1,119 @@
...
+
+    TransmitMgr(int maxTransmits, int maxAlreadyTran)
+        : _maxTransmits(maxTransmits),  _maxAlreadyTran(maxAlreadyTran) {
+        assert(_maxTransmits > 1);
+        assert(_maxAlreadyTran > 1);
+    }
+    TransmitMgr() = delete;
+    TransmitMgr(TransmitMgr const&) = delete;
+    TransmitMgr& operator=(TransmitMgr const&) = delete;
+    virtual ~TransmitMgr() = default;
```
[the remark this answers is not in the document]
jgates108: I try to write classes with the intent that deriving other classes from them is easy. virtual is very cheap. I tried to get this and a couple of other classes to share a common base class, but it didn't really work and this class needs significant modification to handle multiple czars correctly anyway, so I dropped it..

core/modules/wcontrol/TransmitMgr.h
```diff
@@ -0,0 +1,119 @@
...
+    friend class TransmitLock;
+
+private:
+    void _take(bool interactive, bool alreadyTransmitting);
+
+    void _release(bool interactive, bool alreadyTransmitting);
+
+    std::atomic<int> _totalCount{0};
+    std::atomic<int> _transmitCount{0};
+    std::atomic<int> _alreadyTransCount{0};
```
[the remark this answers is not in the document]
jgates108: My feeling is that unsigned int's are a nuisance and should be avoided unless you really need them. Accidentally subtracting a large value from a small value is pretty deadly and the compiler wont save you.

core/modules/wcontrol/TransmitMgr.h
```diff
@@ -0,0 +1,119 @@
...
+    int _maxAlreadyTran;
+    std::mutex _mtx;
+    std::condition_variable _tCv;
+};
+
+
+/// RAII class to support TransmitMgr
+class TransmitLock {
+public:
+    TransmitLock(TransmitMgr& transmitMgr, bool interactive, bool alreadyTransmitting)
```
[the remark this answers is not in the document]
jgates108: Efficiency, easier to read. A RAII lock should have pretty limited scope, so incrementing the values in a shared pointer will just slow things down more by locking another mutex. Holding a reference to a shared pointer (instead of incrementing the pointer) has all the same risks as the reference.

core/modules/wdb/QueryRunner.cc
jgates108: This is old code but there are other process that can move the thread out of the pool. The primary one being moved to the snail scheduler for taking too long.

core/modules/wsched/BlendScheduler.cc
```diff
@@ -357,7 +357,7 @@ util::Command::Ptr BlendScheduler::getCmd(bool wait) {
     if (cmd != nullptr) {
         _infoChanged = true;
         _logChunkStatus();
-        notify(true);
+        notify(false);
```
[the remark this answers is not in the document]
jgates108: I added comments for clarity, but I will keep this in mind when there's more than one argument.

core/modules/wsched/SchedulerBase.h
```diff
@@ -129,10 +131,10 @@ class SchedulerBase : public wcontrol::Scheduler {
...
 
     std::string const _name{}; //< Name of this scheduler.
-    int _maxReserve{1};    //< Number of threads this scheduler would like to have reserved for its use.
-    int _maxReserveDefault{1};
-    int _maxThreads{1};    //< Maximum number of threads for this scheduler to have inFlight.
-    int _maxThreadsAdj{1}; //< Maximum number of threads to have inFlight adjusted for available pool.
+    int _maxReserve = 1;    //< Number of threads this scheduler would like to have reserved for its use.
+    int _maxReserveDefault = 1;
+    int _maxThreads = 1;    //< Maximum number of threads for this scheduler to have inFlight.
+    int _maxThreadsAdj = 1; //< Maximum number of threads to have inFlight adjusted for available pool.
```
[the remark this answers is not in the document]
jgates108: These are used in calculations where negative numbers are possible and needed, it is asking for trouble to use unsigned.

core/modules/xrdsvc/SsiRequest.cc
```diff
@@ -359,7 +359,11 @@ void SsiRequest::Finished(XrdSsiRequest& req, XrdSsiRespInfo const& rinfo, bool
 
     // We can't do much other than close the file.
     // It should work (on linux) to unlink the file after we open it, though.
-    LOGS(_log, LOG_LVL_DEBUG, "RequestFinished " << type);
+    // With the optimizer on '-Og', there was a double free for a SsiRequest.
+    // The likely cause could be keepAlive being optimized out for being unused.
+    // The problem has not reoccurred since adding keepAlive to the following
+    // comment, but having code depend on a comment line is ugly in its own way.
+    LOGS(_log, LOG_LVL_DEBUG, "RequestFinished " << type << " " << keepAlive.use_count());
```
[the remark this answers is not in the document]
jgates108: If that is what happened (and the problem did go away), yeah, it's downright terrifying.

admin/templates/configuration/etc/xrdssi.cnf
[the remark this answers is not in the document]
jgates108: I think the problem is really complicated and may need machine learning to come up with an optimal solution. At least having them in the config files makes it easier to experiment and find decent answers.

admin/templates/configuration/etc/qserv-czar.cnf
```diff
@@ -20,11 +20,24 @@ passwd =
...
 port = 0
+
 # maximum user query result size in MB
 maxtablesize_mb = 5100
 
+# Database engine used for aggreagting results. 
+# Values:   myisam, innodb, memory
+# myisam is serial, innodb and memory are parallel.
+# memory does not support blob or text fields
+engine = innodb
```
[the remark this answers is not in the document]
jgates108: myisam is the default and I added a comment.

admin/templates/configuration/etc/qserv-czar.cnf
```diff
@@ -20,11 +20,24 @@ passwd =
...
+# Database engine used for aggreagting results. 
+# Values:   myisam, innodb, memory
+# myisam is serial, innodb and memory are parallel.
+# memory does not support blob or text fields
+engine = innodb 
+
+# The maximum number of database connections to use for
+# aggregating results. Only meaningful for 'parallel' 
+# engines such as innodb and memory.
+maxconnections = 40
```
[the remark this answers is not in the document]
jgates108: Max per czar, not per query. I added a comment.

PR #519, opened 2020-02-12
DM-23484: qserv column headers in results are wrong with optimization turned on

n8pease:


core/modules/SConscript
```diff
@@ -201,6 +201,8 @@ if not env.GetOption('clean'):
 
     env.Append(CPPPATH=[os.getcwd()])
 
+    env.Append(CCFLAGS = ' -Og ')
```
ktlim: You don't actually need spaces; SCons takes care of separating all CCFLAGS entries.

core/modules/ccontrol/ParseAdapters.h
```diff
@@ -2095,7 +2095,7 @@ class AggregateWindowedFunctionAdapter :
             auto param = std::make_shared<query::ValueExpr>();
             assert_execution_condition(nullptr != _valueFactor, "ValueFactor must be populated.", _ctx);
             param->addValueFactor(_valueFactor);
-            antlr4::tree::TerminalNode * terminalNode;
+            antlr4::tree::TerminalNode * terminalNode = nullptr;
```
iagaponenko: The `TerminalNode` times `terminalNode` is equal to `nullptr`? 😹
```
antlr4::tree::TerminalNode* terminalNode = nullptr;
```

core/modules/query/ValueExpr.h
```diff
@@ -245,7 +245,7 @@ class ValueExpr {
 private:
     std::string _alias;
     FactorOpVector _factorOps;
-    bool _aliasIsUserDefined; /// true if the alias was defined by the user in the select statement.
+    bool _aliasIsUserDefined=false; /// true if the alias was defined by the user in the select statement.
```
iagaponenko: Also, the "LSST Style Guide" requires that "Assignment operators always have spaces around them":  https://developer.lsst.io/cpp/style.html#the-following-white-space-conventions-should-be-followed

core/modules/SConscript
n8pease: just a typo but thx for catching.

PR #521, opened 2020-02-28
Add debug tools (gdb etc.) to base container

fritzm:


fritzm:
Pulling in `devtoolset-8-toolchain` gets `gdb`, `gcore`, `strace`, and `pstack`; `lsof` installed explicitly

PR #522, opened 2020-03-06
Tickets/dm 23086

iagaponenko:


admin/tools/docker/replication/container/tools/Dockerfile.gke
```diff
@@ -1,5 +1,5 @@
 # Using the development toolchain
-FROM centos/devtoolset-6-toolchain-centos7
+FROM qserv/replica:dev
```
fritzm: What builds qserv/replica:dev?  Not part of these changes?

iagaponenko: Unfortunately, there is no such thing as {{centos/devtoolset-8-toolchain-centos7}} (*NOTE* version *8* in the proposed container tag) yet. So, I had to go with  {{qserv/replica:dev}} which was upgraded to {{lsstsqre/centos:7-stackbase-devtoolset-8}}. In the end, it doesn't matter because this is a layered container build. One added benefit of that move is that now I have the stack from the base container.

One question to you: will this affect the memory size of the container?

PR #523, opened 2020-03-06
Refactored the performance test to support threads and flow-control.

iagaponenko:


core/modules/wpublish/qserv-worker-perf-chunks.cc
```diff
@@ -46,101 +160,79 @@ int test() {
...
+                vector<RequestT::Ptr> requests;
+                for (auto&& resourcePath: jobs) {
+                    requests.push_back(makeRequest(numRequestsInFlight, serviceProvider, resourcePath));
+                }
+    
+                // Block while at least one request is in progress
+                util::BlockPost blockPost(1000, 2000);
+                while (numRequestsInFlight.counter() != 0) {
+                    blockPost.wait(200);
+                }
```
jgates108: Why the random wait time? It looks like it is trying to hide a potential race condition with numRequestedInFlight.
I think you would be better off with a separate Counter defined in this thread, say localCounter. localCounter would be incremented in the requests.push_back loop (the increment must happen in this thread or there's a race condition) and then decremented at the same point numRequestsInFlight is decremented (so it needs to be passed to makeRequest). Then just wait for localCounter to reach 0.

iagaponenko: Good point! Thank you!

core/modules/wpublish/qserv-worker-perf-chunks.cc
```diff
@@ -35,7 +67,89 @@ string value;
...
+        }
+    }
+
+    void dec() {
+        if (_maxRequestsAllowed == 0) {
+            --_counter;
+        } else {
+            unique_lock<mutex> lock(_mtx);
+            --_counter;
+            _cv.notify_one();
```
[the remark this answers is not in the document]
iagaponenko: You're right!

core/modules/wpublish/qserv-worker-perf-chunks.cc
jgates108: On the other hand, I think the join below is enough since there isn't any other logic that needs to be run before the join.

PR #527, opened 2020-04-02
Tickets/dm 22745

iagaponenko:


core/modules/replica/IngestServerConnection.cc
```diff
@@ -191,6 +195,13 @@ void IngestServerConnection::_handshakeReceived(boost::system::error_code const&
     _isOverlap       = request.is_overlap();
     _columnSeparator = request.column_separator() == ProtocolIngestHandshakeRequest::COMMA ? ',' : '\t';
 
+    // Check if the client is authorized for the operation
+
+    if (request.auth_key() != _authKey) {
```
ktlim, line 200: But this isn't separated out into a function...

iagaponenko: Good point!

core/modules/replica/HttpModule.cc
```diff
@@ -102,4 +109,13 @@ void HttpModule::sendData(qhttp::Response::Ptr const& resp,
     resp->send(result.dump(), "application/json");
 }
 
+
+void HttpModule::_enforceAuthorization(qhttp::Request::Ptr const& req) const {
+    HttpRequestBody body(req);
+    auto authKey = body.required<string>("auth-key");
+    if (authKey != _processorConfig.authKey) {
```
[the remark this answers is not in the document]
iagaponenko: Method `_enforceAuthorization` was introduces exactly for this reason.

iagaponenko:
@andy-slac (writing it here as, for some reason, I no longer able to put replies next to your questions).

Regarding your comment on the "on implemented" method `HttpExportModule::_getTables`. This class is implemented in a dependent branch `tickets/DM-24053`. When I started working on the current development I was caught at some "wrong" moment of time. I had to add an incomplete version of this class into the current branch to avoid running into some nasty `git` games with rebasing/merging.

As far as your statement on the "authorization plugin" is concerned, note that I'm using a custom version of the `qhttp` server for C++. This server is rather simple, and it doesn't have support for authorization/authentication. So, I tried to confine the auth-enforcement code witin the base class `HttpModule` of the modules. This appeared to me like a reasonable compromise  which let me avoid modifying each subclass (module).

And I 100% agree with your other comments/suggestions (minus having it all done in Python 😸).

PR #530, opened 2020-04-17
Another major refactoring of the REST services. The goal was to

iagaponenko:
allow a dedicated instance of a module class serving a particular
request. In the previous model a static collection of modules was
instantiated ahead of time.

### DIFF
--- a/core/modules/replica/HttpProcessor.cc
+++ b/core/modules/replica/HttpProcessor.cc
@@ -52,10 +52,7 @@ HttpProcessor::Ptr HttpProcessor::create(
                         Controller::Ptr const& controller,
                         HttpProcessorConfig const& processorConfig,
                         HealthMonitorTask::Ptr const& healthMonitorTask) {
-
-    auto ptr = Ptr(new HttpProcessor(
-        controller, processorConfig, healthMonitorTask
-    ));
+    auto ptr = Ptr(new HttpProcessor(controller, processorConfig, healthMonitorTask));
     ptr->_initialize();
     return ptr;
 }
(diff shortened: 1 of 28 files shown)


core/modules/replica/HttpProcessor.cc
```diff
@@ -107,193 +82,302 @@ void HttpProcessor::_initialize() {
     controller()->serviceProvider()->httpServer()->addHandlers({
         {"GET", "/replication/v1/catalogs",
             [self](qhttp::Request::Ptr const& req, qhttp::Response::Ptr const& resp) {
-                self->_catalogsModule->execute(req, resp);
+                HttpCatalogsModule::process(
+                        self->controller(), self->name(), self->_processorConfig,
+                        req, resp);
```
jgates108, line 87: I don't quite understand why you made all the process functions when I think this would have worked
```
{
    HttpCatalogsModule catalogModule(self->controller, taskName. processorConfig);
    catalogueModule.execute(req, resp);
}
```

iagaponenko: The main reason was to reduce the number of statements in lambdas. The requests routing table table in `HttpProcessor` is already quite long. The second reason was just a personal preference.

jgates108: What I've got above is wrong, my bad. But all of the static process functions just do the following with slightly different parameters
```
{
    HttpCatalogsModule module(controller, taskName, processorConfig, req, resp);
    module.execute(subModuleName, authType);
}
```
Getting rid of the static process functions an putting that code directly in the lambda would get rid of dozens , if not hundreds, of lines of code that only defines the process functions.

iagaponenko: What you have just observed was the third reason for the change - hiding implementation details behind the static method `::process`. I may, for example implement a pool of modules to avoid instantiating module objects for each requests as it's in the current implementation. In this case the client code `HttpProcessor` won't be affected.

jgates108: Another way to do that would be to make a class specific version of execute() in those cases where it is needed.

iagaponenko: I didn't get your last idea. Could you post a demo please?

jgates108: I don't really have an example to work with, but all the variables given to the static function are available in the constructor and the call to execute. It is good that all the functions call the master version of execute, since they all want to do the same thing. And, I believe, all of the process functions call execute().

If one of the classes needs to do something entirely different for execute(), say class CrazyCornerCase, then write CrazyCornerCase::execute(std::string const& subModuleName=std::string(), AuthType const authType) to do whatever needs to be done. It doesn't look like there on any real constraints on what's in execute that would not apply to the static process function.

jgates108: Also note, that CrazyCornerCase does not need to use HttpModule as a base class.

iagaponenko: Note that all module classes derive from `HttpModule`. That base class provides a bunch of services for the subclasses: 1) logging services, 2) parsing parameters of the URL and teh body of the HTTP request, 3) verifying and enforcing authorizations, 3) interacting with `qhttp` to sending errors or results back to clients. Some of this is done in the constructor of the base class, some other in method `::execute(..)`. The rest is done by subclasses in the subclass-specific final/overloaded methods `::executeImpl()`.

I should also mention that the forth(!) reason for this refactoring was to make the implementation of the REST services (modules) 100% independent of `qhttp`. The only dependency onto `qhttp` is now within the base class `HttpModule` and the request dispatcher class `HttpProcessor`.

PR #532, opened 2020-04-23
Tickets/dm 24458

iagaponenko:


core/modules/replica/IngestServerConnection.cc
```diff
@@ -404,93 +405,105 @@ void IngestServerConnection::_loadDataIntoTable() {
...
-                    "CREATE TABLE IF NOT EXISTS " + table + " LIKE " + sqlProtoTable
-                );
-                statements.push_back(
-                    "ALTER TABLE " + table + " ADD PARTITION IF NOT EXISTS (PARTITION " + sqlPartition +
-                        " VALUES IN (" + to_string(_transactionId) + "))"
-                );
+            // Note, that the algorithm will create chunked tables for _ALL_ partitioned
+            // tables (not just for the current one) to ensure they have representations
+            // in all chunks touched by the ingest workflows. Missing representations would
+            // case Qserv to fail when processing queries involving these tables.
```
andy-slac, line 438: case -> cause?

core/modules/replica/IngestServerConnection.cc
```diff
@@ -404,93 +405,105 @@ void IngestServerConnection::_loadDataIntoTable() {
...
+
+                // Chunked tables are created from the prototype table which is expected
+                // to exist in the database before attempting data loading.
+
+                bool const overlap = true;
+                string const sqlProtoTable       = sqlDatabase + "." + h.conn->sqlId(table);
+                string const sqlTable            = sqlDatabase + "." + h.conn->sqlId(ChunkedTable(table, _chunk, not overlap).name());
+                string const sqlFullOverlapTable = sqlDatabase + "." + h.conn->sqlId(ChunkedTable(table, _chunk, overlap).name());
+
+                vector<string> const tablesToBeCreated = {
```
andy-slac, line 450: Could do just `string tablesToBeCreated[]`, no need for vector.

core/modules/replica/IngestServerConnection.cc
```diff
@@ -404,93 +405,105 @@ void IngestServerConnection::_loadDataIntoTable() {
...
+                    " FIELDS TERMINATED BY " + h.conn->sqlValue(string() + _columnSeparator)
             );
         }
         for (auto&& statement: statements) {
             LOGS(_log, LOG_LVL_DEBUG, context << __func__ << "  statement: " << statement);
         }
-        conn->execute([&statements](decltype(conn) const& conn_) {
-            conn_->begin();
+        // Note that each statement gets executed in its own transaction to eliminate
+        // a probability of deadlocking on the DML operations creating tables.
```
andy-slac, line 496: Not sure what this means? "DML operations creating tables" -- creating table is a DDL, and though that it implicitly commits transactions. And MyISAM is not transactional.

core/modules/replica/SqlApp.cc
```diff
@@ -221,6 +222,34 @@ SqlApp::SqlApp(int argc, char* argv[])
...
+        _table
+    ).required(
+        "engine",
+        "The name of a MySQL engine for the new tables",
+        _engine
+    ).required(
+        "schema-file",
+        "The name of a file where column definitions of the table schema will be"
+        " read from. If symbol '-' is passed instead of the file name then column"
+        " definitions will be read from the Standard Input File. The file is required"
```
andy-slac, line 243: Standard Input File? :what-is-it???:

core/modules/replica/SqlApp.cc
```diff
@@ -221,6 +222,34 @@ SqlApp::SqlApp(int argc, char* argv[])
...
+        " to have the following format: <column-name> <type>",
+        _schemaFile
+    ).option(
+        "partition-by-column",
+        "The name of a column which is used for creating the tables based on"
+        " the MySQL partitioning mechanism,",
```
andy-slac, line 249: comma should a stop?

core/modules/replica/SqlCreateTablesJob.h
```diff
@@ -0,0 +1,149 @@
...
+// This header declarations
+namespace lsst {
+namespace qserv {
+namespace replica {
+
+/**
+ * Class SqlCreateTablesJob represents a tool which will broadcast batches of
+ * the table creation requests to workers.
+ *
+ * @note the meaning of the 'table' depends on a kind of a table. If this is
```
andy-slac, line 45: the -> The

core/modules/replica/SqlCreateTablesJob.h
```diff
@@ -0,0 +1,149 @@
...
+     * @param database the name of a database from which a table will be created
+     * @param table the name of the base table to be affected by the operation
+     * @param allWorkers engage all known workers regardless of their status.
+     *  If the flag is set to 'false' then only 'ENABLED' workers which are not
+     *  in the 'READ-ONLY' state will be involved into the operation.
+     * @param controller is needed launching requests and accessing the Configuration
+     * @param parentJobId (optional) identifier of a parent job
+     * @param onFinish (optional) callback function to be called upon a completion of the job
+     * @param options (optional) defines the job priority, etc.
+     * @return pointer to the created object
```
andy-slac, line 77: Not all parameters are documented.

core/modules/replica/SqlCreateTablesJob.cc
```diff
@@ -0,0 +1,164 @@
...
+        bool allWorkers,
+        Controller::Ptr const& controller,
+        string const& parentJobId,
+        CallbackType const& onFinish,
+        Job::Options const& options)
+    :   SqlJob(0,
+               allWorkers,
+               controller,
+               parentJobId,
+               "SQL_DROP_TABLE_PARTITION",
```
andy-slac, line 91: Why is it called SQL_DROP_TABLE_PARTITION?

core/modules/replica/SqlCreateTablesJob.cc
```diff
@@ -0,0 +1,164 @@
...
+
+
+list<pair<string,string>> SqlCreateTablesJob::extendedPersistentState() const {
+    list<pair<string,string>> result;
+    result.emplace_back("database", database());
+    result.emplace_back("table", table());
+    result.emplace_back("engine", engine());
+    result.emplace_back("partition_by_column", partitionByColumn());
+    result.emplace_back("num_columns", to_string(columns().size()));
+    result.emplace_back("all_workers", string(allWorkers() ? "1" : "0"));
```
andy-slac, line 109: If you do emplace then there is no point in wrapping it into `string()`.

core/modules/replica/IngestServerConnection.cc
iagaponenko: This optimization will the next block of code to fail:
```
                for (auto&& table: tablesToBeCreated) {
                    statements.push_back(
                        "CREATE TABLE IF NOT EXISTS " + table + " LIKE " + sqlProtoTable
                    );
                    statements.push_back(
                        "ALTER TABLE " + table + " ADD PARTITION IF NOT EXISTS (PARTITION " + sqlPartition +
                            " VALUES IN (" + to_string(_transactionId) + "))"
                    );
                }
```
It would require wrapping table names into `string(table)` to successfully concatenate strings.

andy-slac: I don't think so, `table` would still be a `std::string` (or reference to it).

core/modules/replica/IngestServerConnection.cc
iagaponenko: You're right about the transactions & MyISAM. In this case the proposed modification makes no sense. I just forgot that it's not InnoDB. A reason why I tried this within a transaction was to avoid occasional deadlocks (have seen them a coupe of times when the Ingest servers were under heavy load) on sequences of the table `CREATE`, `ALTER` (to add a MySQL partition) and `LOAD DATA INFILE` statements. It turns out table locking is trickier if using MySQL partitions. So, let me leave this code as it's now. I will investigate this issue as separate effort.

core/modules/replica/IngestServerConnection.cc
iagaponenko: You're right! Fixed.

core/modules/replica/IngestServerConnection.cc
iagaponenko: At the mean time I'm eliminating the confusing comment.

PR #536, opened 2020-04-27
Added exception handling for BOOST ASIO + Protobuf requests

iagaponenko:


core/modules/replica/ExportServerConnection.cc
```diff
@@ -74,9 +74,9 @@ bool isErrorCode(boost::system::error_code const& ec, string const& scope) {
 bool readIntoBuffer(boost::asio::ip::tcp::socket& socket,
                     shared_ptr<ProtocolBuffer> const& ptr,
                     size_t bytes) {
-
-    // The buffer needs to be resized to accommodate
-    // the message.
+    // Make sure the buffer has enough space to accommodate the data
+    // of the message. Note, this call may throw an exception which is
+    // supposed to be caught by the method's caller.
```
n8pease, line 79: but you don't catch the exception. Should you?

core/modules/replica/IngestServerConnection.cc
```diff
@@ -77,9 +77,11 @@ bool isErrorCode(boost::system::error_code const& ec,
 bool readIntoBuffer(boost::asio::ip::tcp::socket& socket,
                     shared_ptr<ProtocolBuffer> const& ptr,
                     size_t bytes) {
+    // Make sure the buffer has enough space to accommodate the data
+    // of the message. Note, this call may throw an exception which is
+    // supposed to be caught by the method's caller.
+    ptr->resize(bytes);
```
n8pease, line 83: again; catch? Or are you saying it should be caught by *this* method's caller?

core/modules/replica/ExportServerConnection.cc
iagaponenko: Not in this method. All exceptions are caught in the only client method:
```
template <class T>
bool readMessage(boost::asio::ip::tcp::socket& socket,
                 shared_ptr<ProtocolBuffer> const& ptr,
                 size_t bytes,
                 T& message) {
    try {
        if (readIntoBuffer(socket, ptr, bytes)) {
            ptr->parse(message, bytes);
            ...
```

core/modules/replica/IngestServerConnection.cc
iagaponenko: Exactly! This method may throw exceptions. The callers are supposed to call them.

PR #545, opened 2020-06-12
Tickets/dm 24307

iagaponenko:


core/modules/replica/HttpIngestModule.cc
```diff
@@ -1034,35 +1034,57 @@ pair<string,size_t> HttpIngestModule::_buildEmptyChunksListImpl(string const& da
     set<unsigned int> uniqueChunks;
     for (auto chunk: chunks) uniqueChunks.insert(chunk);
 
-    auto const file = "empty_" + database + ".txt";
-    auto const filePath = fs::path(config->controllerEmptyChunksDir()) / file;
+    unsigned int const maxChunkAllowed = 1000000;
```
jgates108, line 1037: Magic number. Maybe this can be taken from an existing source? Otherwise, it would be better to get it from a config file or DB entry.

iagaponenko: Thank you for the idea! Actually, this code has access to the required metadata (the number of chunks and sub-chunks), so it would be easy to migrate it to `Chunker`. I will open a separate "bug fix" ticket for that.

PR #546, opened 2020-06-12
Optimized implementation of the empty chunk list generator

iagaponenko:


core/modules/replica/HttpIngestModule.cc
```diff
@@ -1074,16 +1083,17 @@ json HttpIngestModule::_buildEmptyChunksListImpl(string const& database,
         if (not ofs.good()) {
             throw runtime_error("failed to create/open file: " + filePath.string());
         }
-        for (unsigned int chunk = 0; chunk < maxChunkAllowed; ++chunk) {
-            if (not uniqueChunks.count(chunk)) {
+        for (auto&& chunk: allChunks) {
+            if (not ingestedChunks.count(chunk)) {
```
jgates108, line 1087: Should we include invalid chunks in the empty chunk list? Invalid chunks should be empty. I think the logic should prevent qserv from looking for invalid chunks in area searches, but I'm not so sure about queries that mention a specific chunk id. If invalid chunks are put in the list, the last entry should be the largest chunk id + 1 so that any chunk >= to the last entry is known to be empty.

iagaponenko: Thank you for the comment! I didn't know about this problem. I will add "the largest chunk id + 1" to the empty chunks list. What about the lowest chunk number? Should I add it as well? Something like "the lowest chunk id - 1"?

PR #550, opened 2020-06-26
Tickets/dm 25482

iagaponenko:


admin/tools/docker/replication/container/tools/build.sh
```diff
@@ -34,7 +34,7 @@ docker run \
        -v /etc/group:/etc/group:ro \
        -v $HOME:$HOME \
        -v $PWD:$PWD \
-       qserv/replica:dev_conda \
+       qserv/replica:dev_conda_ceb6bb6 \
```
jgates108, line 37: Is linking to a specific version long term?

core/modules/replica/HttpIngestChunksModule.h
```diff
@@ -110,7 +110,7 @@ class HttpIngestChunksModule: public HttpModule {
                            std::string const& database,
                            unsigned int chunk) const;
 
-    util::Mutex _ingestManagementMtx;   /// Synchronized access to the Ingest management operations
+    static util::Mutex _ingestManagementMtx;    /// Synchronized access to the chunk management operations
```
jgates108, line 113: Maybe a few less spaces between the line and the comment.

admin/tools/docker/replication/container/tools/build.sh
iagaponenko: I hope it's not! It's one of the consequences of the Conda migration of the LSST Stack. The hash in the container tag's name relates to some Qserv-specific overide in the LSST Stack environment:
{code}
LSST_SPLENV_REF=ceb6bb6
{code}
I'm using a specific hash to ensure I have a container based on the corresponding version of Stack.
I also hope this all will get much easier after migrating to the CMake based build system.

iagaponenko: And before that, as the LSST Stack will get more stable I may go back to the less elaborate naming convention for naming these containers:
```
qserv/replica:dev
```

core/modules/replica/HttpIngestChunksModule.h
iagaponenko: 👍

PR #552, opened 2020-07-09
Added a REST service returning replica disposition.

iagaponenko:


core/modules/replica/HttpIngestChunksModule.cc
```diff
@@ -467,4 +468,62 @@ void HttpIngestChunksModule::_registerNewChunk(string const& worker,
...
 }
 
+
+json HttpIngestChunksModule::_getChunks() {
+    debug(__func__);
+
+    auto const databaseServices = controller()->serviceProvider()->databaseServices();
+    auto const config = controller()->serviceProvider()->config();
+
+    string const database = body().required<string>("database");
```
andy-slac, line 478: Why does it have to be in a body? I thought for GET request usual way to pass parameters is with the URL query.

core/modules/replica/HttpIngestChunksModule.cc
```diff
@@ -467,4 +468,62 @@ void HttpIngestChunksModule::_registerNewChunk(string const& worker,
...
+
+        // fetch and report actual values of the attributes.
+        for (auto&& file: replica.fileInfo()) {
+            auto& replicaTableResult = replicaResult[file.baseTable()];
+            if (file.isOverlap()) {
+                if      (file.isData())  replicaTableResult["overlap_data_size"]  = file.size;
+                else if (file.isIndex()) replicaTableResult["overlap_index_size"] = file.size;
+            } else {
+                if      (file.isData())  replicaTableResult["data_size"]  = file.size;
+                else if (file.isIndex()) replicaTableResult["index_size"] = file.size;
```
andy-slac, line 520: Multi-line `if` should use curly braces: https://developer.lsst.io/cpp/style.html#the-if-else-class-of-statements-should-have-the-following-form

iagaponenko: Agreed and fixed.

core/modules/replica/HttpIngestChunksModule.cc
iagaponenko: Well, that's how I initially intended to implement the service. Then I ran into a few problems. First of all, the name of a database is not a resource for this service, so putting it into the URL makes things to look a bit awkward. This is just one of (potentially many) filters and output formatters which could be passed into the service. I didn't have then in the initial implementation. But I expect the implementation of the service to be extended over time as we get more experience with (and further requirements for) the service. A more serious reason why I finally chose to pass parameters in the BODY in the body is the optional security information (the so called "auth_key") which is passed with some services. Chances are that I would have to do this for all services. Obvious, no one wants to pollute (Web and cache) servers' log files with the sensitive info.
I have also searched on the Web what others say on this subject. There are claims that in the previous HTTP RFC's sending anything in the request's body wasn't advised since some implementations may strip the bodies. Apparently these statements were removed in the current HTTP protocol 2.*. And body is allowed with any method.
Having said that, I would like to keep the way it's implemented now. However, if you have a strong feeling that I should remove it then ... I will consider it 😹

iagaponenko: I have also found that in some well known products, like `Elastic Search` parameters (queries to) the GET methods are always passed within the BODY. See: https://www.elastic.co/guide/en/elasticsearch/guide/current/_empty_search.html

iagaponenko: Okay, you have convinced me. Moved the parameters into the query string:
```
curl 'http://localhost:25080/ingest/chunks?database=sdss_stripe82_01' -X GET
```

andy-slac: Looks like you convinced yourself 🐈

PR #554, opened 2020-07-10
Added configuration parameter for xrootd "spread" in "czar".

iagaponenko:


admin/templates/configuration/etc/init.d/mysql-proxy
```diff
@@ -78,6 +78,12 @@ if [ "$USER" = "root" ]; then
     proxy_user_option="--user=$proxy_user"
 fi
 
+# This parameter is required for achieving the most optimal performance of
+# the xrootd client. The parameter setting will be moved into a configuration file
+# of the proxy when it would be possible (there will be XrootD API for that) to set
+# a desired value of the parameter programmatically.
+export XRD_PARALLELEVTLOOP=100
```
jgates108, line 85: Maybe stick a TODO in here so it is less likely to be forgotten.

iagaponenko: Good suggestion! Same thought crossed my mind. I will do so.

PR #560, opened 2020-07-30
Fix crash/hang in qhttp unit test teardown

fritzm:
This addresses the intermittent fail in the qhttp unit test.  Two changes were needed:

- check for acceptor shutting down condition in accept_handler chain and tail out appropriately
- add io_service::work in qhttp test fixture, so io_service thread doesn't exit between Server::stop() and Server::start()

### DIFF
--- a/core/modules/qhttp/Server.cc
+++ b/core/modules/qhttp/Server.cc
@@ -129,12 +129,15 @@ void Server::_accept()
     _acceptor.async_accept(
         *socket,
         [self, socket](boost::system::error_code const& ec) {
-            self->_accept(); // start accept for the next incoming connection
+            if (!self->_acceptor.is_open() || ec == asio::error::operation_aborted) {
+                return;
+            }
             if (!ec) {
                 boost::system::error_code ignore;
                 socket->set_option(ip::tcp::no_delay(true), ignore);
                 self->_readRequest(socket);
             }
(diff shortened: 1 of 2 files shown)


core/modules/qhttp/Server.cc
```diff
@@ -129,12 +129,15 @@ void Server::_accept()
     _acceptor.async_accept(
         *socket,
         [self, socket](boost::system::error_code const& ec) {
-            self->_accept(); // start accept for the next incoming connection
+            if (!self->_acceptor.is_open() || ec == asio::error::operation_aborted) {
```
iagaponenko, line 132: Just curious. Do you really need them (tests) both here? The documentation on the method recommends:
```
void accept_handler(const boost::system::error_code& error)
{
  if (!error)
  {
    // Accept succeeded.
  }
}
```
Which, I would assume implies that your may return from the handler if:
```
if (ec == asio::error::operation_aborted) return;
```

fritzm: In practice, both tests are, in fact, needed.  The first condition handles all situations related to acceptor close(); the second situation can and does occur in isolation when a previously closed acceptor is subsequently re-opened (a situation that occurs with the qhttp Server open()/close() protocol, but isn't well explored in the boost documentation, where most examples are single-shot).

PR #562, opened 2020-08-04
Tickets/dm 26034

iagaponenko:


core/modules/replica/IngestServerConnection.cc
```diff
@@ -430,7 +430,13 @@ void IngestServerConnection::_loadDataIntoTable() {
         string const sqlDatabase = h.conn->sqlId(_databaseInfo.name);
         string const sqlPartition = h.conn->sqlPartitionId(_transactionId);
 
-        vector<string> statements;
+        vector<string> tableMgtStatements;
+
+        // Make sure no outstanding table locks exist from prior operations
+        // on persistent database connections.
+        tableMgtStatements.push_back("UNLOCK TABLES");
```
andy-slac, line 437: Why do you need it, do you explicitly lock tables somewhere?

core/modules/replica/IngestServerConnection.cc
```diff
@@ -481,26 +486,63 @@ void IngestServerConnection::_loadDataIntoTable() {
...
+        LOGS(_log, LOG_LVL_DEBUG, context << __func__ << "  statement: " << dataLoadStatement);
+
+        // Allow retries for the table management statements in case of deadlocks.
+        // Deadlocks may happen when two or many threads are attempting to create
+        // or modify partitioned tables, or at a presence of other threads loading
+        // data into these tables.
+        //
+        // TODO: the experimental limit for the maximum number of retries may need
+        //       to be made unlimited, or be limited by some configurable timeout.
+        int const maxRetries = 1;
```
andy-slac, line 510: 3 may be better?

core/modules/replica/IngestServerConnection.cc
```diff
@@ -481,26 +486,63 @@ void IngestServerConnection::_loadDataIntoTable() {
...
             }
+        }
+
+        // Load table contribution
+        if (dataLoadStatement.empty()) {
+            throw std::runtime_error(context + string(__func__) + "  no data loading statement generated");
+        }
+        h.conn->execute([&dataLoadStatement](decltype(h.conn) const& conn_) {
+            conn_->begin();
+            conn_->execute(dataLoadStatement);
```
andy-slac, line 542: Could this deadlock with the DDL ALTER TABLE statements? OR is it always DDL that get that error?

core/modules/replica/IngestServerConnection.cc
iagaponenko: I have a persistent connection pool for MySQL. This statements is to ensure no remaining locks are left on the connection from a previous client.

core/modules/replica/IngestServerConnection.cc
iagaponenko: I haven't see this in the data loading statements. It's always the ALTER TABLE statement which is giving me troubles.

core/modules/replica/IngestServerConnection.cc
iagaponenko: With 1 the problems seems to be gone. I would like to keep it this way to to get some stats on the problem should it happen again. Then I would have to look at more sophisticated lock avoidance techniques.

PR #563, opened 2020-08-06
Fix negative time-remaining bug in qhttp ajax unit test

fritzm:


fritzm:
Thanks, Igor.

> It appears to me there is a possibility that the body of the while loop would never get a chance to be executed. This could happen if a process would get preempted for a duration of time longer than nmsec. I that case none of the handlers would get invoked. Is this okay?

For the purpose of these unit tests, yes.  If you are passing test durations that small, timing out of this loop would be the expected and acceptable result.

Re. cast not needed for `BOOST_TEST` vs. `BOOST_EQUAL` -- interesting...  I assume I had the cast there in the first place because of a compiler complaint.  I wonder if I had changed from `BOOST_EQUAL` to `BOOST_TEST` at some point?

PR #564, opened 2020-08-10
DM-26100: Add application for KPM testing

andy-slac:
README contains more or less complete description of how to run it.

### DIFF
--- a/admin/python/lsst/qserv/testing/README.md
+++ b/admin/python/lsst/qserv/testing/README.md
@@ -161,8 +161,8 @@ option `--slurm`. When this option is passed it uses slurm environment
 variables `SLURM_NTASKS` and `SLURM_NODEID` (see
 https://slurm.schedmd.com/srun.html). For this to work correctly one has to
 make sure that slurm allocates whole node per task, this can be done in
-different ways, one of then is to pass `-n` and `-N` options with the same
-values to sbatch/srun.
+different ways, one of them is to pass `--ntasks` and `--nodes` options with
+the same values to sbatch/srun.
 
 Here is an example of a script for sbatch command:
 
--- a/admin/python/lsst/qserv/testing/README.md
+++ b/admin/python/lsst/qserv/testing/README.md
@@ -1,0 +1,177 @@
+
+# Qserv KPM test harness
+
+This package contains Python code and configuration examples for KPM testing
+of QServ cluster. It is used in implementation of command line application
+`qserv-kraken` (lives in `admin/bin/`) which is the main tool for KPM testing.
+Folder `config` in this package contains an example of the configuration file
+that can be used for running `qserv-kraken`.
+
+`qserv-kraken` runs multiple "clients", each client generating queries of
+specific query class and executing them sequentially. Total number of clients
+is determined by configuration (sum of all `concurrentQueries` values), and it
+is typically around 100. Every client in this implementation runs in a
+separate sub-process, all those sub-processes are started and managed by
+`qserv-kraken`. Depending on hardware it may be better to run the whole
+shebang on several hosts, `qserv-kraken` provides simple way to do that.
+
+
+## Configuration
+
+The queries that need to be executed, number of clients and few other
+parameters are specified in the YAML configuration file, here is the
+description of its contents.
+
+At the top level config file has two keys:
+
+- `queryClasses` -- defines execution policies for different classes of
+  queries,
+- `queries` -- defines queries to be executed.
+
+### `queryClasses` tree
+
+The next level in `queryClasses` defines a bunch of keys corresponding to
+query classes ("LV", "NearN", etc.)
+
+For each query class next level of the tree defines parameters for query
+execution with these keys:
+
+- `concurrentQueries` -- (`int`) number of concurrent queries to run, which is
+  also a number of sub-processes that will be executing them;
+- `targetTime` -- target time for query execution, this is not used currently;
+- `maxRate` -- (`float`, optional) maximum query submission rate in Hz for
+  queries of this class for single client, if not specified or `null` then
+  rate is not limited;
+- `arraysize` -- (`int`, optional) number of result rows to fetch at once with
+  `fetchmany()` DBAPI call, if not specified or `null` then driver default is
+  used.
+
+### `queries` tree
+
+The next level in `queries` tree specifies keys for query classes. These are
+the same classes as defined in `queryClasses` tree, but it is not required to
+be a full set of classes. A subset of the class names can be used in `queries`
+tree and only these classes will be selected for execution.
+
+Next level in `queries` tree contains query definitions. Each query defined
+there has a unique key which can be an arbitrary name (e.g. "q1", "q42"). This
+key is used to identify queries in the log and in monitoring output. The value
+for the query definition is either a fixed query string (e.g. "SELECT * FROM
+Object") or a definition of a template used to generate queries. Template
+definition has these keys:
+
+- `template` -- (`str`) text of a query which include template variables
+  enclosed into curly braces ("SELECT ... WHERE ra > {ra_min}");
+- `variables` -- definition of one or more template variables.
+
+In the definition of template variables the name of a variable is a key, and
+the value is a configuration of how to generate values for this variable.
+Currently there are two supported specifications:
+
+- Random choice of integer numbers that are read from a file, this is going to
+  be used for random objectIds. This specification has configuration with two
+  keys:
+  - `path` -- (`str`) the name of a file to read numbers from, this is a text
+    files with a bunch of numbers separated by spaces and/or newlines.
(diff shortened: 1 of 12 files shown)


admin/python/lsst/qserv/testing/README.md
```diff
@@ -0,0 +1,164 @@
+
+# Qserv KPM test harness
+
+This package contains Python code and configuration examples for PM testing of
+QServ cluster. Main application is `qserv-kraken` (lives in `admin/bin/`) and
```
n8pease, line 5: I'm probably lame but I don't know what PM is. Avoid acronyms in readme?

admin/python/lsst/qserv/testing/README.md
```diff
@@ -0,0 +1,164 @@
...
+  also a number of sub-processes that will be executing them;
+- `targetTime` -- target time for query execution, this is not used currently;
+- `maxRate` -- (`float`, optional) maximum query submission rate in Hz for
+  queries of this class for single client, if not specified or `null` then
+  rate is not limited;
+- `arraysize` -- (`int`, optional) number of result rows to fetch at once with
+  `fetchmany()` DBAPI call, if not specified or `null` then driver default is
+  used.
+
+For `queries` tree the level below query classes contain query definitions.
```
n8pease, line 46: to me, "level below" makes it sound like queries is within queryClasses, but around line 26 it sounds like query and queryClasses are peers. I'm not sure which is right or if I'm misunderstanding in general, but I'm confused by this wording.

admin/python/lsst/qserv/testing/README.md
n8pease: and again, probably I'm dense but maybe it's worth saying "Main application is a command line application called `qserv-kraken`..." I only realized on line 84 that it's a CLI app 🙃

admin/python/lsst/qserv/testing/config.py
```diff
@@ -0,0 +1,299 @@
...
+
+    Parameters
+    ----------
+    txt : `str`
+        If ``variables`` is ``None`` then full query text, otherwise it is a
+        query template.
+    variables : `dict` [`str`, `dict`]
+        Dictionary whose keys are variable names and values are dictionaries
+        with a description of how to generate variable value.
+    """
```
n8pease, line 84: needs Returns doc

admin/python/lsst/qserv/testing/config.py
```diff
@@ -0,0 +1,299 @@
...
+                        generator = _ValueRandomUniformInt(min, max)
+                elif "path" in config:
+                    path = config["path"]
+                    mode = config.get("mode", "random")
+                    generator = _ValueIntFromFile(path, mode)
+                if generator is None:
+                    raise ValueError(f"Cannot parse variable configuration {var} = {config}")
+                self._vars[var] = generator
+
+    def query(self):
```
n8pease, line 109: pls add docstring with Returns doc

n8pease: my specific q is, will the return always be a dict? if it returns self._txt could it be a string that is not in a container?

admin/python/lsst/qserv/testing/config.py
```diff
@@ -0,0 +1,299 @@
...
+
+        if not configs:
+            raise ValueError("empty configurations list")
+
+        # merge them
+        self._config = {}
+        for config in configs:
+            self._config = self._merge(self._config, config)
+
+        classes = set(self._config.get("queryClasses", {}).keys())
```
n8pease, line 137: Here I'm personally right at the edge of my tolerance for nested containers and functional manipulation of them. IMO it may be time now or soon to consider writing a proper config class.

admin/python/lsst/qserv/testing/config.py
```diff
@@ -0,0 +1,299 @@
...
+
+    @classmethod
+    def from_yaml(cls, config_files):
+        """Make configuration from bunch of YAML files
+
+        Parameters
+        ----------
+        config_files : `list` [`str` or file]
+            List of YAML sources, file name or open file object is acceptable.
+        """
```
n8pease, line 167: needs Returns section

admin/python/lsst/qserv/testing/config.py
n8pease: ...well, the api to this class is good so 🤷.

admin/python/lsst/qserv/testing/main.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+def _logConfig(level, slot):
+    levels = {0: logging.WARNING, 1: logging.INFO, 2: logging.DEBUG}
+    if slot is None:
+        simple_format = "%(asctime)s %(levelname)7s %(name)s -- %(message)s"
+    else:
+        simple_format = f"%(asctime)s %(levelname)7s %(name)s [slot:{slot}] -- %(message)s"
+    level = levels.get(level, logging.DEBUG)
+    logging.basicConfig(format=simple_format, level=level)
```
n8pease, line 29: `basicConfig`'s `level` can take a string (tho you have to validate inputs or handle exception if it's wrong)

admin/python/lsst/qserv/testing/main.py
```diff
@@ -0,0 +1,146 @@
...
+            args.slot = int(slot)
+
+    _logConfig(args.verbose, args.slot)
+
+    # build/split config
+    cfg = Config.from_yaml(args.config)
+    if args.num_slots is not None:
+        cfg = cfg.split(args.num_slots, args.slot)
+    if args.dump_config:
+        print("Configuration for this process (n_slots={args.num_slots}, slot={args.slot}):")
```
n8pease, line 106: is this supposed to be an f string?

admin/python/lsst/qserv/testing/main.py
```diff
@@ -0,0 +1,146 @@
...
+            host=args.host,
+            port=args.port,
+            user=args.user,
+            passwd=args.password,
+            db=args.db
+        )
+
+    # monitor
+    tags = None if args.slot is None else {"slot": args.slot}
+    monitor = None
```
n8pease, line 124: AFAIK you don't have to predeclare `montior`.

admin/tests/test_testing.py
```diff
@@ -0,0 +1,436 @@
...
+        rows = cursor.fetchmany(1000)
+        self.assertEqual(len(rows), 22)
+        rows = cursor.fetchmany(1000)
+        self.assertEqual(len(rows), 0)
+
+
+class TestQueryRunner(unittest.TestCase):
+
+    def test_query_runner_count(self):
+
```
n8pease, line 279: could use a docstring: what is getting tested here? I can't see any verification.

n8pease: same for the other test function in this class and the test functions in TestRunnerManager.

admin/python/lsst/qserv/testing/README.md
andy-slac: PM is KPM with missing K. Sorry for that, but my spell checker was OK with PM (it probably assumed I'm talking about Prime Minister or something) 😄 I guess we all should know what KPM is? 

I'll expand on "main application".

admin/python/lsst/qserv/testing/README.md
andy-slac: "queries" and "queryClasses" are two separate configuration trees at the top level, but each of them have query classes as keys at one level. I'll try to fix poor wording.

admin/python/lsst/qserv/testing/config.py
andy-slac: It always returns string - full text of the query. We interpolate query variables directly into the query text, not passing them to DBAPI level. Potentially we could pass variables as separate dictionary, for MySQL (or for how we uset) that does not really matter.

admin/python/lsst/qserv/testing/config.py
andy-slac: "The proper config class" is this class itself. We need to read data from YAML in any case and this is what this class does. Clients use API which does not expose this detail.

admin/python/lsst/qserv/testing/main.py
andy-slac: Hmm, that looks like an undocumented feature.

admin/python/lsst/qserv/testing/main.py
andy-slac: Indeed.

admin/python/lsst/qserv/testing/main.py
andy-slac: Well, true for now, but if/elif below does not cover 100% of space, I want to be defensive in case this gets extended later.

admin/tests/test_testing.py
andy-slac: 👍  This test is for that it does not run forever and eventually stops. I'll add a comment.

PR #565, opened 2020-08-14
Fixed a bug in the Replication System's Messenger

iagaponenko:


fritzm:
LGTM.  Some feedback from Slack conversation:

- consider using error-code versions of asio calls instead of exception-throwing versions, for greater robustness in face of unexpected errors (particularly for calls like close and cancel where its clear you just want to move on anyway)

- Please double-check to make sure there aren't cases where outstanding operations now _aren't_ cancelled in STATE_CONNECTING (these may blow up later of the socket is closed out from under them without cancelling them out)

PR #574, opened 2020-11-04
Tickets/dm 27400

iagaponenko:


core/modules/replica/DatabaseMySQL.cc
```diff
@@ -599,6 +622,20 @@ void Connection::_connectOnce() {
...
+    // This is required by 'LOAD DATA LOCAL INFILE ...' to allow ingesting files which
+    // are not directly seen by the MySQL server. The 'LOCAL' option would make a file
+    // local to a client opening this connection to be transferred to some temporary
+    // directory owned by the server. After that the file will get ingestd into
+    // the destination table.
+    //
+    // NOTES:
+    // - The server must be configured with global variable 'local_infile=1'
+    // - The temporary folder managed by the server is required to have enough space
+    //   to accommodate files received from the client.
```
n8pease, line 634: It seems like these notes would be worth putting in a header or in documentation (if they are duplicated here that's ok, but if they're only here they may be hard to find?)

n8pease:
one minor comment. looks fine.

core/modules/replica/DatabaseMySQL.cc
iagaponenko: Good point! I will put this info into the class's header.

PR #576, opened 2020-11-05
Tickets/dm 27334

jgates108:


core/modules/qdisp/QdispPool.cc
```diff
@@ -228,6 +228,8 @@ QdispPool::QdispPool(int poolSize, int largestPriority, std::vector<int> const&
         LOGS(_log, LOG_LVL_INFO, "creating priQ pri=" << pri << " min=" << minRun << " max=" << maxRun);
         _prQueue->addPriQueue(pri, minRun, maxRun);
     }
+    // This pool does not kick threads out when they take time (but little CPU) to process,
+    // so maxPoolThreads is just slightly larger than poolSize.
```
iagaponenko, line 232: What's a definition of "slightly"? Could it be expressed in a fraction of the {poolSize}}, or in a fixed number? Is there any chance thread allocation may "go wild" and exceed all threads available to a process (which is a little less than 30k)?

core/modules/util/ThreadPool.cc
```diff
@@ -81,6 +83,11 @@ bool PoolEventThread::leavePool(Command::Ptr const& cmd) {
     // This thread will stop accepting commands
     _loop = false;
     LOGS(_log, LOG_LVL_DEBUG, "PoolEventThread::leavePool " << this);
+    if (_threadPool->atMaxThreadPoolCount()) {
+        LOGS(_log, LOG_LVL_WARN, "A new thread cannot be created immediately as there are too many threads");
+        return false;
+    }
+
```
iagaponenko, line 86: This test is counterintuitive as it's done inside method which I think is supposed to delete the thread. Or, I'm mistaken?

My second comment is that. the Doxygen comment before tis method says:
```
// @return false if a different command is running than cmd.
```
The `if` block returns `false`. Is documentation consistent with the behavior of the method?

core/modules/util/ThreadPool.cc
```diff
@@ -137,17 +145,25 @@ PoolEventThread::Ptr CommandThreadPool::getAndNullPoolEventThread() {
 }
 
 
+ThreadPool::Ptr ThreadPool::newThreadPool(unsigned int thrdCount, unsigned int maxThreadCount,
+                                          CommandQueue::Ptr const& q, EventThreadJoiner::Ptr const& joiner) {
+    Ptr thp(new ThreadPool(thrdCount, maxThreadCount, q, joiner)); // private constructor
+    thp->_resize();
```
iagaponenko, line 159: What's a reason not to resize the thread pool in the constructor?

core/modules/util/ThreadPool.cc
```diff
@@ -137,17 +145,25 @@ PoolEventThread::Ptr CommandThreadPool::getAndNullPoolEventThread() {
...
+    return thp;
+}
+
+
 ThreadPool::Ptr ThreadPool::newThreadPool(unsigned int thrdCount,
-      CommandQueue::Ptr const& q, EventThreadJoiner::Ptr const& joiner) {
-    Ptr thp(new ThreadPool(thrdCount, q, joiner)); // private constructor
+                                          CommandQueue::Ptr const& q, EventThreadJoiner::Ptr const& joiner) {
+    Ptr thp(new ThreadPool(thrdCount, thrdCount+1, q, joiner)); // private constructor
```
iagaponenko, line 166: A choice of having the second parameter incremented by 1 could be explained in a comment. It's not clear to a reader why it's +1 and not +2, etc.

core/modules/util/ThreadPool.cc
```diff
@@ -204,6 +220,21 @@ bool ThreadPool::release(PoolEventThread *thrd) {
...
+            std::lock_guard<std::mutex> lockPool(_mxPool);
+            /// This is not expected to be used in cases where low CPU usage threads
+            /// are removed from the pool. In those cases, the targetThrdCount is expected to be
+            /// less than 100 while _maxThreadCount would be several thousand.
+            if (targetThrdCount >= _maxThreadCount) {
+                LOGS(_log, LOG_LVL_WARN, "ThreadPool::resize target count " << targetThrdCount
+                        << " is >= than max " << _maxThreadCount);
+                // Need at least one thread available to deal with released threads.
+                _maxThreadCount = targetThrdCount + 1;
+                _cvPool.notify_all();
```
iagaponenko, line 242: Why notifying all threads instead `notify_one` if increasing the thread count by 1?

core/modules/util/ThreadPool.cc
```diff
@@ -204,6 +220,21 @@ bool ThreadPool::release(PoolEventThread *thrd) {
 /// Change the size of the thread pool.
 void ThreadPool::resize(unsigned int targetThrdCount) {
     {
+        LOGS(_log, LOG_LVL_WARN, "&&&WARN ThreadPool::resize " << targetThrdCount);
```
iagaponenko: Since `resize` is a normal  and probably rather frequent operation then reporting it at level `WARN` is may be a bit of an "over-kill" for the production code. Perhaps `DEBUG` or even `TRACE` would still work?

core/modules/util/ThreadPool.cc
```diff
@@ -258,4 +289,35 @@ void ThreadPool::waitForResize(int millisecs) {
     }
 }
 
+void ThreadPool::_incrPoolThreadCount() {
+    std::lock_guard<std::mutex> lockPool(_mxPool);
+    ++_poolThreadCount;
+    LOGS(_log, LOG_LVL_WARN, "&&&WARN incr _poolThreadCount=" << _poolThreadCount);
```
iagaponenko: Since `_incrPoolThreadCount` is a normal  and probably rather frequent operation then reporting it at level `WARN` is may be a bit of an "over-kill" for the production code. Perhaps `DEBUG` or even `TRACE` would still work?

core/modules/util/ThreadPool.cc
```diff
@@ -258,4 +289,35 @@ void ThreadPool::waitForResize(int millisecs) {
...
+}
+
+void ThreadPool::_decrPoolThreadCount() {
+    {
+        std::lock_guard<std::mutex> lockPool(_mxPool);
+        --_poolThreadCount;
+    }
+    LOGS(_log, LOG_LVL_WARN, "&&&WARN decr _poolThreadCount=" << _poolThreadCount);
```
iagaponenko: Since `_decrPoolThreadCount ` is a normal  and probably rather frequent operation then reporting it at level `WARN` is may be a bit of an "over-kill" for the production code. Perhaps `DEBUG` or even `TRACE` would still work?

core/modules/util/ThreadPool.cc
```diff
@@ -258,4 +289,35 @@ void ThreadPool::waitForResize(int millisecs) {
...
+    _cvPool.notify_all(); // There should not be many threads waiting.
```
iagaponenko, line 312: Why notifying all after changing thread size by 1?

core/modules/util/ThreadPool.cc
```diff
@@ -258,4 +289,35 @@ void ThreadPool::waitForResize(int millisecs) {
...
+}
+
+void ThreadPool::_waitIfAtMaxThreadPoolCount() {
+    std::unique_lock<std::mutex> lockPool(_mxPool);
+    LOGS(_log, LOG_LVL_WARN, "&&&WARN wait A _poolThreadCount=" << _poolThreadCount);
+    _cvPool.wait(lockPool, [this](){return (_poolThreadCount <= _maxThreadCount);});
+    LOGS(_log, LOG_LVL_WARN, "&&&WARN wait B _poolThreadCount=" << _poolThreadCount);
+}
```
iagaponenko, line 323: Same comments as before on reporting as `WARN` instead of `DEBUG` or `TRACE`.

core/modules/util/ThreadPool.cc
```diff
@@ -258,4 +289,35 @@ void ThreadPool::waitForResize(int millisecs) {
...
+
+bool ThreadPool::atMaxThreadPoolCount() {
+    std::unique_lock<std::mutex> lockPool(_mxPool);
+    bool atMax = _poolThreadCount > _maxThreadCount;
+    if (atMax) {
+        LOGS(_log, LOG_LVL_WARN, "atMaxThreadPoolCount current=" << _poolThreadCount << " max=" << _maxThreadCount);
```
iagaponenko, line 329: Same comments as before on reporting as `WARN` instead of `DEBUG` or `TRACE`.

core/modules/util/ThreadPool.h
```diff
@@ -135,6 +158,26 @@ class ThreadPool : public std::enable_shared_from_this<ThreadPool> {
...
+    void _waitIfAtMaxThreadPoolCount();
+
+    /// Increase the count of existing threads.
+    void _incrPoolThreadCount();
+
+    /// Decrease the count of existing threads.
+    void _decrPoolThreadCount();
+
+    unsigned int _poolThreadCount = 0;    ///< Number of threads that exist.
+    unsigned int _maxThreadCount = 20000; ///< Max number of thread allowed &&& use config file value
```
iagaponenko: This fixed number is not consistent with the default set to `10000` in the config file's parameter:
```
max_pool_threads = 10000
```

core/modules/util/testEventThread.cc
```diff
@@ -163,10 +163,10 @@ BOOST_AUTO_TEST_CASE(EventThreadTest) {
         std::mutex goCVMtx;
         // Create more threads than can fit in the pool and don't let any complete.
         // Since they all leave the pool (via peThread->leavePool();), they should all run.
-        int threadsRunning = 2*sz;
+        int threadsRunning = sz*2;
```
iagaponenko, line 166: Would this double the maximum number of threads (right now it's `10000`) specified in the worker's configuration file?

core/modules/util/ThreadPool.cc
```diff
@@ -300,23 +300,26 @@ void ThreadPool::waitForResize(int millisecs) {
...
-    LOGS(_log, LOG_LVL_WARN, "&&&WARN decr _poolThreadCount=" << _poolThreadCount);
+    LOGS(_log, LOG_LVL_DEBUG, "decr _poolThreadCount=" << _poolThreadCount);
     _cvPool.notify_all(); // There should not be many threads waiting.
 }
 
 void ThreadPool::_waitIfAtMaxThreadPoolCount() {
     std::unique_lock<std::mutex> lockPool(_mxPool);
-    LOGS(_log, LOG_LVL_WARN, "&&&WARN wait A _poolThreadCount=" << _poolThreadCount);
+    auto logLvl = LOG_LVL_DEBUG;
+    if (_poolThreadCount >= _maxThreadCount - 10) {
```
iagaponenko, line 318: Why this fixed number `10` here? Does it have any specific meaning?
There is also a chance of getting into an integer overflow since is defined as:
```
unsigned int _maxThreadCount;
```
This would happed if, for example:
```
_maxThreadCount = 9;
```

iagaponenko:
I'm approving this code since it has been tested on the large scale, where it was proven to work. However, I have left a few comments and questions. Some of those may arise from my superficial understanding on the design and behavior of `ThreadPool`.

core/modules/util/ThreadPool.cc
jgates108: This just has to do with leaving a log message when approaching the limit. If max thread count is 9, all the messages will be warnings instead of debug. I'll just remove it.

core/modules/util/ThreadPool.cc
jgates108: I think it is important to know that the system has run out of threads for the pool. Currently, this should be a rare occurrence as is really not easy to test under load. If it isn't a rare thing, we need to know that too.

core/modules/util/testEventThread.cc
jgates108: This is a unit test, a pretty complicated one at that, but it should have no effect on the czar or workers.

PR #578, opened 2020-11-16
Tickets/dm 27258

iagaponenko:


core/modules/qhttp/Server.cc
```diff
@@ -206,7 +206,9 @@ void Server::_readRequest(std::shared_ptr<ip::tcp::socket> socket)
                 request->_parseHeader();
                 request->_parseUri();
                 if (request->version == "HTTP/1.1") {
-                    *reuseSocket = true;
+                    // Temporary disable this option due to a bug in the implementation
```
fritzm, line 209: Could you please put a reference to the associated JIRA ticket in the comment here, as well?  That will make it easier to find more technical detail if reading the code later.

fritzm (approved):

fritzm (approved):

core/modules/qhttp/Server.cc
iagaponenko: Here is the one: https://jira.lsstcorp.org/browse/DM-27396

PR #579, opened 2020-11-16
Tickets/dm 27504

iagaponenko:


fjammes (approved):
LGTM, impressive C++ code!

core/modules/replica/HttpModuleBase.cc
```diff
@@ -51,12 +51,10 @@ HttpModuleBase::HttpModuleBase(string const& authKey,
         _req(req),
         _resp(resp),
         _query(req->query) {
-    LOGS(_log, LOG_LVL_INFO, "HttpModuleBase::" << __func__);
```
fjammes, line 54: :+1:

core/modules/replica/HttpIngestConfigModule.cc
```diff
@@ -0,0 +1,179 @@
...
+    // Ignore missing parameters if they were not set.
+
+    json result({{"database", databaseInfo.name}});
+
+    try {
+        result[HttpFileReaderConfig::sslVerifyPeerKey] = stoi(databaseServices->ingestParam(
+                databaseInfo.name,
+                HttpFileReaderConfig::category,
+                HttpFileReaderConfig::sslVerifyPeerKey).value);
+    } catch (DatabaseServicesNotFound const&) {}
```
andy-slac, line 91: If missing parameter is not an exceptional case then raising an exception is potentially a problem. Maybe return a default value in that case would work better?

core/modules/replica/IngestHttpSvcMod.cc
```diff
@@ -155,19 +167,91 @@ json IngestHttpSvcMod::_readLocal(string const& filename) {
...
+    HttpFileReader reader(method, url, data, headers, sslVerifyPeer, caPath, caInfo);
+    reader.read([this, &numBytes, &numRows](string const& row) {
         this->writeRowIntoFile(row);
         numBytes += row.size() + 1;     // counting the newline character
         ++numRows;
     });
+    if (!caInfoTmp.empty()) {
+        // Make the best effort to delete the file. Ignore any errors.
+        boost::system::error_code ec;
+        fs::remove(fs::path(caInfoTmp), ec);
```
andy-slac, line 253: IF exception happens in `reader.read()` then file is not removed? I'd move all code that creates/removes temporary file into a separate class and use RAII to manage its lifetime.

core/modules/replica/IngestHttpSvcMod.cc
```diff
@@ -155,19 +167,91 @@ json IngestHttpSvcMod::_readLocal(string const& filename) {
...
+                    HttpFileReaderConfig::caPathKey).value;
+        } catch (DatabaseServicesNotFound const&) {}
+
+        // Get a path to an existing file with peer certificate, if any was configured
+        // for this catalog.
+        try {
+            caInfo = databaseServices->ingestParam(
+                    database,
+                    HttpFileReaderConfig::category,
+                    HttpFileReaderConfig::caInfoKey).value;
```
andy-slac, line 209: This is ignored if `caInfoValKey` exists, maybe reverse the order and not try to retrieve it if `caInfoValKey` is there, will save one trip to database?

core/modules/replica/HttpFileReaderApp.h
```diff
@@ -0,0 +1,78 @@
...
+
+// This header declarations
+namespace lsst {
+namespace qserv {
+namespace replica {
+
+/**
+ * Class HttpFileReaderApp implements a tool that reads files from an object store
+ * over the HTTP/HTTPS protocol. If option '--out=<file>' is present the file's content
+ * will be writted into the specified file. Otherwise the content will be printed onto
```
andy-slac, line 35: Typo: writted 

printed onto -> printed to

core/modules/replica/HttpFileReaderApp.cc
```diff
@@ -0,0 +1,144 @@
...
+        "The HTTP header to be sent with a request. Note this application allows"
+        " only one header.",
+        _header
+    ).option(
+        "data",
+        "The data to be sent in the body of a request.",
+        _data
+    ).flag(
+        "no-ssl-verify-peer",
+        "The flag that would disable verifying the peer's SSL certificate.",
```
andy-slac, line 84: would disable -> disables

core/modules/replica/HttpFileReaderApp.cc
```diff
@@ -0,0 +1,144 @@
...
+        _caInfo
+    ).option(
+        "file",
+        "A path to an output file where the content received from a remote source will be written."
+        "  If the option is not specified then the content will be printed onto the standard output"
+        " stream. This option is ignored if flag --silent is specified.",
+        _file
+    ).flag(
+        "silent",
+        "The flag that would disable printing or writing the content received from a remote"
```
andy-slac, line 104: would disable -> disables

core/modules/replica/HttpFileReaderApp.cc
```diff
@@ -0,0 +1,144 @@
...
+        _caPath
+    ).option(
+        "ca-info",
+        "A path to a Certificate Authority (CA) bundle to verify the peer with."
+        " This option is ignored if flag --no-ssl-verify-peer is specified.",
+        _caInfo
+    ).option(
+        "file",
+        "A path to an output file where the content received from a remote source will be written."
+        "  If the option is not specified then the content will be printed onto the standard output"
```
andy-slac, line 99: printed onto -> printed to

core/modules/replica/HttpFileReaderApp.cc
```diff
@@ -0,0 +1,144 @@
...
+        " source.",
+        _silent
+    );
+}
+
+
+int HttpFileReaderApp::runImpl() {
+    string const context = "HttpFileReaderApp::" + string(__func__) + "  ";
+    vector<string> const allowedMethods = {"GET", "POST", "PUT", "DELETE"};
+    if (allowedMethods.cend() == find(allowedMethods.cbegin(), allowedMethods.cend(), _method)) {
```
andy-slac, line 114: I think Python version of this would be shorter :)
In C++ you don't need `vector` to use algorithms, some thing simpler and more efficient like this would do:
```
    const char* allowedMethods[] = {"GET", "POST", "PUT", "DELETE"};
    if (std::count(std::cbegin(allowedMethods), std::cend(allowedMethods), _method) == 0) {
    }
```

andy-slac (approved):
Looks OK, few comments.

core/modules/replica/HttpFileReader.cc
```diff
@@ -81,6 +87,16 @@ void HttpFileReader::read(CallbackType const& onEachLine) {
...
     _errorChecked("curl_easy_setopt(CURLOPT_HTTPHEADER)", curl_easy_setopt(_hcurl, CURLOPT_HTTPHEADER, _hlist));
+    if (_sslVerifyPeer) {
+        if (!_caPath.empty()) {
+            _errorChecked("curl_easy_setopt(CURLOPT_CAPATH)", curl_easy_setopt(_hcurl, CURLOPT_CAPATH, _caPath.c_str()));
+        }
+        if (!_caInfo.empty()) {
+            _errorChecked("curl_easy_setopt(CURLOPT_CAINFO)", curl_easy_setopt(_hcurl, CURLOPT_CAINFO, _caInfo.c_str()));
+        }
+    } else {
+        _errorChecked("curl_easy_setopt(CURLOPT_SSL_VERIFYPEER)", curl_easy_setopt(_hcurl, CURLOPT_SSL_VERIFYPEER, 0L));
```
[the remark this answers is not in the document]
iagaponenko: Disabling `CURLOPT_SSL_VERIFYHOST` looks like a dangerous option to me. According to the documentation on `libcurl`:
* disabling `CURLOPT_SSL_VERIFYHOST` will disable checking if a host's cert belong to that host
* while, disabling `CURLOPT_SSL_VERIFYPEER`will disable checking the authenticity of the host's cert.
How about adding an additional option for `CURLOPT_SSL_VERIFYHOST`?
I would expect that the right logic (on the client's side) here would be:
```
if CURLOPT_SSL_VERIFYHOST is disabled then
  disable CURLOPT_SSL_VERIFYPEER
endif
  ...
```
I can add a configuration option for `CURLOPT_PROXY_SSL_VERIFYPEER` as well.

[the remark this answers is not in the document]
andy-slac: I believe `curl -k` disables both CURLOPT_SSL_VERIFYPEER and CURLOPT_SSL_VERIFYHOST and I though that this is the behavior you want to reproduce.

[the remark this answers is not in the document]
andy-slac: And CURLOPT_SSL_VERIFYHOST is not for host certificate (I do not know what is that), it is checking that host name in service cert is the same as the host name you are connecting to. I think one common use case is when the same cert is used for services running on multiple hosts, so their names do not match.

core/modules/replica/DatabaseServicesMySQL.cc
```diff
@@ -2230,6 +2230,7 @@ void DatabaseServicesMySQL::saveIngestParam(string const& database,
                         conn->sqlEqual("database", database) + " AND " + 
                         conn->sqlEqual("category", category) + " AND " +
                         conn->sqlEqual("param", param);
+                conn->rollback();
```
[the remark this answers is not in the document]
iagaponenko: This bug fix was required because it's done in the following context:
```
        _conn->executeInsertOrUpdate(
            [&](decltype(_conn) conn) {
                conn->begin();
                conn->executeInsertQuery(...);
                conn->commit();
            },
            [&](decltype(_conn) conn) {
                conn->rollback();
                conn->begin();
                conn->executeSimpleUpdateQuery(...);
                conn->commit();
            }
        );
```
Method `executeInsertOrUpdate` requires a user to provide 2 `lambda` functions, the first one for attempting to insert something into a database, the second one - is a failover to update something which already exist in the database instead of inserting. When executing the first lambda, the operation `executeInsertOrUpdate` will catch the `DuplicateKey` exception, and if the one was thrown then it would invoke the second (`update`) lambda. Note that the method requires a user to do his/her own transaction management.That was a reason behind adding the `rollback` to cleanup anything which might be modified in the database by the `insert` lambda.  Otherwise `conn->begin()` in the `update` lambda would throw a "logic error" exception.

Note that `executeInsertOrUpdate` would also do automatic reconnects to MySQL should the connection be lost. This would also invalidate any transactions. That's the whole point in handing lambdas instead of single queries to that method.

core/modules/replica/HttpIngestConfigModule.cc
```diff
@@ -0,0 +1,179 @@
...
+                HttpFileReaderConfig::caInfoKey).value;
+    } catch (DatabaseServicesNotFound const&) {}
+
+    try {
+        result[HttpFileReaderConfig::caInfoValKey] = databaseServices->ingestParam(
+                databaseInfo.name,
+                HttpFileReaderConfig::category,
+                HttpFileReaderConfig::caInfoValKey).value;
+    } catch (DatabaseServicesNotFound const&) {}
+
```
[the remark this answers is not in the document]
iagaponenko: I too thought about this ... initially before I had to address type conversion in:
```
        result[HttpFileReaderConfig::sslVerifyPeerKey] = stoi(databaseServices->ingestParam(
                databaseInfo.name,
                HttpFileReaderConfig::category,
                HttpFileReaderConfig::sslVerifyPeerKey).value);
```
I also thought about adding a class template. Unfortunately, that would add more code than I have now.
On the other hand, if I were two add two more options `CURLOPT_SSL_VERIFYHOST` and `CURLOPT_PROXY_SSL_VERIFYPEER `, like it was discussed  earlier, then adding the class template would be a reasonable solution.

core/modules/replica/HttpIngestConfigModule.cc
```diff
@@ -0,0 +1,179 @@
...
+                caInfo);
+    }
+    if (hasCaInfoVal) {
+        debug(__func__, HttpFileReaderConfig::caInfoValKey + "=" + caInfoVal);
+        databaseServices->saveIngestParam(
+                databaseInfo.name,
+                HttpFileReaderConfig::category,
+                HttpFileReaderConfig::caInfoValKey,
+                caInfoVal);
+    }
```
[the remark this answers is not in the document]
iagaponenko: It's due to the same requirements to do type conversion into `int` for one of the keys. I will investigate a possibility of using class template to address this.

core/modules/replica/HttpIngestConfigModule.cc
iagaponenko: This would violate to the expected contract of the REST service which is expected to return only those parameters that were explicitly set by a user, not any defaults. The very same logic (to do only what user explicitly asked for) is been following in the rest of the code, including a code configuring `libcurl` in the worker ingest service.
So, I politely refuse this suggestion 😸

core/modules/replica/IngestHttpSvcMod.cc
iagaponenko: Good suggestion! I too thought about it. Though, there is nothing wrong in leaving those weirdly (uniquely) named tiny files in the temporary filesystem. Then file system could be cleaned up when the server would be starting next time.

core/modules/replica/HttpFileReaderApp.cc
iagaponenko: 👍

PR #582, opened 2020-11-25
Tickets/dm 27681

iagaponenko:


core/modules/replica/NamedMutexRegistry.h
```diff
@@ -0,0 +1,69 @@
...
+     * @param name The name of a mutex.
+     * @return A reference to the mutex.
+     * @throw std::invalid_argument If the name is empty.
+     */
+    util::Mutex& get(std::string const& name);
+
+private:
+    std::map<std::string, std::unique_ptr<util::Mutex>> _registry;
+    util::Mutex _registryAccessMtx;
+};
```
jgates108, line 65: It looks like the named mutexes are added to the map and not removed until NamedMutexRegistry is destroyed, and it looks like it last until the program is shutdown. There could be a few hundred thousand of these if there's one per table per chunk (if 'Object_231' requires a different mutex from 'Object_150000'). Occasionally calling a cleanup function (say after a transaction finishes) and deleting all mutexes with only one reference may be worth it. I think that requires the use of shared pointers and passing shared pointers to callers and it becomes very important that the shared pointer lives longer than the lock, which may be tricky.

Another, probably much better, option would be to limit the lifetime of the NamedMutexRegistry by making one per database ingest and deleting it when the ingest is finished.

iagaponenko: Your concerns are founded. Even though Linux mutex is a small data structure occupying 40 bytes on the 64-bit machine, having many thousands of those may result in potentially non-negligible growth in memory. Note that the actual amount of per-mutex memory would be possible twice as large due to additional overheads imposed by class {{util::Mutex}} (that adds two data members on top of {{std::mutex}}), extra overheads for storing the name of a mutex ( for example: {{Object_123456}}, the memory to store the {{unique_ptr<std::mutex>>}}, and the internal memory in {{std::map<std::string,std::unique_ptr<std::mutex>>}}. I wouldn't be surprised by seeing up to 100 bytes per each such entry in the registry. On the other hand, the number of such entries per worker is determined by the total number of chunks divided by the number of workers. If we take, as an example, a typical "full-sky" catalog of 150,000 chunks stored within the 30-nodes Qserv instance, we would have 5,000 or 10,000 (x2 is required due to "FullOveralp" tables) mutex objects per worker. This will result in about of {{1MB}} of extra memory allocated for ingesting one partitioned table of the catalog.

Having said that, I have conflicting feelings about introducing extra complexity into a design of the Registry. As you have already mentioned, that would require 1) implementing some "garbage collection" mechanism, and 2) guaranteeing the life span of the shared pointers to the mutex objects.

I also think that any such "garbage collection" could be potentially dangerous w/o reference counting on the shared pointers of each entry in the dictionary. Though, in theory, method {{std::shared_ptr<T>::use_count()}} may help here. Based on this method, the "garbage collection" could be implemented as simple as removing all entries from the registry that have counter equal to 1. That would be done on each request to method {{NamedMutexRegistry::get(mutexName)}}. This would also imply that all interested parties waiting on the mutex have a valid shared pointer to the mutex. In this respect the pain C/C++ references (w/o any such "garbage collection") are much simple and more reliable. The second drawback of the above presented "garbage collection" would be its compute-time overhead as it would imply scanning the content of the dictionary to identify unused (reference count equal to 1) entries. The maximum number of "active" entries in the dictionary is presently equal to the size of a thread pool configured at the ingest worker. This number could be a factor of the number of CPU cores of the worker's machine. In the current configuration of the largest Qserv cluster at NCSA  this parameter is set to 16 (or may be 32). In other setups the number could be larger. Hence, the price to pay for the {{1MB}} of extra memory would be 16, 32, etc. iterations over the internal map of the Registry at each invocation of  method {{NamedMutexRegistry::get(mutexName)}}. Obviously, various compromise solutions are also possible. For example, the garbage collection could be triggered by some threshold (set via the constructor of the Registry) for the "soft" (desired) maximum number of entries in the Registry. The limit could be set high enough to avoid scanning the registry on each invocation of method {{::get()}}, and to be low enough to reduce the memory consumption well (times, an order of magnitude) below the estimated {{1MB}} per database.

As far as the "per-database-registry" suggestion is concerned then it's not possible to implement the one at the ingest worker's level. These workers don't maintain a state of ingested databases. They have a very trivial role - get a client request to ingest a table contribution from some data source (or the data itself) in some context (transaction, table name an a chunk), pull the data, pre-process the data, create the destination table, modify table's schema (add a MySQL partition), and load the data into the table.

core/modules/replica/HttpIngestChunksModule.cc
```diff
@@ -297,20 +308,32 @@ json HttpIngestChunksModule::_addChunk() {
...
+            throw invalid_argument(
+                    context() + "::" + string(__func__) + " this service expects either 'database' or "
+                    " 'transaction_id' to be provided to define a scope of the request.");
+        }
+        TransactionId const transactionId = body().required<TransactionId>("transaction_id");
+        debug(__func__, "transactionId=" + to_string(transactionId));
+        auto const transactionInfo = databaseServices->transaction(transactionId);
+        database = transactionInfo.database;
+    }
+    debug(__func__, "database=" + database);
```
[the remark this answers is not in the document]
iagaponenko: Good point! I will eliminate the code duplication here!

core/modules/replica/NamedMutexRegistry.h
jgates108: I'm not too worried about the number of mutexes from a single ingest, even if the database has 150k chunks. I think it's a mutex per table, so 3 tables would be 450k, plus extras for overlaps. Ingest that a few user tables that have a couple of elements for every chunk, and this could be very large.

You certainly don't what to be garbage collecting every time you call `get`, but it is probably worth it to do it whenever a database ingest finishes, or ingest is cancelled, or every few thousand times `get` is called. At any given time most of them should be unlocked and having a few left behind should not be an issue as they will be garbage collected next time. The goal is to keep this from being millions of unused mutex. Checking `use_count() <= 1` should be safe if the mutexes are always accessed in the following manner. `ptrMutex` should always outlive `lock`, and the `use_count` will be at least 2. `getNamedMutex` must return a shared pointer. 

```   
    auto ptrMutex = _serviceProvider->getNamedMutex(statement.mutexName)
    {
        util::Lock const lock(*ptrMutex, context_);
        func() // do whatever needs to be done
    }
```

What I don't like about that is it requires the mutex to always be called like that. The only way I can think of to force people to call it that way would be to make `getNamedMutex` a private method and have a public method take `std:function func` as an argument and includes the code above. That has other drawbacks, mostly declaring more lamda functions and more difficult to read.

core/modules/util/Mutex.h
```diff
@@ -123,26 +104,11 @@ class Mutex
...
- */
-//Lock&& lock(Mutex& mutex, std::string const& context=std::string());
-
-
 /**
  * Class Lock is designed to completement the above defined class Mutex.
  * The current implementation of the class is very similar to std::lock_guard.
  * In addition Lock would also print out 3 debug messages into the log stream when
- * a state transition occures:
+ * a state transition ocurres:
```
jgates108, line 111: occurs

core/modules/replica/ServiceProvider.h
```diff
@@ -132,6 +133,13 @@ class ServiceProvider : public std::enable_shared_from_this<ServiceProvider> {
     /// @return a reference to worker messenger service (configured for controllers)
     MessengerPtr const& messenger();
 
+    /**
+     * @param database The name of a named mutex.
+     * @return A smart pointer to a mutex for the name.
+     * @throw std::invalid_argument If the name is empty.
```
jgates108, line 139: I'd add a comment to this that the shared pointer must live at least as long as the mutex is locked, or bad things could happen.

iagaponenko: Good suggestion!

PR #583, opened 2020-11-27
Add IN2P3 TLS CA to Qserv's Docker image

fjammes:


iagaponenko:
Is this package the right place for adding site-specific certificates?

PR #584, opened 2020-12-01
Tickets/dm 27781

iagaponenko:


core/modules/qdisp/QdispPool.cc
```diff
@@ -111,14 +111,7 @@ util::Command::Ptr PriorityQueue::getCmd(bool wait){
...
         ++localLogLimiter;
-        // Log this every once in while to INFO so there's some idea of system
-        // load without generating crushing amounts of log messages.
-        if (localLogLimiter % 50 == 0) {
-            LOGS(_log, LOG_LVL_INFO, "priQueGet " << *this);
-        } else {
-            LOGS(_log, LOG_LVL_DEBUG, "priQueGet " << *this);
-        }
-
+        LOGS(_log, LOG_LVL_DEBUG, "priQueGet " << *this);
```
jgates108, line 114: Actually, this is a really important/useful piece of information. It indicates what the czar is most busy trying to do and there's no other way to access it at this time. Reducing it to 1 in 100, or 1 in 500 getting logged to INFO would be better. Having localLogLimiter be a a member variable and replacing 50 with a member variable would be better still.

jgates108 (commented):
Aside from from one message I'd really like left in, it looks good.

core/modules/qdisp/QdispPool.cc
iagaponenko: I chose to downgrade the logging level of the message to `DEBUG` since it was making a lot of noise  in the log files. Note that extensive logging:
* results in huge log files
* badly affects the overall performance of the application because LSST Logger has a synchronized API serializing threads.
If you think seeing this information is very important at the `INFO` level then I will put it back at a level 1 to 500.

Though, in general I think we need to implement a different mechanism for monitoring `czar`. Dumping this info into the log files is not the best option.

iagaponenko: More info on this subject. An intent of printing 1 out of 50 was to avoid "...generating crushing amounts of log messages..." (as in the original code's comment). In reality, with 150,000 chunks per query, we end up with 3,000 such messages per each query. That's a lot! 

Another complication of that idea is that the logging is happening under protection of the lock:
```
util::Command::Ptr PriorityQueue::getCmd(bool wait){
    util::Command::Ptr ptr;
    std::unique_lock<std::mutex> uLock(_mtx);
    while (true) {
        _changed = false;
        ++localLogLimiter;
        // Log this every once in while to INFO so there's some idea of system
        // load without generating crushing amounts of log messages.
        if (localLogLimiter % 500 == 0) {
            LOGS(_log, LOG_LVL_INFO, "priQueGet " << *this);
        } else {
            LOGS(_log, LOG_LVL_DEBUG, "priQueGet " << *this);
        }
        ...
```
This is not a good idea since while the method is under the lock it would also get locked by the the LSST Logger. As a result of this the performance of `PriorityQueue::getCmd` may badly suffer.

Perhaps, we should move the reporting outside the lock?

I have also noticed that the overloaded operator `<<` for class `qdisp::QdispPool` is not so simple:
```
// must hold pq._mtx before calling
std::ostream& operator<<(std::ostream& os, PriorityQueue const& pq) {
    for (auto const& elem : pq._queues)   {
        PriorityQueue::PriQ::Ptr const& que = elem.second;
        os << "(pr=" << que->getPriority()
           << ":sz="  << que->size()
           << ":r="   << que->running << ")";
    }
    return os;
}
```
One can see both iteration over a collection of queues as well as a lot of string manipulations (and writing into a stream) happening under protection of a lock.
Perhaps this code could be refactored into something simpler that would only require locking for data extraction, which doing the rest outside the lock?

iagaponenko: How about adding a synchronized method:
```
struct PriorityQueue::QueuStats {
    int priority;
    size_t size;
    int running;
};
std::vector<PriorityQueue::QueuStats> PriorityQueue::_qStats() {
    std::vector<PriorityQueue::QueuStats> stats;
    std::unique_lock<std::mutex> uLock(_mtx);
    for (auto const& elem : pq._queues)   {
        PriorityQueue::PriQ::Ptr const& que = elem.second;
        stats.emplace_back({que->getPriority(), que->size(), que->running});
    }
    return stats;
}
std::ostream& operator<<(std::ostream& os, std::vector<PriorityQueue:;QueuStats> const& qStats) {
    for (auto const& que : qStats)   {
        os << "(pr=" << que.priority
           << ":sz="  << que.size
           << ":r="   << que.running << ")";
    }
    return os;
}
```
This will allow reordering operations in method `getCmd()` to allow safely printing debugging info w/o keeping the lock for too long, or printing into the Logger while holding the lock:
```
util::Command::Ptr PriorityQueue::getCmd(bool wait){
    // Log this every once in while to INFO so there's some idea of system
    // load without generating crushing amounts of log messages.
    if (localLogLimiter % 500 == 0) {
        LOGS(_log, LOG_LVL_INFO, "priQueGet " << _qStats());
    } else {
        LOGS(_log, LOG_LVL_DEBUG, "priQueGet " << _qStats());
    }
    util::Command::Ptr ptr;
    std::unique_lock<std::mutex> uLock(_mtx);
    while (true) {
        _changed = false;
        ++localLogLimiter;
        ...
```

iagaponenko: Or, even a lighter implementation reducing duration of time needed to keep th elock:
```
std::vector<PriorityQueue::QueuStats> PriorityQueue::_qStats() {
    auto queues;
    {
        std::unique_lock<std::mutex> uLock(_mtx);
        queues = _queues;
    }
    std::vector<PriorityQueue::QueuStats> stats;
    for (auto const& elem : queues)   {
        auto const& que = elem.second;
        stats.emplace_back({que->getPriority(), que->size(), que->running()});
    }
    return stats;
}
```
Note that no lock is required for calling `que->getPriority()` which is not mutable, and `que->size()` is already a synchronized method that acquire a lock and having another one is just asking for troubles such a potential deadlock!!! And the last call to `que->running()` gets a value of an `atomic` data member that doesn't require any additional synchronization.
All together, the lock is only needed to make a copy of a map of a few elements.

jgates108: Copying _queues isn't particularly cheap. Each shared pointer has it's own mutex, and it's reference counter needs to be incremented when the copy is made, and then decremented when `queues` is destroyed at the end of the function call.

`running` is atomic, but if you access it without having the _mtx lock, the value could be newer than value for `que->size()` which could be confusing if you're trying to sort out a bug in the logs.

Another approach would be to:
1. Decide if the logging will be done before entering the mutex.
2. If the log message is needed, put it in a string while in the mutex.
3. log the the string after releasing the mutex.

iagaponenko: Ok. This makes sense, except I disagree that with a statement that "...Each shared pointer has it's own mutex...". This class does not have the one. It has a pointer to the pointed object and another pointer on some other object containing atonic counter. Therefore, making a copy of the shared pointer should be done under protection of a lock. Meanwhile using different instances of shared pointers (to the same object) from different threads is a safe operation. If one instance of the shared pointer gets destroyed by the atomic counter gets safely decremented by 1. And so on.

core/modules/wdb/QueryRunner.h
```diff
@@ -126,8 +138,14 @@ class QueryRunner : public wbase::TaskQueryRunner, public std::enable_shared_fro
...
     util::MultiError _multiError; // Error log
 
+#ifdef QSERV_USE_PROTO_ARENA
+    std::unique_ptr<google::protobuf::Arena> _arena;
+    proto::ProtoHeader* _protoHeader = nullptr;
+    proto::Result* _result = nullptr;
+#else
     std::shared_ptr<proto::ProtoHeader> _protoHeader;
     std::shared_ptr<proto::Result> _result;
+#endif
```
jgates108, line 148: I think it's safe to just get rid of the code that doesn't use the arena and the QSERV_USE_PROTO_ARENA define.

iagaponenko: Fixed.

PR #587, opened 2020-12-08
Fix xrootd startup script pidfile issue

fritzm:
Since update to XrootD 5.0, cmsd pidfile no longer has ".mangr"
element.  Remove legacy special plumbing for this in xrootd
init.d script.


iagaponenko:
Looks good to me.

PR #593, opened 2021-01-20
Tickets/dm 27855

iagaponenko:


core/modules/replica/MessengerConnector.cc
```diff
@@ -144,34 +148,41 @@ void MessengerConnector::stop() {
...
-
-    if (_currentRequest and (_currentRequest->id() == id)) {
-        _currentRequest = nullptr;
-        if (_state == STATE_COMMUNICATING) _restart(lock);
+    // communications with a worker.
+    if (_state == STATE_COMMUNICATING) {
+        if ((_currentRequest != nullptr) and (_currentRequest->id() == id)) {
+            // Make sure the request gets eliminated before restarting the connection.
+            // Otherwise it will be picked up again.
+            _currentRequest = nullptr;
```
jgates108, line 171: I think this could easily lead to a race condition. If this happens in one thread while another thread has just passed the 'if' part of `(_currentRequest ? _currentRequest->id() : ""` the id() is invalid because _currentRequest is now 'nullptr'

If this bit of code can only be called when no other threads can possibly access _currentRequest, then you're ok.

Otherwise, there are 2 options. (they are basically the same thing)
- Make a copy of _currentRequest and use the copy everywhere else. Since you're using the copy, _currentRequest being set to nullptr wont hurt you, since you have a valid local copy 
- The other option is use a weak pointer to access it. One normal shared pointer holds onto the real object and all access is done by locking the weak pointer. Again, having the original set to nullptr wont hurt you since there's a valid local copy.

iagaponenko: Fortunately that's not going to happen since this operation is being performed in the synchronized method holding a lock on any modifications on the internal state of an object:
```
void MessengerConnector::cancel(string const& id) {
    util::Lock lock(_mtx, _context() + __func__);
    ...
```

PR #594, opened 2021-01-21
Tickets/dm 28006

iagaponenko:


core/modules/replica/HttpExceptions.h
```diff
@@ -65,6 +65,21 @@ class HttpError: public std::runtime_error {
...
+ * Report an error as an exception.
+ * @note This method always throws and it never returns control back to a caller.
+ *   The method is meant to be used by Ingest system workers to report error conditions
+ *   that won't require aborting a transaction.
+ * @param scope A scope of the error.
+ * @param error A human readable error message.
+ * @param (optional) HTTP code of a error if applies.
+ * @throws HttpError
+ */
+void retryAllowedError(std::string const& scope,
```
andy-slac, line 79: MAybe call this method as `raiseRetryAllowedError`, leading `retry` reads like an action verb.

core/modules/replica/HttpExceptions.h
```diff
@@ -65,6 +65,21 @@ class HttpError: public std::runtime_error {
...
 
+
+/**
+ * Report an error as an exception.
+ * @note This method always throws and it never returns control back to a caller.
+ *   The method is meant to be used by Ingest system workers to report error conditions
+ *   that won't require aborting a transaction.
+ * @param scope A scope of the error.
+ * @param error A human readable error message.
+ * @param (optional) HTTP code of a error if applies.
```
andy-slac, line 76: `an error`

core/modules/replica/IngestFileSvc.cc
```diff
@@ -95,11 +96,11 @@ void IngestFileSvc::openFile(TransactionId transactionId,
     try {
         auto transactionInfo = _serviceProvider->databaseServices()->transaction(_transactionId);
         if (transactionInfo.state != TransactionInfo::STARTED) {
-            throw logic_error(context_ + "transaction " + to_string(_transactionId) + " is not active");
+            retryAllowedError(context_, "transaction " + to_string(_transactionId) + " is not active");
```
andy-slac, line 99: I'm not sure I understand this. If client retries this again then result will be the same error unless transaction is started somehow. It feels like you are probably abusing "retry" to mean "do not kill this transaction but correct some other condition first" which is different from "retry immediately". Maybe you need separate error to indicate that transaction is safe to continue but error is an error? This applies to other cases in this file.

core/modules/replica/IngestHttpSvcMod.cc
```diff
@@ -139,7 +140,7 @@ json IngestHttpSvcMod::executeImpl(string const& subModuleName) {
     string const table = body().required<string>("table");
     string const columnSeparatorStr = body().required<string>("column_separator");
     if (columnSeparatorStr.empty() || columnSeparatorStr.size() != 1) {
-        throw invalid_argument(string(__func__) + " column separator must be a single character string");
+        retryAllowedError(__func__, "column separator must be a single character string");
```
andy-slac, line 143: This also does not feel like retrying would work.

andy-slac (approved):
Approved but you may want to check that retry is going to work for some operations.

core/modules/replica/HttpExceptions.h
iagaponenko: Good point! I'll rename the function.

core/modules/replica/IngestFileSvc.cc
iagaponenko: In theory you're right. There is no way for an ingest workflow to determine if the transaction needs to be started ... unless the client will parse the error message. And the later is very impractical. Introducing another error code to elaborate the problem may not help ether since most ingest workflows may do transaction management in a separate process from the one that's initiated the actual ingest. So, perhaps, the best course of action here would be to roll back the change to the previous code.

core/modules/replica/IngestHttpSvcMod.cc
iagaponenko: Agreed. See my previous comment.

PR #595, opened 2021-01-26
Fixed a bug in the MySQL Connection class

iagaponenko:


core/modules/replica/DatabaseMySQL.cc
```diff
@@ -287,6 +287,7 @@ void Connection::_processLastError(string const& context,
         case CR_SERVER_LOST:
         case CR_SERVER_LOST_EXTENDED:
         case CR_TCP_CONNECTION:
+        case CR_UNKNOWN_HOST:
```
andy-slac, line 290: This looks like a random list of error codes. Not sure how to improve this though, mysql is a mess.

Also commenting on the code below (can't comment on code that did not change). Raising an exception after re-connect is potentially problematic. I guess your intent is that client catches this exception and retries. Problem is that this implies that client code (and your code) has to be exception-safe, which is hard to verify.

andy-slac (approved):
This change is OK, but existing code has some potential issues.

core/modules/replica/DatabaseMySQL.cc
iagaponenko: For the collection of errors addressed in this section of the code, they all are related to various problems that may occur either when establishing a new connection to MySQ, or while communicating with MySQL. In all those cases a connection needs to be established or re-established. I've added these coded after analyzing the corresponding MySQL & MariaDB documentation to cover the relevant failure modes. The new one `CR_UNKNOWN_HOST` was missing since the code was developed in the pre-Kubernetes era while a missing DNS entry was a MySQL server was seen as something "unimaginable". These days (with Kubernetes) with have to program everything as if we were walking on the shaky swampland, by watching each step and not trusting the infrastructure.
 
The exception `Reconnected` thrown by the method (if reconnection is allowed) is for internal consumption only, and it's caught only at one location within the implementation of class `Connection`. It's thrown only after successfully (re-)establishing a connection. Client code shall not see it. The exception is expected by a method that processes user queries (including transaction management, result set extraction, etc., everything that's one by user script - lambda function):
```
Connection::Ptr Connection::execute(function<void(Connection::Ptr)> const& script,
                                    unsigned int maxReconnects,
                                    unsigned int timeoutSec) {
    while (true) {
        try {
            script();
            return self;
        } catch (Reconnected const& ex) {
            ...
        }
    }
```
That code is only interested in this specific catching exception `Reconnected`. The rest is passed on to a client trying to execute a query. It's supposed to be up to the client to deal with those exceptions.
I agree that there may be various caveats in this logic. I have seen none yet.

BTW, I don't mind discussing options for improving this code beyond a scope of tis (or other) tickets. My goal is to make this code as reliable as possible.

iagaponenko: The caveat here is that the above mentioned user script shall correctly manage its transactions to avoid leaving side effects should the script be re-tried due to reconnects. Right now all Replication and Ingest systems' code does so. And I have no plans to expand using this MySQL API beyond a scope of its present use. 

Unfortunately, this won't guarantee that someone else (than me) would make a mistake and don't do transaction management correctly. To prevent this I recently extended the API with method `Connection::executeInOwnTransaction(script)` that would automatically manage transactions "around" the script. Most of the existing code of the Replication/Ingest system has already been migrated to use that method.

PR #596, opened 2021-02-01
Tickets/dm 26101

iagaponenko:


core/modules/replica/DatabaseServices.cc
```diff
@@ -167,6 +167,25 @@ json TransactionInfo::toJson() const {
...
 
+json TransactionContribInfo::toJson() const {
+    json info;
+    info["id"] = id;
+    info["transaction_id"] = transactionId;
+    info["worker"]     = worker;
+    info["database"]   = database;
+    info["table"]      = table;
+    info["chunk"]      = chunk;
+    info["overlap"]    = isOverlap ? 1 : 0;
```
n8pease, line 178: just curious, where does `isOverlap` come from? I don't see it anywhere else in this file.

core/modules/replica/DatabaseServicesMySQL.h
```diff
@@ -348,6 +366,12 @@ class DatabaseServicesMySQL : public DatabaseServices {
     std::vector<TransactionInfo> _findTransactionsImpl(util::Lock const& lock,
                                                        std::string const& predicate);
 
+    TransactionContribInfo _transactionContribImpl(util::Lock const& lock,
+                                                   std::string const& predicate);
+
+    std::vector<TransactionContribInfo> _transactionContribsImpl(util::Lock const& lock,
+                                                                 std::string const& predicate);
```
n8pease, line 373: are these Impl functions still needed? I can't tell that they're being called anywhere.

core/modules/replica/DatabaseServices.cc
iagaponenko: This is a data member of the class (see the class's header `core/modules/replica/DatabaseServices.h`):
```
class TransactionContribInfo {
public:
    ...
    bool isOverlap = false;     /// (optional) A flavor of the chunked table (partitioned tables only)
```

core/modules/replica/DatabaseServicesMySQL.h
iagaponenko: The implementation is in the corresponding `CC` file `core/modules/replica/DatabaseServicesMySQL.cc`:
```
vector<TransactionContribInfo> DatabaseServicesMySQL::_transactionContribsImpl(
        util::Lock const& lock,
        string const& predicate) {

    string const context = _context(__func__) + "predicate=" + predicate + " ";
    ...
```

core/modules/replica/DatabaseServices.cc
n8pease: lol, thx. I've got 🐍-brain I guess.

core/modules/replica/DatabaseServicesMySQL.h
n8pease: huh, I thought I looked there. Ok, thx.

PR #598, opened 2021-02-04
Added code to detect oversize results on the czar.

jgates108:


core/modules/rproc/InfileMerger.cc
```diff
@@ -115,8 +115,8 @@ InfileMerger::InfileMerger(InfileMergerConfig const& c,
       _mysqlConn(_config.mySqlConfig),
       _databaseModels(dm),
       _jobIdColName(JOB_ID_BASE_NAME),
-      _maxResultTableSizeMB(_config.czarConfig.getMaxTableSizeMB()),
       _maxSqlConnectionAttempts(_config.czarConfig.getMaxSqlConnectionAttempts()),
+      _maxResultTableSizeBytes(_config.czarConfig.getMaxTableSizeMB()*1000000),
```
iagaponenko, line 119: Should this rather be `1024*1024` instead of `1000000`?

core/modules/rproc/InfileMerger.cc
```diff
@@ -206,21 +192,24 @@ void InfileMerger::_setQueryIdStr(std::string const& qIdStr) {
 }
 
 
-bool InfileMerger::merge(std::shared_ptr<proto::WorkerResponse> response) {
+bool InfileMerger::merge(std::shared_ptr<proto::WorkerResponse> response, bool last) {
     if (!response) {
         return false;
     }
     // TODO: Check session id (once session id mgmt is implemented)
+    int jobId = response->result.jobid();
```
iagaponenko, line 200: For the sake of code safety, anything that gets initialized once and never changes later, has to be declared like:
```
int const obId = ...
```

core/modules/rproc/InfileMerger.cc
```diff
@@ -206,21 +192,24 @@ void InfileMerger::_setQueryIdStr(std::string const& qIdStr) {
...
     std::string queryIdJobStr =
-        QueryIdHelper::makeIdStr(response->result.queryid(), response->result.jobid());
+        QueryIdHelper::makeIdStr(response->result.queryid(), jobId);
     if (!_queryIdStrSet) {
         _setQueryIdStr(QueryIdHelper::makeIdStr(response->result.queryid()));
     }
+    size_t resultSize = response->result.transmitsize();
```
iagaponenko, line 206: If you're not going to modify a value of this var then please do:
```
size_t const resultSize = 
```

core/modules/rproc/InfileMerger.cc
```diff
@@ -258,6 +247,23 @@ bool InfileMerger::merge(std::shared_ptr<proto::WorkerResponse> response) {
         return true;
     }
 
+    size_t tResultSize;
```
iagaponenko, line 250: Should this be rather called `taskResultSize`? Right now one has to guess what's the meaning of `t`, whether it stands for "temporary", "table", or "total". There are too many variable containing "ResultSize" in this code.

core/modules/wdb/QueryRunner.h
```diff
@@ -121,6 +121,8 @@ class QueryRunner : public wbase::TaskQueryRunner, public std::enable_shared_fro
 
     wbase::Task::Ptr const _task; ///< Actual task
 
+    size_t const _maxResultSize = 5500000000; ///< &&& remove ???
```
iagaponenko, line 124: This limit has to be loaded from a configuration of "czar". If you don't want to set any specific limit (I may see scenarios when that could be the case) then I'd rather initialize this attribute with 0 (or with `std::numeric_limits<size_t>::max()`. Then you would have to check in the rest of your code if it's set to 0 and don't enforce the limit. Just post a warning in the Logger stream.

core/modules/wdb/QueryRunner.cc
```diff
@@ -283,6 +283,10 @@ bool QueryRunner::_fillRows(MYSQL_RES* result, int numFields, uint& rowCount, si
                 LOGS_ERROR("Message single row too large to send using protobuffer");
                 return false;
             }
+            auto totalSize = _task->addTransmitSize(tSize);
+            if (totalSize > _maxResultSize) { // &&& just forget this part on the worker ???
+                LOGS_ERROR("Result set of single query too large totalSize=" << totalSize);
+            }
```
iagaponenko, line 289: Don't you want to begin cancelling query execution right away once Qserv "czar" runs into this condition?
There is another problem you would need to think about - if someone accidentally does `SELECT *` on a large table w/o conditions then czar's log file will bee flooded with thousand identical messages of this kind. Perhaps you need to downgrade the logging level for this message to `TRACE`?

core/modules/rproc/InfileMerger.h
```diff
@@ -311,12 +311,14 @@ class InfileMerger {
...
     bool _deleteInvalidRows(std::set<int> const& jobIdAttempts);
-
-    int _sizeCheckRowCount = 0; ///< Number of rows read since last size check.
-    int _checkSizeEveryXRows = 1000; ///< Check the size of the result table after every x number of rows.
-    size_t _maxResultTableSizeMB = 5000; ///< Max result table size.
     int const _maxSqlConnectionAttempts = 10; ///< maximum number of times to retry connecting to the SQL database.
 
+    /// Variable to track result size. Each
+    size_t const _maxResultTableSizeBytes; ///< Max result table size in bytes.
+    size_t _totalResultSize{0}; ///< Size of result so far in bytes.
```
iagaponenko, line 318: Has to be:
```
size_t _totalResultSize = 0;
```

iagaponenko (approved):

core/modules/wdb/QueryRunner.cc
jgates108: I thought about this a bit. It wouldn't much faster than having the czar do it, it wound up making things a lot more complicated, and it should be a rare occurrence (i.e. I'd expect most results too large errors to come from the results of multiple chunks being too large). 

Whether the cancellation  starts at this line or on the czar when it receives this transmit, most of the result from this task will have already been sent to the czar as the worker only reads out what will be sent in the next transmit. I'll probably just get rid of this check as trying to keep configuration values consistent for both the czar and the worker is not simple.

core/modules/wdb/QueryRunner.h
jgates108: Deleted. I had thought I'd already gotten rid of all of this, but I missed some.

PR #601, opened 2021-02-10
Added and removed configuration parameters.

jgates108:
I had thought there were more items to add to the configuration files, but I din't see any.

### DIFF
--- a/core/modules/qproc/QuerySession.h
+++ b/core/modules/qproc/QuerySession.h
@@ -79,8 +79,9 @@ public:
 
     QuerySession(std::shared_ptr<css::CssAccess> const& css,
                  qproc::DatabaseModels::Ptr const& dbModels,
-                 std::string const& defaultDb)
-        : _css(css), _defaultDb(defaultDb), _databaseModels(dbModels) {}
+                 std::string const& defaultDb,
+                 int const interactiveChunkLimit)
+        : _css(css), _defaultDb(defaultDb), _databaseModels(dbModels), _interactiveChunkLimit(interactiveChunkLimit) {}
 
     std::shared_ptr<query::SelectStmt> parseQuery(std::string const & statement);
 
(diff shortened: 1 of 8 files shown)


core/modules/qproc/QuerySession.h
```diff
@@ -79,8 +79,9 @@ class QuerySession {
 
     QuerySession(std::shared_ptr<css::CssAccess> const& css,
                  qproc::DatabaseModels::Ptr const& dbModels,
-                 std::string const& defaultDb)
-        : _css(css), _defaultDb(defaultDb), _databaseModels(dbModels) {}
+                 std::string const& defaultDb,
+                 int const interactiveChunkLimit)
```
iagaponenko, line 83: Extra `const` for parameters passed by value isn't strictly required. But it won't hurt either.

core/modules/qproc/QuerySession.h
```diff
@@ -244,19 +245,18 @@ class QuerySession {
...
-    int _isFinal{0}; ///< Has query analysis/optimization completed?
+    int _isFinal =0; ///< Has query analysis/optimization completed?
 
     ChunkSpecVector _chunks; ///< Chunk coverage
     std::shared_ptr<QueryPluginPtrVector> _plugins; ///< Analysis plugin chain
 
-    /// Maximum number of chunks in an interactive query. TODO: DM-10273 put in config file.
-    int const _interactiveChunkLimit{10};
-    bool _scanInteractive{true}; ///< True if the query can be considered interactive.
+    int const _interactiveChunkLimit = 10; ///< Maximum number of chunks in an interactive query.
```
iagaponenko, line 258: The same default is set in `CzarConfig`. Do you think it would worth an effort to define it in one location? If not then it's probably okay.

iagaponenko (approved):
Looks good to me.

core/modules/qproc/QuerySession.h
jgates108: Unit tests use it. i added a comment to that affect.

PR #602, opened 2021-02-11
Tickets/dm 28603

iagaponenko:


core/modules/replica/SqlJob.h
```diff
@@ -210,6 +210,34 @@ class SqlJob : public Job {
...
+     *   Note, this parameter is only taken into consideration if the previous
+     *   parameter 'allTables' was set to 'false'.
+     * @return A collection of tables found.
+     * @throw std::invalid_argument If the database or a table isn't valid.
+     */
+    std::vector<std::string> workerTables(std::string const& worker,
+                                          TransactionId const& transactionId,
+                                          std::string const& table,
+                                          bool allTables=true,
+                                          bool overlapTablesOnly=false) const;
```
jgates108, line 239: I find the allTables and overlapTablesOnly variables a bit confusing. Maybe get rid of allTables and use:
```
 reportNonOverlapTables = true,
 reportOverlapTables = true
```

jgates108 (commented):
Looks good, just minor things.

core/modules/replica/SqlJob.h
iagaponenko: A "cascading" use of two options `allTables` and `overlapTablesOnly` allows implementing the 3-state logic by unambiguously representing the following table section expression:
```
if (allTables) {
    tables = {"chunk_123", "chunk_123_overlap"};
} else {
    if (overlapTablesOnly) {
        tables = {"chunk_123_overlap"};
    } else {
        tables = {"chunk_123"};
    }
}
```
What you're proposing has 4 states, of which one is undetermined since it would be equivalent to:
```
if (reportNonOverlapTables && reportOverlapTables) {
    tables = {"chunk_123", "chunk_123_overlap"};
} else {
    if (reportOverlapTables && !reportNonOverlapTables) {
        tables = {"chunk_123_overlap"};
    } else if (!reportOverlapTables && reportNonOverlapTables) {
        tables = {"chunk_123"};
   } else {
        // PROBLEM! What are we going to do about this?
        // (reportOverlapTables = false) && (reportNonOverlapTables==false))
   }
}
```

PR #605, opened 2021-02-26
Tickets/dm 28750

jgates108:


core/modules/wbase/Task.cc
```diff
@@ -95,10 +98,15 @@ IdSet Task::allIds{};
 /// available to define the action to take when this task is run, so
 /// Command::setFunc() is used set the action later. This is why
 /// the util::CommandThreadPool is not called here.
-Task::Task(Task::TaskMsgPtr const& t, SendChannel::Ptr const& sc)
-    : msg(t), sendChannel(sc),
-      _qId(t->queryid()), _jId(t->jobid()), _attemptCount(t->attemptcount()),
-      _idStr(QueryIdHelper::makeIdStr(_qId, _jId)) {
+Task::Task(TaskMsgPtr const& t, std::string const& query, int fragmentNumber,
+                  std::shared_ptr<SendChannel> const& sc)
+: msg(t), sendChannel(sc),
```
abh3, line 103: This looks like a text alignment error.

core/modules/wcontrol/Foreman.cc
```diff
@@ -75,15 +77,15 @@ Foreman::Foreman(Scheduler::Ptr                  const& scheduler,
...
+    _backend = make_shared<wdb::SQLBackend>(_mySqlConfig);
     _chunkResourceMgr = wdb::ChunkResourceMgr::newMgr(_backend);
 
     assert(_scheduler); // Cannot operate without scheduler.
 
     LOGS(_log, LOG_LVL_DEBUG, "poolSize=" << poolSize << " maxPoolThreads=" << maxPoolThreads);
     _pool = util::ThreadPool::newThreadPool(poolSize,  maxPoolThreads, _scheduler);
 
-    _workerCommandQueue = std::make_shared<util::CommandQueue>();
+    _workerCommandQueue = make_shared<util::CommandQueue>();
```
abh3, line 88: So, how come "std::" is explicitly use but here it is not? That seems inconsistent.

core/modules/wdb/QueryRunner.cc
```diff
@@ -344,7 +359,7 @@ void QueryRunner::_transmit(bool last, unsigned int rowCount, size_t tSize) {
 
 
 void QueryRunner::_sendBuf(xrdsvc::StreamBuffer::Ptr& streamBuf, bool last,
-                           util::TimerHistogram& histo, std::string const& note) {
+                           util::TimerHistogram& histo, string const& note) {
```
abh3, line 362: In fact, inconsistent use of "std::" seems to pervade this update across a wide swath of classes!

core/modules/xrdsvc/SsiRequest.cc
```diff
@@ -129,25 +129,19 @@ void SsiRequest::execute(XrdSsiRequest& req) {
...
-                            " chunkId=" + std::to_string(ru.chunk()));
+                        " chunkId=" + std::to_string(ru.chunk()));
                 return;
             }
 
-            // Now that the request is decoded (successfully or not), release the
-            // xrootd request buffer. To avoid data races, this must happen before
-            // the task is handed off to another thread for processing, as there is a
-            // reference to this SsiRequest inside the reply channel for the task,
-            // and after the call to BindRequest.
```
abh3, line 140: Why is this comment block no longer relevant? It seems the basic code has not changed to make it irrelevant.

core/modules/util/EventThread.cc
```diff
@@ -42,6 +42,17 @@ namespace lsst {
...
 
+
+void CommandQueue::queCmd(std::vector<Command::Ptr> const& cmds) {
+    {
+        std::lock_guard<std::mutex> lock(_mx);
+        for (auto const& cmd:cmds) {
+            _qu.push_back(cmd);
+        }
+    }
+    notify(cmds.size() > 1); // notify all if more than 1 command, otherwise notify 1.
```
abh3, line 53: I'm still convinced that this should be handled via a semaphore that gives you precise control of how many threads will be usefully engaged. My presumption is that more often than note the number of tasks will be >1 but practically always less that the number of waiting threads.

abh3 (approved):
There is a lot to grok here and frankly I can't comment much on functionality. I did leave some comments about various things that you may wish to address.

core/modules/wdb/QueryRunner.cc
jgates108: 'using namespace std' in .cc files is removes a lot of unnecessary characters making the code easier to read. It makes sense to remove 'std::' on definitions everyone is familiar with, but there's no reason not to leave it on less commonly used functions. It's a change @iagaponenko and I have been doing as we come across them. The change adds a lot of chaff to pull requests.

core/modules/xrdsvc/SsiRequest.cc
jgates108: Good point. I put it back.

core/modules/wcontrol/Foreman.cc
```diff
@@ -95,42 +97,61 @@ Foreman::~Foreman() {
...
+void Foreman::_setRunFunc(shared_ptr<wbase::Task> const& task) {
     auto func = [this, task](util::CmdData*){
         proto::TaskMsg const& msg = *task->msg;
         int const resultProtocol = 2; // See proto/worker.proto Result protocol
         if (!msg.has_protocol() || msg.protocol() < resultProtocol) {
             LOGS(_log, LOG_LVL_WARN, "processMsg Unsupported wire protocol");
-            if (!task->getCancelled()) {
+            if (!task->isCancelled()) {
                 // We should not send anything back to xrootd if the task has been cancelled.
+                mutex* streamMutex = task->sendChannel->getStreamMutexPtr();
```
iagaponenko, line 109: I'm sorry to say so, but exposing internals mutexes of an object (SendChannel) just to call some method of an object smells like "spaghetti code". This reveals too much details on the internal implementation of `SendChannel` to the outside code and it will for sure make the code more difficult to debug and migrate in the future. Why not to follow a simple design pattern of the "synchronized" methods and acquire the lock inside the method implementation?
```
void SendChannel::sendError(std::string const& msg) {
    std::lock_guard<std::mutex> const lock(_mtx);
    ...
}
```
In case if the external locking is absolutely necessary then my second comment is about the pointer:
```
mutex* streamMutex = ...
```
Wby not to just pass a reference (assuming there is the accessor method returning the reference `std::mutex& sendChannel->getStreamMutex()`):
```
std::lock_guard<std::mutex> streamLock(task->sendChannel->getStreamMutex());
```

core/modules/wcontrol/Foreman.cc
```diff
@@ -95,42 +97,61 @@ Foreman::~Foreman() {
...
-                                                       _sqlConnMgr, _transmitMgr);
+                    _sqlConnMgr, _transmitMgr);
             bool success = false;
             try {
                 success = qr->runQuery();
             } catch (UnsupportedError const& e) {
                 LOGS(_log, LOG_LVL_ERROR, "runQuery threw UnsupportedError " << e.what() << *task);
             }
             if (not success) {
+                mutex* streamMutex = task->sendChannel->getStreamMutexPtr();
```
iagaponenko, line 123: Please, see my earlier comment on preferring the "synchronized" methods in favor of exposing internal data structures (such as this mutex) to outside codes).

core/modules/util/EventThread.cc
```diff
@@ -42,6 +42,21 @@ namespace lsst {
...
 namespace util {
 
+
+void CommandQueue::queCmd(std::vector<Command::Ptr> const& cmds) {
+    {
+        LOGS(_log, LOG_LVL_WARN, "&&& queCmd a");
+        std::lock_guard<std::mutex> lock(_mx);
+        LOGS(_log, LOG_LVL_WARN, "&&& queCmd b");
+        for (auto const& cmd:cmds) {
+            LOGS(_log, LOG_LVL_WARN, "&&& queCmd c");
```
iagaponenko: This may produce the hell of the messages for the shared scan queries. Does it really need to be the `WARN` level rather than (preferably) `TRACE`?

core/modules/wbase/Task.cc
```diff
@@ -127,6 +130,38 @@ Task::~Task() {
...
 
+std::vector<Task::Ptr> Task::createTasks(std::shared_ptr<proto::TaskMsg> const& taskMsg,
+                                         std::shared_ptr<wbase::SendChannel> sendChannel) {
+    QSERV_LOGCONTEXT_QUERY_JOB(taskMsg->queryid(), taskMsg->jobid());
+    std::vector<Task::Ptr> vect;
+
+    ///&&&; // Determine if there should be sub-tasks.
+    int fragmentCount = taskMsg->fragment_size();
+    if (fragmentCount < 1) {
+        throw Bug("QueryRunner: No fragments to execute in TaskMsg");
```
iagaponenko, line 146: The exception (should it be really thrown) may mislead readers that it happened in the class `QueryRunner` instead of `Task::createTasks`.

core/modules/wdb/QueryRunner.cc
```diff
@@ -322,29 +324,33 @@ void QueryRunner::_transmit(bool last, unsigned int rowCount, size_t tSize) {
...
-        // StreamBuffer::create invalidates resultString by using std::move()
-        xrdsvc::StreamBuffer::Ptr streamBuf(xrdsvc::StreamBuffer::createWithMove(resultString));
-        _sendBuf(streamBuf, last, transmitHisto, "body");
-    } else {
-        LOGS(_log, LOG_LVL_DEBUG, "_transmit cancelled");
+    {
+        // Nothing else can use this channel until this transmit is done.
+        lock_guard<mutex> streamLock(task->sendChannel->getStreamMutexRef());
+        _transmitHeader(resultString);
+        LOGS(_log, LOG_LVL_DEBUG, "_transmit last=" << last
```
iagaponenko: C++ has a built-in function `__func__` that captures the name of the current context:
```
LOGS(_log, LOG_LVL_DEBUG, __func__ << " last=" << last
```
It may help you in the future in two scenarios:
* if you will decide to rename the function `_transmit`) to something else. Then you won't need to chance all occurrences of the function names in the debug messages.
* if you will cut-and-paste the printout message with an intent to make a similar printout from some other method and forget to rename `_transmit` to match the name of a new host method.
The predefined compiler-time constant `__func__` prevents these and many other potential problems.

core/modules/wdb/QueryRunner.cc
```diff
@@ -322,29 +324,33 @@ void QueryRunner::_transmit(bool last, unsigned int rowCount, size_t tSize) {
...
+                << " resultString=" << util::prettyCharList(resultString, 5));
+
+        if (!_cancelled) {
+            // StreamBuffer::create invalidates resultString by using std::move()
+            xrdsvc::StreamBuffer::Ptr streamBuf(xrdsvc::StreamBuffer::createWithMove(resultString));
+            _sendBuf(streamBuf, last, transmitHisto, "body");
+        } else {
+            LOGS(_log, LOG_LVL_DEBUG, "_transmit cancelled");
```
iagaponenko: See my previous comment on suggested use of the predefined compiler-time constant `__func__` in favor of the actual function names.

core/modules/wdb/QueryRunner.cc
```diff
@@ -430,25 +436,51 @@ class ChunkResourceRequest {
...
 
     unsigned int rowCount = 0;
     size_t tSize = 0;
 
     try {
+        if (_cancelled) {
+            break;
+        }
+
+        proto::TaskMsg_Fragment const& fragment(m.fragment(fragNum));
```
iagaponenko: It's better to use the assignment operator to initialize fererences:
```
proto::TaskMsg_Fragment const& fragment = m.fragment(fragNum);
```

core/modules/wsched/BlendScheduler.cc
```diff
@@ -137,6 +137,134 @@ void BlendScheduler::_sortScanSchedulers() {
...
+    queCmd(vect);
+}
+
+
+/// &&& Change this to accept a vector of util::Command::Ptr const&
+void BlendScheduler::queCmd(std::vector<util::Command::Ptr> const& cmds) {
+    // Do the book keeping and determine which queue the tasks in 'cmds'
+    // should be added to. All tasks in 'cmds' must belong to the same
+    // user query and will go to the same scheduler.
+    LOGS(_log, LOG_LVL_WARN, "&&& BlendScheduler::queCmd a");
```
iagaponenko: Please, consider using the predefined compiler-time constant `__func__` here in everywhere:
```
LOGS(_log, LOG_LVL_WARN, "&&& BlendScheduler::" << __func__ << " a");
```

core/modules/wsched/BlendScheduler.cc
```diff
@@ -137,6 +137,134 @@ void BlendScheduler::_sortScanSchedulers() {
...
+    bool first = true;
+    SchedulerBase::Ptr s = nullptr;
```
iagaponenko: If this is the shared pointer then you don't need to explicitly initialize it with `nullptr`. This is done by the default c-tor of class `std::shared_ptr<T>`.
My second comment is that that name `s` is pretty much meaningless. Could some reasonable name be given to it to reflect a role of the variable?

core/modules/wsched/BlendScheduler.cc
```diff
@@ -137,6 +137,134 @@ void BlendScheduler::_sortScanSchedulers() {
...
+    bool onInteractive = false;
+    for (auto const& cmd : cmds) {
+        LOGS(_log, LOG_LVL_WARN, "&&& BlendScheduler::queCmd b");
+        wbase::Task::Ptr task = dynamic_pointer_cast<wbase::Task>(cmd);
+        if (task == nullptr) {
+            LOGS(_log, LOG_LVL_WARN, "&&& BlendScheduler::queCmd c");
+            // These should be few and far between.
+            LOGS(_log, LOG_LVL_INFO, "BlendScheduler::queCmd got control command");
+            if (cmds.size() > 1) {
+                throw Bug("cmds.size() > 1 when no task was set.");
```
iagaponenko, line 161: When throwing exceptions it's always good to provide more info on the context from which the exception was thrown like:
```
throw Bug("BlendScheduler::" + string(__func__) + " cmds.size() > 1 when no task was set.");
```
This will provide additional info when debugging problems.

core/modules/wsched/BlendScheduler.cc
```diff
@@ -137,6 +137,134 @@ void BlendScheduler::_sortScanSchedulers() {
...
+            }
+            notify(true); // notify all=true
+            LOGS(_log, LOG_LVL_WARN, "&&& BlendScheduler::queCmd c99");
+            return;
+        }
+
+        if (first) {
+            QSERV_LOGCONTEXT_QUERY_JOB(task->getQueryId(), task->getJobId());
+        }
+        LOGS(_log, LOG_LVL_WARN, "&&& BlendScheduler::queCmd d");
```
iagaponenko: These single letter messages should be posted at the level `WARN` since they don't contain anything resembling warnings. I think the right level should be `DEBUG` or `TRACE`.

core/modules/wsched/BlendScheduler.cc
```diff
@@ -137,6 +137,134 @@ void BlendScheduler::_sortScanSchedulers() {
...
+                                break;
+                            }
+                        }
+                    }
+                }
+                LOGS(_log, LOG_LVL_WARN, "&&& BlendScheduler::queCmd k");
+                // If the user query for this task has been booted, put this task on the snail scheduler.
+                auto queryStats = _queries->getStats(task->getQueryId());
+                if (queryStats && queryStats->getQueryBooted()) {
+                    s = _scanSnail;
```
iagaponenko: It looks like a value of `s` that might be (if certain conditions were met) previously set in the `for` loop above would be overwritten here. Would this cause any issues?

core/modules/wsched/ScanScheduler.cc
```diff
@@ -220,6 +220,46 @@ util::Command::Ptr ScanScheduler::getCmd(bool wait)  {
...
+
+void ScanScheduler::queCmd(std::vector<util::Command::Ptr> const& cmds) {
+
+    {
+        std::lock_guard<std::mutex> lock(util::CommandQueue::_mx);
+
+        for (auto const& cmd:cmds) {
+            wbase::Task::Ptr t = std::dynamic_pointer_cast<wbase::Task>(cmd);
+            if (t == nullptr) {
+                LOGS(_log, LOG_LVL_WARN, getName() << " queCmd could not be converted to Task or was nullptr");
```
iagaponenko, line 238: This looks scarier to me than just a warning. Why not to throw an exception right away as this could be a logic problem within the code?

core/modules/wbase/SendChannel.cc
```diff
@@ -139,6 +139,23 @@ SendChannel::Ptr SendChannel::newStringChannel(std::string& d) {
...
+bool SendChannel::transmitTaskLast(bool inLast) {
+    /// _caller must have locked _streamMutex before calling this.
+    if (not inLast) {
+        // This wasn't the last message buffer for this task, so it doesn't matter.
+        return false;
+    }
+    ++_lastCount;
+    bool lastTaskDone = _lastCount >= _taskCount;
+    return lastTaskDone;
+}
```
iagaponenko, line 164: The code could be rewritten to be as short as this:
```
bool SendChannel::transmitTaskLast(bool inLast) {
    /// _caller must have locked _streamMutex before calling this.
    if (not inLast) inLast;  // This wasn't the last message buffer for this task, so it doesn't matter.
    return _lastCount >= ++_taskCount;
}
```
I would also suggest passing the lock as a parameter into the method to indicate that the method requires a valid lock. Then you don't need to make the comment - it will become obvious in both definition and declarations of the method:
```
bool SendChannel::transmitTaskLast(lock_guard<mutex> const& lock, bool inLast) {
    if (not inLast) inLast;  // This wasn't the last message buffer for this task, so it doesn't matter.
    return _lastCount >= ++_taskCount;
}

File: core/modules/wbase/Task.cc [hunk 20]
```diff
@@ -150,18 +150,29 @@ std::vector<Task::Ptr> Task::createTasks(std::shared_ptr<proto::TaskMsg> const&
...
             vect.push_back(task);
         }
     }
+    sendChannel->setTaskCount(fragmentCount);
     return vect;
 }
 
 
+void Task::setQueryString(std::string const& qs) {
+    _queryString = qs;
```
iagaponenko: Is this safe to make this modification w/o doing any locking on the object state?

core/modules/wbase/Task.cc
```diff
@@ -150,18 +150,29 @@ std::vector<Task::Ptr> Task::createTasks(std::shared_ptr<proto::TaskMsg> const&
...
+}
+
+
+void Task::setQueryFragmentNum(int fragNum) {
+    _queryFragmentNum = fragNum;
```
iagaponenko: Is this safe to make this modification w/o doing any locking on the object state?

core/modules/wbase/Task.h
```diff
@@ -178,10 +182,12 @@ class Task : public util::CommandForThreadPool {
     std::chrono::milliseconds finished(std::chrono::system_clock::time_point const& now);
 
 private:
-    QueryId  const    _qId{0}; //< queryId from czar
-    int      const    _jId{0}; //< jobId from czar
-    int      const    _attemptCount{0}; // attemptCount from czar
+    QueryId  const    _qId = 0; ///< queryId from czar
+    int      const    _jId = 0; ///< jobId from czar
+    int      const    _attemptCount = 0; // attemptCount from czar
     std::string const _idStr{QueryIdHelper::makeIdStr(0, 0, true)}; // < for logging only
```
iagaponenko, line 189: Would the assignment operator work here to initialize the member like you did for 3 primitive data members above?

core/modules/wdb/QueryRunner.cc
```diff
@@ -441,7 +441,7 @@ class ChunkResourceRequest {
 };
 
 bool QueryRunner::_dispatchChannel() {
-    int fragNum = task->getQueryFragmentNum();
+    int fragNum = _task->getQueryFragmentNum();
```
iagaponenko, line 453: If you are not going to modify the variable then it should be declared as:
```
int const fragNum = _task->getQueryFragmentNum();
```

core/modules/wdb/QueryRunner.cc
```diff
@@ -441,7 +441,7 @@ class ChunkResourceRequest {
...
     proto::TaskMsg& m = *_task->msg;
```
iagaponenko, line 454: It looks like all uses of this variable `m` are gone now. Perhaps it's time to get rid of it as well?

core/modules/wdb/QueryRunner.cc
```diff
@@ -459,27 +459,25 @@ bool QueryRunner::_dispatchChannel() {
...
-        _fillSchema(res);
-        numFields = mysql_num_fields(res);
-        // TODO fritzm: revisit this error strategy
-        // (see pull-request for DM-216)
-        // Now get rows...
-        if (!_fillRows(res, numFields, rowCount, tSize)) {
-            erred = true;
+        if (!_cancelled) {
+            //&&& proto::TaskMsg_Fragment const& fragment(m.fragment(fragNum));
+            string const& query = _task->getQueryString();
```
iagaponenko, line 471: Is it safe to keep a reference to an internal string of a tasks? Tasks may disappear from queries for other reasons and result in mysterious crashes that will require debugging.

core/modules/wbase/SendChannel.cc
jgates108: I'll shorten it up a bit but I'm leaving lastTaskDone in there as it describes the return value without needing a comment.
Generally, I don't think forcing the mutex to be passed as an argument is that helpful, but the use of the mutex is in this class is weird enough to merit it.

core/modules/wcontrol/Foreman.cc
jgates108: The locking  of the mutex needs to happen outside of the SendChannel class to use the reliable lock_guard class. I don't think using a ponter over a reference makes much difference. I can try making a function that returns a lock_guard object.

core/modules/wdb/QueryRunner.cc
jgates108: It's still in use.

core/modules/wdb/QueryRunner.cc
jgates108: The _task object will be good until execution leaves QueryRunner::runQuery(). Everything else is called by runQuery().

core/modules/wsched/ScanScheduler.cc
jgates108: There are valid commands that are not of type Task. They are usually only seen when the program is shutting down.

core/modules/wbase/Task.cc
```diff
@@ -127,6 +130,38 @@ Task::~Task() {
...
+    ///&&&; // Determine if there should be sub-tasks.
+    int fragmentCount = taskMsg->fragment_size();
+    if (fragmentCount < 1) {
+        throw Bug("QueryRunner: No fragments to execute in TaskMsg");
+    }
+    for (int fragNum=0; fragNum<fragmentCount; ++fragNum) {
+        proto::TaskMsg_Fragment const& fragment(taskMsg->fragment(fragNum));
+        for (const std::string queryStr: fragment.query()) {
+            std::string qs(queryStr);
+            if (fragment.has_subchunks() && false == fragment.subchunks().id().empty()) {
```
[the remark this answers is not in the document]
jgates108: Actually, fragment.has_subchunks() == true and fragment.subchunks().id().empty() == false is valid and it's important that it goes to the else clause. I've added a comment to that affect.

core/modules/wbase/SendChannel.cc
jgates108: I added a SendChannelShared class as adding lock_guard<mutex> const& lock to one function is pointless if it isn't added to all of the functions where it needs to be locked. There are a lot of cases where SendChannel is used and no locking is needed. Polluting those with unneeded locks would be undesirable. 
SendChannelShared holds the mutex as well as other shared only member variables, and has a SendChannel member set at construction.

core/modules/wdb/QueryRunner.cc
jgates108: For the most part, including the function name doesn't make that much difference as the log includes file name and line number. However, using __func__ can certainly be useful in many cases where the message context is vague.

core/modules/wsched/BlendScheduler.cc
jgates108: This must be from and intermediate commit as it has already been removed.

core/modules/wsched/BlendScheduler.cc
jgates108: It is supposed to be overwritten. The point of the snail scheduler is that user queries that take too long get moved to the snail  scheduler, which is low priority.

core/modules/wbase/Task.cc
jgates108: From an earlier commit before this was moved into the constructor.

core/modules/wbase/Task.cc
jgates108: From an earlier commit before this was moved into the constructor.

core/modules/wbase/SendChannel.cc
iagaponenko: I know this isn't your fault as tis code has been around for "ages" (in Qserv life). But I still find it really scary that the calling the same method from one context may require some external locking while, doing so from some other context - won't. This is one of the things that makes the core hard to debug, and that may result in mysterious crashes. I'm not really suggesting that any significant effort was spent on reinforcing and streamlining locking behavior right now as it may be a significant. However, that needs to be done at some point in a context of some major code overhaul/refactoring project.

jgates108: The SendChannelShared class always requires locking, but the SendChannel class and it's children do not require locking. So I don't really understand the question.

PR #606, opened 2021-03-08
Added code to track/limit worker transmits with czar ids.

jgates108:


core/modules/qdisp/JobDescription.h
```diff
@@ -58,12 +59,13 @@ class ResponseHandler;
 class JobDescription {
 public:
     using Ptr =  std::shared_ptr<JobDescription>;
-    static JobDescription::Ptr create(QueryId qId, int jobId, ResourceUnit const& resource,
+    static JobDescription::Ptr create(qmeta::CzarId czarId, QueryId qId, int jobId,
```
iagaponenko, line 62: Since you're modifying the class then could you also declare the deleted copy constructor and the assignment operator for the class?
```
JobDescription(JobDescription const&) = delete;
JobDescription& operator=(JobDescription const&) = delete;
```

core/modules/wcontrol/TransmitMgr.h
```diff
@@ -62,51 +62,72 @@ class TransmitMgr {
     TransmitMgr& operator=(TransmitMgr const&) = delete;
     virtual ~TransmitMgr() = default;
 
-    int getTotalCount() { return _totalCount; }
-    int getTransmitCount() { return _transmitCount; }
-    int getAlreadyTransCount() { return _alreadyTransCount; }
+    int getTotalCount(int czarId);
```
iagaponenko, line 65: Normally the "get" methods are declared as `const`. If the only reason you don't do that is because of acquiring a lock onto the mutex then the standard approach is to declare the mutex as:
```
mutable std::mutex _mtx;
```
The `mutable' property would allow modifying the mutex's state from the `const` methods.

core/modules/wcontrol/TransmitMgr.h
```diff
@@ -62,51 +62,72 @@ class TransmitMgr {
...
-    void _take(bool interactive, bool alreadyTransmitting);
+    void _take(bool interactive, bool alreadyTransmitting, int czarId);
+
+    void _release(bool interactive, bool alreadyTransmitting, int czarId);
+
+    /// This class is used to store transmit information for a czar
+    class TransmitInfo {
+        using Ptr = std::shared_ptr<TransmitInfo>;
+        TransmitInfo() = default;
+        TransmitInfo(TransmitInfo const&) = default;
```
iagaponenko, line 94: If you  still go with the full-blown class then declaring the assignment operator  alongside the copy c-tor is the usual practice:
```
TransmitInfo& operator=(TransmitInfo const&) = default;
```

core/modules/wdb/QueryRunner.cc
```diff
@@ -318,8 +318,11 @@ void QueryRunner::_transmit(bool inLast, unsigned int rowCount, size_t tSize) {
         return;
     }
 
+    // Get the czar id number, if there is one.
+    int czarId = (_task->msg->has_czarid()) ? _task->msg->czarid() : -1;
```
iagaponenko, line 322: (Just curious) And what will happen if the identifier will be assigned to `-1`?

core/modules/wcontrol/TransmitMgr.h
```diff
@@ -62,51 +62,72 @@ class TransmitMgr {
...
 
-    void _release(bool interactive, bool alreadyTransmitting);
+        friend class TransmitMgr;
+        int _totalCount{0};
```
iagaponenko, line 97: Normally this needs to be done:
```
int _totalCount = 0;
```

core/modules/wcontrol/TransmitMgr.h
```diff
@@ -62,51 +62,72 @@ class TransmitMgr {
...
+    std::string dump() const;
+
 private:
-    void _take(bool interactive, bool alreadyTransmitting);
+    void _take(bool interactive, bool alreadyTransmitting, int czarId);
+
+    void _release(bool interactive, bool alreadyTransmitting, int czarId);
+
+    /// This class is used to store transmit information for a czar
+    class TransmitInfo {
```
iagaponenko, line 91: Since this is a private class within the class `TransmitMgr` then you may make all member `public` and avoid things like `friend class TransmitMgr;`. You may also drop all constructors because you are initializing counters to zeroes.

core/modules/wcontrol/TransmitMgr.h
```diff
@@ -62,51 +62,72 @@ class TransmitMgr {
...
+        using Ptr = std::shared_ptr<TransmitInfo>;
```
iagaponenko, line 92: The class has tiny footprint (3 primitive data members). I believe it makes very little sense to handle objects of the class via `shared_ptr` rather than passing them by values. The pointer management would just complicate things w/o any added benefits.

core/modules/proto/worker.proto
```diff
@@ -73,6 +73,7 @@ message TaskMsg {
     required int32 jobid = 11;
     required bool scaninteractive = 12;
     required int32 attemptcount = 13;
+    optional uint32 czarid = 14;
```
iagaponenko, line 76: I've noticed that you have type mismatch in the `proto` definition where you use `unsigned int` versus the rest of the code where you use `int`. My understanding is that you do tis because you're making `czarid` the optional field in the protocol definition. This complicates the rest of your code. I would suggest to avoid this and make it mandatory. In the end we do have such thing as `czarId` in Qserv.
`

core/modules/wcontrol/TransmitMgr.cc
```diff
@@ -37,63 +37,134 @@ namespace qserv {
 namespace wcontrol {
 
 
-void TransmitMgr::_take(bool interactive, bool alreadyTransmitting) {
-    LOGS(_log, LOG_LVL_DEBUG, "TransmitMgr::_take locking " << *this);
+int TransmitMgr::getTotalCount(int czarId) {
```
iagaponenko, line 40: `czarId` has to be of this type:
```
qmeta::CzarId
```

core/modules/wcontrol/TransmitMgr.cc
```diff
@@ -37,63 +37,134 @@ namespace qserv {
...
-    if (not interactive || _transmitCount >= _maxTransmits || _alreadyTransCount >= _maxAlreadyTran) {
+    auto const& iter = _czarTransmitMap.find(czarId);
+    if (iter != _czarTransmitMap.end()) {
+        return iter->second->_totalCount;
+    }
+    return 0;
+}
+
+
+int TransmitMgr::getTransmitCount(int czarId) {
```
iagaponenko, line 50: `czarId` has to be of this type:
```
qmeta::CzarId
```

core/modules/wcontrol/TransmitMgr.cc
```diff
@@ -37,63 +37,134 @@ namespace qserv {
...
+    unique_lock<mutex> uLock(_mtx);
+    auto const& iter = _czarTransmitMap.find(czarId);
+     if (iter != _czarTransmitMap.end()) {
+         return iter->second->_transmitCount;
+     }
+     return 0;
+}
+
+
+int TransmitMgr::getAlreadyTransCount(int czarId) {
```
iagaponenko, line 60: `czarId` has to be of this type:
```
qmeta::CzarId
```

core/modules/wcontrol/TransmitMgr.cc
```diff
@@ -37,63 +37,134 @@ namespace qserv {
...
+    unique_lock<mutex> uLock(_mtx);
+    auto const& iter = _czarTransmitMap.find(czarId);
+     if (iter != _czarTransmitMap.end()) {
+         return iter->second->_alreadyTransCount;
+     }
+     return 0;
+}
+
+
+void TransmitMgr::_take(bool interactive, bool alreadyTransmitting, int czarId) {
```
iagaponenko, line 70: `czarId` has to be of this type:
```
qmeta::CzarId
```

core/modules/wcontrol/TransmitMgr.cc
```diff
@@ -37,63 +37,134 @@ namespace qserv {
...
-    ++_transmitCount;
-    LOGS(_log, LOG_LVL_DEBUG, "TransmitMgr::_take locking done " << *this);
+    ++(info->_transmitCount);
+    LOGS(_log, LOG_LVL_DEBUG, "TransmitMgr take locking done " << dump());
 }
 
 
-void TransmitMgr::_release(bool interactive, bool alreadyTransmitting) {
-    LOGS(_log, LOG_LVL_DEBUG, "TransmitMgr::_release locking " << *this);
+void TransmitMgr::_release(bool interactive, bool alreadyTransmitting, int czarId) {
```
iagaponenko, line 107: `czarId` has to be of this type:
```
qmeta::CzarId
```

iagaponenko (changes requested):
Please, review my comments regarding inconsistent declarations of `czarId` in the code. It has to be all the same type `qproc::CzarId` everywhere.

core/modules/wcontrol/TransmitMgr.h
jgates108: I disagree. Doing arithmetic, specifically subtraction, with unsigned integers is very dangerous. While they are not supposed to go negative, if they ever do, it can be disastrous with unsigned.

core/modules/wdb/QueryRunner.cc
jgates108: -1 is perfectly valid as this is only on the worker where czarId is int. -1 was useful in that it indicated the czarId was not from a real source. 0 should work just as well, provided czarIds start at 1.

core/modules/wcontrol/TransmitMgr.h
jgates108: It is possible for the TransmitMgr::_release function to see a value of zero for at least _transmitCount if TransmitMgr::_take is waiting. I'm using the shared pointer to tell if _take is waiting before erasing the item. I can probably use a flag in TransmitInfo for the same purpose.

core/modules/wdb/QueryRunner.cc
jgates108: That slipped through. I changed the type to qmeta::CzarId and removed the if statement, since czarId is now required.

If there's a czarId == 0, it doesn't really make much difference, all czar's with czarId == 0 will share the same set of counts/restrictions. This shouldn't happen as long as the database is incrementing the czarId numbers. From this code's point of view, the only thing that matters is that all active czar's have unique id numbers.

PR #609, opened 2021-03-15
Changed to send header information using metadata.

jgates108:


core/modules/ccontrol/MergingHandler.cc
```diff
@@ -76,93 +78,84 @@ MergingHandler::~MergingHandler() {
...
     LOGS(_log, LOG_LVL_DEBUG, "From:" << _wName << " flush state="
          << getStateStr(_state) << " blen=" << bLen << " last=" << last);
-    if ((bLen < 0) || (bLen != (int)bufPtr->size())) {
-        if (_state != MsgState::RESULT_EXTRA) {
-            LOGS(_log, LOG_LVL_ERROR, "MergingRequester size mismatch: expected " <<
-                 bufPtr->size() << " got " << bLen);
-            // Worker sent corrupted data, or there is some other error.
-        }
+
+    if (bLen < 0) {
```
abh3, line 96: Can blen be zero? Is that a valid state?

core/modules/ccontrol/MergingHandler.cc
```diff
@@ -248,11 +242,11 @@ void MergingHandler::_setError(int code, std::string const& msg) {
     _error = Error(code, msg);
 }
 
-bool MergingHandler::_setResult(BufPtr const& bufPtr) {
+bool MergingHandler::_setResult(BufPtr const& bufPtr, int blen) {
     auto start = std::chrono::system_clock::now();
     std::lock_guard<std::mutex> lg(_setResultMtx);
     auto& buf = *bufPtr;
-    if (!ProtoImporter<proto::Result>::setMsgFrom(_response->result, &(buf[0]), buf.size())) {
+    if (!ProtoImporter<proto::Result>::setMsgFrom(_response->result, &(buf[0]), blen)) {
```
abh3, line 249: Since I don't see what BufPtr is now it clearly is no longer a container, right? So, I assume it needs an external blen. Any reason for that change?

core/modules/proto/ProtoHeaderWrap.h
```diff
@@ -50,8 +50,9 @@ class ProtoHeaderWrap {
     ProtoHeaderWrap() {};
     virtual ~ProtoHeaderWrap() {};
 
-    static std::string wrap(std::string& protoHeaderString);
+    static std::string wrap(std::string const& protoHeaderString);
     static bool unwrap(std::shared_ptr<WorkerResponse>& response, std::vector<char>& buffer);
+    static size_t getProtoHeaderSize();
```
abh3, line 55: Out of curiosity, where is it getting a header to tell you the size? This method, being static, makes it totally unclear.

core/modules/qdisp/QueryRequest.cc
```diff
@@ -299,20 +302,46 @@ bool QueryRequest::_importStream(JobQuery::Ptr const& jq) {
...
-    _queueAskForResponse(_askForResponseDataCmd, jq, true);
+
+    // Get the first header from metadata.
+    int expectedLen = proto::ProtoHeaderWrap::PROTO_HEADER_SIZE;
+    int len = expectedLen;
+    const char* buff = GetMetadata(len);
+    if (len != expectedLen) {
+        throw Bug("_importStream wrong header size=" + to_string(len) + " expected=" + to_string(expectedLen));
+    }
+    ResponseHandler::BufPtr bufPtr = make_shared<vector<char>>(len);
```
abh3, line 313: So, now I am confused. It seems that BufPtr is indeed a container. Since a vector carries the length with it, I would suggest not using an external length as you will constantly have to check for consistency.

core/modules/qdisp/QueryRequest.cc
```diff
@@ -414,20 +443,49 @@ void QueryRequest::_processData(JobQuery::Ptr const& jq, int blen, bool last) {
...
+        make_shared<vector<char>>(bufPtr->begin() + respSize, bufPtr->end());
+
+    // Read the result
     bool largeResult = false;
     int nextBufSize = 0;
-    bool flushOk = jq->getDescription()->respHandler()->flush(blen, bufPtr, last, largeResult, nextBufSize);
+    bool flushOk = jq->getDescription()->respHandler()->flush(respSize, bufPtr, last,
+                                                              largeResult, nextBufSize);
+    if (last) {
+        throw Bug("_processData result had 'last' true, which cannot be allowed.");
```
abh3, line 462: Explain to me why this can no longer be true. What prevents SSI from setting last? An explanatory comment here would be good.

core/modules/wbase/SendChannelShared.cc
```diff
@@ -0,0 +1,299 @@
...
+
+void SendChannelShared::_run() {
+    if (_threadStarted.exchange(true)) {
+        return;
+    }
+    if (isDead()) {
+        return;
+    }
+    std::thread thrd(&SendChannelShared::_transmitLoop, this);
+    _thread = std::move(thrd);
```
abh3, line 144: This likely is explained somewhere but under what cases is the channel joinable vs not joinable? Usually it's one or the other by design.

core/modules/wbase/SendChannelShared.cc
```diff
@@ -0,0 +1,299 @@
...
+        // no previous message it could attach its header to.
+        if (_firstTransmit.exchange(false)) {
+            // Put the header for the first message in metadata
+            // _metaDataBuf must remain valid until Finished() is called.
+            proto::ProtoHeader* thisPHdr = thisTransmit->header;
+            string thisHeaderString;
+            thisPHdr->SerializeToString(&thisHeaderString);
+            _metadataBuf = proto::ProtoHeaderWrap::wrap(thisHeaderString);
+            lock_guard<mutex> streamLock(streamMutex);
+            bool metaSet = _sendChannel->setMetadata(_metadataBuf.data(), _metadataBuf.size());
```
abh3, line 248: I presume my reading of the definition of setMetadata was in the wrong context and this setMetadata is actually calling the SSI version. right?

core/modules/wbase/SendChannelShared.cc
```diff
@@ -0,0 +1,299 @@
...
+            if (!metaSet) {
+                LOGS(_log, LOG_LVL_ERROR, "Failed to setMeta " << idStr);
```
abh3, line 250: So, this really makes debugging difficult. The SSI SetMedata returns the reason for the failure and that appears to get lost. So, when this error occurs it next to impossible to figure out what went wrong. Could you pass through all related return codes so they actually get in a log message? This request applies to all SSI methods as I see other method calls also drop the reason for a failure.

core/modules/xrdsvc/SsiRequest.cc
```diff
@@ -436,6 +436,18 @@ bool SsiRequest::replyStream(StreamBuffer::Ptr const& sBuf, bool last) {
...
 
+bool SsiRequest::sendMetadata(const char *buf, int blen) {
+    Status stat = SetMetadata(buf, blen);
+    if (stat == XrdSsiResponder::wasPosted) return true;
+    if (stat == XrdSsiResponder::notActive) {
+        LOGS(_log, LOG_LVL_ERROR, "failed to setMetadata notActive");
+    } else {
+        LOGS(_log, LOG_LVL_ERROR, "failed to setMetadata notPosted blen=" << blen);
+    }
+    return false;
```
abh3, line 447: Well, this will work as long as a new return code is not introduced. While that is highly unlikely, programming for the case would be a good conservative approach.

abh3 (approved):
This is a fairly large change so it's difficult to ascertain it's impact. I left a few suggestions. One of which is to not drop the SSI failure status codes in messages. I see in one place that a message is actually pout in the log when a failure occurs with the reason for the failure. However that message is also repeated without the code in another code section. I strongly suggest consistency here which will also reduce the number of similar messages related to the same failure. Otherwise, OK.

core/modules/ccontrol/MergingHandler.cc
jgates108: Yes, 0 indicates that there's nothing following this header and the message is done.

core/modules/ccontrol/MergingHandler.cc
jgates108: The result for this message and the header for the next message are in one buffer. The header is small, so it's just copied to another buffer. The result could be big, so it stays in the buffer and blen is the length of the buffer minus the header.

core/modules/proto/ProtoHeaderWrap.h
jgates108: It's a constant defined in the cc file.

core/modules/qdisp/QueryRequest.cc
jgates108: The 2 buffers (this result and next header) are put in one buffer so they can be sent as a single transmit. Once received, the buffer needs to be split into the separate parts so protobufs can decipher them and figure out how big the next bufer should be.

core/modules/qdisp/QueryRequest.cc
jgates108: The 'last' must show up as part of reading a header, not a result. If last is true while reading in the result, something really bad happened. I'll add a comment.

jgates108: It also looks like some shadowing is happening.

core/modules/wbase/SendChannelShared.cc
jgates108: The unit test tend to segv when you don't join threads. 

It looks like the threads need to go away anyway. When the system is under load, they just sit there doing nothing, even though they have something to do, and the process runs out of threads. Between the threads and unsent messages, a bunch of memory is eaten up as well.

core/modules/wbase/SendChannelShared.cc
jgates108: It should be.

core/modules/wbase/SendChannelShared.cc
jgates108: I'll try to fix that.

core/modules/wbase/SendChannelShared.cc
```diff
@@ -0,0 +1,223 @@
...
+
+    // If this is reallyLast or at least 2 items are in the queue, the transmit can happen
+    if (reallyLast || _transmitQueue.size() >= 2) {
+        bool scanInteractive = tData->scanInteractive;
+        // If there was an error, give this high priority.
+        if (erred) scanInteractive = true;
+        int czarId = tData->czarId;
+        // Don't wait if cancelled or error.
+        if (erred || cancelled) {
+            return _transmit(erred, largeResult, czarId);
```
abh3, line 117: I simply don't understand the logic here. _transmit() seems hell bent on sending something irrespective of erred or cancelled being set. Even by passing only erred doesn't give it a clue that it might have been cancelled. Why a) pass this at all, b) if you are erred or cancelled why is being called at all? I would think you just want to clean up by quickly deep-sixing the messages.

core/modules/wbase/SendChannelShared.h
```diff
@@ -0,0 +1,161 @@
...
+    /// many member functions require a StreamGuard argument.
+    using StreamGuard = std::lock_guard<std::mutex> const&;
+
+    SendChannelShared()=delete;
+    SendChannelShared(SendChannelShared const&) = delete;
+    SendChannelShared& operator=(SendChannelShared const&) = delete;
+
+    static Ptr create(SendChannel::Ptr const& sendChannel, wcontrol::TransmitMgr::Ptr const& transmitMgr);
+
+    ~SendChannelShared() = default;
```
abh3, line 60: This is also weird. Prior code made sure the channel was not active should the destructor be called and actually tried to kill it so that bad things wouldn't happen. Now it seems it doesn't matter yet the code hasn't changed from the "active" perspective. What am I missing?

core/modules/wbase/SendChannelShared.h
```diff
@@ -0,0 +1,161 @@
...
+    /// This should not be changed once set.
+    void setTaskCount(int taskCount);
+
+
+    /// Try to transmit the data in tData.
+    /// If the queue already has at least 2 TransmitData objects, addTransmit
+    /// may wait before returning. It's more efficient use of memory to
+    /// collect results from MariaDB as they are sent to the czar than
+    /// to read them all in at once.
+    bool addTransmit(bool cancelled, bool erred, bool last, bool largeResult,
```
abh3, line 101: OK, the comment doesn't really tell me how this interplays with removing the thread. Not sure what the comment really has to say but given your changes it makes me anyone else looking at the code wonder.

core/modules/wbase/SendChannelShared.cc
```diff
@@ -0,0 +1,223 @@
...
+}
+
+
+bool SendChannelShared::addTransmit(bool cancelled, bool erred, bool last, bool largeResult,
+                                    TransmitData::Ptr const& tData, int qId, int jId) {
+    QSERV_LOGCONTEXT_QUERY_JOB(qId, jId);
+    assert(tData != nullptr);
+
+    // This lock may be held for a very long time.
+    std::unique_lock<std::mutex> qLock(_queueMtx);
```
abh3, line 84: I know you said you removed the thread but this makes me think that there can be more threads that call this and they will simply pile up if it goes into a long wait. What makes it wait "long"; waiting for mySQL results?

core/modules/wbase/SendChannelShared.cc
```diff
@@ -0,0 +1,223 @@
...
+        bool scanInteractive = tData->scanInteractive;
+        // If there was an error, give this high priority.
+        if (erred) scanInteractive = true;
+        int czarId = tData->czarId;
+        // Don't wait if cancelled or error.
+        if (erred || cancelled) {
+            return _transmit(erred, largeResult, czarId);
+        }
+        // Limit the number of concurrent transmits.
+        wcontrol::TransmitLock transmitLock(*_transmitMgr, scanInteractive, largeResult, czarId);
```
abh3, line 120: I would guess this is another place where a wait will be imposed, yes?

abh3 (commented):
OK, I left some more questions about the code. I can't say I understand what it's actually doing other than moving messages around, sometimes sometimes not.

core/modules/wbase/SendChannelShared.cc
jgates108: If the czar is being flooded with results, this thread will sit and wait while holding onto the mutex. All the other threads waiting on this mutex would also need to write results back to the same czar, so they'd be stuck too even if there was no mutex. There's really not much that can be done if the czar is busy except wait for it to catch up. 

From past experience, if the workers don't wait for the czar, they crash as they run out of memory holding onto results they can't send.

core/modules/wbase/SendChannelShared.cc
jgates108: This is the only place. There are lots of different SendChannelShared objects all fighting to send their results back to the czars. TranmitLock/TransmitMgr try to keep the amount of traffic from this worker at reasonable levels. Basically 50 per czar, with each czar having its own count.

core/modules/wbase/SendChannelShared.cc
jgates108: If there was an error, there is probably an error message that needs to sent to the czar. It's probably best if the czar starts retrying this query, or more likely squashing the entire user query asap.  If it was cancelled, the SendChannel is likely dead, so there's probably nothing that can be done. Just finish up asap to free up this this thread and resources.

core/modules/wbase/SendChannelShared.h
jgates108: I'll take look. Mostly the changes here had to do with stopping a thread that no longer exists but it does seem like the SendChannel should be checked.

jgates108: I put the kill() back, but it doesn't do much besides set a flag. When there was a thread, this would cause the thread to finish quickly without trying to send anything. Since the thread is gone, it doesn't have much effect anymore. SendChannel cleans up itself when it is deleted.

core/modules/xrdsvc/SsiRequest.cc
jgates108: Changed to a switch statement.

PR #612, opened 2021-04-16
Tickets/dm 29717

iagaponenko:


core/modules/replica/Configuration.cc
```diff
@@ -39,79 +39,163 @@ namespace {
...
-void Configuration::setQservMasterDatabasePassword(string const& newPassword) {
-    string result = newPassword;
-    swap(result, _qservMasterDatabasePassword);
+// ---------------
+// The static API.
+// ---------------
+
+void Configuration::setQservCzarDbUrl(string const& url) {
+    util::Lock const lock(_classMtx, _context(__func__));
+    if (url.empty()) {
```
andy-slac, line 78: This check can be done before locking, not that it matters much in this case but generally it's a good practice.

core/modules/replica/Configuration.cc
```diff
@@ -39,79 +39,163 @@ namespace {
...
 void Configuration::setDatabaseAllowReconnect(bool value) {
-    swap(value, _databaseAllowReconnect);
+    util::Lock const lock(_classMtx, _context(__func__));
+    _databaseAllowReconnect = value;
+}
+
+
+bool Configuration::databaseAllowReconnect() {
+    util::Lock const lock(_classMtx, _context(__func__));
+    return _databaseAllowReconnect;
```
andy-slac, line 126: For simple data types you could just user `atomic<>` instead of a global mutex, mutex-level synchronization is probably an overkill in this case. Same is probably true for URL strings.

core/modules/replica/HttpModule.cc
```diff
@@ -66,31 +66,23 @@ string HttpModule::context() const {
...
 
 
 shared_ptr<css::CssAccess> HttpModule::qservCssAccess(bool readOnly) const {
     auto const config = controller()->serviceProvider()->config();
+    // Use all parmeters of the connection from the czar's MySQL connection parameters object.
+    auto const connectionParams = Configuration::qservCzarDbParams("qservCssData");
     map<string, string> cssConfig;
     cssConfig["technology"] = "mysql";
-    // Address translation is required because CSS MySQL connector doesn't set
+    // Address translation is requird because CSS MySQL connector doesn't set
```
andy-slac, line 79: It was correct before the change

andy-slac (approved):
Looks good, few minor comments/questions.

core/modules/replica/Configuration.cc
```diff
@@ -39,79 +39,163 @@ namespace {
...
 // These (static) data members are allowed to be changed, and they are set
 // globally for an application (process).
 bool         Configuration::_databaseAllowReconnect = true;
 unsigned int Configuration::_databaseConnectTimeoutSec = 3600;
 unsigned int Configuration::_databaseMaxReconnects = 1;
 unsigned int Configuration::_databaseTransactionTimeoutSec = 3600;
-string       Configuration::_qservMasterDatabasePassword = "";
-string       Configuration::_qservWorkerDatabasePassword = "";
+string       Configuration::_qservCzarDbUrl = "mysql://qsreplica@localhost:3306/qservMeta";
+string       Configuration::_qservWorkerDbUrl = "mysql://qsreplica@localhost:3306/qservw_worker";
```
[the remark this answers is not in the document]
n8pease: my experience is that port number does not force it to use TCP; IMO it would be better for my use case if the default was '127.0.0.1` (which does use TCP)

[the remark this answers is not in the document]
iagaponenko: The current implementation of the class `replica::DatabaseMySQL` forces TCP by:
```
    // Only allow TCP/IP, no UNIX sockets for now
    enum mysql_protocol_type const prot_type = MYSQL_PROTOCOL_TCP;
    mysql_options(_mysql, MYSQL_OPT_PROTOCOL, &prot_type);
```
Sockets are pretty much useless in Kubernetes due to various storage locality constraints.

admin/tools/docker/replication/container/tools/README
```diff
@@ -6,54 +6,27 @@
...
+#   ~/group
+#
+# User definition is needed to run the Docker container collecting dependencies
+# under the current user.
 
 TAG="qserv/replica:tools-$(git describe --dirty --always)" 
-admin/tools/docker/replication/container/tools/build.sh $PWD gke "$TAG"
-
+admin/tools/docker/replication/container/tools/build.sh $PWD "$TAG"
+admin/tools/docker/replication/container/tools/build.sh $PWD "$TAG" --local-user
```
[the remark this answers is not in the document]
iagaponenko: One of them, depending on if there is a user entry in the local `/etc/passwd` file.
I should also note that this mechanism for building the binary containers is only used at NCSA. I was deployed 2 years ago to avoid dealing with monstrous 5 GB containers. It's supposed to be replaced with Nate's `lite` version.

core/modules/replica/Configuration.cc
iagaponenko: Good point.

admin/tools/docker/replication/container/tools/build.sh
```diff
@@ -2,47 +2,64 @@
...
-esac
-TAG="$3"
-if [ -z "$TAG" ]; then
-    TAG="qserv/replica:tools"
 fi
-
+PASSWD=$HOME/passwd
+GROUP=$HOME/group
+LOCAL_USER_FLAG="$3"
+if [ ! -z "$LOCAL_USER_FLAG" ]; then
```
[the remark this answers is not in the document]
iagaponenko: 👍

core/modules/replica/Configuration.cc
iagaponenko: I too was considering atomics. Then decided not to because I would still have to use the mutex for strings. I wish operations over strings in C++ were thread-safe.

PR #613, opened 2021-04-23
Tickets/dm 29840

jgates108:


core/modules/qdisp/Executive.cc
```diff
@@ -372,7 +375,10 @@ void Executive::_setup() {
     if (_config.serviceUrl.compare(_config.getMockStr()) == 0) {
         _xrdSsiService = new XrdSsiServiceMock(this);
     } else {
-        _xrdSsiService = XrdSsiProviderClient->GetService(eInfo, _config.serviceUrl.c_str()); // Step 1
+        if (_xrdSsiServiceStatic == nullptr) {
+            _xrdSsiServiceStatic = XrdSsiProviderClient->GetService(eInfo, _config.serviceUrl.c_str());
+        }
+        _xrdSsiService = _xrdSsiServiceStatic; // Step 1
     }
```
iagaponenko, line 382: Is there a chance to run into a race condition with the static initialization in case if there will be more than one instance of `Executive`?
Perhaps getting rid of `_xrdSsiService` and replacing it with a call to the static method returning `_xrdSsiServiceStatic` would be a more reliable option:
```
# File: Executive.h
class Executive {
public:
    static XrdSsiService* getXrdSsiService() {
        std::lock_guard<std::mutex> lock(_staticInitMtx);
        if (_xrdSsiServiceStatic == nullptr) {
            _xrdSsiServiceStatic = XrdSsiProviderClient->GetService(eInfo, _config.serviceUrl.c_str());
        }
        return _xrdSsiServiceStatic;
    }

private:
    static XrdSsiService* _xrdSsiServiceStatic;
    static std::mutex _staticInitMtx;
};
Note that this will do the lazy initialization of the static pointer before its first use.

File: core/modules/wbase/SendChannel.cc [hunk 2]
```diff
@@ -188,6 +188,14 @@ bool SendChannel::kill() {
 }
 
 
+bool SendChannel::isDead() {
+    if (_dead) return true;
+    if (_ssiRequest == nullptr) return true;
```
iagaponenko, line 193: This question is for my own education - how it's possible that `_ssiRequest` was the `nullptr`? Is this because of a chance of constructing `SendChannel` using the default constructor? If that's the case then why one would use the default constructor at all? My understanding is that such a constructor would produce a completely useless object.

core/modules/wbase/Task.h
```diff
@@ -138,10 +138,12 @@ class Task : public util::CommandForThreadPool {
     char timestr[100]; ///< ::ctime_r(&t.entryTime, timestr)
     // Note that manpage spec of "26 bytes"  is insufficient
 
+    /// Cancel the query in progress and set set _cancelled.
```
iagaponenko, line 141: Extra `set set` in the comment

core/modules/wdb/QueryRunner.cc
```diff
@@ -110,6 +110,9 @@ QueryRunner::QueryRunner(wbase::Task::Ptr const& task,
 bool QueryRunner::_initConnection() {
     mysql::MySqlConfig localMySqlConfig(_mySqlConfig);
     localMySqlConfig.username = _task->user; // Override with czar-passed username.
+    if (_mysqlConn != nullptr) {
+        LOGS(_log, LOG_LVL_ERROR, "QueryRunner::_initConnection _mysqlConn not nullptr _mysqlConn=" << _mysqlConn.get());
+    }
```
iagaponenko, line 115: Is this really possible, and if so then why? If this is happening because of a logic error in the implementation then the process must be aborted right away. If this is happening because of a memory corruption - the same thing, abort right away to investigate the problem rather than hiding it "under the carpet". If this is because of race condition - then a reason for the race condition had to be investigated.  In aby case, aborting the application would be the only right step after discovering and reporting this condition.

core/modules/wdb/QueryRunner.cc
```diff
@@ -435,6 +443,7 @@ bool QueryRunner::_dispatchChannel() {
     // Collect the result in _transmitData. When a reasonable amount of data has been collected,
     // or there are no more rows to collect, pass _transmitData to _sendChannel.
     try {
+        _initTransmit(); // set _transmit
```
iagaponenko, line 446: Is this a bug fix? Why it wasn't here before, and what effect it will cause? It's not mentioned in the description for the ticket.

core/modules/wdb/QueryRunner.cc
```diff
@@ -494,13 +502,10 @@ bool QueryRunner::_dispatchChannel() {
...
-    // validation is needed.
-    auto sChannel = _task->sendChannel;
-    if (sChannel != nullptr) {
-        lock_guard<mutex> streamLock(sChannel->streamMutex);
-        sChannel->kill(streamLock);
-    }
+
+    // QueryRunner::cancel() should only be called by Task::cancel().
+    // Task::cancel() should handle cleaning up the sendChannel().
+
```
iagaponenko, line 508: This explanation is better to be put into the documentation section of the method in the class's header.

iagaponenko (approved):
I have reviewed and approved the change. Please, consider my comments & questions.

core/modules/wbase/SendChannel.cc
jgates108: It seems to show up in unti tests.

core/modules/wdb/QueryRunner.cc
jgates108: It isn't supposed to be possible, but it, or something similar happened in one of the cluster crashes. The following line called the destructor, which shouldn't happen. So this is looking for some insight into what went wrong if it happens again.

core/modules/wdb/QueryRunner.cc
jgates108: It's a bug fix for a problem that only occurs when a query is cancelled and its very hard to reproduce on a local installation, but the cluster notices right away. If the query was cancelled, _initTransmit() wouldn't be called, and there's a segv (this killed a lot of workers on the cluster) when this function tries to build an error message. Making sure _transmit is valid prevents the crash.

jgates108: Also, before this ticket, _cancelled was never true, so this crash would never happen.

core/modules/qdisp/Executive.cc
jgates108: It does need mutex, good catch. I'm not sure adding a function really improves anything, _setup is already pretty small.

core/modules/wdb/QueryRunner.cc
jgates108: QueryRunner::cancel() has comments in the header saying it should only be called by Task::cancel(). The other comment line is not completly accurate, so I removed it.

PR #615, opened 2021-04-29
Tickets/dm 10929

jgates108:


core/modules/wdb/QueryRunner.h
```diff
@@ -148,7 +148,7 @@ class QueryRunner : public wbase::TaskQueryRunner, public std::enable_shared_fro
     std::atomic<bool> _cancelled{false};
     std::atomic<bool> _removedFromThreadPool{false};
     mysql::MySqlConfig const _mySqlConfig;
-    std::unique_ptr<mysql::MySqlConnection> _mysqlConn;
+    std::unique_ptr<mysql::MySqlConnection> _mysqlConn{nullptr};
```
iagaponenko, line 151: That is not needed because the `unique_ptr` is initialized with `nullptr`.

core/modules/wdb/QueryRunner.h
```diff
@@ -158,7 +158,7 @@ class QueryRunner : public wbase::TaskQueryRunner, public std::enable_shared_fro
 
     /// Used to limit the number of open MySQL connections.
     std::shared_ptr<wcontrol::SqlConnMgr> const _sqlConnMgr;
-    std::atomic<bool> _runQueryCalled{false};
+    std::atomic<bool> _runQueryCalled{false}; ///< If runQuery gets called twice, the scheduler messed up.
```
iagaponenko, line 161: I'm seeing this in the context of the commit called "Added log messages to indicate possible cause for crash". However, I'm not seeing anything in the commit that would do what the commit message says. Was there any intermediate code that got later removed?

core/modules/wdb/QueryRunner.h
jgates108: There was a crash where gdb said the destructor was called for _mysqlConn when it should have been nullptr. The only ways I can see that happening are:
1) memory corruption
2) Somehow the scheduler scheduled this task to run twice. (log messages)
3) It wasn't initialized to nullptr.

I suspect 1 is most likely, but the log message and other changes are to help rule out other possibilities, if that crash ever happens again.

core/modules/wdb/QueryRunner.h
jgates108: Could have been.

jgates108: I think the other changes were moved  to a different ticket that has already been merged and these are the differences that remained.

iagaponenko (approved):
Looks good to me.

PR #616, opened 2021-05-05
Follow up the migration of configurations in the package qserv_testdata

iagaponenko:


admin/python/lsst/qserv/admin/dataDuplicator.py
```diff
@@ -69,15 +70,15 @@ def _runIndex(self):
         """
 
         for table in self._tables:
-            if os.path.isfile(os.path.join(self._cfgDirname, table + '.json')) is False:
+            if os.path.isfile(os.path.join(self._cfgDirname, "partition", table + '.json')) is False:
```
n8pease, line 73: I'd cache the joined file name and use it again on line 79.

n8pease:
one small nit but looks fine otherwise.

PR #618, opened 2021-05-09
Fixed a bug in using CSS for deleting tables

iagaponenko:


core/modules/replica/HttpIngestModule.cc
```diff
@@ -621,7 +621,7 @@ json HttpIngestModule::_deleteTable() {
 
     if (cssAccess->containsDb(databaseInfo.name)) {
         if (cssAccess->containsTable(databaseInfo.name, table)) {
-            cssAccess->dropTable(databaseInfo.name, databaseInfo.name);
+            cssAccess->dropTable(databaseInfo.name, table);
```
andy-slac, line 624: Is not there still a race condition between checking and deleting in case of concurrent deletes? Or is it harmless?

andy-slac (approved):
Looks reasonable, just one question out of curiosity.

core/modules/replica/HttpIngestModule.cc
iagaponenko: Very good point! In theory, it's possible that many clients would call the service to delete the very same table. In reality, it's very unlikely. In general, these (and similar) REST services don't have protection against race conditions. The problem may affect this and other services. This is a topic I should revisit at some point. I have registered the following JIRA ticket for that: https://jira.lsstcorp.org/browse/DM-30102.

In the meantime, I'm going to reinforce the code to catch the exception `NoSuchTable` (defined in the header `css/CssError.h`) thrown by the method `CssAccess::dropTable` in case if the entry wasn't found in the CSS due to a possibility of the race condition.

PR #619, opened 2021-05-14
Tickets/dm 26574

iagaponenko:


core/modules/wpublish/ResourceMonitor.cc
```diff
@@ -78,16 +67,8 @@ unsigned int ResourceMonitor::count(int chunk,
...
-    ResourceCounter result = _resourceCounter;
-    return result;
-}
-
-
 json ResourceMonitor::statusToJson() const {
+    lock_guard<mutex> lock(_mtx);
     json result = json::array();
     for (auto&& entry: _resourceCounter) {
         result.push_back({entry.first, entry.second});
```
jgates108, line 74: I'm guessing that the race condition was caused by another thread altering _resourceCounter while in the loop.

jgates108 (approved):
The change looks good.

core/modules/wpublish/ResourceMonitor.cc
iagaponenko: Exactly!

PR #620, opened 2021-05-26
Tickets/dm 29880

iagaponenko:


core/modules/replica/ConfigDatabase.h
```diff
@@ -67,8 +67,14 @@ class DatabaseInfo {
     /// @return The names of all tables.
     std::vector<std::string> tables() const;
 
-    std::string directorTable;      // The name of the Qserv "director" table if any.
-    std::string directorTableKey;   // The name of the primary key column in the "director" table.
+    std::string directorTable;          // The name of the Qserv "director" table if any.
+    std::map<std::string,               // The table name (partitioned tables only!).
+        std::string> directorTableKey;  // The name of the table's key representing object identifiers.
```
andy-slac, line 72: I am a bit confused by language, what does `table's key` mean here? Is it a column name or FK constraint name?

core/modules/replica/ConfigDatabase.h
```diff
@@ -67,8 +67,14 @@ class DatabaseInfo {
...
+                                        // NOTES: (1) In the "dependent" tables the key represents the FK
+                                        // associated with the corresponding PK of the "director" table.
+                                        // (2) The key is allowed to be empty for the "dependent" tables
+                                        // that don't have any objectId-based association with
+                                        // the "director" table.
```
andy-slac, line 77: If there is no association with object table how the table can be dependent?

core/modules/replica/Configuration.h
```diff
@@ -544,11 +544,14 @@ class Configuration {
      * @param isDirectorTable An (optional) flag indicating if this is the "director"
      *   table of the catalog. Note there could be only one such table in a catalog,
      *   and this table must be "partitioned".
-     * @param directorTableKey The (optional) name of a column in the "director"
-     *   table which is used as an "object" identifier for astronomical objects.
-     *   This parameter applies to the director tables only. And if provided (and
-     *   allowed) the column must be found among the names of columns in a value
-     *   of parameter "columns".
+     * @param directorTableKey The (optional) name of a column representing object identifiers.
```
andy-slac, line 547: `directorTableKey` is also confusing, if this is a column name then it's probably better to call it a column, "key" has some meaning in the database context which does not help here.

andy-slac (approved):
Looks good, but I am confused by use of the word "key" when it actually means "column".

core/modules/replica/ConfigDatabase.h
iagaponenko: Accepted. It was a mistake to use this term. It was introduced in the very first version of the Replication system. Since there are many uses of this term in the code and in the comments to the code then I'd rather fix it in a separate refactoring effort. I have registered it in: https://confluence.lsstcorp.org/display/DM/Fix+incorrect+naming+convention+for+the+object+identifiers

core/modules/replica/ConfigDatabase.h
iagaponenko: This is an interesting question. I always thought that any partitioned table in Qserv is either the `director` or a `dependent` table of some `director`. The tables are partitioning is based either on the spatial coordinates or on the object identifiers leading to the corresponding entries of some other partitioned table (the `director` table). Does this mean I have a logic flow in the implementation of the Replication/Ingest system that (presently) allows a partitioned table that's is not explicitly presented to the system as the `director` not to have the objectId-based association to some  `director`. The current implementation still leaves this door "open". Should I fix this?

andy-slac: No idea. But I think that partitioned table that is partitioned based on its own ra/dec is a director table and it is supposed to have unique/PK objectId. I'm not sure that we had any example of partitioned table without objectId. But I would ask @fritzm for definitive answer.

PR #621, opened 2021-06-03
Tickets/dm 30095

iagaponenko:


core/modules/replica/IngestSvcConn.h
```diff
@@ -241,6 +242,17 @@ class IngestSvcConn: public IngestFileSvc,
     /// Buffer management class facilitating serialization/de-serialization
     /// of data sent over the network
     std::shared_ptr<ProtocolBuffer> const _bufferPtr;
+
+    /// A value of this flag is sent to a client in a response message in case
+    /// of a failure to indicate if the contribution could be retried. The default
+    /// state of the flag is flipped when irreversible changes to the destination
```
n8pease, line 248: "is flipped" - I think you want to say "set to false" (unless it's true that it can be flipped back to true the next time irreversible changes can be made?)

iagaponenko: Good point! I will fix it.

PR #622, opened 2021-06-03
Tickets/dm 30236

jgates108:


core/modules/qdisp/QueryRequest.cc
```diff
@@ -476,8 +483,7 @@ void QueryRequest::_processData(JobQuery::Ptr const& jq, int blen, bool xrdLast)
     }
 
     // Read the next header
-    largeResult = false;
-    nextBufSize = 0;
+    // Values for largeResult and nextBufSize will be filled in by flush().
```
iagaponenko, line 486: Does this apply to `last` as well? If so then it should be mentioned here as well.

core/modules/wbase/SendChannelShared.cc
```diff
@@ -180,19 +178,21 @@ bool SendChannelShared::_transmit(bool erred, bool largeResult, qmeta::CzarId cz
...
+                kill(streamLock, "metadata");
                 return false;
             }
         }
 
         // Put the data for the transmit in a StreamBuffer and send it.
         auto streamBuf = xrdsvc::StreamBuffer::createWithMove(thisTransmit->dataMsg);
         {
+            // Limit the number of concurrent transmits.
+            wcontrol::TransmitLock transmitLock(*_transmitMgr, scanInteractive, largeResult, czarId);
```
iagaponenko, line 190: Where the limit is set?
It would be good to leave some info on this subject here.

core/modules/wcontrol/SqlConnMgr.cc
```diff
@@ -53,7 +53,15 @@ void SqlConnMgr::_take(bool scanQuery) {
...
-    _tCv.notify_one();
+    // All threads threads must be checked as nothing will happen if one thread is
+    // notified and it is waiting for _maxScanSqlConnections, but a different
+    // thread could use _maxSqlConnections.
+    // This shouldn't  hurt performance too much, since at any given time,
+    // very few threads should be waiting. (They can only wait when first scheduled
+    // and the scheduler is limited to about 20-30 threads.)
+    // If things are backed up, it's terribly important to run any runable
+    // threads found.
+    _tCv.notify_all();
```
iagaponenko, line 64: How many threads would be notified by this in the worst-case scenario? How often this synchronization call would be made? I'm just curious about the potential performance impact of this change.

core/modules/wcontrol/TransmitMgr.cc
```diff
@@ -57,49 +57,23 @@ int TransmitMgr::getTransmitCount(qmeta::CzarId czarId) const {
...
+        // Interactive transmits never need to wait.
+        if (info._transmitCount >= _maxTransmits) {
             _tCv.wait(uLock, [this, &info](){ return info._transmitCount < _maxTransmits; });
-        } else {
-            _tCv.wait(uLock, [this, &info](){
-                return (info._transmitCount < _maxTransmits)
-                        && (info._alreadyTransCount < _maxAlreadyTran);
-            });
         }
+        ++(info._transmitCount);
```
iagaponenko, line 73: In the previous version of the code the interactive transmits were also being counted. The new version seems to exclude them. Was this done on purpose?

core/modules/wdb/QueryRunner.cc
```diff
@@ -445,6 +441,11 @@ bool QueryRunner::_dispatchChannel() {
     // or there are no more rows to collect, pass _transmitData to _sendChannel.
     try {
         _initTransmit(); // set _transmit
+        ChunkResourceRequest req(_chunkResourceMgr, tMsg);
+        ChunkResource cr(req.getResourceFragment(fragNum));
```
iagaponenko, line 445: These two lines were not present in the older version of the code. And I'm not seeing any use for `cr` in the rest of the method. What's the purpose of these statements?

core/modules/wpublish/QueriesAndChunks.cc
```diff
@@ -219,6 +219,8 @@ void QueriesAndChunks::removeDead() {
     for (auto const& dead : dList) {
         removeDead(dead);
     }
+    lock_guard<mutex> gdend(_deadMtx);
```
iagaponenko, line 222: What's the purpose of the lock? Could it be commented on here?

core/modules/xrdsvc/ChannelStream.cc
```diff
@@ -39,16 +39,19 @@ namespace {
...
 }
 
+using namespace std;
 
 namespace lsst {
 namespace qserv {
 namespace xrdsvc {
 
+/// Provide each Channel stream with a unique identifier.
+static atomic<uint32_t> localSeq(0);
```
iagaponenko, line 49: Why this is here, rather than being declared as a static member of class `ChannelStream` where it belongs to?

core/modules/xrdsvc/ChannelStream.cc
```diff
@@ -61,30 +64,30 @@ void ChannelStream::append(StreamBuffer::Ptr const& streamBuffer, bool last) {
...
-        LOGS(_log, LOG_LVL_DEBUG, "Waiting, no data ready");
+        LOGS(_log, LOG_LVL_DEBUG, "seq=" << to_string(_seq) << " Waiting, no data ready");
         _hasDataCondition.wait(lock);
     }
-    if (_msgs.empty() && _closed) { // We are closed and no more
-        // msgs are available.
-        LOGS(_log, LOG_LVL_DEBUG, "Not waiting, but closed");
+    if (_msgs.empty() && _closed) {
+        // It's closed and no more msgs are available.
+        LOGS(_log, LOG_LVL_INFO, "seq=" << to_string(_seq) << " Not waiting, but closed");
```
iagaponenko, line 90: I've noticed a change from the `DEBUG` level to `INFO`. How frequently this message is supposed to be printed? Does it report any useful information that has to be seen at the `INFO` level?

iagaponenko (approved):

admin/templates/configuration/etc/xrdssi.cnf
```diff
@@ -78,7 +78,5 @@ reservedinteractivesqlconn = 20
 
 [transmits]
 # Maximum number of concurrent transmits to a czar.
-maxtransmits = 50
-# If more than this number of large transmits is happening at once, wait to
-# start more transmits until some are done.
-maxalreadytransmitting = 10
+maxtransmits = 40
```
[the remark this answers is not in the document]
jgates108: No, there is nothing specific about it. I tried value from 3 to 50, and queries commpleted with all of them. Setting it too really large values would probably cause problems for the czar.

I expect this value is inversly proportional to the size of the cluster. The more workers, the smaller this number should be. This is also per czar, so if there are 2 czars, the worker can have 40 outgoing transmits for each czar.

core/modules/ccontrol/UserQuerySelect.cc
```diff
@@ -365,7 +365,7 @@ QueryState UserQuerySelect::join() {
     }
     if (successful) {
         _qMetaUpdateStatus(qmeta::QInfo::COMPLETED);
-        LOGS(_log, LOG_LVL_DEBUG, "Joined everything (success)");
+        LOGS(_log, LOG_LVL_INFO, "Joined everything (success)");
```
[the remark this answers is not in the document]
jgates108: This should be printed once per user query, so very little log spam.

core/modules/qdisp/Executive.cc
```diff
@@ -435,13 +436,17 @@ void Executive::_unTrack(int jobId) {
             incompleteJobs = _incompleteJobs.size();
             if (_incompleteJobs.empty()) _allJobsComplete.notify_all();
         }
-        if (!untracked || LOG_CHECK_LVL(_log, LOG_LVL_DEBUG)) {
+        auto sz = _incompleteJobs.size();
+        logSome = (sz < 50) || (sz%1000 == 0) || !untracked;
+        if (logSome || LOG_CHECK_LVL(_log, LOG_LVL_DEBUG)) {
```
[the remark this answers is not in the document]
jgates108: It list 1 out of every 1000 messages until it gets to the last 50, then it prints all of them. It prints a short list of unfinished job id numbers. It's very useful if the query doesn't complete, as you know at least some of the stuck job ids. With 180,000 chunks, that's about 230 log lines per full table scan.

core/modules/qdisp/Executive.cc
```diff
@@ -435,13 +436,17 @@ void Executive::_unTrack(int jobId) {
...
             // Log up to 5 incomplete jobs. Very useful when jobs do not finish.
             s = _getIncompleteJobsString(5);
         }
     }
-    LOGS(_log, (untracked ? LOG_LVL_DEBUG : LOG_LVL_WARN),
-         "Executive UNTRACKING " << (untracked ? "success":"failed") << "::" << s);
+    auto logLvl = logSome ? LOG_LVL_INFO : LOG_LVL_DEBUG;
+    LOGS(_log, (untracked ? logLvl : LOG_LVL_WARN),
+         "Executive UNTRACKING job=" << jobId << (untracked ? " success":" failed") << "::" << s);
```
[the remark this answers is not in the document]
jgates108: The executive will 'TRACK' new jobs as it creates them and sends them out to workers. It then 'UNTRACK' each job when the worker has finished with it and the result has been added to the result table. 

Success, means the worker successfully answered the query and returned a valid result. Failure means there was an error of some kind and user query has failed.

core/modules/wcontrol/SqlConnMgr.cc
jgates108: The maximum number is a result of the scheduler configuration, and that is largely a function of how many cores the worker has. This is currently about 20 on all of the systems (12 cores and the scheduler probably wont function well without at least 20 threads). If we get some systems with 30 cores, then the scheduler configuration values need to change to run more threads.

If the scheduler isn't running anything, it will wake up 20 threads. If the scheduler is busy, and all threads are running, there is nothing to wake up. If all 20 threads are waiting for an SQL connection, 20 will be woken up and 19 will go back to sleep, which is probably the worst case for us.

jgates108: Nothing will be waiting on this unless transits back to the czar are taking a long time. The SQL connections cannot be closed until all the results have been. When the worker reads all the results first, then sends them, then the worker tends to run out of memory and crash. So, the results are read and transmited in reasonable sized pieces, and the SQL connection needs to be held open until all transmits for the task are done.

core/modules/wcontrol/TransmitMgr.cc
jgates108: Yes, this was the source of the lockup problem. There was an effort to prioritize large results, but either the logic here was faulty and releases were not happening correctly, and/or the interactions between TransmitMgr, SqlConnectionMgr, and double buffering could cause the system could lock up.

After simplifying the logic here, I've been unable to reproduce the lockup. Prioritizing large results was probably of limited value.

core/modules/wdb/QueryRunner.cc
jgates108: They were there. They have always been there. Exactly where, I'd have to look. I moved them into the try block from line 435, but I don't remeber where they were before that.

These 2 lines create the temporary subchunk tables need for this task. Eventually, through a maze of code, they call the load function that creates the database and in memory tables that we have been discussing in another ticket the past couple of weeks.

The TODO here has to do with the fact that any other query in this chunk, using the subchunks, should be able to reuse these tables, but these tables only last until the end of the try block. They should really last until thes scheduler is done with all the tasks in this chunk, which is a bit tricky to do.

core/modules/xrdsvc/ChannelStream.cc
jgates108: Currently, maybe once per job. This function is what xrootd calls when it wants data. This message is printed when there is no more data to send.

core/modules/wpublish/QueriesAndChunks.cc
jgates108: It just protects _deadQueries. I put the lock and the log line in debug log lvl check so it only locks when it is logging the line.

iagaponenko: Should the lock be acquired before iterating over `dList` and calling `removeDead`?

jgates108: removeDead(list) operates on different lists, locking the appropriate mutexes as it needs.

PR #623, opened 2021-06-03
Redefined the default behavior of the unit test testNamedMutexRegistry

iagaponenko:


core/modules/replica/testNamedMutexRegistry.cc
```diff
@@ -171,6 +172,18 @@ BOOST_AUTO_TEST_CASE(NamedMutexRegistryTest2) {
 
 BOOST_AUTO_TEST_CASE(NamedMutexRegistryTest3) {
 
+    bool testMaxCpuUsage = false;
+    for (int i = 1; i < framework::master_test_suite().argc; ++i) {
+        string const arg = string(framework::master_test_suite().argv[1]);
```
n8pease, line 177: is `1` supposed to be `i` here?

core/modules/replica/testNamedMutexRegistry.cc
```diff
@@ -171,6 +172,18 @@ BOOST_AUTO_TEST_CASE(NamedMutexRegistryTest2) {
 
 BOOST_AUTO_TEST_CASE(NamedMutexRegistryTest3) {
 
+    bool testMaxCpuUsage = false;
+    for (int i = 1; i < framework::master_test_suite().argc; ++i) {
```
n8pease, line 176: please somehow document the use of this command line argument. (IMO it doesn't have to be formal but it should be easy to understand.)

core/modules/replica/testNamedMutexRegistry.cc
iagaponenko: I'm not sure I understood the question. In C/C++ the `argv` array is zero-based, where the index `[0]` represents a path to the application itself. Actual command-line arguments start with index `1`.

PR #624, opened 2021-06-09
Tickets/dm 30666

iagaponenko:


core/modules/replica/HttpModule.cc
```diff
@@ -112,4 +113,33 @@ bool HttpModule::localLoadSecondaryIndex(string const& database) const {
...
+        auto const transactionInfo = databaseServices->transaction(transactionId);
+        database = transactionInfo.database;
+    }
+    debug(func, "database=" + database);
+
+    auto const databaseInfo = config->databaseInfo(database);
+    if (databaseInfo.isPublished) {
+        throw HttpError(
+                context() + "::" + func, "database '" + databaseInfo.name + " is already published.");
+    }
```
n8pease, line 141: why is it an error to get database info for a published database?

iagaponenko: The services that are using the method require that the database was still open for ingests (not _published_).

n8pease:
The optional parameter to throw if published while getting database info is an improvement, but since it's defaulted to true it's still going to be non-obvious from where it's called that the method may throw for a not-get related reason. But I said I'd be ok with the option so I'll let it be.

n8pease:
I thought there was a way to approve a PR but I can't find the button. Maybe because it's still failing travis? Anyway, I'm done, will mark the ticket.

PR #627, opened 2021-06-22
DM-30803: DP0.1 JOIN fail between object and truth_match tables

n8pease:


core/modules/qproc/testQueryAnaDuplSelectExpr.cc
```diff
@@ -136,18 +136,5 @@ BOOST_AUTO_TEST_CASE(Simple) {
     BOOST_CHECK(context);
 }
 
-BOOST_AUTO_TEST_CASE(SameNameDifferentTable) {
```
iagaponenko, line 139: What was the reason for removing this unit test?

iagaponenko (approved):
LGTM

core/modules/qproc/testQueryAnaDuplSelectExpr.cc
n8pease: we discussed this in the cowork meeting but to follow up here; this ticket adds support for having the same column name in different tables, and this test verifies that same condition fails, so this test is no longer needed. I did try to change the logic on this test to verify that having the same condition does _not_ fail, but because the mock is incomplete for further query analysis the test still fails (IIRC it's a problem related to having necessary info to join the tables). I added an integration test (case01 query 1104) for this test case so we are covered there.

PR #630, opened 2021-06-30
Changed to prevent select * and count(*) from being interactive.

jgates108:


admin/templates/configuration/etc/qserv-czar.cnf
```diff
@@ -99,6 +99,11 @@ xrootdSpread = 0
 # This is per user query and important milestones ignore this limit.
 qMetaSecsBetweenChunkCompletionUpdates = 59
 
+# Set queryDistributionTestVer to non-zero to treat simple queries,
+# such as "SELECT COUNT(*) FROM <table>" as interactive queries to stress test
```
iagaponenko, line 103: And `SELECT * ` as well?

core/modules/qana/ScanTablePlugin.cc
```diff
@@ -182,13 +182,19 @@ ScanTablePlugin::_findScanTables(query::SelectStmt& stmt,
...
-    // plugin's applyFinal
-    if (hasSelectColumnRef || hasSelectStar) {
+    // Even trivial queries need to use full table scans to avoid crippling the czar
+    // with heaps of high priority, but very simple queries. Unless there are
+    // factors that greatly restrict how many chunks need to be read, it needs to be
+    // a table scan.
+    // For system testing, it is useful to see how the system handles large numbers
+    // trivial queries as the amount of work done by the workers is minimal. This
+    // highlights the cost of czar-worker communications.
+    if (!queryDistributionTestVer || hasSelectColumnRef) {
```
iagaponenko, line 197: What happened to `hasSelectStar`? Will they be now treated as interactive queries?

jgates108: It never did anything. It was just a placeholder (see comment at line 185). Filling in an accurate value for that is not easy. Since 'queryDistributionTestVer' should never be true in a production environment, it shouldn't matter.

admin/templates/configuration/etc/qserv-czar.cnf
jgates108: Treating SELECT COUNT(*) and SELECT * separately would take real effort. As stated in the other comment, that configuration value should always be false outside of specific testing needs.

core/modules/qana/ScanTablePlugin.cc
iagaponenko: Thanks! Then it's okay to merge the PR and get it tested.

core/modules/qana/ScanTablePlugin.cc
```diff
@@ -184,7 +184,7 @@ ScanTablePlugin::_findScanTables(query::SelectStmt& stmt,
     }
 
     auto czar = czar::Czar::getCzar();
-    bool queryDistributionTestVer = czar->getQueryDistributionTestVer();
+    bool queryDistributionTestVer = (czar == nullptr) ? 0 : czar->getQueryDistributionTestVer();
```
iagaponenko, line 187: Is this meant to address issues with unit testing?

iagaponenko (approved):

core/modules/qana/ScanTablePlugin.cc
jgates108: Yes

PR #631, opened 2021-07-09
Tickets/dm 30074

iagaponenko:


core/modules/replica/Configuration.h
```diff
@@ -589,6 +596,22 @@ class Configuration {
...
 
+    /**
+     * Make sure this worker is known in the configuration
+     * @param name The name of a worker.
+     * @throws std::invalid_argument if the worker is unknown
+     */
+    void assertWorkerIsValid(std::string const& name);
+
+    /**
+     * Make sure workers are now known in the configuration and they're different
```
andy-slac, line 607: `now`?

core/modules/replica/Configuration.h
```diff
@@ -589,6 +596,22 @@ class Configuration {
...
+     * @param workerOneName The name of the first worker in the comparison.
+     * @param workerTwoName The name of the second worker in the comparison.
+     * @throws std::invalid_argument if either worker is unknown
```
andy-slac, line 610: What happens if they are both known but identical?

core/modules/replica/FileUtils.cc
```diff
@@ -279,6 +279,44 @@ string FileUtils::createTemporaryFile(string const& baseDir,
...
+                }
+
+            }
+        }
+
+        // Create a temporary file to test write permissions set for the folder.
+        // The file will get removed upon the completion of the test.
+        try {
+            string const tmpFileName = createTemporaryFile(folder, ".test-write-permissions-");
+            fs::remove(fs::path(tmpFileName), ec);
```
andy-slac, line 310: This will not throw if you pass `ec`? The message below implies that you also expecting exception from `remove`?

core/modules/replica/testFileUtils.cc
```diff
@@ -156,4 +156,89 @@ BOOST_AUTO_TEST_CASE(FileUtils_createTemporaryFile) {
...
+        });
+
+        // Make another run w/o attempting creating a folder
+        BOOST_REQUIRE_NO_THROW({
+            FileUtils::verifyFolders("TEST", folders, !createMissingFolders);
+        });
+
+        // Now, make the best attempt to delete the folder. Ignore any errors.
+        try {
+            fs::remove(fs::path(uniqueFolderPath), ec);
```
andy-slac, line 230: Same comment, I think this does not throw if `ec` is passed?

andy-slac (approved):
Looks good, few minor comments.

core/modules/replica/FileUtils.cc
iagaponenko: Actually, I don't care about any exceptions thrown by `remove as there is nothing that could be done here. Leaving undeleted files in a folder is not going to be a problem since there is no such file created in the relevant folders when the system is starting. The only possible, yet highly unlikely, scenario (apart from the underlying file system going down at a time of the file removal request or some other system error) would be another process deleting the same file. This may happen if two instances of the Replication ingest system were running on the same set of folders AND if two processes managed to generate exactly the same file name using the temporary file generator `createTemporaryFile`.
The actual reason for the `try`  block here is the previous line - method `createTemporaryFile` that is making some the temporary file.

core/modules/replica/testFileUtils.cc
iagaponenko: You're right - I had no reason to put the file removal operation into the `try` block. Error processing for the operation is not needed.

core/modules/replica/FileUtils.cc
andy-slac: this is fine with me, the exception message below says `" failed to create/delete the temporary file ..."` which I interpreted as "create OR delete". Maybe just don't mention "delete" there?

iagaponenko: 👍

PR #634, opened 2021-07-21
Tickets/dm 31123

jgates108:


core/modules/wconfig/WorkerConfig.cc
```diff
@@ -75,6 +75,7 @@ WorkerConfig::WorkerConfig(const util::ConfigStore& configStore)
       _maxTasksBootedPerUserQuery(configStore.getInt("scheduler.maxtasksbootedperuserquery", 5)),
       _maxSqlConnections(configStore.getInt("sqlconnections.maxsqlconn", 800)),
       _ReservedInteractiveSqlConnections(configStore.getInt("sqlconnections.reservedinteractivesqlconn", 50)),
+      _bufferMaxTotalGB(configStore.getInt("transmit.buffermaxtotalgb", 41)),
```
iagaponenko, line 78: Just curious, why `41`? Is this just a random number?

core/modules/xrdsvc/SsiService.cc
```diff
@@ -102,6 +102,9 @@ SsiService::SsiService(XrdSsiLogger* log, wconfig::WorkerConfig const& workerCon
         throw wconfig::WorkerConfigError("Unrecognized memory manager.");
     }
 
+    int64_t bufferMaxTotalBytes = workerConfig.getBufferMaxTotalGB() * 1'000'000'000;
```
iagaponenko, line 105: This is prone to int32 overflow unless you add `ULL` in the end like in:
```
1'000'000'000ULL;
```

core/modules/xrdsvc/StreamBuffer.cc
```diff
@@ -38,17 +38,42 @@ LOG_LOGGER _log = LOG_GET("lsst.qserv.xrdsvc.StreamBuffer");
...
 
-std::atomic<size_t> StreamBuffer::_totalBytes(0);
+
+atomic<int64_t> StreamBuffer::_maxTotalBytes{40'000'000'000};
+atomic<int64_t> StreamBuffer::_totalBytes(0);
+mutex StreamBuffer::_createMtx;
+condition_variable StreamBuffer::_createCv;
+
+
+static void StreamBuffer::setMaxTotalBytes(int64_t maxBytes) {
```
iagaponenko: Is there a reason not to set this via the constructor of the class?

core/modules/xrdsvc/StreamBuffer.cc
```diff
@@ -38,17 +38,42 @@ LOG_LOGGER _log = LOG_GET("lsst.qserv.xrdsvc.StreamBuffer");
...
+using namespace std;
+
+
 namespace lsst {
 namespace qserv {
 namespace xrdsvc {
 
-std::atomic<size_t> StreamBuffer::_totalBytes(0);
+
+atomic<int64_t> StreamBuffer::_maxTotalBytes{40'000'000'000};
```
iagaponenko, line 49: Why signed integer? anything that it's about sizes has to be unsigned. The best type is `size_t`.

core/modules/xrdsvc/StreamBuffer.cc
```diff
@@ -38,17 +38,42 @@ LOG_LOGGER _log = LOG_GET("lsst.qserv.xrdsvc.StreamBuffer");
...
+atomic<int64_t> StreamBuffer::_totalBytes(0);
```
iagaponenko, line 50: Same question about a rationale behind using the signed integer type here.

core/modules/xrdsvc/StreamBuffer.cc
```diff
@@ -38,17 +38,42 @@ LOG_LOGGER _log = LOG_GET("lsst.qserv.xrdsvc.StreamBuffer");
...
+mutex StreamBuffer::_createMtx;
+condition_variable StreamBuffer::_createCv;
+
+
+static void StreamBuffer::setMaxTotalBytes(int64_t maxBytes) {
+    LOGS(_log, LOG_LVL_INFO, "StreamBuffer::setMaxTotalBytes maxBytes=" << maxBytes);
+    if (maxBytes < 0) {
+        throw invalid_argument("StreamBuffer::setMaxTotalBytes negative " + to_string(maxBytes));
+    }
+    if (maxBytes < 1.0e9) {
```
iagaponenko, line 60: Mixing integer and floating pointing comparison looks confusing (to me). Why not compare with?
```
1'000'000'000
```

core/modules/xrdsvc/StreamBuffer.cc
```diff
@@ -38,17 +38,42 @@ LOG_LOGGER _log = LOG_GET("lsst.qserv.xrdsvc.StreamBuffer");
...
-std::atomic<size_t> StreamBuffer::_totalBytes(0);
+
+atomic<int64_t> StreamBuffer::_maxTotalBytes{40'000'000'000};
+atomic<int64_t> StreamBuffer::_totalBytes(0);
+mutex StreamBuffer::_createMtx;
+condition_variable StreamBuffer::_createCv;
+
+
+static void StreamBuffer::setMaxTotalBytes(int64_t maxBytes) {
+    LOGS(_log, LOG_LVL_INFO, "StreamBuffer::setMaxTotalBytes maxBytes=" << maxBytes);
```
iagaponenko, line 56: I suggest using (here and elsewhere) the predefined constant `__func__` instead:
```
"StreamBuffer::" + string(__func__) + " setMaxTotalBytes maxBytes=" << maxBytes);
```
This would simplify the code refactoring effort should you (or someone else) decide to rename the metod.

core/modules/xrdsvc/StreamBuffer.cc
```diff
@@ -38,17 +38,42 @@ LOG_LOGGER _log = LOG_GET("lsst.qserv.xrdsvc.StreamBuffer");
...
+    if (maxBytes < 0) {
+        throw invalid_argument("StreamBuffer::setMaxTotalBytes negative " + to_string(maxBytes));
```
iagaponenko, line 58: Another optimization that makes this code a bit more readable/shorter is to define this variable at the beginning of each method:
```
string const context = "StreamBuffer::" + string(__func__) + " ";
```
After that you will just use this in the rest of the method:
```
throw invalid_argument(context + "negative " + to_string(maxBytes));
```
This technique helps with refactoring the names of methods or classes. It also makes exception throwing (and LSST logging) statements  less nosier.

core/modules/xrdsvc/StreamBuffer.cc
```diff
@@ -38,17 +38,42 @@ LOG_LOGGER _log = LOG_GET("lsst.qserv.xrdsvc.StreamBuffer");
...
+
+atomic<int64_t> StreamBuffer::_maxTotalBytes{40'000'000'000};
+atomic<int64_t> StreamBuffer::_totalBytes(0);
+mutex StreamBuffer::_createMtx;
+condition_variable StreamBuffer::_createCv;
+
+
+static void StreamBuffer::setMaxTotalBytes(int64_t maxBytes) {
+    LOGS(_log, LOG_LVL_INFO, "StreamBuffer::setMaxTotalBytes maxBytes=" << maxBytes);
+    if (maxBytes < 0) {
```
iagaponenko, line 57: You wouldn't have this check should the `maxBytes` be defined as an unsigned integer.

core/modules/xrdsvc/StreamBuffer.cc
```diff
@@ -38,17 +38,42 @@ LOG_LOGGER _log = LOG_GET("lsst.qserv.xrdsvc.StreamBuffer");
...
+}
+
 
 // Factory function, because this should be able to delete itself when Recycle() is called.
 StreamBuffer::Ptr StreamBuffer::createWithMove(std::string &input) {
-     Ptr ptr(new StreamBuffer(input));
-     ptr->_selfKeepAlive = ptr;
-     return ptr;
+    unique_lock<mutex> uLock(_createMtx);
+    if (_totalBytes >= _maxTotalBytes) {
```
iagaponenko, line 78: What if they both are equal to `0`, say because `_maxTotalBytes` was misconfigured and set to `0`?

core/modules/xrdsvc/StreamBuffer.h
```diff
@@ -88,7 +94,11 @@ class StreamBuffer : public XrdSsiStream::Buffer {
     Ptr _selfKeepAlive; ///< keep this object alive until after Recycle() is called.
     util::InstanceCount _ic{"StreamBuffer"}; ///< Useful as it indicates amount of waiting for czar.
 
-    static std::atomic<size_t> _totalBytes;
+    // Members associated with limiting memory use.
+    static std::atomic<int64_t> _totalBytes; ///< Total bytes currently in use by all StreamBuffer instances.
```
iagaponenko, line 98: Once again. An unsigned integer type like `size_t` is a more suitable type for `_totalBytes` and for `_maxTotalBytes`.

core/modules/wcontrol/TransmitMgr.h
```diff
@@ -29,15 +29,55 @@
...
+    LockCount() = default;
+
+    friend QidMgr;
+private:
+    int _take(); /// @return _totalCount.
+    int _release(); /// @return _totalCount.
+
+    std::atomic<int> _totalCount{0};
+    std::atomic<int> _count{0};
+    std::atomic<int> _maxCount{1};
```
iagaponenko, line 56: What's a reason for using the signed integer type for counters?

iagaponenko (approved):
Looks good to me

core/modules/wconfig/WorkerConfig.cc
jgates108: I'm not sure what the best number is. The value of 10 in the configuration file is also a guess (backed by Andy H.), but at least it seems to work. More experimentation is needed to find the optimal value, but it isn't essential to do this immediately.

core/modules/wcontrol/TransmitMgr.h
jgates108: I'm doing subtraction with these. Unsigned are inherently dangerous in subtraction and the int range is far more than I would ever expect to need for this purpose.

core/modules/xrdsvc/StreamBuffer.cc
jgates108: I'm doing subtraction, this is never compared to to an actual size, and never used to access a memory addres. Really bad things can happen if unsigned wraps around 0 due to some race conditon, so unless I really need to use unsigned, I avoid it.

core/modules/xrdsvc/StreamBuffer.cc
jgates108: Perhaps, but if someone puts a -1 in the config file, what happens? ConfigStore only has getInt(), nothing for unsigned. Making the check is pretty trivial.

core/modules/xrdsvc/StreamBuffer.cc
jgates108: Having _maxTotalBytes = 0 would be an error, hence the check above, and techncally, if was zero, it would have used all the memory it was allowed.

core/modules/xrdsvc/StreamBuffer.h
jgates108: I respectfully disagree.

core/modules/xrdsvc/StreamBuffer.cc
jgates108: I have some mixed feeling about this. If there's just one or two log statements, I don't worry about it. I like to cut and paste comments from the log file and then use that to search the code. If I search for " setMaxTotalBytes maxBytes=" in the code, I wont get any hits after making the above change. Once there gets to be more log statements in the function, it's more worthwhile. In this case, I made the change.

PR #636, opened 2021-07-30
Tickets/dm 31185

jgates108:


core/modules/wcontrol/TransmitMgr.h
```diff
@@ -45,6 +45,9 @@ class QidMgr;
 class LockCount {
 public:
     LockCount() = default;
+    LockCount(LockCount const&) = delete;
+    LockCount& operator=(LockCount const&) = delete;
+    ~LockCount() = default;
 
     friend QidMgr;
```
iagaponenko, line 52: This header has 2 "friends". Too many "friends" is a sign of the "spaghetti" code since this is a way to bypass the `private` protection barrier to get uncontrolled access to the class's data members. Perhaps, this class (`LockCount`) could be moved into either the CC file or into the private section of its only user - the class `QidMgr`?. Then all members of `LockCount` could be made public?

iagaponenko (approved):
LGTM.

core/modules/wcontrol/TransmitMgr.cc
```diff
@@ -66,12 +66,37 @@ int LockCount::_release() {
...
+}
+
 
 void QidMgr::_take(QueryId const& qid) {
     LockCount* lockCount;
     {
         lock_guard<mutex> uLock(_mapMtx);
         lockCount = &(_qidLocks[qid]);
+        lockCount->_maxCount.store(_maxCount);
+        _setMaxCount(_qidLocks.size());
```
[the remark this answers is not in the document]
jgates108: I think the previous example was just a cut and paste of this. 
I think the block of code is a bit more clear about the mutex duration than unlock(). Not much difference.

core/modules/wcontrol/TransmitMgr.h
jgates108: Nested classes is fine in this case. I'll make the change.

PR #637, opened 2021-08-17
Tickets/dm 31309

iagaponenko:


fritzm (approved):
Okay, but see question about ref to temp in string const& defaults?

PR #644, opened 2021-09-18
Tickets/dm 31682

iagaponenko:


core/modules/replica/DatabaseServicesMySQL.cc
```diff
@@ -2243,57 +2243,64 @@ vector<TransactionContribInfo> DatabaseServicesMySQL::transactionContribs(
...
-TransactionContribInfo DatabaseServicesMySQL::beginTransactionContrib(
-        TransactionId transactionId,
-        string const& table,
-        unsigned int chunk,
-        bool isOverlap,
-        string const& worker,
-        string const& url) {
+TransactionContribInfo DatabaseServicesMySQL::createdTransactionContrib(
+        TransactionContribInfo const& info,
+        bool failed) {
```
jgates108, line 2248: I think it would be better to modify ` TransactionContribInfo& info` instead of returning TransactionContribInfo, but it doesn't matter much either way.

core/modules/replica/schema/migrate-0-to-1.sql
```diff
@@ -0,0 +1,509 @@
...
+CREATE TABLE IF NOT EXISTS `controller` (
+
+  `id`  VARCHAR(255) NOT NULL ,
+
+  `hostname`  VARCHAR(255) NOT NULL ,
+  `pid`       INT          NOT NULL ,
+
+  `start_time`  BIGINT UNSIGNED NOT NULL ,
+
+  PRIMARY KEY (`id`)
```
jgates108, line 177: I don't think the extra empty lines are helpful in these create table statements.

core/modules/replica/SConscript
```diff
@@ -38,3 +38,7 @@ for f in env.Glob(os.path.join(path, "qserv-*.cc"), source=True, strings=True):
 # the above discovered binary sources
 
 standardModule(env, bin_cc_files=bin_cc_files, unit_tests="", test_libs="curl")
+
+# install schema files
+build_data = DefaultEnvironment()['build_data']
+build_data['install'] += env.Install("$prefix/share/qserv/schema/replica", env.Glob("schema/*.sql"))
```
jgates108, line 44: No newline at eof.

jgates108 (approved):
Looks good, just minor things.

core/modules/replica/IngestSvcConn.cc
```diff
@@ -178,29 +185,92 @@ void IngestSvcConn::_handshakeReceived(boost::system::error_code const& ec,
...
+    auto const trans = databaseServices->transaction(_contrib.transactionId);
+    _contrib.database = trans.database;
+
+    if (!config->databaseInfo(_contrib.database).hasTable(_contrib.table)) {
+        _failed("no such table '" + _contrib.table + "' in database '" + _contrib.database + "'.");
+    }
+
+    // Prescreen parameters of the request to ensure they're valid in the given
+    // contex. Check the state of the transaction. Refuse to proceed with the request
+    // should any issues were detected.
```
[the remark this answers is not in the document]
iagaponenko: 👍

core/modules/replica/IngestHttpSvcMod.cc
```diff
@@ -133,100 +134,199 @@ string IngestHttpSvcMod::context() const {
...
+        resource.reset(new Url(contrib.url));
+        switch (resource->scheme()) {
+            case Url::FILE:
+            case Url::HTTP:
+            case Url::HTTPS:
+                break;
+            default:
+                throw invalid_argument(context + " unsupported url '" + contrib.url + "'");
+        }
+    } catch (exception const& ex) {
```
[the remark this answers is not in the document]
iagaponenko: Exceptions are being caught in this particular `try -> catch` loop to sanitize values of the input parameters of the REST service and report them (if any) as a failed attempt to ingest a table contribution. If the exceptions weren't caught they would be intercepted by some high-level code (the REST service processor and be reported back to a client).

The constructor of class `Url` may throw other exceptions when parsing its input. Exception `invalid_argument` is thrown for the correctly parsed `url` that is not suitable in the given context. This module only works with file-based URLs.

The algorithm won't be setting `system_error` in this `try -> catch` block because operations in the loop don't invoke any system libraries. Availability of the meaning full value for `errno` in the rest of the code depends on a particular operation. I understand the motivation behind your suggestion to generalize handling `errno`. So, I have tried to look at how to do this in the code. Unfortunately, I don't see a way to do this cleanly.

core/modules/replica/DatabaseServicesMySQL.cc
iagaponenko: Unfortunately, method `DatabaseServicesMySQL::createdTransactionContrib` can't return the `non-const` reference to the input object (of class `TransactionContribInfo`) that is passed into the method via the `const` reference. And it's probably better to keep the code this way by separating the non-mutable input from the output returned by value. Though, I don't have a strong preference for that.

core/modules/replica/IngestHttpSvcMod.cc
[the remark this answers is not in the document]
iagaponenko: I have simplified the code of the module by merging two `try -> catch` blocks into one.

PR #646, opened 2021-10-04
Tickets/DM-23308: Qserv container-based builds

n8pease:


src/admin/python/cli/script.py
```diff
@@ -0,0 +1,660 @@
...
+    """Entrypoint script for the xrootd container that runs on worker nodes."""
+
+    # TODO This sets the amount of data that can be locked into memory to
+    # almost the entire amount of memory on the machine. I think in a dev-env
+    # (docker-only on a single machine) we don't want to grab quite so much
+    # memory? TBD WTD here - do/don't set this? set it to how much? needs a way
+    # to set env (dev/prod/etc)
+    # # Increase limit for locked-in-memory size
+    # MLOCK_AMOUNT=$(grep MemTotal /proc/meminfo | awk '{printf("%.0f\n", $2 - 1000000)}')
+    # ulimit -l "$MLOCK_AMOUNT"
```
n8pease, line 329: What is the correct value for this: in production, on a developer machine?
I still need a way to pass it in via cfg I think.

n8pease:
Some comments about commits that may need some explanation: 

Commits of note:

Reorganize directory layout - mainly `qserv/core/modules` is moved to `qserv/src`. Adds an `qserv/admin` folder for stuff like Dockerfiles and the `qserv` CLI.

"Remove obsolete files" and "Remove wmgr" - wmgr is no longer used by qserv, and other old/unused files are removed.

Add Submodules - adds git submodules for the requirement "Manage co-developed or LSST-provided packages (XRootD, log, sphgeom, partition, etc.) by some mechanism other than eups"

Add Dockerfiles - these are the Dockerfiles for the lite images. 
* lite-build (in `base`) contains most of the requirements for the build container. 
* lite-build-user (in `build-user`) specializes the lite-build container for a developer/builder
* lite-run-base (in `base`) contains most of the requirements for a lite-qserv container
* lite-qserv (in `run`) adds the qserv artifacts
* lite-mariadb (in `mariadb`) is the db image used by qserv.

Add GitHub CI - this is for github actions: 
* builds base images and tags & pushes them to dockerhub if needed, or pulls them if available.
* builds user build image, qserv, and the lite-qserv image, runs unit tests.
* runs integration tests.

Unit test fixes - work was done to get unit tests running & passing. There are still a few unit tests that don't pass and need attention but are currently marked as "expected to fail" (in the cmake files), fixing these may be a good "intro to lite development" for John and Igor (and Andy but he may be busy on other things). 

Qserv CLI command - this implements the CLI command for building qserv and related containers, running qserv in docker-compose, and examining the current qserv dev "environment" - environment variable values and CLI option values that are deduced from environment variables and the location of files (e.g. QSERV_ROOT can be discovered by the location of the CLI file)

Add entrypoint CLI command - this implements the CLI command that is executed _inside_ the qserv container to launch a particular qserv service. The "flavor" of qserv container (e.g. czar, worker-cmsd, reel-controller, etc) is controlled by which `entrypoint` subcommand is run.


Schema migration changes for lite environment - adds schema initialization from a totally uninitialized state (adds from `None` scripts, like `migrate-None-to-N.sql`) and adds optional jinja templating to the sql files. 

Add watcher CLI command - This is the qserv watcher script that runs at IDF. This is somewhat out-of-band for lite, except that the watcher runs in a lite-run container. It could get removed from this PR and added in a separate ticket.

andy-slac:
> Some comments about commits that may need some explanation:
> 
Thanks for extra info! To simplify reviewer's miserable life, could you add that info to corresponding commit messages? 🙂

n8pease:
> Thanks for extra info! To simplify reviewer's miserable life, could you add that info to corresponding commit messages? 🙂
Good idea, it is done.

src/replica/CMakeLists.txt
```diff
@@ -0,0 +1,500 @@
...
+    WorkerRequest.cc
+    WorkerRequest.h
+    WorkerRequestFactory.cc
+    WorkerRequestFactory.h
+    WorkerServer.cc
+    WorkerServer.h
+    WorkerServerConnection.cc
+    WorkerServerConnection.h
+    WorkerSqlRequest.cc
+    WorkerSqlRequest.h
```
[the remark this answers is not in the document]
fritzm: I (personally) prefer explicit lists to globbing in this application.  Let's discuss at the next cowork -- if the majority opinion is for globbing, will change on a subsequent ticket.

admin/local/cli/qserv_cli.py
```diff
@@ -0,0 +1,432 @@
...
+    "mysql://root:CHANGEME@worker-db-0:3306",
+    "mysql://root:CHANGEME@worker-db-1:3306",
+]
+repl_connection_default = "mysql://root:CHANGEME@repl-mgr-db:3306/qservw_worker"
+
+
+@qserv.command()
+@czar_connection_option(
+    default=czar_connection_default,
+    help=f"{czar_connection_option.keywords['help']} The default value works with the default docker-compose file: {click.style(czar_connection_default, fg='green', bold=True)}",
```
andy-slac, line 423: This line (and two others below) is too long.

admin/local/cli/qserv_log.py
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env python3
```
andy-slac, line 1: You don't need this line

admin/local/cli/qserv_log.py
```diff
@@ -0,0 +1,28 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+"""Logging definitions and utilities for qserv CLI.
+"""
+
+log_level_flag = "--log-level"
+default = "INFO"
```
andy-slac, line 27: `default` -> `default_log_level`?

andy-slac (commented):
Bunch of comments for a "Add qserv CLI command" commit.

One general comment - maybe try to avoid importing things from `src/admin/` and keep all that you need in `admin/cli` (but I do not know how much stuff you need from `src/`)?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,239 @@
...
+      - name: Shut down qserv
+        run: |
+          ./admin/local/cli/qserv --log-level DEBUG down \
+          -v \
+
+
+
+
+
+#       - name: Declare Version Variables
```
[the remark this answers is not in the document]
n8pease: Yes; The GitHub actions CI code was originally written by @fjammes. I changed it, basically entirely, but I didn't want to lose what he had written, at least not right away. As part of our process to keep commits organized & targeted I was afraid it would get lost in a fixup or squash during a rebase.

Probably it should get removed before going to master, or at least an explanation added in code about why the commented code is there. Will do one or the other.

src/admin/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/python/ DESTINATION ${CMAKE_INSTALL_PREFIX}/python/lsst/qserv/admin)
```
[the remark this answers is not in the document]
n8pease: seems like not.

src/admin/templates/wmgr-secret/etc/wmgr.cnf.jinja
```diff
@@ -0,0 +1 @@
+USER:{{ wmgr_secret.user }}
```
andy-slac, line 1: I thought we abandoned wmgr? If yes then this file should not be needed.

src/util/ConfigStore.cc
```diff
@@ -95,25 +95,33 @@ std::string ConfigStore::get(std::string const& key,
     }
 }
 
+
 int ConfigStore::getInt(std::string const& key, int const& defaultValue) const {
+    try {
+        return getIntRequired(key);
+    } catch (util::KeyNotFoundError& e) {
```
andy-slac, line 102: Catch by const reference. I'd generally try to avoid using exceptions for non-exceptional cases in C++.

src/wconfig/WorkerConfig.cc
```diff
@@ -78,6 +77,18 @@ WorkerConfig::WorkerConfig(const util::ConfigStore& configStore)
       _bufferMaxTotalGB(configStore.getInt("transmit.buffermaxtotalgb", 41)),
       _maxTransmits(configStore.getInt("transmit.maxtransmits", 40)),
       _maxPerQid(configStore.getInt("transmit.maxperqid", 3)) {
+    try {
+        _mySqlConfig = mysql::MySqlConfig(configStore.getRequired("mysql.username"),
+                                          configStore.get("mysql.password"),
+                                          configStore.getRequired("mysql.socket"));
+    } catch (util::KeyNotFoundError& e) {
```
andy-slac, line 84: Is this supposed to handle the case of missing  "mysql.socket"? I'd again try to avoid relying on exceptions for this expected failure.

andy-slac (commented):
Bunch of comments for "Add entrypoint CLI command" commit.

src/query/testColumnRef.cc
```diff
@@ -30,7 +30,7 @@
 #include <fstream>
 
 // list must be included before boost/test/data/test_case.hpp, because it is used there but not included.
-// (or that file could be included after boost/test/included/unit_test.hpp, which does cause list to be
+// (or that file could be included after boost/test/unit_test.hpp, which does cause list to be
 // included. But, we like to include our headers alphabetically so I'm including list here.
```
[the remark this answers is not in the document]
n8pease: no, removing it and the list include.

andy-slac:
> there are options that are used in the entrypoint command (inside containers, source is under src/admin) and can be passed through from qserv to entrypoint (via docker run command), so they're used by qserv as well

Are these options actually used by qserv or are they just passed to entrypoint? If it is latter then you could just pass them without parsing on qserv side (maybe like `qserv comand [command-options] -- [entrypoint-options]`). If they actually needed by qserv then you can probably move that module to `admin/local/cli/` and then install it in `python/lsst/qserv/admin`.

src/admin/python/mysql_connection.py
```diff
@@ -0,0 +1,64 @@
...
+    -------
+    connection : `mysql.connection`
+        The connected connection object.
+    """
+    if uri:
+        parsed = urlparse(uri)
+        hostname = parsed.hostname
+        port = parsed.port
+        user = parsed.username
+        pw = parsed.password
```
andy-slac, line 55: To make URI more consistent with sqlalchemy you could also extract database name from `parsed.path` (`dialect://username:password@host:port/database`). Or maybe we should switch to sqlalchemy instead of using native interfaces.

src/admin/python/itest_load.py
```diff
@@ -0,0 +1,516 @@
...
+    cnx = mysql_connection(uri)
+    _log.debug(f"_create_ref_table schema_file:{schema_file}")
+    if not cnx.is_connected():
+        cnx.connect()
+    cursor = cnx.cursor()
+    cursor.execute(f"USE {db}")
+    with open(schema_file) as f:
+        cursor.execute(f.read(), multi=True)
+    cursor.close()
+    cnx.close()
```
andy-slac, line 114: For consistency I'd suggest `closing` context manager for cursor and connection.

src/admin/python/itest_load.py
```diff
@@ -0,0 +1,516 @@
...
+    if not cnx.is_connected():
+        cnx.connect()
+    with closing(cnx.cursor()) as cursor:
+        cursor.execute(f"USE {db}")
+        cursor.execute(sql)
+        cnx.commit()
+    cnx.close()
+
+
+class LoadTable:
```
andy-slac, line 152: This is probably a good candidate for `@dataclass`

src/admin/python/itest_load.py
```diff
@@ -0,0 +1,516 @@
...
+        "sph-partition",
+    ]
+    for config_file in partition_config_files:
+        args.append("--config-file")
+        args.append(config_file)
+    args.extend(
+        [
+            f"--in.path={data_file}",
+            "--verbose",
+            "--mr.num-workers=6",
```
andy-slac, line 296: 6 is my favorite number 😹

andy-slac (commented):
Comments for "Add integration test support" commit.

src/qmeta/python/schema_migration.py
```diff
@@ -3,216 +3,109 @@
...
+    @backoff.on_exception(
+        exception=(mysql.connector.errors.OperationalError, mysql.connector.errors.ProgrammingError),
+        wait_gen=backoff.expo,
+        on_backoff=on_backoff(log=_log),
+        max_time=max_backoff_sec,
+    )
+    def _set_version(self, version):
+        """Set the version number stored in QMetadata."""
+        # make sure that current version is updated in database
+        query = "UPDATE QMetadata SET value = {} WHERE metakey = 'version'".format(version)
```
andy-slac, line 66: I believe `version` is an `int` here but `QMetadata.value` is a string. Mysql is very liberal when mixing types, but it may be better to convert version to string before saving, e.g.:

             query = f"UPDATE QMetadata SET value = '{version}' WHERE metakey = 'version'"

src/schema/python/schemaMigMgr.py
```diff
@@ -1,32 +1,171 @@
...
+            f"name={self.name}, filepath={self.filepath})"
+        )
 
 
 class SchemaMigMgr(metaclass=ABCMeta):
     """Abstract base class for schema migration managers.
+
+    Parameters
+    ----------
+    TODO
```
andy-slac, line 168: TODO or not TODO, that is the ❔

src/schema/python/schemaMigMgr.py
```diff
@@ -35,45 +174,124 @@ def current_version(self):
...
         """Perform schema migration from current version to given version.
 
         Parameters
         ----------
-        version : int or None
+        to_version : int or None
             If None then migrate to latest known version, otherwise only
             migrate to given version.
         do_migrate : bool
             If True performa migration, otherwise only print steps that
```
andy-slac, line 293: Could you fix "performa" while you are at it?

andy-slac (commented):
Few comments for "Schema migration changes for lite environment" commit.

.github/workflows/doc.yml
```diff
@@ -0,0 +1,17 @@
+# Qserv CI workflow
+# FIXME: Github action runner are too weak to run e2e tests
```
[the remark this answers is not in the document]
n8pease: 👍 ...we should discuss (maybe at the next cowork?) where integration tests should be run; one of the benefits of lite is a much faster build time, and so the integration tests can be run on GitHub actions during the lite build. It might make sense to run them on the operator as well, to test for correctness.

admin/local/cli/images.py
```diff
@@ -0,0 +1,366 @@
...
+        "docker",
+        "push",
+        image_name,
+    ]
+    if dry:
+        print(" ".join(args))
+        return
+    _log.debug(f"Running {' '.join(args)}")
+    result = subprocess.run(args)
+    result.check_returncode()
```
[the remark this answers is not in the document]
n8pease: looks like fancy newer api is `subprocess.run(args, check=True)`

admin/local/watcher/watcher_int.yaml
```diff
@@ -0,0 +1,27 @@
...
+      labels:
+        app: watcher
+    spec:
+      volumes:
+        - name: slack-webhook-url
+          secret:
+            secretName: secret-watcher
+      containers:
+        - name: watcher
+          image: qserv/lite-watcher:c1973a8b
```
iagaponenko, line 21: Why using the specific tag in this config? Would it be possible to rely on something like `qserv/lite-watcher:latest` (`:current`, etc.)?

admin/local/watcher/watcher_prod.yaml
```diff
@@ -0,0 +1,27 @@
...
+      labels:
+        app: watcher
+    spec:
+      volumes:
+        - name: slack-webhook-url
+          secret:
+            secretName: secret-watcher
+      containers:
+        - name: watcher
+          image: qserv/lite-watcher:c1973a8b
```
iagaponenko, line 21: Same question as before regarding the tag.

src/admin/tests/test_watcher.py
```diff
@@ -0,0 +1,203 @@
...
+    CheckFailure,
+    first_check_msg_t,
+    msg_prefix_t,
+    notify_data,
+    notify_headers,
+    query_msg_t,
+    show_query_msg_t,
+    start_msg_t,
+    stop_msg_t,
+    Watcher,
```
iagaponenko, line 40: Asking mostly out of curiosity. I'm seeing the extra comma at the last element in nearly all lists and tuples of this file. AFAIK it's not formally required by the language standard. In other languages, this would be a syntax error. Why it's here? Is there any specific requirement imposed by the LSST Style Guide, or this is just a "good practice" based on "how people usually do things"?

src/admin/tests/test_smig.py
```diff
@@ -0,0 +1,127 @@
...
+import os
+import unittest
+from unittest.mock import patch
+
+from lsst.qserv.admin.cli import script
+from lsst.qserv.qmeta.schema_migration import QMetaMigrationManager
+from lsst.qserv.schema.schemaMigMgr import SchemaUpdateRequired, Uninitialized
+
+
+latest_qmeta_schema_version = 4
```
iagaponenko, line 35: Why this number is here? Does it refer to the real version? Or this it's just for testing purposes? If not then it may become a management burden to update this variable here after updating the schema. I'd rather have it where the schema of the `qservMeta` database is defined.

src/css/schema/migrate-None-to-0.sql
```diff
@@ -0,0 +1,39 @@
+-- MySQL Script generated by MySQL Workbench
+-- Thu Jul 30 11:58:38 2015
+-- Model: New Model    Version: 1.0
+-- MySQL Workbench Forward Engineering
+
+SET @OLD_UNIQUE_CHECKS=@@UNIQUE_CHECKS, UNIQUE_CHECKS=0;
+SET @OLD_FOREIGN_KEY_CHECKS=@@FOREIGN_KEY_CHECKS, FOREIGN_KEY_CHECKS=0;
+SET @OLD_SQL_MODE=@@SQL_MODE, SQL_MODE='TRADITIONAL,ALLOW_INVALID_DATES';
```
iagaponenko: We should probably get rid of these lines and the comment above that  "MySQL Script generated by MySQL Workbench", etc.

src/css/schema/migrate-None-to-0.sql
```diff
@@ -0,0 +1,39 @@
...
+
+-- -----------------------------------------------------
+-- Schema qservCssData
+-- -----------------------------------------------------
+
+-- -----------------------------------------------------
+-- Schema qservCssData
+-- -----------------------------------------------------
```
iagaponenko: Duplicate comment lines.

src/css/schema/migrate-None-to-0.sql
```diff
@@ -0,0 +1,39 @@
...
+  CONSTRAINT `kvData_parent_key`
+    FOREIGN KEY (parentKvId)
+    REFERENCES kvData(kvId),
+  UNIQUE INDEX `kvData_kvKey_UNIQUE` (`kvKey` ASC)
+) ENGINE = InnoDB;
+
+
+SET SQL_MODE=@OLD_SQL_MODE;
+SET FOREIGN_KEY_CHECKS=@OLD_FOREIGN_KEY_CHECKS;
+SET UNIQUE_CHECKS=@OLD_UNIQUE_CHECKS;
```
iagaponenko: These 3 lines should be removed

src/qmeta/schema/migrate-None-to-4.sql.jinja
```diff
@@ -0,0 +1,196 @@
...
+-- created by integration test script/loader for now
+CREATE DATABASE IF NOT EXISTS qservMeta;
+GRANT ALL ON qservMeta.* TO '{{ mysqld_user_qserv }}'@'localhost';
+GRANT ALL ON qservMeta.* TO '{{ mysqld_user_qserv }}'@'%';
+
+
+SET @OLD_UNIQUE_CHECKS=@@UNIQUE_CHECKS, UNIQUE_CHECKS=0;
+SET @OLD_FOREIGN_KEY_CHECKS=@@FOREIGN_KEY_CHECKS, FOREIGN_KEY_CHECKS=0;
+SET @OLD_SQL_MODE=@@SQL_MODE, SQL_MODE='ALLOW_INVALID_DATES';
+
```
iagaponenko, line 18: These 3 lines need to be removed.

src/qmeta/schema/migrate-None-to-4.sql.jinja
```diff
@@ -0,0 +1,196 @@
...
+-- Version 0 corresponds to initial QMeta release and it had no
+-- QMetadata table at all.
+-- Version 1 introduced QMetadata table and altered schema for QInfo table
+-- Version 2 added query progress data to ProcessList tables.
+-- Version 3 added storing the result query in QMeta.
+INSERT INTO `QMetadata` (`metakey`, `value`) VALUES ('version', '3');
+
+SET SQL_MODE=@OLD_SQL_MODE;
+SET FOREIGN_KEY_CHECKS=@OLD_FOREIGN_KEY_CHECKS;
+SET UNIQUE_CHECKS=@OLD_UNIQUE_CHECKS;
```
iagaponenko, line 196: These 3 lines need to be removed.

src/replica/schema/migrate-None-to-0.sql
```diff
@@ -0,0 +1,660 @@
...
+
+GRANT ALL ON qservReplica.* TO  qsreplica@localhost;
+GRANT ALL ON qservReplica.* TO  qsreplica@'%';
+
+FLUSH PRIVILEGES;
+
+
+SET @OLD_UNIQUE_CHECKS=@@UNIQUE_CHECKS, UNIQUE_CHECKS=0 ;
+SET @OLD_FOREIGN_KEY_CHECKS=@@FOREIGN_KEY_CHECKS, FOREIGN_KEY_CHECKS=0 ;
+SET @OLD_SQL_MODE=@@SQL_MODE, SQL_MODE='TRADITIONAL' ;
```
iagaponenko, line 15: These 3 lines need to be removed.

src/replica/schema/migrate-None-to-0.sql
```diff
@@ -0,0 +1,660 @@
...
+COMMENT = 'Metadata about database as a whole, bunch of key-value pairs';
+
+
+INSERT INTO `ReplicaMetadata` (`metakey`, `value`) VALUES ('version', '0');
+
+
+
+SET SQL_MODE=@OLD_SQL_MODE ;
+SET FOREIGN_KEY_CHECKS=@OLD_FOREIGN_KEY_CHECKS ;
+SET UNIQUE_CHECKS=@OLD_UNIQUE_CHECKS ;
```
iagaponenko, line 660: These 3 lines need to be removed.

src/rproc/python/schema_migration.py
```diff
@@ -0,0 +1,78 @@
...
+import logging
+
+from lsst.qserv.schema import SchemaMigMgr, Uninitialized
+
+
+_log = logging.getLogger(__name__)
+qservResultDb = "qservResult"
+
+
+class RprocMigrationManager(SchemaMigMgr):
```
iagaponenko, line 37: I'm still seeing these classes in Qserv modules. Each does nearly exactly the same. At one point I heard a comment from @andy-slac that we need to have a generic class for that. And for those cases when the standard behavior of that class wouldn't be suitable (non-trivial data migration required by the new schema) then we should have an option to set up an overload.
I'm not sure what would be the best way to implement this in Python.

src/replica/python/schema_migration.py
```diff
@@ -1,209 +1,123 @@
...
-        Sqlalchemy engine instance.
+        Module name, e.g. "admin"
+    connection : dbapi connection
+        Database connection instance.
     scripts_dir : `str`
         Path where migration scripts are located, this is system-level directory,
         per-module scripts are usually located in sub-directories.
+    set_initial_configuration : function or `None`
+        A function to be called to set the initial configuration of the repl database.
+        Will only be called if the schema is migrated from None to version 1. Optional.
```
iagaponenko, line 121: This idea on migrating from version `None` to version `1` is not consistent with the above-defined migration of `admin` from `None` of `0`? Or, I'm missing some important concept of the version `0`.

src/admin/python/template.py
```diff
@@ -0,0 +1,118 @@
...
+
+_log = logging.getLogger(__name__)
+
+
+default_cfg_file_path = "/config-etc/qserv-template.cfg"
+
+cfg_file_path = os.environ.get("QSERV_CFG", default_cfg_file_path)
+
+
+def save_template_cfg(values: Dict[str, str]) -> None:
```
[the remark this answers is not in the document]
n8pease: I did not expect two entrypoint scripts to be run in the same container (so, no?)

src/admin/tests/test_watcher.py
n8pease: [pep-8](https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas) has this to say about trailing commas:
> they are often helpful when a version control system is used, when a list of values, arguments or imported items is expected to be extended over time. The pattern is to put each value (etc.) on a line by itself, always adding a trailing comma, and add the close parenthesis/bracket/brace on the next line

I've been using a code formatting tool called black ("The Uncompromising Code Formatter") that Just Does all the annoying formatting that the pythonistas would otherwise snark at, and its rule is no trailing comma=list on one line, and trailing comma=each element on a line. If the list is too long for one line it'll even rewrite the list to be one element per line (with trailing comma). So, wookie/win, & I get to 🍺 that much sooner 😄

src/admin/schema/migrate-None-to-0.sql.jinja
```diff
@@ -0,0 +1,29 @@
...
+GRANT ALL ON qservMeta.* TO '{{ mysqld_user_qserv }}'@'localhost';
+GRANT ALL ON qservMeta.* TO '{{ mysqld_user_qserv }}'@'%';
+
+-- CSS database
+GRANT ALL ON qservCssData.* TO '{{ mysqld_user_qserv }}'@'localhost';
+GRANT ALL ON qservCssData.* TO '{{ mysqld_user_qserv }}'@'%';
+
+-- TODO FIGURE OUT HOW TO HANDLE THE POD NAME IN CONTAINER-LAND
+-- -- Grant root access to replication controller pod
+-- GRANT ALL ON *.* TO 'root'@'{{ replication_controller_FQDN }}' IDENTIFIED BY 'CHANGEME' WITH GRANT OPTION;
```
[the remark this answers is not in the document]
n8pease: aren't the passwords stored in GitHub already, in `qsrev-meta.conf`? I'm not sure the replication controller pw is there, but others are. I agree it's not good. Making it a template parameter is likely part of the solution, but how to define these secrets for the system needs a little design?

src/rproc/python/schema_migration.py
n8pease: In python, inheritance is specified in the parens after the class, so `RprocMigrationManager` inherits from `SchemaMigMgr`. The common code is in the latter class as of a year or so ago IIRC.

src/replica/python/schema_migration.py
n8pease: Sorry, it's a typo, it should say "Will only be called if the schema is migrated from None" or maybe "Will only be called if the repl database is uninitialized.". Does that make more sense now? (🤞)

iagaponenko: Yes, it would make sense. Though, it may not be needed as I have a PR that has been reviewed and approved by @andy-slac. The ticket brings support for the `snig`-based schema migration of the Replication system's database. This should make your life easier. I'm still looking through the comments on the ticket to make some changes. However, it's very likely that the ticket will be merged into the `master` branch before yours. When you will be rebasing it then you may need to make some changes to the way you do schema migration for module `replica`.

src/rproc/python/schema_migration.py
iagaponenko: I still don't understand why you need a special class for each nodule even if all module-specific classes essentially do the same. The only difference between them is that they print slightly different messages. I can imagine having the schema migration tool which would simply look at the `schema/<module>/` folder to locate schema versions and do the job. And if some module has a special procedure for migrating its schema then perhaps some optional module-specific class would be pulled and used. Otherwise, having these (essentially) empty classes make no sense to me.

admin/local/cli/images.py
```diff
@@ -0,0 +1,366 @@
...
+        A git sha.
+    b : `str`
+        A git sha.
+    cwd : `str`
+        The directory to execute the git command in.
+
+    Returns
+    -------
+    log : `list` [`str`]
+        A list of commits rechable from b but not from a.
```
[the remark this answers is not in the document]
andy-slac: I am still confused, for me "reachable" means going forward in history. Do you mean that reachable is going back to parent commits?

src/admin/python/replicationInterface.py
```diff
@@ -0,0 +1,529 @@
...
+        """Get the names of the existing databases from the replication system.
+
+        Returns
+        -------
+        databases : List[str]
+            The database names.
+        """
+        url = (
+            url
+        ) = f"http://{self.repl_ctrl.hostname}:{self.repl_ctrl.port}/replication/config"
```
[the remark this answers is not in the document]
andy-slac: No I'm perplexed by:

    url = (
        url
    ) = ...

src/replica/python/schema_migration.py
n8pease: Sure. If you give me a heads up when you do the merge that would be great, for when we rebase this ticket branch onto master again after that.

src/parser/MySqlParser.g4
```diff
@@ -25,7 +25,7 @@ THE SOFTWARE.
 
 parser grammar MySqlParser;
 
-options { tokenVocab=MySqlLexer; }
+// options { tokenVocab=MySqlLexer; }
```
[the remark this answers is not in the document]
fritzm: Done

src/replica/schema/migrate-None-to-0.sql
```diff
@@ -0,0 +1,660 @@
+
+CREATE USER IF NOT EXISTS qsreplica@localhost;
+CREATE USER IF NOT EXISTS qsreplica@'%';
+
+CREATE DATABASE IF NOT EXISTS qservReplica;
+
+GRANT ALL ON qservReplica.* TO  qsreplica@localhost;
+GRANT ALL ON qservReplica.* TO  qsreplica@'%';
+
+FLUSH PRIVILEGES;
```
iagaponenko, line 10: If I understand correctly the idea of the `None-to-0` stage schema migration the only these lines should stay in this file(?). These are just for initializing the database. Anything else below line 10 needs to be put into the actual schema migration.

My other comment is that for the sake of consistency (and for other good reasons), the account name and the (optional) password need to be injected into the database initialization file as you did it in `src/admin/schema/migrate-None-to-0.sql.jinja`.

This will help to unify and simplify the site-specific initialization of the MySQL instances (czar's, Replication system's and workers').

src/worker/schema/migrate-None-to-0.sql.jinja
```diff
@@ -0,0 +1,71 @@
+-- WARN all '<user>'@'%' must be removed in production
+
+CREATE USER IF NOT EXISTS 'qsmaster'@'localhost';
+CREATE USER IF NOT EXISTS 'qsmaster'@'%';
+GRANT ALL ON `q\_memoryLockDb`.* TO 'qsmaster'@'localhost';
```
iagaponenko, line 5: MySQL accounts and the corresponidng (optional) passwords need to be injected into this file like it's done in `src/admin/schema/migrate-None-to-0.sql.jinja`. In general, anything that's either site-specific or security-related has to be injected into the database initialization sequences.

src/rproc/python/schema_migration.py
n8pease: Currently there's no requirement that modules manage their schema version the same way, subclassing allows specialization of getting & setting the version.

admin/tools/docker/build-user/Dockerfile
```diff
@@ -0,0 +1,10 @@
+ARG QSERV_BUILD_BASE=qserv/lite-build:latest
```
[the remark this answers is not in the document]
fritzm: Additional comments on motivation and use added to all Dockerfiles

src/replica/CMakeLists.txt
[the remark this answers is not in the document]
fritzm: Philosophy question: if it doesn't throw a link error, and the unit and integration tests pass, was it really needed anyway? 🤔

admin/local/cli/images.py
```diff
@@ -0,0 +1,366 @@
...
+# ImageTagger init args as needed. Currently they're set at member variables in
+# init so they can be changed on a class instance.
+manifest_header = "application/vnd.docker.distribution.manifest.v2+json"
+auth_header = "Bearer {token}"
+
+
+_log = logging.getLogger(__name__)
+
+
+def get_tag(dockerfiles: List[str], cwd: str) -> str:
```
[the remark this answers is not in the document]
fritzm: Who is really going to argue with you and Andy re. Python practices?  :-)

src/replica/CMakeLists.txt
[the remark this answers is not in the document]
iagaponenko: From my point of view, I don't see any significant benefits in "globbing". To me, this is just a matter of taste. And when it comes to taste a decision should be up to the developer .. not to the "reviewing committee".
By the way, mentioning files explicitly does have one benefit - an ability to exclude (if needed) some files from the compilation. This may be especially important during the prototyping or active development phase of some new features.

admin/local/cli/launch.py
```diff
@@ -0,0 +1,797 @@
...
+from collections import namedtuple
+import getpass
+import grp
+import logging
+import os
+import pwd
+import subprocess
+from typing import Dict, List, Optional
+
+from opt import (
```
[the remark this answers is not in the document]
n8pease: yeah, I'll fix all the relative import stuff when we reorganize the "admin" python files (all into "src")

admin/local/cli/qserv
```diff
@@ -0,0 +1,56 @@
...
+_log = logging.getLogger(__name__)
+
+
+# Adding the src/admin/python path to sys.path allows options declared in
+# src/admin/cli/options.py to be imported and used by the qserv command.
+sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), "../../../src/admin/python")))
+
+
+if __name__ == "__main__":
+    log_level = sys.argv[sys.argv.index(log_level_flag)+1].upper() if log_level_flag in sys.argv else default
```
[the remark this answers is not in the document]
n8pease: noted for a future ticket.

src/admin/python/replicationInterface.py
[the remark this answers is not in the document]
n8pease: oh, haha.

PR #655, opened 2021-11-04
make post-review lite-qserv improvements: move files, log level option parse, linter prep: DM-32465

n8pease:


src/schema/python/schemaMigMgr.py
```diff
@@ -67,45 +93,59 @@ def cast(other):
...
+    @staticmethod
+    def _otherVal(other: object) -> Union[int, Type[Uninitialized]]:
         """Get the value of the other value. May be a ``Version`` instance,
         ``Uninitialized``, or an ``int``.
         """
-        if type(other) is int or other is Uninitialized:
-            return other
-        if type(other) is Version:
+        if not Version._validateOther(other):
+            return NotImplemented
```
andy-slac, line 105: `NotImplemented` is not in the return type annotation.

src/schema/python/schemaMigMgr.py
```diff
@@ -67,45 +93,59 @@ def cast(other):
...
+        if isinstance(other, Version):
             return other.value
-        raise TypeError(f"Unsupported value {other}.")
+        if isinstance(other, int):
+            return int(other)
```
andy-slac, line 109: It's already an int?

src/schema/python/schemaMigMgr.py
```diff
@@ -67,45 +93,59 @@ def cast(other):
...
+        if other is Uninitialized:
+            return Uninitialized
+        raise NotImplementedError(f"_otherVal is not implemented for the type of {other}")
 
-    def __lt__(self, other):
+    def __lt__(self, other: object) -> bool:
         otherVal = self._otherVal(other)
```
andy-slac, line 115: Do you want to add

        if otherVal  is NotImplemented:
            return NotImplemented

here similarly to other methods?

src/schema/python/schemaMigMgr.py
```diff
@@ -116,41 +156,57 @@ class Migration:
...
     @property
-    def dirname(self):
+    def dirname(self) -> str:
         return os.path.dirname(self.filepath)
 
-    def __init__(self, from_version, to_version, name, filepath):
+    def __init__(
+        self,
+        from_version: Union[int, Type[Uninitialized]],
+        to_version: Union[int, type[Uninitialized]],
```
andy-slac, line 165: type - > Type

Will we ever want to migrate _to_ Uninitialized?

src/schema/python/schemaMigMgr.py
```diff
@@ -116,41 +156,57 @@ class Migration:
...
+            return NotImplemented
         return (self.from_version, self.to_version, self.name, self.filepath) == (
             other.from_version,
             other.to_version,
             other.name,
             other.filepath,
         )
 
-    def __lt__(self, other):
+    def __lt__(self, other: "Migration") -> bool:
```
andy-slac, line 184: I think you don't need quotes if you do `from __future__ import annotations`

src/schema/python/schemaMigMgr.py
```diff
@@ -210,9 +264,9 @@ def max_migration_version(self):
         on_backoff=on_backoff(log=_log),
         max_time=max_backoff_sec,
     )
-    def _connect(self, connection):
+    def _connect(self, connection: str) -> None:
```
andy-slac, line 267: It returns a connection, not None?

src/schema/python/schemaMigMgr.py
```diff
@@ -278,9 +335,9 @@ def apply_migrations(self, migrations):
                     )  # empty migration files are used for testing.
                 self.connection.commit()
                 _log.info("+++ Script %s completed successfully", migration.name)
-        return migration.to_version
+        return migration.to_version.value
 
-    def migrate(self, to_version=None, do_migrate=False):
+    def migrate(self, to_version: Union[int, None] = None, do_migrate: bool = False) -> Union[None, int]:
```
andy-slac, line 340: `Union[int, None]` -> `Optional[int]`?

src/schema/python/schemaMigMgr.py
```diff
@@ -46,19 +71,20 @@ def __init__(self):
```
andy-slac, line 71: I wonder if it would make code much simpler if you just do `Uninitialized = -1`

src/schema/python/schemaMigMgr.py
```diff
@@ -365,27 +430,28 @@ def migration_path(from_version, to_version, migrations):
...
         Returns
         -------
-        versions : `None` or `tuple` [`int` or `None`, `int`]
-            None if no match. Pair of versions if there is a match, the first
-            item (the "from" value) may be `None` to indicate "from an
-            uninitialized database", or an integer to indicate the version the
-            script can upgrade from, and the second value will be an integer
-            indicating the version the script will upgrade to.
+        versions : `None` or `MigMatch`
+            None if no match. A `MigMatch` isntance if there is a match, the
```
andy-slac, line 439: Typo: isntance

src/schema/python/schemaMigMgr.py
```diff
@@ -365,27 +430,28 @@ def migration_path(from_version, to_version, migrations):
...
         """
         match = cls.mig_name_re.match(fname)
         if match:
-            v_from = match.group("from")
-            v_from = Uninitialized if v_from == "None" else int(v_from)
+            v_from_ = match.group("from")
+            v_from: Union[int, Type[Uninitialized]] = Uninitialized if v_from_ == "None" else int(v_from_)
             v_to = int(match.group("to"))
-            return (v_from, v_to)
+            _log.info(f"from {v_from} to {v_to}")
```
andy-slac, line 450: Should this be `debug`? And `%s` for `logging` instead of f-strings.

src/schema/tests/test_SchemaMigMgr.py
```diff
@@ -158,7 +159,10 @@ def test_migrate(self):
         class TestMigMgr(SchemaMigMgr):
             def __init__(self, current_version, scripts_dir):
                 self._current_version = current_version
-                super().__init__(scripts_dir)
+                super().__init__(scripts_dir, "connection-uri")
+
+            def _connect(self, connection: str) -> None:
+                self.connection = None
```
andy-slac, line 165: I think `_connect()` should return a connection instead of setting `self.connection`?

src/admin/python/lsst/qserv/admin/qservCli/qserv_log.py
```diff
@@ -22,6 +22,42 @@
...
 """
 
+from typing import Any, List, Tuple
+
 log_level_flag = "--log-level"
 default_log_level = "INFO"
 log_level_choices = ["CRITICAL", "ERROR", "WARNING", "INFO", "DEBUG"]
+
+
+def log_level_from_args(args: List[Any], default: str = default_log_level) -> Tuple[bool, str]:
```
andy-slac, line 32: Should it be `List[str]` instead of `List[Any]`?

src/admin/python/lsst/qserv/admin/qservCli/qserv_log.py
```diff
@@ -22,6 +22,42 @@
...
+    ----------
+    args : list [Any]
+        A list of arguments a script was launched with.
+
+    Returns
+    -------
+    (success, value) : Tuple[bool, str]
+        If the first item is true True then a log level option was found or not
+        found, and then the second item is the log level value that was found or
+        the default value. If the second item is False then a log level option
```
andy-slac, line 45: "second item" -> "first item"

src/admin/python/lsst/qserv/admin/qservCli/qserv_log.py
```diff
@@ -22,6 +22,42 @@
...
+        a user-readable description of the problem.
+    """
+    split_args = []
+    for arg in args:
+        if "=" in arg:
+            split_args.extend(arg.split("="))
+        else:
+            split_args.append(arg)
+    val = (
+        split_args[split_args.index(log_level_flag) + 1].upper() if log_level_flag in split_args else default
```
andy-slac, line 56: This may generate `IndexError`

admin/local/cli/qserv
```diff
@@ -33,20 +33,23 @@ import sys
...
-        format="%(asctime)s %(levelname)-8s %(message)s",
-        datefmt="%Y-%m-%d %H:%M:%S"
-    )
+    success, levelMsg = log_level_from_args(sys.argv)
+    if success:
+        logging.basicConfig(
+            level=levelMsg,
+            format="%(asctime)s %(levelname)-8s %(message)s",
+            datefmt="%Y-%m-%d %H:%M:%S",
+        )
```
andy-slac, line 45: I agree that initializing logging early is a good idea, but I would still prefer to leave parsing to `click`. Is there a reason why do you need this even before parsing command lie with `click`?

src/admin/tests/test_qserv_log.py
```diff
@@ -0,0 +1,50 @@
...
+
+class QservLogTestCase(unittest.TestCase):
+    """Tests features of the qserv_log helper function(s)."""
+
+    def test_log_level_from_args(self):
+        self.assertEqual(log_level_from_args(["--foo", "--log-level", "DEBUG", "--bar=baz"]), (True, "DEBUG"))
+        self.assertEqual(log_level_from_args(["--log-level", "debug"]), (True, "DEBUG"))
+        self.assertEqual(log_level_from_args(["--log-level=debug"]), (True, "DEBUG"))
+        self.assertEqual(log_level_from_args(["--log-level=DEBUG"]), (True, "DEBUG"))
+        res = log_level_from_args(["--log-level", "debugg"])
```
andy-slac, line 38: Can you add a test like

    log_level_from_args(["--log-level"])

andy-slac (approved):
Looks good, bunch of minor comments. Did you re-ran `black` and `mypy` on all latest changes?

src/schema/python/schemaMigMgr.py
n8pease: did you mean add these lines like in other methods?
```
if not self._validateOther(other):
    return NotImplemented```

Comment (andy-slac) [reply src/schema/python/schemaMigMgr.py:115]: `self._otherVal(other)` already calls `_validateOther` and returns `NotImplemented`, so you can use the result of that call. I would probably remove `_validateOther` call from `_otherVal()` and call it explicitly here, because in other places you call `_validateOther` first and `_otherVal` after that, which repeats call to `_validateOther`.

File: src/schema/python/schemaMigMgr.py (hunk 4 above)
Comment (andy-slac) [reply src/schema/python/schemaMigMgr.py:165]: Yes, I think `to_version` wants to be an `int`, I doubt we ever want to migrate to `Uninitialized`.

File: admin/local/cli/qserv (hunk 15 above)
Comment (n8pease) [reply admin/local/cli/qserv:45]: I have code that does work while click options are being loaded & initialized, and some of it emits debug log messages. To see those I have to init logging before click is able to parse & init logging. :sadpanda:

File: src/schema/python/schemaMigMgr.py (hunk 3 above)
Comment (n8pease) [reply src/schema/python/schemaMigMgr.py:115]: good call. thx.

File: admin/local/cli/qserv (hunk 15 above)
Comment (andy-slac) [reply admin/local/cli/qserv:45]: Maybe comment out those debugging messages? :happypanda: I think for cases like this it should be fine to use some alternative debugging mechanism instead of logging.

File: src/schema/python/schemaMigMgr.py [hunk 17]
```diff
@@ -306,36 +369,48 @@ def migrate(self, to_version=None, do_migrate=False):
...
+        to_version: Version,
+        migrations: List[Migration],
+    ) -> Optional[List[Migration]]:
         """Look in migrations and find the shortest possible path between
         versions.
 
         Parameters
         ----------
-        from_version : ``Version`` or one of its supported types.
+        from_version : Version
```
andy-slac, line 411: Backticks around `Version`?

src/schema/python/schemaMigMgr.py
```diff
@@ -368,27 +443,28 @@ def migration_path(from_version, to_version, migrations):
...
-        versions : `None` or `tuple` [`int` or `None`, `int`]
-            None if no match. Pair of versions if there is a match, the first
-            item (the "from" value) may be `None` to indicate "from an
-            uninitialized database", or an integer to indicate the version the
-            script can upgrade from, and the second value will be an integer
-            indicating the version the script will upgrade to.
+        versions : `None` or `MigMatch`
+            None if no match. A `MigMatch` instance if there is a match, the
+            from_version may be `None` to indicate "from an uninitialized
+            database", or an integer to indicate the version the script can
```
andy-slac, line 454: `from_version` is `Uninitialized` if database is not initialized.

andy-slac (approved):
RE-reviewed, looks OK, couple of minor comments.

PR #656, opened 2021-11-06
DM-32473: restore qserv replica schema version table to its previous name

n8pease:


iagaponenko:
I'm a bit confused with the comment in the first commit that says "rename init script to be usable by smig". In actuality, the init script got removed from the package.

src/admin/python/cli/script.py
```diff
@@ -195,6 +196,10 @@ def smig_replication_controller(
     ----------
     connection : `str`
         Connection string in format mysql://user:pass@host:port/database
+    repl_connection : `str`, optional
+        The connection string for the replication manager database for the
+        non-admin user. Required when initializing the database, not needed
```
iagaponenko, line 201: This statement (regarding "non-admin") may be incorrect because this connection is used later by method `_create_database` to do the typical database administrator operations. These actions require administrative privileges in MySQL (or any other database):
```
            f"CREATE USER IF NOT EXISTS {user}@localhost;",
            f"CREATE USER IF NOT EXISTS {user}@'%';",
            f"CREATE DATABASE IF NOT EXISTS qservReplica;",
            f"GRANT ALL ON qservReplica.* TO  {user}@localhost;",
            f"GRANT ALL ON qservReplica.* TO  {user}@'%';",
            f"FLUSH PRIVILEGES;",
```

iagaponenko: Remind me, please, why do you have two connection strings in this app?
```
connection=
repl_connection=
```

src/replica/python/schema_migration.py
```diff
@@ -80,6 +92,47 @@ def _set_version(self, version):
...
+        user = make_url(self.repl_connection).username
+        if not user:
+            raise RuntimeError(
+                "To initialize the replication database, the non-admin connection uri must contain a user "
+                "name."
+            )
+        for stmt in [
+            f"CREATE USER IF NOT EXISTS {user}@localhost;",
+            f"CREATE USER IF NOT EXISTS {user}@'%';",
+            f"CREATE DATABASE IF NOT EXISTS qservReplica;",
```
iagaponenko, line 123: The name of the database is not predefined. It could be different in some test environments as well as in a possible case when the same instance of MySQL would be shared. In this case, one would need the flexibility to name these databases differently.

src/replica/python/schema_migration.py
```diff
@@ -80,6 +92,47 @@ def _set_version(self, version):
...
+            )
+        user = make_url(self.repl_connection).username
+        if not user:
+            raise RuntimeError(
+                "To initialize the replication database, the non-admin connection uri must contain a user "
+                "name."
+            )
+        for stmt in [
+            f"CREATE USER IF NOT EXISTS {user}@localhost;",
+            f"CREATE USER IF NOT EXISTS {user}@'%';",
```
iagaponenko, line 122: Always creating accounts w/o passwords is not safe in some environments, especially because this account is granted all privileges for the Replication database. Here I'm more concerned about accidents (misconfigurations, etc) rather than malicious activities. Accidents of this kind may happen from time to time.

I think Fabrice always initializes database accounts with passwords, including this one.

Would it be possible to add the template placeholder for the password as well?

n8pease:
> I'm a bit confused with the comment in the first commit that says "rename init script to be usable by smig". In actuality, the init script got removed from the package. Did you mean to say "remove"? Or, it was actually re-introduced in a different commit?

Ah. Well, two things happened. The init script that was removed was added for lite, in 7453e20e7fa5bf5c56b10895daae9f78bba1cc36. It had a bug (the wrong table name for schema version) It's not really useful anymore, because any new database can be initialized to the later version. And the `replication.sql` init script was renamed with the migration script naming pattern. I'll update the comment to be more clear.

src/replica/python/schema_migration.py
n8pease: Yes, but let's do it in a different ticket. I don't mind doing it next even, if you think it's high priority. But I want to defer, because this ticket, as-is unblocks Fabrice & will allow him to smig his existing databases to the latest version on Monday. This code, which is missing passwords, won't be run by him on Monday, because it is only run on a brand new, "Uninitialized" instance of qserv. So, it's not needed for what he will do.

Also, I'm not sure if template, or uri option is the correct source for the pw. Currently the {user} comes from the `--repl-connection` URI. Let's discuss it more Monday?

src/replica/python/schema_migration.py
n8pease: it's an error that this line got left in. (The database is created in the function above this one). I'll remove this line.

src/admin/python/cli/script.py
n8pease: connection is the admin URI to the replication database. 

repl_connection is as the comment describes. The `connection` is used to create the user named in `repl_connection`

iagaponenko: Could it be also used as a source of the user's password? The password would be used in:
```
f"CREATE USER IF NOT EXISTS {user}@'%' IDENTIFIED BY '{password}';",
```
It's still fine to have the passwordless entries for logins from `localhost` as in:
```
f"CREATE USER IF NOT EXISTS {user}@localhost;",
```

src/replica/python/schema_migration.py
iagaponenko: 👍

src/replica/python/schema_migration.py
iagaponenko: I think it's easy to extract the password from the same connection string?
Though, if you need to defer this it's fine with me.

iagaponenko (approved):
LGTM.

src/admin/python/cli/script.py
n8pease: > Could it be also used as a source of the user's password?
yes, exactly. As discussed below. Will defer, I've added it to The List.

PR #657, opened 2021-11-09
Serve web dashboard from replication controller

fritzm:
This is largely a subtractive change to the entrypoint cli and docker compose tooling for `lite` – with this approach, the initialization of a separate volume to contain the dashboard static web content and the deployment of a separate nginx container to serve it are no longer required.


src/admin/python/cli/entrypoint.py
```diff
@@ -250,21 +250,6 @@ def replication_controller(**kwargs):
...
-@click.option(
-    "--dashboard-port",
-    help="The port the dashboard will serve on.",
-)
-@click.option(
-    "--dashboard-html",
-    help="The path to the folder with the html sources for the nginx dashboard.",
-)
-@repl_ctrl_domain_name_option()
-@repl_ctrl_port_option()
```
n8pease, line 263: it appears this option can be deleted from options.py, I think this is the only use of it. (repl_ctrl_domain_name_option *is* used in other places tho)

n8pease (approved):
one comment/cleanup item, otherwise LGTM

src/admin/python/cli/entrypoint.py
fritzm: Hmm -- don't see it listed in options.py?

fritzm: I _do_ see it on the host side cli, but I think it is still wanted there -- it is plumbed to the `port` directive in compose.yaml, and still necessary as a default override for users to avoid port collisions in shared environments like `master02`.

PR #658, opened 2021-11-10
DM-32541: Add a revised documentation GitHub Actions workflow

jonathansick:
This PR shows a way towards simpler documentation configuration, builds, and deployments. This workflow is essentially how we're setting up documentation builds for projects that aren't Pipelines projects. The key idea here is to break the Qserv docs away from the pipelines.lsst.io build patterns because they're not relevant here. The other key idea is to drop use of the `lsst-dm/doc-container` repo, which I don't support or endorse the use of.

- New `doc/Makefile` file provides basic commands for building the docs. See `make help`.
- New `doc/conf.py` file that provides key configurations and allows you to easily customize the Qserv docs
- Revised GitHub Actions workflow that directly builds the docs
- I've deleted some doc pages that link to non-existent example files. You can always re-add these pages later with updated info.

One to-do item from this is setting the version info in the `conf.py` file. As a start, the version is hard-coded, but if there's an environment variable, git tag heuristic, etc, then we can update `conf.py` to get that data and provide actual version info.


fritzm (approved):
This looks great; thank you very much!  I will merge presently.

PR #659, opened 2021-11-15
Tickets/dm 32395

iagaponenko:


src/replica/SqlRowStatsJob.cc
```diff
@@ -0,0 +1,332 @@
...
+        for (auto&& row: resultSet.rows) {
+            if (row.nulls[0] || row.nulls[1]) {
+                reportResultThatHas("unexpected NULL values");
+                return !succeeded;
+            }
+            try {
+                TransactionId const transactionId = lsst::qserv::replica::stoui(row.cells[0]);
+                size_t const numRows = std::stoul(row.cells[1]);
+                counters[internalTable][transactionId].push_back(numRows);
+            } catch (...) {
```
andy-slac, line 320: This is a very generic catch, but the code seems to assume it's due to badly formatted numbers?

src/replica/HttpIngestModule.h
```diff
@@ -125,6 +129,48 @@ class HttpIngestModule: public HttpModule {
      */
     nlohmann::json _deleteTable();
 
+    /**
+     * @brief Scan internal tables of a given table to collect row counters
+     *   and return this information to a caller.
+     * 
+     * Depeniding on parameters of the method, the service may also update
```
andy-slac, line 136: Typo: Depeniding

andy-slac (approved):
Looks OK, few minor comments. I do not quite agree with the need to change formatting style, I understand that is your personal preference but would be better to have a common set of rules to avoid unnecessary changes (and reviewing them). I'll post something to Slack on that.

src/replica/ConfigParserMySQL.cc
```diff
@@ -62,6 +68,28 @@ void ConfigParserMySQL::parse() {
...
 
+void ConfigParserMySQL::_parseVersion() {
+    string const table = "QMetadata";
+    string const colname = "value";
+    string version;
+    bool const isNotNull = _conn->executeSingleValueSelect(
+        "SELECT " + _conn->sqlId(colname) + " FROM " + _conn->sqlId(table)
+        + " WHERE " + _conn->sqlEqual("metakey", "version"),
+        colname,
+        version);
```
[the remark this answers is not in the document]
iagaponenko: Good suggestion. Thank you!

src/replica/ConfigParserMySQL.cc
```diff
@@ -62,6 +68,28 @@ void ConfigParserMySQL::parse() {
...
+    if (!isNotNull) {
+        throw ConfigError(_context + " no schema version info found in the metadata table '" + table);
```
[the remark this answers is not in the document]
iagaponenko: 👍

src/replica/DatabaseServicesPool.cc
```diff
@@ -532,30 +548,24 @@ DatabaseServices::Ptr DatabaseServicesPool::_allocateService() {
...
-        [&numRemoved, &service] (DatabaseServices::Ptr const& ptr) {
-            if (ptr == service) {
-                numRemoved++;
-                return true;
-            }
-            return false;
+    _usedServices.remove_if([&numRemoved, &service] (DatabaseServices::Ptr const& ptr) {
+        if (ptr == service) {
+            numRemoved++;
+            return true;
```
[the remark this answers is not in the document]
iagaponenko: My bad. I was tempted to fix the format while adding a new method within the same commit.

src/replica/DatabaseServices.h
```diff
@@ -403,6 +403,53 @@ class DatabaseIngestParam {
...
+ */
+class TableRowStatsEntry {
+public:
+    TransactionId transactionId = 0;
+    unsigned int chunk = 0;     ///< The optional parameter for the "regular" tables.
+    bool isOverlap = false;     ///< The optional parameter for the "regular" tables.
+    size_t numRows = 0;
+    uint64_t updateTime = 0;
+
+    explicit TableRowStatsEntry(
```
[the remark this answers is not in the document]
iagaponenko: Agreed. Not needed. I had this keyword in the original version of the c-to that had default values for all parameters. Then forgot to eliminate the keyword.

src/replica/SqlRowStatsJob.cc
iagaponenko: Yes, because the only method where it's expected to fail is calling:
```
lsst::qserv::replica::stoui
std::stoul
```
Dig into fine details of a problem makes very little sense in this context. The catch was just meant to ensure the result set received from a worker is not corrupt.

andy-slac: This is not quite true even for current code, `operator[]` and `push_back` may throw if memory allocation fails.

iagaponenko: 👍

PR #660, opened 2021-11-16
DM-32606: qserv integration test case01 fails to load object table

n8pease:


src/admin/etc/integration_tests.yaml
```diff
@@ -7,7 +7,7 @@ czar-db-admin-uri: mysql://root:CHANGEME@czar-db:3306
 # the folder where the testdata repo will be downladed:
 qserv-testdata-dir: /qserv/data/qserv_testdata
 # location where a tarball can be downloaded from:
-qserv-testdata-repo: https://github.com/lsst/qserv_testdata/tarball/master
+qserv-testdata-repo: https://github.com/lsst/qserv_testdata/tarball/tickets/DM-32606 # REMOVE BEFORE MASTER
```
fritzm, line 10: "REMOVE BEFORE MASTER"?

fritzm (approved):
Looks okay, but: are we sure we don't care whether we create tables as "root" vs "qsmaster"?  Could, for example, the grants be behaving differently in here now?  Perhaps we don't care for now, but if we ever had integration tests that tested the grants themselves (qserv vs. standalong mariadb) we probably would care...

src/admin/etc/integration_tests.yaml
n8pease: thanks for the reminder. Chicken/egg problem here - need this (actually, not today, but in theory) for PR to pass CI itest. OTOH don't want it to go into master. Prolly would want an algorithm: "get the branch name, and if it exists in qserv_testdata, use that repo, else use master". But that's for another day.

PR #662, opened 2021-11-22
DM-32677: improve itest_load connection management, and remove -j default from qserv build

n8pease:


src/admin/python/itest_load.py
```diff
@@ -54,22 +54,29 @@ def unzip(source: str, destination: str) -> None:
         shutil.copyfileobj(_source, _target)
 
 
-def execute(cursor: MySQLCursor, stmt: str, multi: bool = False) -> Sequence[Sequence[Union[str, int]]]:
-    """Execute a statement in a cursor, and clean up the cursor
-    so it can be closed."""
+def execute(cursor: MySQLCursor, stmt: str, multi: bool = False):
```
fritzm, line 57: would be nice to eliminate repeated code here if we can?

fritzm (approved):
One minor comment, for your consideration, otherwise LGTM and thanks!

src/admin/python/itest_load.py
n8pease: done

PR #663, opened 2021-11-27
DM-32708: Fix integration test scripts after master -> main transition

fritzm:
Pull test data from main instead of master


n8pease:
LGTM

PR #668, opened 2021-12-10
DM-32900: add an options-file option to qserv entrypoint commands

n8pease:


n8pease:
help output for entrypoint commands now looks like 
```
$ entrypoint cmsd-manager --help
Usage: entrypoint cmsd-manager [OPTIONS]

Options:
  --cms-delay-servers TEXT  The value for 'cms.delay servers' in the cmsd-
                            manager.cf file.
  -@, --options-file TEXT   Path to a YAML file containing overrides of
                            command line options. The YAML should be organized
                            as a hierarchy with subcommand names at the top
                            level, options names and values (as key-value
                            pairs) for that subcommand below. The option name
                            should NOT include prefix dashes.
  --help                    Show this message and exit.
```

PR #670, opened 2021-12-12
DM-32945: Reinforce qhttp static handler for paths with embedded nulls

fritzm:
Return a 404 for any resource path routed to a static handler that contains embedded nulls, before letting the resource path string hit any `boost::filesystem` calls.

This PR also fixes and re-enables the `qhttp` unit tests, which had been disabled during the port to `lite`.

### DIFF
--- a/src/qhttp/CMakeLists.txt
+++ b/src/qhttp/CMakeLists.txt
@@ -26,6 +26,4 @@ target_link_libraries(testqhttp PUBLIC
     Threads::Threads
 )
 
-add_test(NAME test_qhttp COMMAND testqhttp)
-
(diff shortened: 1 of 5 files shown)


iagaponenko (approved):
LGTM as a temporary workaround for an immediate problem. Though, this specific problem and a wider class of similar problems need to be addressed differently. Specifically, I propose:
- putting the `try ... catch` block in the whole body of the static handler
- adding two optional user-specific callbacks (lambdas) to the constructor of the class `qhttp` to allow reporting incoming requests (before they're processed) and reporting errors on exceptions or any problems detected within the body if the static handler.

fritzm:
Okay, thanks -- per comments in Slack I will merge this, then begin immediately on another ticket to add exception handling and logging.

PR #671, opened 2021-12-16
Tickets/dm 32885

iagaponenko:


src/admin/python/lsst/qserv/admin/cli/script.py
```diff
@@ -706,7 +705,42 @@ def enter_replication_controller(
         print the command and arguments that would have been run.
     """
 
-    def set_initial_configuartion(workers: Sequence[str], xrootd_manager: str) -> None:
+    def get_worker_id(qserv_worker_db: str) -> str:
```
n8pease, line 708: It seems like this may be more generally useful, could probably be in the utils file (in the same dir as this file). Also, the multiple nested functions in entery_proxy seems like too much (the other one made sense to me because it was being used as a lambda, but maybe even it belongs somewhere else). But since getting the worker id may be useful other places, move it anyway... agree?

src/admin/python/lsst/qserv/admin/schema_migration.py
```diff
@@ -34,8 +34,7 @@
 class AdminMigrationManager(SchemaMigMgr):
     """Class implementing schema migration for QMeta database."""
 
-    # scripts are located in qmeta/ sub-dir
-    def __init__(self, name: str, connection: str, scripts_dir: str):
```
n8pease, line 38: nice. I've been putting off removing "name", glad you did it.

src/qmeta/schema/migrate-3-to-4.sql
```diff
@@ -8,3 +8,6 @@ ALTER TABLE `QInfo`
     MODIFY COLUMN `qTemplate`   MEDIUMTEXT NOT NULL     COMMENT 'Query template, string used to build final per-chunk query.',
     MODIFY COLUMN `qMerge`      MEDIUMTEXT DEFAULT NULL COMMENT 'Merge (or aggregate) query to be executed on results table, result of this query is stored in merge table. If NULL then it is equivalent to SELECT *.',
     MODIFY COLUMN `resultQuery` MEDIUMTEXT DEFAULT NULL COMMENT 'Query to be used by mysqlproxy to get final results.';
+
+-- Update record for schema version, migration script expects this record to exist.
+UPDATE `QMetadata` SET `value` = '4' WHERE `metakey` = 'version';
```
n8pease, line 13: I think this should not be done here. the SchemaMigMgr subclass applies the migration files, and is allowed to do other work (which is why the version should not be set), and setting the version should be the very last thing it does.

n8pease: (same comment applies other `migrate-` files where inserting the version has been added.

src/replica/python/schema_migration.py
```diff
@@ -53,19 +51,14 @@ class MasterReplicationMigrationManager(SchemaMigMgr):
...
         self.repl_connection = repl_connection
-        self.set_initial_configuartion = set_initial_configuration
+        self.set_initial_configuration = set_initial_configuration
         super().__init__(scripts_dir, connection)
-        if not self.database:
-            raise RuntimeError(
-                "The name of the replication database must be provided in the connection URI."
-            )
 
     def current_version(self) -> Version:
```
n8pease, line 63: it's nice that you are standardizing the pattern, `current_version` and `_set_version` can probably be moved down to the base class?

src/admin/python/lsst/qserv/admin/cli/script.py
iagaponenko: Good suggestion! I will move it.

src/replica/python/schema_migration.py
```diff
@@ -99,21 +98,21 @@ def _set_version(self, version: int) -> None:
...
         """Create the replication controller database.
         """
         with closing(self.connection.cursor()) as cursor:
-            cursor.execute(f"CREATE DATABASE {self.database};")
+            cursor.execute(f"CREATE DATABASE {database};")
             warnings = cursor.fetchwarnings()
             if warnings:
-                _log.warn("Warnings were creating database %s", self.database)
-        _log.info(f"Created database {self.database}.")
+                _log.warn(f"Warnings were creating database {database}.")
```
n8pease, line 114: f strings should not be used in _log statements (the previous way is the correct the way). This is because it's more efficient to delay substituting the parameters into the string as much as possible, in case the string won't be used.

src/qmeta/schema/migrate-3-to-4.sql
iagaponenko: Well, I myself was not certain if putting the version number into the SQL files was the right thing to do. I just see this in some, and not in other files. So, if you're sure this shouldn't be done in the `migrate-<from>-<to>.sql` files then perhaps we should do it consistently on all files?
@andy-slac What do you think about this?

src/worker/python/schema_migration.py
```diff
@@ -19,23 +19,29 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-"""Module defining methods used in schema migration of Worker database.
+"""Module defining methods used in schema migration of Worker metadata database.
```
n8pease, line 22: I think we figured out that we need wdb or worker schema management, but not both, right? Can one of them be removed? (Or, why make the changes to wdb and how are they tested?)

src/replica/python/schema_migration.py
iagaponenko: Exactly. And not only that. I was thinking about introducing thie class:
```
class StandardReplicationMigrationManager(SchemaMigMgr):
   ...
```
Which would implement the standard way of managing migrations without reductions and complications. After that in the module-specific `schema_migration` we could have the following class factory function:
```
from lsst.qserv.schema import StandardReplicationMigrationManager

database = "qservMeta"

def make_migration_manager(
    connection: str,
    scripts_dir: str,
    set_initial_configuration: Optional[Callable[[], None]] = None
) -> SchemaMigMgr:
 
    return StandardReplicationMigrationManager(
        database,
        connection,
        scripts_dir,
        set_initial_configuration
    )
```
The next step would be to migrate the schema migration files of the relevant modules to allow using the "standard" migration protocol.

IN those special cases where we can't do we could override relevant methods of `StandardReplicationMigrationManager` as:
```
class QMetaMigrationManager(StandardReplicationMigrationManager):
    ...
```

iagaponenko: ^ sorry, I had to fix s "bug" in the example above.

src/replica/python/schema_migration.py
iagaponenko: Thanks! I didn't know that. I will put the original code back.

src/worker/python/schema_migration.py
iagaponenko: Well, that's is one of the questions I too wanted to ask you. I'm leaning toward removing the `worker` module entirely in favor of the existing module `wdb` because the latter has C++ classes depending on the database schema. And the former (f I understood you correctly) was introduced purely for schema management purposes.
I can do this if you agree (?)

src/qmeta/schema/migrate-3-to-4.sql
n8pease: I think removing it from migrate- files is correct. I may have added it in some cases before I had learned more/developed an opinion. But I'm also curious what Andy will say. (Probably something I haven't thought of entirely haha)

src/replica/python/schema_migration.py
n8pease: I don't see why we need a "standard" migration manager. Just implement default behaviors in `SchemaMigMgr` and as you suggested with `StandardMigrationManager`, allow subclasses to override default behavior as needed.

src/worker/python/schema_migration.py
n8pease: yes, I agree.

src/replica/python/schema_migration.py
iagaponenko: That's a very good option too. In this case, we may need to add the name of a database as a parameter of class `SchemaMigMgr`. We can't rely on the database name in the connection string because we may use the same connection string to manage different databases. A good example would be:
```
mysql://root:CHANGEME@127.0.0.1:3306?socket=/qserv/mariadb/run/mysqld.sock
```
or this one:
```
mysql://root:CHANGEME@czar-db:3306/qservMeta
```
Each could be used for managing schemas of databases:
```
qservMeta
qservCssData
qservResults
...
```
Or any other databases at czar's MariaDB.

src/admin/python/lsst/qserv/admin/cli/script.py
iagaponenko: Well, I have to correct myself - moving this method into a different script is not possible because its implementation depends on data members and other methods defined within the file `scipt.py`. So, I would just leave it here for now.

src/qmeta/schema/migrate-3-to-4.sql
iagaponenko: I've inspected the schema migration code. and it looks like the version management is done by the package-level `schema_management.py` files, not by the main `smig` tool. I've cleaned up the schema version update from the SQL files except for those cases where the initial version is inserted (SQL's `INSERT INTO QMetadata ...`) right after creating that table.

src/replica/python/schema_migration.py
iagaponenko: Upon further thinking on this improvement, I decided not to do it in this ticket. This will be postponed before we'll decided on a comprehensive solution for schema management (which I expect will also affect the schema migration tools).

src/worker/python/schema_migration.py
iagaponenko: This will be postponed before we'll decided on a comprehensive solution for schema management (which I expect will also affect the schema migration tools).

src/replica/testCommonStr.cc
```diff
@@ -0,0 +1,88 @@
...
+        words = strsplit("_a_bc_def_", '_');
+        BOOST_CHECK_EQUAL(words.size(), 3U);
+        BOOST_CHECK_EQUAL(words[0], "a");
+        BOOST_CHECK_EQUAL(words[1], "bc");
+        BOOST_CHECK_EQUAL(words[2], "def");
+    });
+    LOGS_INFO("CommonStrTest test ends");
+}
+
+BOOST_AUTO_TEST_SUITE_END()
```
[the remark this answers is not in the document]
iagaponenko: 👍

n8pease:
LGTM, thanks!

PR #674, opened 2022-01-06
Add error reporting for operations with qhttp

iagaponenko:


src/replica/HttpSvc.cc
```diff
@@ -64,10 +64,11 @@ HttpSvc::~HttpSvc() {
 
 
 void HttpSvc::run() {
-    LOGS(_log, LOG_LVL_INFO, context() << __func__);
+    string const context_ = context() + " " + string(__func__) + " ";
+    LOGS(_log, LOG_LVL_INFO, context_);
```
fritzm, line 68: If these log messages with just `context_` are all over the place (presumably for tracing call execution) should they maybe be `LOG_LVL_TRACE` instead of `LOG_LVL_INFO`?

fritzm (approved):
LGTM -- one minor question re. TRACE vs. INFO, but leave it up to your judgement.

src/replica/HttpSvc.cc
iagaponenko: Good catch! Fixed.

PR #675, opened 2022-01-08
Tickets/dm 33145

iagaponenko:


src/replica/Application.cc
```diff
@@ -105,6 +108,20 @@ int Application::run() {
             "Change the default value limiting a duration of each attempt to execute"
             " a database transaction before to fail.",
             _databaseTransactionTimeoutSec
+        ).option(
+            "schema-upgrade-wait",
+            "If a value of the option is 0 and the schema version of the Replication/Ingest system's"
```
n8pease, line 113: "If a value" -> "If the value"

src/replica/Configuration.cc
```diff
@@ -172,6 +180,33 @@ unsigned int Configuration::databaseTransactionTimeoutSec() {
...
+    util::Lock const lock(_classMtx, _context(__func__));
+    return _schemaUpgradeWaitTimeoutSec;
+}
+
+
+void Configuration::setSchemaUpgradeWaitTimeoutSec(unsigned int value) {
+    util::Lock const lock(_classMtx, _context(__func__));
+    if (0 == value) {
+        throw invalid_argument("Configuration::" + string(__func__) + "  0 is not allowed.");
+    }
```
n8pease, line 205: I suppose you could get rid of the option to wait or not, and just use "upgrade wait time"==0 to mean "don't wait for upgrade? Seems a little simpler, but either way seems ok to me.

src/replica/Configuration.cc
```diff
@@ -341,13 +376,47 @@ void Configuration::_load(util::Lock const& lock, string const& configUrl, bool
...
+            });
+            break;
+        } catch (ConfigVersionMismatch const& ex) {
+            if (Configuration::schemaUpgradeWait()) {
+                if (ex.version > ex.requiredVersion) {
+                    LOGS(_log, LOG_LVL_ERROR, _context() << "Database schema version is newer than"
+                        << " the one required by the application, ex: " << ex.what());
+                        throw;
+                }
+                schemaUpgradeTimer.stop();
```
n8pease, line 401: do you need to start the timer again?

src/replica/Configuration.cc
iagaponenko: I too had the very same idea before I realized that I would prefer keeping knobs for "enabling something" and "setting a timeout for waiting for some event to happen" semantically separate both in the interface and the dependent code.

n8pease:
overall LGTM. I don't know the timer API but there might be one small problem, I'll leave it to you to investigate.

src/replica/Configuration.cc
iagaponenko: Not in this case. The name of the class `util::Time` that's I'm using is confusing. In reality, it's not a "timer" (the method won't start any system timer, like asynchronous timers in BOOST ASIO). This is just a simple utility class for capturing and measuring time intervals by calling methods:
```
start() - capture the current time and store it in a member variable
stop() - capture the current time and store it in another member variable
```
And there is another method that computes a difference (an interval passed) between those two member variables:
```
timeDiff()
```
That's why I'm calling `start()` just once at the very beginning of waiting for the schema upgrade tracking loop. Then I'm calling `stop()` after each failed check in order to "extend" the interval of time stored in that "Timer" object. And this is going on until the "timer" "expires".
Norte details can be found in the header of the class:
https://github.com/lsst/qserv/blob/main/src/util/Timer.h

src/replica/Application.cc
iagaponenko: 👍

PR #676, opened 2022-01-13
Tickets/dm 33215

iagaponenko:


src/replica/ConfigurationSchema.h
```diff
@@ -131,6 +135,42 @@ class ConfigurationSchema {
...
+     * @param param The name of the parameter within its category.
+     * @param attr The name of the attribute.
+     * @param defaultValue The default value to be returned if no such parameter
+     *   or its attribute was fount.
+     * @return T The value of the attribute (or the default value).
+     */
+    template <typename T>
+    static T _attributeValue(std::string const& category, std::string const& param,
+                             std::string const& attr, T const& defaultValue) {
+        if (_schemaJson.count(category)) {
```
andy-slac, line 152: Maybe `contains()` instead of `count()`? Though in this case `find()` should be better.

src/replica/ConfigurationSchema.h
```diff
@@ -131,6 +135,42 @@ class ConfigurationSchema {
...
 private:
+    /**
+     * @brief Retreive a value of the parameter's attribute (allow default value).
+     * 
+     * @tparam T The type of the attribute's value.
+     * @param category The name of the parameter's category.
+     * @param param The name of the parameter within its category.
+     * @param attr The name of the attribute.
+     * @param defaultValue The default value to be returned if no such parameter
+     *   or its attribute was fount.
```
andy-slac, line 146: typo: fount

src/replica/ConfigurationSchema.cc
```diff
@@ -435,12 +446,30 @@ string ConfigurationSchema::json2string(string const& context, json const& obj)
...
 json ConfigurationSchema::_restrictor(string const& category, string const& param) {
-    return _attributeValue<json>(_schemaJson, category, param, "restricted", json());
+    return _attributeValue<json>(category, param, "restricted", json());
+}
+
+
+json ConfigurationSchema::_attributeValueJson(string const& category, string const& param,
+                                              string const& attr) {
+    if (_schemaJson.count(category)) {
+        json const& categoryJson = _schemaJson.at(category);
```
andy-slac, line 461: `find()` should be better here too.

src/replica/Application.cc
```diff
@@ -138,20 +135,19 @@ int Application::run() {
...
-        );
+        // Inject options for th egeneral configuration parameters.
+        for (auto&& itr: ConfigurationSchema::parameters()) {
+            string const& category = itr.first;
+            for (auto&& param: itr.second) {
+                // The read-only parameters can't be updated programmatically.
+                if (ConfigurationSchema::readOnly(category, param)) continue;
+                _general[category][param] = ConfigurationSchema::defaultValueAsString(category, param);
+                parser().option(
+                    category + "." + param,
```
andy-slac, line 146: So instead of `--xrootd-allow-reconnect` you now have `--xrootd.allow-reconnect`? That looks a bit odd, can you keep dash separator?

src/replica/Application.cc
```diff
@@ -182,20 +178,27 @@ int Application::run() {
         Configuration::setSchemaUpgradeWaitTimeoutSec(_schemaUpgradeWaitTimeoutSec);
     }
     if (_enableServiceProvider) {
+        _serviceProvider = ServiceProvider::create(_config, _instanceId);
 
-        // Create and then start the provider in its own thread pool before
-        // performing any asynchronous operations via BOOST ASIO.
-        //
+        // Update general configuration parameters.
+        // Note that options specified by a user will have non-empty values.
```
andy-slac, line 184: Does this imply that options not specified by user have an empty value? Cannot user say `--option=""` which will make an empty value?

src/replica/Application.h
```diff
@@ -152,14 +149,15 @@ class Application: public std::enable_shared_from_this<Application> {
...
-    unsigned int _xrootdAllowReconnect;
-    unsigned int _xrootdConnectTimeoutSec;
 
     // Schema upgrade waiting options
 
     unsigned int _schemaUpgradeWait;
     unsigned int _schemaUpgradeWaitTimeoutSec;
 
+    /// General parameters
+    std::map<std::string, std::map<std::string, std::string>> _general;
```
andy-slac, line 159: Maybe name it `_generalParams` instead of `_general`?

andy-slac (approved):
Looks OK, few minor comments. I find `--xrootd.host=...` command line option looking very odd.

iagaponenko:
> Looks OK, few minor comments. I find `--xrootd.host=...` command line option looking very odd.

You may be correct - this may look odd from the aesthetical point of view. This is a byproduct of the logical grouping of the parameters into "categories". In this example, `xrootd` represents the category. When building names of the parameters I'm concatenating the name of the category and the name of the parameter within its category. I could `-` instead. I will look at the code to see how to improve this.

There is a related issue - the names of the parameters in the internal dictionary have underscores, like in this example:
```
---worker.svc_port=
```
If I change the separator then it will look like this:
```
---worker-svc_port=
```
which would be odd as well. I reason why I chose underscores was because I was trying to make the life of the Python and JavaScript clients easier.:
```
obj.svc_port
```
vs:
```
obj["svc-port"]
```
Perhaps, I was wrong. I should revisit this topic and rename the parameters.

src/replica/ConfigurationSchema.h
iagaponenko: Yes, that one would work: https://json.nlohmann.me/api/basic_json/contains/

src/replica/ConfigurationSchema.h
iagaponenko: 👍

src/replica/ConfigurationSchema.cc
iagaponenko: Good idea!

src/replica/Application.cc
iagaponenko: I will fix that.

src/replica/Application.cc
iagaponenko: Yes to both questions. Right now, none of the existing parameters has an empty value. Please, see `json const ConfigurationSchema::_schemaJson` in https://github.com/lsst/qserv/blob/main/src/replica/ConfigurationSchema.cc

Besides, the command line parser that I'm using here supports:
- positional parameters (with a possibility to specify some trailing parameters as optional)
- options (which must have non-empty value)
- flags

Examples:
```
required-1 required-2 [optional-1 [ optional2 ... ]]
--option=val
--flag
```
Configuration parameters in the "general" category are represented by the command-line "options" (as defined above). If I have a logical flag then it would be still represented with an option that would have a numeric value:
```
--logical-option={0|1}
```

PR #677, opened 2022-01-14
DM-33277: improve qserv entrypoint for configurability

n8pease:
I started a confluence page that [describes entrypoint, including extended args and templates](https://confluence.lsstcorp.org/display/DM/Qserv-Lite+entrypoint).

I did not apply the new templating system to schema migration files, since we're going to be refactoring that system soon anyway, and it will need design work first.

Similarly, the replication controller init code is not (yet) configurable. We should discuss that and put it into a separate ticket if needed. (we can do it next if it's urgent).


src/admin/python/lsst/qserv/admin/qservCli/launch.py
```diff
@@ -335,7 +336,7 @@ def mypy(
         "--exclude",
         "testing",
     ]
-    mypy_ini_file = os.path.join(qserv_build_root.format(user=user), "mypy.ini")
+    mypy_ini_file = os.path.join(qserv_build_root.format(user=user), mypy_cfg_file)
```
iagaponenko, line 339: Besides "pleasing the boss" (which is probably a good justification by itself 😹 ), what's a reason for changing an expectation from finding this file in this root folder of the build tree to its sub-folder `src/admin/python/mypy.ini`? In the older version of the code, the file was expected to be found at:
```
<QSERV_BUILD_ROOT>/mypy.ini
```
And now it's expected to be at:
```
<QSERV_BUILD_ROOT>/src/admin/python/mypy.ini
```
Is the source tree still available within the binary container?

src/admin/python/lsst/qserv/admin/cli/entrypoint.py
```diff
@@ -318,31 +359,42 @@ def worker_xrootd(
     )
 
 
-@entrypoint.command(help=f"Start as a replication worker node.\n\n{socket_option_description}")
-@instance_id_option(required=True)
+@entrypoint.command(
+    help="Start as a replication worker node.\n\n"
+         "{socket_option_description}\n\n"
```
iagaponenko, line 570: The string `f` (format type) was dropped in the previous line. Was this done on purpose? Would the parameter substitution still work for `{socket_option_description}`?

admin/local/docker/compose/docker-compose.yml
```diff
@@ -348,11 +354,13 @@ services:
...
           --db-admin-uri mysql://root:CHANGEME@repl-mgr-db:3306/qservReplica
-          --qserv-czar-db mysql://root:CHANGEME@czar-db:3306/qservMeta
-          --instance-id qserv_proj
           --worker qserv_worker_db=mysql://qsmaster@worker-db-0:3306/qservw_worker,host=worker-repl-0
           --worker qserv_worker_db=mysql://qsmaster@worker-db-1:3306/qservw_worker,host=worker-repl-1
           --xrootd-manager manager-xrootd
+          --
+          --instance-id=qserv_proj
+          --qserv-czar-db=mysql://root:CHANGEME@czar-db:3306/qservMeta
+          --http-root=/usr/local/qserv/www
```
iagaponenko, line 363: In practice, this option (`--http-root`) is de-facto mandatory in all Qserv deployments. Hence it should be present in the command template and (normally, unless one is experimenting with some non-standard configuration) be not present in the Docker compose config because a location of the root folder is fixed within the `lite` container: `/usr/local/qserv/www`.

If you think a user should have an option to override its value then you could move this option "to the left", before `--`.

admin/local/docker/compose/docker-compose.yml
```diff
@@ -348,11 +354,13 @@ services:
...
           --db-uri mysql://qsreplica@repl-mgr-db:3306/qservReplica
           --db-admin-uri mysql://root:CHANGEME@repl-mgr-db:3306/qservReplica
-          --qserv-czar-db mysql://root:CHANGEME@czar-db:3306/qservMeta
-          --instance-id qserv_proj
           --worker qserv_worker_db=mysql://qsmaster@worker-db-0:3306/qservw_worker,host=worker-repl-0
           --worker qserv_worker_db=mysql://qsmaster@worker-db-1:3306/qservw_worker,host=worker-repl-1
           --xrootd-manager manager-xrootd
+          --
+          --instance-id=qserv_proj
+          --qserv-czar-db=mysql://root:CHANGEME@czar-db:3306/qservMeta
```
iagaponenko, line 362: This option (`--qserv-czar-db=`) is de-facto mandatory in all Qserv deployments because it's used by the Controller to update `czar`'s database during catalog ingests. I would recommend keeping it in the list of the entry point's options (before`--`).

iagaponenko (approved):
Looks good to me (meaning - the code seems to implement what it's supposed to do). Though, I have a couple of general observations. First of all, the complexity of the code keeps growing, mostly because it has to handle more command-line options which are passed around (within the Python code) as individual parameters. Perhaps, there is a way to package those parameters into a dictionary and pass the dictionary around? The same comment for the decorators. The list of decorators are getting very long for some functions.

My second comment is regarding database connector parameters in entry points for the Replication system's controller and workers, including:
```
--db-uri=<>
--repl-connection=<>
--czar-db-uri=<>
--worker-db-uri=<>
--config=<>
```
One problem with that is that passing these connectors as the entry point option poses a security risk by exposing passwords. Thought, a bigger problem (from my point of view) is that this would complicate using these in the Kubernetes-based deployments where some components of the connection strings could be injected into the entry points from other sources (config maps, or secrets). This includes user names and MySQL passwords. Besides, some parts of the connection strings should be constant (like the names of databases). Therefore, I think an alternative (a better) approach would be to build the connection strings inside the entry points using components harvested from various sources (config maps and secrets in Kubernetes, env variables passed into the containers via the option `-e`, the new feature - the template argument option `--targs`, or explicit parameters of the entry points).

For example, instead of this URI:
```
--repl-connection=<uri>
```
The entry point would offer two options:
```
--repl-db-host=<default: 127.0.0.1>
--repl-db-port=<default: 3306>
```
And assume that the connection string has the fixed name of the user `qsreplica`, the fixed port `3306`, and the fixed name of the database `qservReplica`. Database password (if any) would have to be injected into the entry point from some secure source.

src/admin/python/lsst/qserv/admin/qservCli/launch.py
n8pease: it was and will be available in the build container (when qserv sources are mounted in). It never was and is not part of the run image.
The reason is to keep the top-level directory of the qserv sources free of clutter, it's a style preference on @fritzm 's part.

src/admin/python/lsst/qserv/admin/cli/entrypoint.py
n8pease: It was done on purpose. This is not an f string, because `socket_option_description` is not a valid variable when this string is being created (which is how an f-string would get the value of the var). Instead, this string can be rendered later using `.format` e.g.
```
s = "I'd like to say {something}"
...
print(s.format(something="hello world")
```

src/admin/python/lsst/qserv/admin/cli/options.py
```diff
@@ -237,3 +237,36 @@
 below. The option name should NOT include prefix dashes.""",
     callback=yaml_presets,
 )
+
+
+cmd_option = partial(
+    click.option,
+    "--cmd",
```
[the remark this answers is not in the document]
n8pease: We are migrating toward as much execution-time configurability as possible. This allows more configurability without having to rebuild the run image. We hope that this will make development and debugging easier in some cases, and enable more kinds of emergency patches.

src/admin/templates/xrootd/etc/cmsd-manager.cf.jinja
```diff
@@ -35,7 +35,7 @@ xrd.network nodnr
 xrd.network dyndns
 xrd.network cache 0
 
-all.manager {{ cmsd_manager }}:2131
+all.manager UNUSED:2131
```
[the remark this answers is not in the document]
n8pease: I was only trying to reduce unnecessary templatized parameters, as you and I have discussed as being a goal for simplicity. But, if it can be removed that would be nice. I think we must specify the port number for configuration, and in the docs `name:port` is given as the example. Perhaps I can remove the name, I'll try it.

src/admin/python/lsst/qserv/admin/cli/entrypoint.py
```diff
@@ -346,8 +539,16 @@ def worker_xrootd(
     repl_ctl_dn: str,
     mysql_monitor_password: str,
     db_qserv_user: str,
+    cmsd_worker_cfg_file: str,
+    cmsd_worker_cfg_path: str,
+    xrdssi_cfg_file: str,
+    xrdssi_cfg_path: str,
+    targs: Dict[str, str],
+    targs_file: str,
+    cmd: str,
```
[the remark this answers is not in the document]
n8pease: unforts, dict-instead-of-args is an anti-pattern. It's opaque, very hard to know what's being passed around and why.

src/admin/python/lsst/qserv/admin/cli/entrypoint.py
```diff
@@ -371,20 +577,27 @@ def worker_xrootd(
     help=f"{repl_connection_option.keywords['help']} {socket_option_help}"
 )
 @debug_option()
+@cmd_option(default="qserv-replica-worker --qserv-worker-db={{db_admin_uri}} {% for arg in extended_args %}{{arg}}  {% endfor %}")
```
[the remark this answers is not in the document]
n8pease: @iagaponenko I don't see a comment re. `db_admin_url`, is it missing or can you point me to a file & function?

`entrypoint` does not do anything with the `--config` option right now. What are you imagining it might need to do, or how would it need to be part of the `entrypoint` template processing?

admin/local/docker/compose/docker-compose.yml
n8pease: `entrypoint` does not need it, and we'd have to reproduce the help and so forth. Isn't this inconsistent with what you've said about `entrypoint` not getting in-between the CLI and the options used by the repl binary? We should discuss this during cowork?

admin/local/docker/compose/docker-compose.yml
n8pease: All options at this level are known to & settable by the user, (IMO) the `--` does not factor into that. (As above, `--` separates options known to `entrypoint` and those passed unchanged to the binary).

If you think `--http-root` belongs in the command template we can do that, although at that point, would it make sense to just make "/usr/local/qserv/www" a default value for `--http-root` in `replication-controller`?

n8pease:
WRT your second point, I'm hearing several sub-items:
1. pass values by 
 - config file (secrets, config maps)
 - environment variable
2. separate URIs, make options for variable parts (e.g. hostname) and bake-in non-variable parts (username qsmaster).
3. security issues

re 1: we can pass options via config file right now, using the `entrypoint -@/--options-file` option. We can also add pass-by environment variable very easily, though it's not done right now.

re 2: I don't mind evolving toward this approach, tho I think it's separate from this ticket and I'd want to be sure that other team members (@fritzm) agree.

re. 3 - are the perceived security issues because passwords would be saved in checked-in sources?
In GitHub actions, secrets can be substituted into the source, I'm pretty sure k8s has a similar feature. For example, in the checked in yaml I use `${{ secrets.FM_DOCKERHUB_TOKEN_USER }}` and that's replaced by named secret before executing.

admin/local/docker/compose/docker-compose.yml
iagaponenko: Regarding setting some reasonable default value for the folder in the controller's implementation. That would result in starting the `qhttp` service. Right now the service would only start if the non-empty option `--http-root` is provided. Perhaps we should fix that.

admin/local/docker/compose/docker-compose.yml
iagaponenko: The idea that I had in mind here was to build (or amend) that URI by the entry point based on the security tokens (password, etc.) injected into the containers via Kubernetes config-maps, security, or environment variables, or some other mechanism. The same applies to all database connectors.
Let's discuss this topic at the group meeting!

src/admin/python/lsst/qserv/admin/cli/entrypoint.py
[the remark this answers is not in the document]
iagaponenko: It's in the same line for which I made the comment (look for `--qserv-worker-db`):
```
@cmd_option(default="qserv-replica-worker --qserv-worker-db={{db_admin_uri}} ...
```

PR #680, opened 2022-01-24
DM-33349 add spawned-app --help to "qserv entrypoint-help"

n8pease:


src/admin/python/lsst/qserv/admin/qservCli/launch.py
```diff
@@ -1166,22 +1168,26 @@ def entrypoint_help(
...
-    if command:
-        args.append(command)
-    args.append("--help")
-    if dry:
-        print(" ".join(args))
-    else:
-        _log.debug('Running "%s"', " ".join(args))
-        subprocess.run(args, check=True)
+    if entrypoint:
+        print(f"Help output for 'entrypoint':\n")
```
iagaponenko, line 1179: Does this need to be the `f`-string?
I also have an impression this printout is not neederd.

iagaponenko:
One quick observation. After I built and tested the new command, it seems to report some internal debuging message that I don't think s supposed to be here:
```
% qserv entrypoint-help
Help output for 'entrypoint':

Usage: entrypoint [OPTIONS] COMMAND [ARGS]...

Options:
...
```

iagaponenko:
Next issue. I don't understand what the following command does:
```
% qserv entrypoint-help
Help output for 'entrypoint':

Usage: entrypoint [OPTIONS] COMMAND [ARGS]...

Options:
  ...
Commands:
  ...
  spawned-app-help        Print the help output of a spanwed command for...
```
After invoking the new command, I'm seeing this:
% qserv entrypoint-help spawned-app-help
Help output for 'entrypoint':

Usage: entrypoint spawned-app-help [OPTIONS] COMMAND

  Print the help output of a spanwed command for a given entrypoint subcommand
  or a message indicating why help is not avialble.

  Help is available for entrypoint commands that spawn an app. If the named
  command does not exist or is misspelled is misnamed the message will say it
  does not exist. Some entrypoint commands perform work without spawning an
  app, in which case it will say that no spawned app help is available for
  that command.
  ...
```
And now I'm trying to use it w/o any success:
```
% qserv entrypoint-help spawned-app-help replication-controller
Usage: qserv entrypoint-help [OPTIONS] [COMMAND]
Try 'qserv entrypoint-help --help' for help.

Error: Got unexpected extra argument (replication-controller)
```
```
 % qserv entrypoint-help spawned-app-help qserv-replica-master-http
Usage: qserv entrypoint-help [OPTIONS] [COMMAND]
Try 'qserv entrypoint-help --help' for help.

Error: Got unexpected extra argument (qserv-replica-master-http)
```

File: src/admin/python/lsst/qserv/admin/cli/entrypoint.py [hunk 2]
```diff
@@ -748,3 +767,26 @@ def smig_update(czar_connection: str, worker_connections: List[str], repl_connec
...
         repl_connection=repl_connection,
     )
+
+
+@entrypoint.command()
+@click.argument("COMMAND")
+def spawned_app_help(
+    command: str,
+) -> None:
+    """Print the help output of a spanwed command for a given
```
iagaponenko, line 777: Typo in `spanwed`

src/admin/python/lsst/qserv/admin/cli/entrypoint.py
```diff
@@ -748,3 +767,26 @@ def smig_update(czar_connection: str, worker_connections: List[str], repl_connec
...
+    entrypoint subcommand or a message indicating why help is not avialble.
```
iagaponenko, line 778: Typo in `avialble`

src/admin/python/lsst/qserv/admin/cli/entrypoint.py
```diff
@@ -409,7 +433,7 @@ def proxy(
     show_default=True,
 )
 @targs_options()
-@cmd_option(default="cmsd -c {{cmsd_manager_cfg_path}} -n manager -I v4")
+@cmd_option(default=commands["cmsd-manager"].default_cmd)
```
iagaponenko, line 436: As far as I can see, the only for the `cmd_option` decorator is to inject the default command template. All uses are in the current nodule. Why don't do this here and in the rest of the module?
```
@cmd_option("cmsd-manager")
```
The implementation of the decorator would pull the collection of the commands and find the right default for the specified command....in case if your implementation permits it.

src/admin/python/lsst/qserv/admin/qservCli/launch.py
n8pease: it doesn't. thx.

n8pease:
`qserv entrypoint-help --help` may help your understanding; 

```$ qserv entrypoint-help --help
Usage: qserv entrypoint-help [OPTIONS] [COMMAND]

  Show the entrypoint CLI help output.

  COMMAND is the entrypoint subcommand to get help for. If not provided, shows
  help output for the entrypoint command.

  Shows help output for the entrypoint subcommand. If the subcommand spawns
  another process, shows help output from the app that the subcommand spawns.```

so invoking `qserv entrypoint-help replication-controller` shows you first output for `entrypoint replication-controller --help` (executed inside the run container) and then the help output for the spawned app, `qserv-replica-master-http`.

```$ qserv entrypoint-help
Help output for 'entrypoint':
...
Help for qserv-replica-master-http:
...
```

I suppose it would be more clear to change 
```Help output for 'entrypoint':``` 
to 
```Help for 'entrypoint replication-controller'``` 
or maybe just 
```entrypoint replication-controller --help```

What do you think, is it making more sense now?

File: src/admin/python/lsst/qserv/admin/cli/entrypoint.py (hunk 4 above)
Comment (n8pease) [reply src/admin/python/lsst/qserv/admin/cli/entrypoint.py:436]: `cmd_option` is simply a `click.option` with some vars predefined/bound (much like C++ `std::bind`).
But it's an interesting idea to move the use of `commands` into the implementation. I _can_ inherit from `click.Option`, I'll see if your idea can be implemented in a simple way.

**Comment by iagaponenko** [comment]
(After you posted your example) A specific command like the one shown below works:
```
% qserv entrypoint-help replication-controller
```
I ran into the confusing suggestion in the very end of the list:
```
qserv entrypoint-help
Help output for 'entrypoint':

Usage: entrypoint [OPTIONS] COMMAND [ARGS]...

Options:
  --log-level [CRITICAL|ERROR|WARNING|INFO|DEBUG]
                                  The logging level. Supported levels are
                                  [CRITICAL|ERROR|WARNING|INFO|DEBUG]
  --help                          Show this message and exit.

Commands:
  ...
  watcher                 Run a watcher algorithm that sends...
  spawned-app-help        Print the help output of a spanwed command for...
```
My impression was that I'm supposed to use that last command to get app-specific help like this:
```
% qserv entrypoint-help spawned-app-help replication-controller
% qserv entrypoint-help spawned-app-help qserv-replica-master-http
...
```
What's the purpose of that command?

n8pease:
primarily `spawned-app-help` is used by `qserv entrypoint-help` to do the part where it prints the spawned app help.
I considered hiding it from the `entrypoint --help` output, but it *could* be used by a user inside a running container to print the same help output (for the app that the entrypoint command spawns), without having to know the app name.

n8pease:
I was able to get the default for the `cmd_option` similar to how you suggested. Actually, now I don't even have to pass a string to `@cmd_option()` - it's able to do everything programmatically because it can look up the value using the command function name. It's in a separate commit "use an option to get the --cmd default".

iagaponenko:
@n8pease the new decorator looks good!

iagaponenko (approved):
LGTM

PR #681, opened 2022-01-28
DM-33350: add support for templated entrypoint command options

n8pease:
Resolves templates in all arguments in entrypoint.py, before
forwarding work on to script (implementation) fucntions.
This allows code in impl fucntions to be simpler, and
also fails early if a template can not be resolved.

### DIFF
--- a/admin/local/docker/compose/docker-compose.yml
+++ b/admin/local/docker/compose/docker-compose.yml
@@ -92,7 +92,7 @@ services:
           entrypoint worker-xrootd
           --db-uri mysql://qsmaster@127.0.0.1:3306
           --db-admin-uri mysql://root:CHANGEME@127.0.0.1:3306
-          --vnid-config "@/usr/local/lib64/libreplica.so mysql://qsmaster@127.0.0.1:3306/qservw_worker 0 0"
+          --vnid-config "@/usr/local/lib64/libreplica.so {{db_uri}}/qservw_worker 0 0"
           --cmsd-manager-name manager-xrootd
           --cmsd-manager-count 1
           --mysql-monitor-password CHANGEME_MONITOR
@@ -171,12 +171,13 @@ services:
         << : *worker-xrootd
         command: >
           entrypoint --log-level DEBUG worker-xrootd
-          --db-uri mysql://qsmaster@127.0.0.1:3306?socket=/qserv/mariadb/run/mysqld.sock
-          --db-admin-uri mysql://root:CHANGEME@127.0.0.1:3306?socket=/qserv/mariadb/run/mysqld.sock
+          --db-uri mysql://qsmaster@127.0.0.1:3306?socket={{db_socket}}
+          --db-admin-uri mysql://root:CHANGEME@127.0.0.1:3306?socket={{db_socket}}
           --vnid-config "@/usr/local/lib64/libreplica.so mysql://root:CHANGEME@localhost:3306/qservw_worker 0 0"
           --cmsd-manager-name manager-xrootd
           --cmsd-manager-count 1
           --mysql-monitor-password CHANGEME_MONITOR
+          --targs db_socket=/qserv/mariadb/run/mysqld.sock
         environment:
             << : *log-environment
         volumes:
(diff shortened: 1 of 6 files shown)


src/admin/python/lsst/qserv/admin/render.py
```diff
@@ -0,0 +1,77 @@
...
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+
```
iagaponenko, line 20: LSST Python Style guide at https://developer.lsst.io/python/style.html#docstrings-should-be-written-for-all-public-modules-functions-classes-and-methods requires that:

> Write docstrings for all public modules, functions, classes, and methods. 

This means that you may need to have a docstring right here to explain the purpose and use of this module (in addition to what you already have for each function/class/method downstream.

Otherwise, code readers would have to guess what this module with a very generic name `renderer` (BTW, why not call it `targs_proc`?)  is doing.

src/admin/tests/test_render.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def testNone(self):
+        """Test that a dict with None as a value resolves correctly."""
+        self.assertEqual(
+            render({"a": None}),
+            {"a": None}
+        )
+
+    def testBool(self):
+        """Test that a dict with None as a value resolves correctly."""
```
iagaponenko, line 75: Did you mean to say here?
```
"""Test that a dict with Bool as a value resolves correctly."""
```

iagaponenko (approved):
Looks good. Please, consider my comments.

admin/local/docker/compose/docker-compose.yml
```diff
@@ -171,12 +171,13 @@ services:
         << : *worker-xrootd
         command: >
           entrypoint --log-level DEBUG worker-xrootd
-          --db-uri mysql://qsmaster@127.0.0.1:3306?socket=/qserv/mariadb/run/mysqld.sock
-          --db-admin-uri mysql://root:CHANGEME@127.0.0.1:3306?socket=/qserv/mariadb/run/mysqld.sock
+          --db-uri mysql://qsmaster@127.0.0.1:3306?socket={{db_socket}}
+          --db-admin-uri mysql://root:CHANGEME@127.0.0.1:3306?socket={{db_socket}}
```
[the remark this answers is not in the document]
n8pease: --db-uri is used for more than one thing; entrypoint blocks until it can connect to the db at the uri. Later it's used in the czar cfg file. The latter can use the socket, but the former requires a host & port right now. I created DM-33455 to add socket support there.

src/admin/python/lsst/qserv/admin/cli/entrypoint.py
```diff
@@ -596,39 +559,23 @@ def worker_cmsd(
...
+        vnid_config=targs["vnid_config"],
+        repl_ctl_dn=targs["repl_ctl_dn"],
+        mysql_monitor_password=targs["mysql_monitor_password"],
+        db_qserv_user=targs["db_qserv_user"],
+        cmsd_worker_cfg_file=targs["cmsd_worker_cfg_file"],
+        cmsd_worker_cfg_path=targs["cmsd_worker_cfg_path"],
+        xrdssi_cfg_file=targs["xrdssi_cfg_file"],
+        xrdssi_cfg_path=targs["xrdssi_cfg_path"],
+        cmd=targs["cmd"],
     )
```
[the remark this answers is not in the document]
n8pease: It doesn't pass in _each_ element of the targs dict though, it passes in _some_ elements. It passes in those elements because they are in the signature of the called function. `targs` is passed because the function does template expansion on config files (it does say so in the documentation of `enter_worker_xrootd`), and all CLI args (which is where targs comes from) are to be available for template expansion.

src/admin/python/lsst/qserv/admin/render.py
n8pease: a clearer name is a good idea. I think I like `render_targs` tho (?)

iagaponenko: 👍

admin/local/docker/compose/docker-compose.yml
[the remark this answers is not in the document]
iagaponenko: I'm sorry, I might not be sufficiently specific in my prior comment. It wasn't about the role of that URI. It was about its format, which I believe should have one of these forms:
```
mysql://<db_user>:<db_pass>@<host>:<port>/<db_name>
mysql://<db_user>:<db_pass>@/<db_name>?unix_socket=<socket_path>
```
These are mutually exclusive. The first form (`host` and `port`) is used for connecting over the TCP protocol. The second one (`socket` file) - is for connecting over the local UNIX socket.

fritzm (approved):
LGTM, thanks!

PR #685, opened 2022-02-04
Move qserv_web into project

fritzm:
qserv_web is no longer a submodule; sources are included in this
repo directly in src/www

users with existing qserv repositories can update their repo by
running:

rm -rf extern/qserv_web
rm -rf .git/modules/extern/qserv_web


n8pease:
it may be helpful to add a README in `www` to explain its use in qserv. otherwise LGTM!

PR #686, opened 2022-02-06
DM-33088: qhttp logging

fritzm:
This `qhttp` PR: 

* adds logging (`lsst::log`)
* reverts library API to exceptions (logged with details at point of occurrence) instead of error codes
* follows RFCs 7230 and 3986 more closely with respect to allowed inputs, and rejects invalid requests with a 400 early rather than deferring to handlers

It may be easiest to review by commit.  To see examples of the new log messages, try:

`build/src/qhttp/testqhttp -- --data=src/qhttp/testdata/ --log-level=DEBUG`

### DIFF
--- a/src/qhttp/AjaxEndpoint.cc
+++ b/src/qhttp/AjaxEndpoint.cc
@@ -24,23 +24,31 @@
 #include "qhttp/AjaxEndpoint.h"
 
 // Local headers
+#include "lsst/log/Log.h"
+#include "qhttp/LogHelpers.h"
 #include "qhttp/Request.h"
 #include "qhttp/Response.h"
 #include "qhttp/Server.h"
+
+namespace {
+    LOG_LOGGER _log = LOG_GET("lsst.qserv.qhttp");
+}
 
 namespace lsst {
 namespace qserv {
 namespace qhttp {
 
 
-AjaxEndpoint::AjaxEndpoint()
+AjaxEndpoint::AjaxEndpoint(std::shared_ptr<Server> server)
+:
+    _server(server)
 {
 }
 
 
 AjaxEndpoint::Ptr AjaxEndpoint::add(Server& server, std::string const& path)
 {
-    auto aep = std::shared_ptr<AjaxEndpoint>(new AjaxEndpoint);
+    auto aep = std::shared_ptr<AjaxEndpoint>(new AjaxEndpoint(std::shared_ptr<Server>(&server)));
     server.addHandler("GET", path, [aep](Request::Ptr request, Response::Ptr response) {
         std::lock_guard<std::mutex> lock{aep->_pendingResponsesMutex};
         aep->_pendingResponses.push_back(response);
--- a/src/qhttp/AjaxEndpoint.cc
+++ b/src/qhttp/AjaxEndpoint.cc
@@ -20,6 +20,9 @@
  * see <https://www.lsstcorp.org/LegalNotices/>.
  */
 
+// System headers
+#include <utility>
+
 // Class-header
 #include "qhttp/AjaxEndpoint.h"
 
@@ -39,16 +42,16 @@ namespace qserv {
 namespace qhttp {
 
 
-AjaxEndpoint::AjaxEndpoint(std::shared_ptr<Server> server)
+AjaxEndpoint::AjaxEndpoint(std::shared_ptr<Server> const server)
 :
-    _server(server)
+    _server(std::move(server))
 {
 }
 
 
 AjaxEndpoint::Ptr AjaxEndpoint::add(Server& server, std::string const& path)
 {
-    auto aep = std::shared_ptr<AjaxEndpoint>(new AjaxEndpoint(std::shared_ptr<Server>(&server)));
+    auto const aep = std::shared_ptr<AjaxEndpoint>(new AjaxEndpoint(std::shared_ptr<Server>(&server)));
     server.addHandler("GET", path, [aep](Request::Ptr request, Response::Ptr response) {
         std::lock_guard<std::mutex> lock{aep->_pendingResponsesMutex};
         aep->_pendingResponses.push_back(response);
--- a/src/qhttp/AjaxEndpoint.cc
+++ b/src/qhttp/AjaxEndpoint.cc
@@ -52,6 +52,8 @@ AjaxEndpoint::Ptr AjaxEndpoint::add(Server& server, std::string const& path)
     server.addHandler("GET", path, [aep](Request::Ptr request, Response::Ptr response) {
         std::lock_guard<std::mutex> lock{aep->_pendingResponsesMutex};
         aep->_pendingResponses.push_back(response);
+        LOGLS_DEBUG(_log, logger(aep->_server) << logger(aep)
+            << "deferring response (" << aep->_pendingResponses.size() << " total)");
     });
     return aep;
 }
@@ -60,6 +62,8 @@ AjaxEndpoint::Ptr AjaxEndpoint::add(Server& server, std::string const& path)
 void AjaxEndpoint::update(std::string const& json)
 {
     std::lock_guard<std::mutex> lock(_pendingResponsesMutex);
+    LOGLS_DEBUG(_log, logger(_server) << logger(this)
+        << "sending " << _pendingResponses.size() << " deferred response(s)");
     for(auto& pendingResponse : _pendingResponses) {
         pendingResponse->headers["Cache-Control"] = "no-cache";
         pendingResponse->send(json, "application/json");
--- a/src/qhttp/AjaxEndpoint.h
+++ b/src/qhttp/AjaxEndpoint.h
@@ -58,9 +58,9 @@ public:
 
 private:
 
-    AjaxEndpoint(std::shared_ptr<Server> server);
+    AjaxEndpoint(std::shared_ptr<Server> const server);
 
-    std::shared_ptr<Server> _server;
+    std::shared_ptr<Server> const _server;
 
     std::vector<Response::Ptr> _pendingResponses;
     std::mutex _pendingResponsesMutex;
--- a/src/qhttp/AjaxEndpoint.h
+++ b/src/qhttp/AjaxEndpoint.h
@@ -58,7 +58,9 @@ public:
 
 private:
 
-    AjaxEndpoint();
+    AjaxEndpoint(std::shared_ptr<Server> server);
+
+    std::shared_ptr<Server> _server;
 
     std::vector<Response::Ptr> _pendingResponses;
     std::mutex _pendingResponsesMutex;
--- a/src/qhttp/LogHelpers.h
+++ b/src/qhttp/LogHelpers.h
@@ -1,0 +1,89 @@
+/*
+ * LSST Data Management System
+ * Copyright 2022 AURA/LSST.
+ *
+ * This product includes software developed by the
+ * LSST Project (http://www.lsst.org/).
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the LSST License Statement and
+ * the GNU General Public License along with this program.  If not,
+ * see <https://www.lsstcorp.org/LegalNotices/>.
+ */
+
+#ifndef LSST_QSERV_QHTTP_LOGHELPERS_H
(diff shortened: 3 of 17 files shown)


src/qhttp/AjaxEndpoint.cc
```diff
@@ -24,23 +24,31 @@
...
+:
+    _server(server)
 {
 }
 
 
 AjaxEndpoint::Ptr AjaxEndpoint::add(Server& server, std::string const& path)
 {
-    auto aep = std::shared_ptr<AjaxEndpoint>(new AjaxEndpoint);
+    auto aep = std::shared_ptr<AjaxEndpoint>(new AjaxEndpoint(std::shared_ptr<Server>(&server)));
```
iagaponenko, line 51: Although this is not strictly required, it's a good practice to add `const` to variables that are not going to be changed:
```
auto const aep = ...
```
Reasons: self-defense (prevents accidental modifications of the variable) and code documentation (tells a reader that the variable is read-only).

src/qhttp/AjaxEndpoint.h
```diff
@@ -58,7 +58,9 @@ class AjaxEndpoint
 
 private:
 
-    AjaxEndpoint();
+    AjaxEndpoint(std::shared_ptr<Server> server);
+
+    std::shared_ptr<Server> _server;
```
iagaponenko, line 63: Due to a design (static construction, no copy semantic) of the class this needs to be:
```
std::shared_ptr<Server> const _server;
```

src/qhttp/AjaxEndpoint.h
```diff
@@ -58,7 +58,9 @@ class AjaxEndpoint
 
 private:
 
-    AjaxEndpoint();
+    AjaxEndpoint(std::shared_ptr<Server> server);
```
iagaponenko, line 61: The parameter needs to be defined as:
```
std::shared_ptr<Server> const& server
```

src/qhttp/LogHelpers.h
```diff
@@ -0,0 +1,86 @@
...
+//
+//----- Helper ouput stream manipulators for logging.  These let you say things like:
+//      logger(self) << logger(sock) << "log message..." in logger macro calls
+//
+
+struct ServerLogger
+{
+    Server *server;
+    ServerLogger(Server *server) : server(server) {};
+    friend std::ostream& operator<<(std::ostream &str, ServerLogger logger) {
```
iagaponenko, line 51: Why `friend`? This is not needed because members of the `struct` are already public.

src/qhttp/LogHelpers.h
```diff
@@ -0,0 +1,86 @@
...
+};
+
+inline ServerLogger logger(Server* server) { return ServerLogger(server); }
+inline ServerLogger logger(Server::Ptr server) { return ServerLogger(server.get()); }
+
+struct SocketLogger
+{
+    boost::asio::detail::socket_type sockno;
+    SocketLogger(boost::asio::ip::tcp::socket* socket) : sockno(socket->native_handle()) {};
+    friend std::ostream& operator<<(std::ostream &str, SocketLogger logger) {
```
iagaponenko, line 63: `friend`?

src/qhttp/LogHelpers.h
```diff
@@ -0,0 +1,86 @@
...
+inline SocketLogger logger(std::shared_ptr<boost::asio::ip::tcp::socket> socket)
+{
+    return SocketLogger(socket.get());
+}
+
+struct AjaxLogger
+{
+    AjaxEndpoint *aep;
+    AjaxLogger(AjaxEndpoint *aep) : aep(aep) {};
+    friend std::ostream& operator<<(std::ostream &str, AjaxLogger logger) {
```
iagaponenko, line 77: `friend`?

src/qhttp/LogHelpers.h
fritzm: I had been following standard output manipulator pattern here.  In this case, though, since struct members are public, the operator specialization could be implemented without `friend` if moved out of the struct def'n and declared `inline`.  That is only slightly more verbose, but I can make that change here if you think it is stylistically preferable?

src/qhttp/AjaxEndpoint.cc
fritzm: Sensible; will do

src/qhttp/AjaxEndpoint.h
fritzm: Will do

src/qhttp/AjaxEndpoint.h
fritzm: Hmm... Passing `shared_ptr` by `const&` is perennially controversial (see https://stackoverflow.com/questions/3310737/should-we-pass-a-shared-ptr-by-reference-or-by-value and many others).  I have tended to favor the "take shared_ptr by value if you intend to store and share" guideline.  I will poll the group and we'll go with the tyranny of the majority :-)

src/qhttp/LogHelpers.h
iagaponenko: The `friend` is only needed for the operator to access protected or private members (data or methods) of the class. Hence the pattern you're referring to. In case if you have `struct` you simply move the operator definition out of the class as:
```
struct ServerLogger
{
    Server *server;
    ServerLogger(Server *server) : server(server) {};
};
std::ostream& operator<<(std::ostream &str, ServerLogger logger) { ... }
```
My second comment is that since `ServerLogger` is not a primitive data type then you want to pass it into the operator by `const&` reference:
```
std::ostream& operator<<(std::ostream &str, ServerLogger const& logger) ...
```

src/qhttp/Response.cc
```diff
@@ -182,13 +164,41 @@ std::string Response::_headers() const
...
 
     return headerst.str();
 }
 
+
+void Response::_write()
+{
+    if (_transmissionStarted.test_and_set()) {
+        LOGLS_WARN(_log, logger(_server) << logger(_socket)
+            << "handler logic error: multiple responses sent");
```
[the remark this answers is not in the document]
fritzm: This would occur only if a user of the `qhttp` library programmed their custom handler code incorrectly (e.g. calling a response send method multiple times, or throwing an exception after already calling a response `send` method.)  It should not ever occur in practice, but a `qhttp` user could always make mistakes when writing their handlers.

The atomic is here specifically to provide a safeguard against bad user code, and prevents `qhttp` from sending an invalid "double response."  The (new) warning means we can now see it in the logs if it ever _does_ happen.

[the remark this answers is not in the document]
fritzm: Perhaps this should be `LOGLS_ERROR` instead of `LOGLS_WARN`?  If it ever did occur in, say, qserv, it would indicate a programming error we'd want to notice and fix.

src/qhttp/LogHelpers.h
fritzm: Yes, exactly; see comment above.  If you would prefer to see operator defined outside the struct this way, I am happy to change this -- it's only an extra line or two per manipulator.

fritzm: acknowledged on `const&` here -- will do

src/qhttp/Response.cc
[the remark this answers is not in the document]
iagaponenko: Thank you for the explanation! I think it's a good idea to log this condition as `ERROR`.

src/qhttp/LogHelpers.h
iagaponenko: We've been told that `friend` is evil in C++, pretty much like `goto` 😄 That's why it caught my attention. Besides, there are always ways to avoid the "forced friendship", for example by implementing the public method `toString()` like in:
```
class MyClass {
public:
    std::string toString() const;
private:
    int _foo;
    std::string _bar;
};
std::ostream& operator<<(std::ostream &str, MyClass const& obj) {
    str << obj.toString();
    return str;
} 
```
A GOD think about this approach is that the string serialization logic is now a part of the class, not a separate piece of code that would have to be kept in SYNC sith the class should the class implementation evolve over time (new data members, removed data members, etc.).

iagaponenko: By the way, why do you need `ServerLogger`? You might as well overload the operator `<<` for the target class `Server`. In your implementation, the class `ServerLogger` is used solely to store a pointer to `Server` and to have the overloaded operator ... that in the end would go straight to the stored pointer. What are the benefits for this extra layer of indirection?

fritzm: The benefit is being able to say `log_ << logger(server) << "something happened with the server"` and have that result in a log-specific presentation of the server.   Also, `logger()` can be made to work on both regular pointers and smart pointers without having to worry which is which.  The output we want for the log wrt. one of these objects is not necessarily what we generally would want for output streaming (for example, has "srv=" for context, hex addr pointers, etc., logs identity and not value, etc. etc.)

src/qhttp/Response.cc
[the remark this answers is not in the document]
fritzm: will do

src/qhttp/LogHelpers.h
fritzm: In any case, I'll go with "the kids say `friend` is evil" -- not wanting to be all boomer about it, I will rewrite the operator overloads outside the struct def'n :-)

src/qhttp/Request.cc
```diff
@@ -29,65 +29,124 @@
...
+        R"(:[ \t]*)"
+        R"(([\x20-\x7e \t]*?))"
+        R"([ \t]*\r$)"
+    };
+
+    boost::smatch headerMatch;
+
+    while(getline(content, line)) {
+        if (!boost::regex_match(line, headerMatch, headerRe)) {
+            if (line == "\r") break;
```
iagaponenko, line 95: Why not check for this condition before matching the regular expression? My understanding of the protocol is that the empty line indicates the end of the header (and the header may not have any lines at all).

src/qhttp/Server.cc
```diff
@@ -303,8 +307,12 @@ void Server::_readRequest(std::shared_ptr<ip::tcp::socket> socket)
                                 << "request body read failed: " << ec.message());
                             return;
                         }
-                        if (request->header["Content-Type"] == "application/x-www-form-urlencoded") {
-                            request->_parseBody();
+                        if ((request->header["Content-Type"] == "application/x-www-form-urlencoded")
```
iagaponenko, line 310: In reality, all existing applications in Qserv expect a different content-type:
```
application/json
```
Please, see this class https://github.com/lsst/qserv/blob/main/src/replica/HttpRequestBody.cc
It would be natural if `qhttp` would provide the built-in support for that. Though, it would mean making `qhttp` dependent on `nlohmann::json`.

iagaponenko (approved):
Looks good to me.

src/qhttp/Server.cc
iagaponenko: Perhaps in a separate ticket ... if you will find this suggestion reasonable in the context of `qhttp`?

fritzm: I like that idea!  It's a very general capability, so makes sense in the context of qhttp.  And `nlohmann::json` is a pretty light dependency.  Let's do this in a separate ticket.

src/qhttp/LogHelpers.h
```diff
@@ -81,6 +82,31 @@ struct AjaxLogger
...
 inline AjaxLogger logger(AjaxEndpoint::Ptr aep) { return AjaxLogger(aep.get()); }
 
+//
+//----- Ouput stream manipulator that translates embedded control characters to "caret notation".  Used for
+//      logging suspect inputs that fail the legal input parsing regexps.
+//
+
+struct CtrlQuoter
+{
+    std::string toquote;
```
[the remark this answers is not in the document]
fritzm: will do

src/qhttp/AjaxEndpoint.h
fritzm: Per discussion in db team room, went with pass by value here and added `std::move` where ctor takes ownership of the `shared_ptr` (did same systematically through the rest of the `qhttp` code as well)

iagaponenko: 👍

src/qhttp/LogHelpers.h
```diff
@@ -81,6 +82,31 @@ struct AjaxLogger
...
+    CtrlQuoter(std::string const& toquote) : toquote(toquote) {}
+    friend std::ostream& operator<<(std::ostream &str, CtrlQuoter quoter) {
+        for(char const& c: quoter.toquote) {
```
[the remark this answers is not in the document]
iagaponenko: This algorithm won't work if you switch from `std::string toquote;` to `std::string const& toquote;` because it expects the mutable collection.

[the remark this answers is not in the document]
fritzm: I prefer `for` in this case for readability

src/qhttp/Request.cc
fritzm: `"\r"` line on its own, in the no-headers case, will still match the regexp.  However, I think your suggest is still good for code clarity, so will do.

PR #687, opened 2022-02-07
DM-33510: when tagging images use the release tag if it is newer than dockerfile changes

n8pease:
When creating lite qserv image tags, the description of the sha
that has the most recent changes to the relevant dockerfile(s)
is used to tag the image.
This can be confusing if the most recent changes are from a
previous release.
To make it more clear, if the most recent git tag (qserv release)
is newer then the most recent dockefile changes then use a
description of the SHA of the release tag instead.

### DIFF
--- a/src/admin/python/lsst/qserv/admin/qservCli/images.py
+++ b/src/admin/python/lsst/qserv/admin/qservCli/images.py
@@ -41,9 +41,10 @@ auth_header = "Bearer {token}"
 _log = logging.getLogger(__name__)
 
 
-def get_tag(dockerfiles: Optional[List[str]], cwd: str) -> str:
-    """Get the git tag of the commit that contains the most recent change to any
-    of the given dockerfiles.
+def get_description(dockerfiles: Optional[List[str]], cwd: str) -> str:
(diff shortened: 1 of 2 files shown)


src/admin/python/lsst/qserv/admin/qservCli/images.py
```diff
@@ -63,13 +64,52 @@ def get_tag(dockerfiles: Optional[List[str]], cwd: str) -> str:
...
+    res = subprocess.run(
+        "git describe --abbrev=0".split(),
+        stdout=subprocess.PIPE,
+        cwd=cwd,
+    )
+    tag = res.stdout.decode().strip()
+    if res.returncode != 0:
+        raise RuntimeError(f"Failed to get most recent tag from repo at {cwd}.")
+    res = subprocess.run(
+        ["git", "rev-list", "-n", "1", tag],
```
iagaponenko, line 104: I'm curious why you are not doing the same as before?
```
f"git rev-list -n 1 {tag}".split(),
```

iagaponenko (approved):
Looks good to me.

PR #688, opened 2022-02-12
DM-33665: targs can get stuck in a loop if a var combines text with itself

n8pease:
This contains a few small fixes to entrypoint args, the fix for the bug is in the commit "fail if targs contains a self-reference"


fritzm:
So... _must_ it be the case that any globally non-terminating expansion requires at least one variable to eventually loop through itself?  If so, then this is okay.

I suspect the answer is "yes", though I do not have a proof.  On the other hand, I haven't had success in constructing a counter-example...

Okay to go with this for now.  _Should_ we find the need to tighten this up in the future, an approach I'd find more convincingly bullet-proof would be to maintain a cumulative list for each variable, of variables expanded so far in its value.  If we ever find a reference to anything already in that list, we've hit a loop.

fritzm (approved):
Looks okay for now -- see caveat comment

PR #690, opened 2022-02-18
DM-33748: Use multiple jobs in CI Github action

fritzm:
Using multiple jobs in the GHA CI workflow (instead of just multiple steps in a single job) sets the stage for being able to put documentation and Qserv container builds downstream of a single shared base container build job, for parallelized jobs, for parallelized integration tests (compose vs. k8s+operator), etc.  

This will be used as a basis for upcoming changes to containerize the documentation builds and to integrate qserv operator into the qserv repo.

See https://github.com/lsst/qserv/actions/runs/1862847776 for example of new workflow on this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,10 +7,15 @@ on:
     branches:
       - main
 jobs:
-  build:
-    name: Build image
+
+  update-base-images:
+    name: Update base images
     runs-on: ubuntu-20.04
+    outputs:
+      build-rebuilt: ${{ steps.rebuild.outputs.build-rebuilt }}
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -32,45 +37,107 @@ jobs:
           username: ${{ secrets.FM_DOCKERHUB_TOKEN_USER }}
           password: ${{ secrets.FM_DOCKERHUB_TOKEN }}
 
-      - name: Prepare lite-build image
+      - name: Rebuild and push build image
```
n8pease, line 40: probably should say "if needed" like the mariadb job. Same for run base image on line 54.

n8pease:
LGTM!

PR #693, opened 2022-03-02
DM-33866: qhttp exception on invalid Content-Length header

fritzm:
Qualys scored another goal against `qhttp` by passing requests with non-numeric `Content-Type:` headers, triggering an unhandled parse exception ⚽

- Added a unit test for this, and exception handling around the vexed parse.

- Some other minor unit test cleanup while in there: moved nulls-in-path tests from static content suite to more appropriate exception-handling suite, and fixed a few unit tests which should have been using the general `curl` client instead of the special-purpose `asioHttpGet` fixture helper.


iagaponenko (approved):
Looks good

PR #694, opened 2022-03-05
DM-33618: move qserv_testdata contents to qserv repo

n8pease:


src/admin/python/lsst/qserv/admin/qservCli/launch.py
```diff
@@ -948,18 +955,18 @@ def itest(
...
     """
-    ref_db_container_name = itest_ref(qserv_root, itest_file, itest_volume, project, mariadb_image, dry)
+    itest_volumes = make_itest_volumes(project)
+    ref_db_container_name = itest_ref(qserv_root, itest_file, itest_volumes, project, mariadb_image, dry)
     try:
         returncode = integration_test(
             qserv_root,
             itest_container,
-            itest_volume,
+            itest_volumes.exe,
```
fritzm, line 971: .exe?

n8pease: yes, itest_volumes is a namedtuple, with 3 members, one of them is `exe`, this is the execution volume (the others are the mariadb data volume, and the mariadb lib volume)

PR #695, opened 2022-03-08
Tickets/dm 33346

jgates108:


src/wbase/SendChannelShared.h
```diff
@@ -47,8 +50,32 @@ class TransmitLock;
 namespace wbase {
 
 /// A class that provides a SendChannel object with synchronization so it can be
-/// shared by across multiple threads. Due to what may be sent, the synchronization locking
-/// is needs to be available outside of the class.
+/// shared by across multiple threads. Due to what may be sent, the synchronization
+/// locking needs to be available outside of the class.
```
iagaponenko, line 54: I have a question regarding "synchronization locking outside of the class" `wbase::SendChannelShared`. How does this ("external synchronization") work with the class that still has the private atomics, one private condition variable, and three private mutexes?
```
    std::mutex _streamMutex;
    std::mutex _transmitLockMtx;
    std::condition_variable _transmitLockCv;
    std::atomic<bool> _schemaColsSet{false};
    std::atomic<uint32_t> _scsSeq{0};
    std::atomic<bool> _firstChannelSqlConn{true};
    std::atomic<bool> _schemaColsSet{false};
    mutable std::mutex _tMtx;
    ...
```
I'm just concerned that debugging the run-time behavior of the class would be very hard.

I have an impression that considering many (8) synchronization primitives, the class represents "multiple faces" (roles, etc.). Is there any way to restructure the class into something safer?

src/wbase/SendChannelShared.cc
```diff
@@ -148,23 +150,27 @@ bool SendChannelShared::_addTransmit(bool cancelled, bool erred, bool last, bool
...
-        bool sendNow = tData->getScanInteractive();
-        // If there was an error, give this high priority.
-        if (erred || cancelled) sendNow = true;
-        int czarId = tData->getCzarId();
-        return _transmit(erred, sendNow, largeResult, czarId);
+    //&&&if (_lastRecvd || _transmitQueue.size() >= 2) {
+    bool sendNow = tData->getScanInteractive();
+    // If there was an error, give this high priority.
+    if (erred || cancelled) sendNow = true;
+    int czarId = tData->getCzarId();
```
iagaponenko: Probably:
```
CzarId const czarId = ... ;
```
or:
```
auto const czarId = ... ;
```

src/qdisp/QueryRequest.cc
```diff
@@ -329,7 +329,7 @@ bool QueryRequest::_importStream(JobQuery::Ptr const& jq) {
     int len = expectedLen;
     const char* buff = GetMetadata(len);
     if (len != expectedLen) {
-        throw Bug("_importStream wrong header size=" + to_string(len) + " expected=" + to_string(expectedLen));
+        throw Bug("_importStream metadata wrong header size=" + to_string(len) + " expected=" + to_string(expectedLen));
```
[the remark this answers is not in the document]
andy-slac: I would instead recommend using an `Issue` as a base class for all exceptions and `ERR_LOC` macro to avoid any explicit names or `string(__func__)`: https://github.com/lsst/qserv/blob/main/src/util/Issue.h

src/wbase/SendChannelShared.cc
```diff
@@ -148,23 +150,27 @@ bool SendChannelShared::_addTransmit(bool cancelled, bool erred, bool last, bool
...
     }
 
     // If this is reallyLast or at least 2 items are in the queue, the transmit can happen
-    if (_lastRecvd || _transmitQueue.size() >= 2) {
-        bool sendNow = tData->getScanInteractive();
-        // If there was an error, give this high priority.
-        if (erred || cancelled) sendNow = true;
-        int czarId = tData->getCzarId();
-        return _transmit(erred, sendNow, largeResult, czarId);
+    //&&&if (_lastRecvd || _transmitQueue.size() >= 2) {
```
iagaponenko: is there a point for keeping the commented code here? Why was it wrong? The comment on the previous line is still mentioning "at least 2 items".

src/qdisp/QueryRequest.cc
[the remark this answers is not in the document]
iagaponenko: Yes, this would be a better option.

src/wbase/SendChannelShared.cc
```diff
@@ -293,4 +273,153 @@ bool SendChannelShared::_sendBuf(lock_guard<mutex> const& streamLock,
...
+        // from the existing _transmitData object
+        lock_guard<mutex> lock(_tMtx);
+        if (_transmitData != nullptr) {
+            _transmitData->addSchemaCols(_schemaCols);
+        }
+    }
+}
+
+bool SendChannelShared::buildAndTransmitResult(MYSQL_RES* mResult, int numFields, Task& task, bool largeResult,
+        util::MultiError& multiErr, std::atomic<bool>& cancelled, bool &readRowsOk) {
```
iagaponenko, line 307: What's a reason for passing `cancelled` as a non-const reference? I'm not seeing anything body of this method as well as methods called from here that would modify a value of this atomic. The following should be safer:
```
std::atomic<bool> const& cancelled,
```
My second question in this context is - do you really need to pass into the method the atomic or just its value captured at a time of the call? I have noticed that the loop below is checking:
```
    bool more = true;
    while (more && !cancelled) {
        more = !_transmitData->fillRows(mResult, numFields, tSize);
        ...
```
Could the cancellation condition be changed  (externally) in between iterations of the loop?

src/wbase/SendChannelShared.h
jgates108: It looks like the synchronization outside of the class is no longer necessary, but I want to be very careful about removing it. _streamMutex is the only member that could be accessed externally.

src/wbase/SendChannelShared.h
```diff
@@ -155,10 +169,28 @@ class SendChannelShared {
...
+
+    /// @return a log worthy string describing _transmitData.
+    std::string dumpTr() const;
+
 private:
     /// Private constructor to protect shared pointer integrity.
     SendChannelShared(SendChannel::Ptr const& sendChannel,
-                      std::shared_ptr<wcontrol::TransmitMgr> const& transmitMgr);
+                      std::shared_ptr<wcontrol::TransmitMgr> const& transmitMgr,
+                      qmeta::CzarId const& czarId);
```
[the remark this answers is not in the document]
jgates108: It is currently a used defined as a uint32_t, but there isn't anything saying it wont be changed to a string in the future.

src/wbase/SendChannelShared.h
```diff
@@ -228,6 +284,7 @@ class SendChannelShared {
     std::mutex _transmitLockMtx;
     std::condition_variable _transmitLockCv;
     std::shared_ptr<wcontrol::TransmitMgr> _transmitMgr; ///< Pointer to the TransmitMgr
+    qmeta::CzarId _czarId; ///< id of the czar that requested this task(s).
```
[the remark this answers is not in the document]
jgates108: It wasn't set in the constructor until a fairly recent commit. Fixed.

src/wbase/SendChannelShared.h
jgates108: I've removed any need to externally lock _streamMutex. The other mutexes protect different things, which is essential to prevent deadlock and for performance reasons. It is probably possible to refactor the code so there's only one or two mutex per class, but its complicated enough that it should go in another ticket.

src/wbase/SendChannelShared.h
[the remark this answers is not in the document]
iagaponenko: Well, that's fine. However at line 164 of `SendChanneklShared.cc` there is something very different:
```
int czarId = tData->getCzarId();
```
which is not "unsigned", nether a string. I'd argue that the there should be consistency in using this information.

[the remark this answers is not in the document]
jgates108: There would be other ramifications if the type changed, mostly reguarding protobufs. I removed the const&.

src/wbase/SendChannelShared.cc
jgates108: This line has been removed.

iagaponenko (approved):
To the best of my understanding of the code, this looks good to me.

fritzm (approved):
Complicated in here!  Nothing stuck out at me beyond stuff that had already been mentioned addressed, so LGTM.  Thanks, John!

PR #696, opened 2022-03-08
Implemented ASYNC client for communicating with HTTP/REST services

iagaponenko:


src/replica/HttpAsyncClient.cc
```diff
@@ -0,0 +1,399 @@
...
+    switch (_state) {
+        case State::CREATED:
+            _finish(lock, State::CANCELLED);
+            return;
+        case State::IN_PROGRESS:
+            _finish(lock, State::CANCELLED);
+            return;
+        default:
+            // Nothing to be done in the final states of the request.
+            break;
```
andy-slac, line 151: So, this does not actually throw, contrary to what doxygen comment says.

src/replica/HttpAsyncClient.h
```diff
@@ -0,0 +1,312 @@
...
+    /**
+     * @brief Verify the desired state against the current one.
+     * 
+     * @param lock The lock to be held of _mtx.
+     * @param context The calling context.
+     * @param desiredStates The desired states (may be more than one) to be verified.
+     * @throw std::logic_error If the current state didn't match the desired one.
+     */
+    void _assertState(util::Lock const& lock, std::string const& context,
+                      std::set<State> const& desiredStates) const;
```
andy-slac, line 205: `std:set` is a heavy structure, maybe you want to use `std::initializer_list`?

src/replica/HttpAsyncClient.cc
```diff
@@ -0,0 +1,399 @@
...
+void HttpAsyncClient::cancel() {
+    string const context = "HttpAsyncClient::" + string(__func__) + " ";
+    util::Lock lock(_mtx, context);
+    switch (_state) {
+        case State::CREATED:
+            _finish(lock, State::CANCELLED);
+            return;
+        case State::IN_PROGRESS:
+            _finish(lock, State::CANCELLED);
+            return;
```
andy-slac, line 148: The code in both `case`s is identical, consider merging.

src/replica/HttpAsyncClient.h
```diff
@@ -0,0 +1,312 @@
...
+    /**
+     * @brief Parse the URL data member and populate _host, _port and _target.
+     * @note This method needs to be called just once.
+     * The syntax expected by the parser:
+     * @code
+     * http://<host>[:<port>][<target>]
+     * @code 
+     * @throw std::invalid_argument If the URL is empty or if it couldn't be parsed.
+     */
+    void _parsUrl(util::Lock const& lock);
```
andy-slac, line 216: _parsUrl -> _parseUrl?

src/replica/HttpAsyncClient.cc
```diff
@@ -0,0 +1,399 @@
...
+    string const context = "HttpAsyncClient::" + string(__func__) + " ";
+    util::Lock lock(_mtx, context);
+    _assertState(lock, context, {State::FINISHED});
+    return _res.get().body().size();
+}
+
+
+void HttpAsyncClient::_parsUrl(util::Lock const& lock) {
+    string const context = "HttpAsyncClient::" + string(__func__) + " ";
+    regex re("^[ ]*http://([^:^/]+)?(:([0-9]+))?(/([^ ]*))?[ ]*$", regex::extended);
```
andy-slac, line 198: This should be made static, no reason to compile regex on every call.

More importantly, your regex is really bad (sorry). Does not boost have a method for parsing URLs? No one should be using regexes for URL parsing, if you prefer I can show you how to call Python from C++ to do that 🙂 

Or at least write a unit test for this method with 100 or more different and legal URLs.

src/replica/HttpAsyncClient.cc
```diff
@@ -0,0 +1,399 @@
...
+    smatch match;
+    if (!regex_search(_url, match, re)) {
+        throw invalid_argument(context + "incorrect syntax of the URL '" + _url + "'.");
+    }
+    if (match.size() != 6) {
+        throw invalid_argument(context + "problem with the regular expression when parsing URL '" + _url + "'.");
```
andy-slac, line 204: This sort of message is not super-helpful, saying "you have a problem" does not really identify what the problem is.

src/replica/HttpAsyncClient.cc
```diff
@@ -0,0 +1,399 @@
...
+    }
+    auto const parseVal = [](string const& in, string& out) {
+        if (!in.empty()) out = in;
+    };
```
andy-slac, line 208: This is not the lambda that I can approve, even if it's short.

src/replica/HttpAsyncClient.cc
```diff
@@ -0,0 +1,399 @@
...
+    parseVal(match[1].str(), _host);
+    parseVal(match[3].str(), _port);
+    parseVal(match[4].str(), _target);
+}
+
+
+void HttpAsyncClient::_prepareRequest(util::Lock const& lock) {
+    _req.version(11);   // HTTP/1.1
```
andy-slac, line 216: Grrr. You should definitely switch to Python.

src/replica/HttpAsyncClientApp.cc
```diff
@@ -0,0 +1,204 @@
...
+    ).option(
+        "file",
+        "A path to an output file where the response body received from a remote source will"
+        "  be written. This option is ignored if the flag --body is not specified.",
+        _file
+    ).flag(
+        "result2json",
+        "If specified the flag will cause the application to interpret the response body as"
+        " a JSON object.",
+        _result2json
```
andy-slac, line 114: Is this actually used?

src/replica/HttpAsyncClientApp.cc
```diff
@@ -0,0 +1,204 @@
...
+    );
+}
+
+
+int HttpAsyncClientApp::runImpl() {
+    string const context = "HttpAsyncClientApp::" + string(__func__) + "  ";
+    const char* allowedMethods[] = {"GET", "POST", "PUT", "DELETE"};
+    if (count(cbegin(allowedMethods), cend(allowedMethods), _method) == 0) {
+        throw invalid_argument(context + "unknown HTTP method: " + _method);
+    }
```
andy-slac, line 135: Command line parser should do this check.

src/replica/HttpAsyncClientApp.cc
```diff
@@ -0,0 +1,204 @@
...
+void HttpAsyncClientApp::_dump(shared_ptr<HttpAsyncClient> const& ptr, ostream* osPtr) const {
+    if (_verbose) {
+        cout << "Request completion state: " << HttpAsyncClient::state2str(ptr->state())
+                << ", error: " << ptr->error() << endl;
+    }
+    set<HttpAsyncClient::State> const headerAvailableStates = {
+        HttpAsyncClient::State::FINISHED,
+        HttpAsyncClient::State::BODY_LIMIT_ERROR
+    };
+    if (headerAvailableStates.count(ptr->state()) != 0) {
```
andy-slac, line 185: This is a very convoluted and inefficient way to write

    if (ptr->state() == HttpAsyncClient::State::FINISHED or ptr->state() == HttpAsyncClient::State::BODY_LIMIT_ERROR) {

andy-slac (approved):
I'm going to approve this, but I have to say that I have no confidence in the code that I saw. The logic in the async state machine looks unverifiable, and there is no unit test for it (and it's probably impossible to write a unit test for it). I would hate to debug this code when it breaks.

src/replica/HttpAsyncClient.h
```diff
@@ -0,0 +1,312 @@
...
+        IN_PROGRESS,        ///< The request is still in progress
+        FINISHED,           ///< Final state: the operation succeeded, result is available 
+        FAILED,             ///< Final state: the operation failed
+        BODY_LIMIT_ERROR,   ///< Final state: the operation failed because the response's body is larger than requested
+        CANCELLED,          ///< Final state: the operation was explicitly cancelled after it was started
+        EXPIRED             ///< Final state: the operation was aborted due to the timeout expiration
+    };
+
+    /// @return The string representation of the state.
+    /// @throw std::invalid_argument For unknown values of the input parameter.
```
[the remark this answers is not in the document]
iagaponenko: This often happens (I have seen this many times) if there is silent memory corruption. Tracking the out-of-bound input values is very helpful in the early detection of problems.

src/replica/HttpAsyncClient.h
```diff
@@ -0,0 +1,312 @@
...
+     * @brief Static factory for creating objects of the class.
+     *
+     * Objects created by the method will be in the state State::CREATED. The actual
+     * execution of the request has to be initiated by calling the method start().
+     *
+     * @param io_service BOOST ASIO services.
+     * @param onFinish The callback function to be called upon a completion
+     *   (successfull or not) of a request.
+     * @param method The case-sensitive name of the HTTP method ('GET', 'POST', 'PUT', 'DELETE').
+     * @param url A location of the remote resoure.
```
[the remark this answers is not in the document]
iagaponenko: I need to parse URL by myself since we don't have a ready-to-use parser available in the current development container. There is an excellent Boost library that I would like to use https://master.url.cpp.al/. Though, it's not a part of the official Boost releases. The library is relatively new, and it requires some recent version of Boost that we don't have here. Ours is 1.66 (that was released circa 2017). The very same library could be used for parsing database connection strings. @fritzm is aware of the library.

src/replica/HttpAsyncClient.cc
iagaponenko: What's wrong with this lambda? Look at the next 3 lines of code where it's used. A value of the `in` parameter is extracted from the result of matching a regular expression. Using the lambda here simplifies the value analysis and prevets dereferencing the input value from the above mentioned result twice.

src/replica/HttpAsyncClient.cc
iagaponenko: Well, `11` is what the `Boost.Beast` expects from me. I wish we had (in C++) a good library that would be as convenient as Python's `requests` (AKA "HTTP for humans"). In theory, `Boost.Beast` provides a good fundation for building such library. However, this was not my intent here as developing such universally useful high-level library would take a much big effort.

src/replica/HttpAsyncClientApp.cc
iagaponenko: It was _meant_ to be used. I missed implementing a small piece of code that would do it. Thank you for noticing!

src/replica/HttpAsyncClientApp.cc
iagaponenko: You're 100% right. The command-line parser supports this. I will fix this.

src/replica/HttpAsyncClientApp.cc
iagaponenko: May be.

src/replica/HttpAsyncClient.h
```diff
@@ -0,0 +1,312 @@
...
+
+/**
+ * @brief Class HttpAsyncClient represents a simple asynchronous interface for
+ * communicating over the HTTP protocol.
+ * 
+ * The implementation of the class invokes a user-supplied callback (lambda) function
+ * upon a completion or a failure of the request.
+ * 
+ * Here is an example of using the class to pull a file and dump its content on
+ * to the standard output stream:
```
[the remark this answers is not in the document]
iagaponenko: Good point. I will fix this.

src/replica/HttpAsyncClient.h
```diff
@@ -0,0 +1,312 @@
...
+ *   will be failing, or if attempts to send a request will be failing. A short delay
+ *   will be introduced after each unsuccessfull attempt. The request will go into
+ *   the FAILED state if 1) there will be a failure in receiving server's response,
+ *   2) in case of incorrect data received from the server, or 3) the result is too
+ *   large.
+ *
+ * @note The implementation will open and close a new connection for each request.
+ * @note The implementation doesn't support TLS/SSL-based HTTPS protocol.
+ */
+class HttpAsyncClient: public std::enable_shared_from_this<HttpAsyncClient> {
```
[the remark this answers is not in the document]
iagaponenko: This is fine with me. I will rename the class.

src/replica/HttpAsyncClient.h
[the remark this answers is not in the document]
iagaponenko: Upon furthe rinspection of the code, I have realized that I already have the class `replica::Url` that could be extended to suit needs of the async client. Class `replica::Url` alredy has a unit test. I will update the code to use `replica::Url` before we have a better solution, like the previously mentioned `Boost.Url`.

src/replica/HttpAsyncClient.h
```diff
@@ -0,0 +1,312 @@
...
+
+    /// @return The current state of the request.
+    State state() const { return _state; }
+
+    /**
+     * @brief Begin processing a request.
+     *
+     * Upon successful completion of the method the object's state will be set
+     * to State::IN_PROGRESS. Otherwise the object's state will be set to
+     * State::FAILED and an exception will be thrown.
```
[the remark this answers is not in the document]
iagaponenko: Let me explain the logic here. If a problem occurs in the constructor of the class (or its factory method `create`) then no object would be created. Once created, the object goes into the initial state `CREATED`. Both scenarios have no ambiguities. Things are getting a bit more complicated when the method `start` is attempting to initiate the request by setting up parameters of the request, initiating the expiration timer, and initiating resolve on the host's host/port. A failure to initiate the request will not automatically destroy the object (and all shared pointers to it). An exception is thrown to indicate a problem. At this point, it's not quite clear what should be done next and how to treat the object. The first possibility would be to leave the request in the state `CREATED` and assume that another call to the method `start` would "magically" resolve the issue. That's very unlikely because there is no way to "repair" the parameters of the object once it's constructed. Besides, it wouldn't be clear what exactly needs to be rapaired. The second approach that I chose here is to turn the object into the final state `FAILED` ... which also leaves the request object in an unusable state. On the surface, both are nearly the same. The only difference between those is that in the last approach the failed attempt to use the request object is now recorded in the object's state. Therefore, if this object was a part of some collection of requests, or if it was passed to some other context then the other context would be able to determine the new state of the object and take appropriate actions.

src/replica/HttpAsyncClient.h
```diff
@@ -0,0 +1,312 @@
...
+     * 
+     * @throw std::logic_error If the operation already started or finished.
+     */
+    void start();
+
+    /**
+     * @brief Cancel the ongoing request.
+     * @throw std::logic_error If the operation has already finished (successfully or not).
```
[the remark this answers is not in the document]
iagaponenko: Agreed

src/replica/HttpAsyncClient.h
```diff
@@ -0,0 +1,312 @@
...
+     */
+    void cancel();
+
+    /// @return The last error in case of a failure.
```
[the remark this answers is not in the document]
iagaponenko: The later. The HTTP code is only available once the response header was parsed. And this is only in the final states `FINISHED` or `BODY_LIMIT_ERROR`. In those cases the `HTTP` could be obtained by calling the method:
```
    /// @return The completion code (HTTP).
    /// @throw std::logic_error If the operation didn't start or it was still in progress.
    /// @note The header is available in states State::FINISHED and State::BODY_LIMIT_ERROR.
    int responseCode() const;
```

src/replica/HttpAsyncClient.h
```diff
@@ -0,0 +1,312 @@
...
+class HttpAsyncClient: public std::enable_shared_from_this<HttpAsyncClient> {
+public:
+    /// The function type for notifications on the completion of the operation.
+    typedef std::function<void(std::shared_ptr<HttpAsyncClient> const&)> CallbackType;
+
+    enum class State: int {
+        CREATED = 0,        ///< The object was created and no request was sent
+        IN_PROGRESS,        ///< The request is still in progress
+        FINISHED,           ///< Final state: the operation succeeded, result is available 
+        FAILED,             ///< Final state: the operation failed
```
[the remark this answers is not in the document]
iagaponenko: The former. The request gets into the `FINISHED` state if both the response header and the body were received and parsed. It's up to a caller to inspect the response code by calling:
```
    /// @return The completion code (HTTP).
    /// @throw std::logic_error If the operation didn't start or it was still in progress.
    /// @note The header is available in states State::FINISHED and State::BODY_LIMIT_ERROR.
    int responseCode() const;
```
I should perhaps make this more clear in the documentation for the class.

src/replica/HttpAsyncClient.h
```diff
@@ -0,0 +1,312 @@
...
+     */
+    void cancel();
+
+    /// @return The last error in case of a failure.
+    /// @note The method should be used in one of the final states of the request.
+    std::string error() const;
+
+    /// @return The completion code (HTTP).
+    /// @throw std::logic_error If the operation didn't start or it was still in progress.
+    /// @note The header is available in states State::FINISHED and State::BODY_LIMIT_ERROR.
```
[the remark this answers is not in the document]
iagaponenko: My bad.

src/replica/HttpAsyncClient.h
```diff
@@ -0,0 +1,312 @@
...
+    int responseCode() const;
+
+    /// @return The content of response header represented by the key/value map.
+    /// @throw std::logic_error If the operation didn't start or it was still in progress.
+    /// @note The header is available in states State::FINISHED and State::BODY_LIMIT_ERROR.
+    std::map<std::string, std::string> const& responseHeader() const;
```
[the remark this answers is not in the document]
iagaponenko: Very good suggestion! I missed to address this complication.

src/replica/HttpAsyncClient.cc
```diff
@@ -0,0 +1,399 @@
...
+        case State::IN_PROGRESS: return "IN_PROGRESS";
+        case State::FINISHED: return "FINISHED";
+        case State::FAILED: return "FAILED";
+        case State::BODY_LIMIT_ERROR: return "BODY_LIMIT_ERROR";
+        case State::CANCELLED: return "CANCELLED";
+        case State::EXPIRED: return "EXPIRED";
+    }
+    throw invalid_argument(
+            "HttpAsyncClient::" + string(__func__) + " unknown state: "
+            + to_string(static_cast<int>(state)) + ".");
```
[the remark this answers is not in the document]
iagaponenko: I have already explained the motivation for catching the out-of-band values in another comment posted before.

src/replica/HttpAsyncClient.cc
```diff
@@ -0,0 +1,399 @@
...
+
+
+void HttpAsyncClient::start() {
+    string const context = "HttpAsyncClient::" + string(__func__) + " ";
+    util::Lock lock(_mtx, context);
+    _assertState(lock, context, {State::CREATED});
+    try {
+        _parsUrl(lock);
+        _prepareRequest(lock);
+        _resolve(lock);
```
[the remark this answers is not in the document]
iagaponenko: I agree regarding the first two methods. It does make sense to move both to the c-tor. The third method should be initiated here because once created, the request object might be stored in some queue and be initiated at the right time (once it's needed). That was the conscious decision to separate object construction from any network operations. Hence its has two states in its FSA:
```
CREATE -> IN_PROGRESS -> 
```

src/replica/HttpAsyncClient.cc
```diff
@@ -0,0 +1,399 @@
...
+        if (_expirationIvalSec) {
+            _expirationTimer.expires_from_now(boost::posix_time::seconds(_expirationIvalSec));
+            _expirationTimer.async_wait(bind(&HttpAsyncClient::_expired, shared_from_this(), _1));
```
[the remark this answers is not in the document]
iagaponenko: You're next to @fritzm who doesn't like `bind`. I still don't get why. What's wrong with those? Is this just your personal opinion, or advice from the C++ "gurus" (Alexandersku, Sutter, Meyer, etc.)? In any case, I would be curious to see the very source of this opinion. Then, perhaps, I will reconsider my "obsession" with `bind` (I use it quite a bit in other classes).

src/replica/HttpAsyncClient.cc
iagaponenko: Good point.

src/replica/HttpAsyncClient.cc
iagaponenko: This question was addressed in response to the relevant comment.

src/replica/HttpAsyncClient.cc
iagaponenko: Noted. I will improve the message.

iagaponenko:
@andy-slac regarding your general comment on the unit test, I think it's possible to implement the one here. I could launch the in-process `qhttp` on some free port and use this server for testing the new class.

src/replica/HttpAsyncClient.h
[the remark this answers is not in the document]
andy-slac: That's an odd argument. If you have a memory corruption I think you cannot even reason about what your program does.

src/replica/HttpAsyncClient.h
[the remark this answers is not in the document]
andy-slac: This is fine, my question was mostly that it's not clear what can cause this method to fail, as I thought that most checks and hostname/port parsing/resolution should be done in constructor.

src/replica/HttpAsyncClient.h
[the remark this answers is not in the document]
iagaponenko: At least, I will see the problem earlier. This helped in the past. The out-of-bound enumerator values retrieved from the data members of some class usually indicate that an object of that class got destroyed, corrupt, or never created. This is never a problem in Python because Python doesn't know anything about such things as "threads" or "event-driven" programming models 😄 
What I could do here is to remove the notion of the exception from the class contract while still leaving this extra check in the body of the method. This may prevent confusion.

src/replica/HttpAsyncClient.h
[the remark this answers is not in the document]
andy-slac: Maybe change "error" to "error message"?

src/replica/HttpAsyncClient.cc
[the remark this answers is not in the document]
andy-slac: Frankly, that comment does not make sense.

src/replica/HttpAsyncClient.h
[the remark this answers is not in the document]
iagaponenko: I will fix the documentation on the method along with pushing parameter validation to the c-tor.

src/replica/HttpAsyncClient.h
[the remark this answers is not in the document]
iagaponenko: Good suggestion!

src/replica/HttpAsyncClient.cc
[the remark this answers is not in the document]
andy-slac: We all C++ gurus, no need to call other authorities 🙂  Bind sucks because it makes code unreadable and its implementation is just terrible.

src/replica/HttpAsyncClient.cc
andy-slac: Couple of points:
- I don't see this as an improvement, it makes code longer but not clearer
- the name is confusing, `parseVal` does not parse anything, just assigns something to something
- if you care about performance you should probably write something like `if (match[1].matched) _host = match[1];` to avoid creating strings to just check that they are empty.

src/replica/HttpAsyncClient.cc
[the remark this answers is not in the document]
fritzm: Yup, strongly agree w/ Andy on this one, per previous ~arguments~ "discussions".  I don't understand why you embrace modern C++ in so many other ways, but still cling so stubbornly to `bind`?

src/replica/HttpAsyncClient.cc
fritzm: Doxygen tag-line: "ugly, useless, and frequently a pack of lies"

fritzm (changes requested):
Regarding the unit tests:

I still don't see a real need for the AsyncTimer class here, and consider it a dangerous addition to our codebase because it contains an asio anti-pattern (proliferation of io_service instances in class-private threads) and has a high potential for misuse.

At the very least, before merging I would want to see the timer class moved into the unit test cpp/h anonymous namespace instead of separate files.

Much better in my opinion would be to just use a single `io_service` in each test (shared for server, client, and timeout timer) and use the existing asio timer API out of the box, without any need for an additional wrapper class at all.  We did discuss some of this previously, but after seeing the tests in their final form I'm less convinced than previously that the new AsyncTimer class is well-justified.

src/replica/testHttpAsyncReq.cc
```diff
@@ -0,0 +1,546 @@
...
+
+    LOGS_INFO("HttpAsyncReq_simple");
+
+    asio::io_service io_service;
+
+    // The deadline timer limits the duration of the test to prevent the test from
+    // being stuck for longer than expected.
+    ::AsyncTimer testAbortTimer(io_service, 10000, [] (auto expirationIvalMs) {
+        LOGS_INFO("HttpAsyncReq_simple: test exceeded the time budget of " << expirationIvalMs << " ms");
+        std::exit(1);
```
fritzm, line 202: Do you really want `exit()` here?  If one test fails, shouldn't the others still be attempted?

src/replica/testHttpAsyncReq.cc
```diff
@@ -0,0 +1,546 @@
...
+        LOGS_INFO("HttpAsyncReq_expired: test exceeded the time budget of " << expirationIvalMs << " ms");
+        std::exit(1);
+    });
+    testAbortTimer.start();
+
+    // Set up and start the server
+    ::HttpServer httpServer;
+    httpServer.server()->addHandler("POST", "/delayed_response",
+        [] (qhttp::Request::Ptr const& req, qhttp::Response::Ptr const& resp) {
+            this_thread::sleep_for(chrono::milliseconds(2500));
```
fritzm, line 336: This `sleep` is probably jamming the server (which has one thread) from processing further requests (and if you had run the entire test on a single thread would jam the entire test).  Since you are in an asio thread here, instead of sleep, it would probably be better to use an async timer (capture the response pointer in the completion lambda)

src/replica/testHttpAsyncReq.cc
```diff
@@ -0,0 +1,546 @@
...
+        LOGS_INFO("HttpAsyncReq_delayed_server_start: test exceeded the time budget of " << expirationIvalMs << " ms");
+        std::exit(1);
+    });
+    testAbortTimer.start();
+
+    // Set up the server on the fixed port. The server start will be delayed by the timer.
+    // Since we don't have any guarantee that the port is not in use by some other process,
+    // and should this port be not available the timer will cancel the request to allow
+    // this test to pass.
+    uint16_t const fixedPort = 8080;
```
fritzm, line 492: Fixed port is not great, but probably okay.  You might want to pick something less commonly used than 8080?

An alternative might be to create a new asio socket here and `bind()` it to an address with port 0.  This will allocate an unused port.  `qhttp` uses SO_REUSEADDR option already on its listen port, so you could leave the socket open and just pass that newly allocated port to the `qhttp::Server` constructor, and I think it would work and avoid the potential collision problem?

fritzm (approved):
Okay, thanks for changes per previous discussion.  A few additional comments left for your consideration.

src/replica/testHttpAsyncReq.cc
iagaponenko: I too thought about this. I will see how to improve this.

src/replica/testHttpAsyncReq.cc
iagaponenko: Yes, I'm aware of this risk. Fortunatelly, this is not a problem here because `qhttp` has its own instance of `io_service` and a dedicated thread. Besides, each of the test cases initiates no more than one request.

src/replica/testHttpAsyncReq.cc
iagaponenko: I will pick some other (large number) port.

As far as your `bind` idea is concerned, I have already explored a kind of similar track by starting and stopping `qhttp` using this series of calls:
```
auto ptr = qhttp::create(io_service, 0);
ptr->start();
ptr->stop();
...
// Then delayed start in the ASYNC timer callback hoping the server would get the same port
{
  ptr->start()
}
```
This all ended up with some really bad crash in ... `XRootD` stack 😸 

I could also modify this algorithm by dropping the first instance of `qhttp` and replacing it with another one. Would give me the same result as your `bind` proposal?

PR #698, opened 2022-04-05
Fixed a bug that was causing subseconds to be truncated in the timest…

iagaponenko:
…amps

### DIFF
--- a/src/mysql/SchemaFactory.cc
+++ b/src/mysql/SchemaFactory.cc
@@ -60,12 +60,12 @@ struct ColTypeFactory {
           case MYSQL_TYPE_DOUBLE: sqlType = "DOUBLE"; break; // n,m
           case MYSQL_TYPE_NULL: sqlType = "NULL"; break;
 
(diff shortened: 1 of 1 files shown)


src/mysql/SchemaFactory.cc
```diff
@@ -107,6 +107,16 @@ struct ColTypeFactory {
         sqlType = os.str();
     }
 
+    void _setDateTime(char const* baseType, int length) {
+        if (length == 0) {
+            sqlType = baseType;
+        } else {
+            std::ostringstream os;
+            os << baseType << "(" << length << ")";
+            sqlType = os.str();
```
andy-slac, line 116: C++ streams are rather expensive, you could probably make it more efficient with the code like:

    sqlType = baseType;
    sqlType += '(';
    sqlType += std::to_string(length);
    sqlType += ')';

andy-slac (approved):
Looks good, one minor comment.

src/mysql/SchemaFactory.cc
iagaponenko: Fixed

PR #700, opened 2022-04-12
Fix incorrect comparison operator for the worker descriptors

iagaponenko:


src/replica/ConfigWorker.h
```diff
@@ -135,7 +135,7 @@ class WorkerInfo {
     bool operator ==(WorkerInfo const& other) const;
 
     /// @return 'true' if workers objects don't have the same values of attributes
-    bool operator !=(WorkerInfo const& other) const { return operator ==(other); }
+    bool operator !=(WorkerInfo const& other) const { return operator !=(other); }
```
andy-slac, line 138: Should it be `! operator==(other)`? Or do you like an infinite recursion?

andy-slac (approved):
Looks good.

src/replica/ConfigWorker.h
andy-slac: too much != not too much

andy-slac: And I also mentioned unit test in our Slack chat...

PR #702, opened 2022-04-13
DM-32709: rework Qserv doc infrastructure for the lite era

fritzm:
This PR freshens Qserv documentation infrastructure as a starting point for subsequent cleanup/update pass.  Things done here:

- Make HTML documentation buildable in the lite build container (`make docs-html` results in a built and browsable html doc tree at `build/doc/html`)
- Add a `build-docs` subcommand to the `qserv` CLI for developer convenience; this invokes the make command above in a one-shot build container
- Update CI github action to build and publish (via `ltd-conveyor`) documentation at `qserv.lsst.io`
- Reorganize existing documentation into top-level hierarchy per discussions at DAX cowork, and remove/relocate a lot of detritus from the doc directory
- Move to read-the-docs doc theme instead of the former astropy-inspired one

This PR does _not_ attempt to refresh/rewrite/revise/extend the existing (very oudated!) documentation.  That is to follow on separate individual tickets.


doc/conf.py
```diff
@@ -19,37 +20,41 @@
...
-project = "Qserv"
-
-# TODO
-copyright = ""
-
-author = "Rubin Observatory"
+# General information about the project.
+project = u'Qserv'
+copyright = u'2016-2022 Association of Universities for Research in Astronomy, Inc. (AURA)'
+author = u'LSST Data Management'
```
n8pease, line 30: AFAIK python3 strings are unicode by default, the leading `u` is left in the language for portability with python2. It probably should be removed here.

doc/conf.py
```diff
@@ -19,37 +20,41 @@
...
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
 # built documents.
-# TODO either set these manually here, or extract from e.g. env variables.
-version = "X.Y.Z"
-release = "X.Y.Z"
+
+# The short X.Y version.
+github_ref = os.getenv("GITHUB_REF", os.popen('git symbolic-ref HEAD').read().strip())
```
n8pease, line 37: two things:
one is that os.popen is going to run every time, which you don't really need. The other is that I think you need to close the handle returned by popen. So:
```
github_ref = os.getenv("GITHUB_REF")
if github_ref is None:
    with contextlib.closing(os.popen('git symbolic-ref HEAD')) as p:
        github_ref = p.read().strip()
```

.github/workflows/ci.yml
```diff
@@ -108,6 +108,54 @@ jobs:
...
+                - '**'
+
+      - name: Prepare user build image
+        if: needs.update-base-images.outputs.build-rebuilt == 'True' || steps.filter.outputs.docs == 'true'
+        run: |
+          ./admin/local/cli/qserv --log-level DEBUG build-user-build-image \
+          --group docker_outer
+
+      - name: Build and publish
+        if: needs.update-base-images.outputs.build-rebuilt == 'True' || steps.filter.outputs.docs == 'true'
```
n8pease, line 148: only build docs if the build image was changed? That seems like a strange trigger.

fritzm: There's an "or" ("||") on that condition -- rebuild docs if base image (e.g. sphinx tools) changed, OR any files in the the docs filter have changed.

doc/conf.py
fritzm: will do

doc/conf.py
fritzm: will do

PR #705, opened 2022-05-02
DM-34624: Implement optimization for simple SELECT COUNT(*) in qserv

n8pease:


src/admin/python/lsst/qserv/admin/replicationInterface.py
```diff
@@ -370,6 +369,25 @@ def ingest_data_file(
...
+                data=json.dumps(
+                    dict(
+                        database=database,
+                        table=table,
+                        row_counters_deploy_at_qserv=1,
+                        row_counters_state_update_policy="ENABLED",
+                        force_rescan=1,
+                        auth_key="",
+                        admin_auth_key="",
+                        instance_id="qserv_proj",
```
iagaponenko, line 386: Did you add this method `build_table_stats` for the integration test only? In theory, values of the last three properties (`auth_key`, `admin_auth_key`, and `instance_id`) should not be hard-coded because they depend on a specific setup (of Qserv and its Replication/Ingest system).

src/admin/python/lsst/qserv/admin/cli/entrypoint.py
```diff
@@ -441,6 +442,7 @@ def proxy(ctx: click.Context, **kwargs: Any) -> None:
         czar_cfg_file=targs["czar_cfg_file"],
         czar_cfg_path=targs["czar_cfg_path"],
         cmd=targs["cmd"],
+        log_cfg_file=targs["log_cfg_file"]
```
iagaponenko, line 445: (Perhaps) missing `,` in the end of the line?

src/proxy/mysqlProxy.lua
```diff
@@ -197,7 +197,6 @@ function queryType()
     -- Detects if query can be handled locally without sending it to qserv
     local isLocal = function(qU)
         if (string.find(qU, "^SHOW ") and not string.find(qU, "^SHOW .*PROCESSLIST")) or
-           string.find(qU, "^SET ") or
```
andy-slac, line 200: So, this forwards all "SET" queries to qserv instead of mysql? I'm a bit worried about that, there _may be_ some mysql options that people may want to change with the SET queries, and stuff like JDBC could also issue SET queries to change some session-level behavior. Maybe it's better to still forward all SET queries to mysql but send qserv-specific options to qserv. Probably similarly to the line above change this one to:

    (string.find(qU, "^SET ") and not string.find(qU, "^SET .*QSERV")) or

andy-slac (commented):
One suggestion for proxy LUA code.

src/admin/python/lsst/qserv/admin/replicationInterface.py
n8pease: right now this interface is only used by itest. It could in theory be more generally useful, though auth_key="" and admin_auth_key="" are used pretty generally in this file, though I could make them arguments to the itest entrypoint command. Where are the values for these set in the RI system?

qserv_proj is passed as an argument to the replication system executables via extended args to entrypoint commands in the docker-compose file, and so that should definitely be an argument to the itest entrypoint command as well. I'll do that.

src/admin/python/lsst/qserv/admin/cli/entrypoint.py
n8pease: Hmm I'm surprised `black` didn't flag that. It's not strictly necessary but yes, it indicates to black that I want one item per line. And it's how I normally write it so it is better for consistency to add it.

iagaponenko (approved):
Looks good to me

PR #706, opened 2022-05-05
DM-34462 Qserv CI does not see a release tag the first time it's pushed

n8pease:


n8pease:
@fritzm I think the doc loader pushes to different places based on something, is it branch? How does it know what location to push to?

fritzm:
@n8pease see here: https://github.com/lsst/qserv/blob/main/doc/conf.py#L37

The `GITHUB_REF` env variable is used, if set (this is the case in a GHA), then passed to documenteer `form_ltd_edition_name()`

fritzm:
@n8pease actually, previous comment relates to how version is established _within_ the generated docs.  For upload/publish, see here: https://github.com/lsst/qserv/blob/16d077d89e62db8a3a1ddd9abd544f6f7ffdf9d2/src/admin/python/lsst/qserv/admin/qservCli/launch.py#L603

fritzm (approved):
LGTM -- thanks!

PR #707, opened 2022-05-06
Tickets/dm 34464

jgates108:


src/qmeta/schema/migrate-4-to-5.sql
```diff
@@ -21,3 +21,12 @@ CREATE TABLE IF NOT EXISTS `QMessages` (
...
 COMMENT = 'Table of messages generated during queries.';
 
+
+--
+-- Add the chunkCount column to QInfo.
+--  
+ALTER TABLE `QInfo` 
+  ADD `chunkCount` INT,
+  ADD `resultBytes` INT DEFAULT 0,
+  ADD `resultRows` INT DEFAULT 0;
```
iagaponenko, line 31: These need to be consistent with what was put into the default schema initialization file `migrate-None-to-5.sql.jinja`. The following should do the trick:
```
ALTER TABLE `QInfo`
  ADD `chunkCount` INT NOT NULL COMMENT 'number of chunks needed by the query',
  ADD `resultBytes` INT DEFAULT 0 COMMENT 'number of bytes in the result',
  ADD `resultRows` INT DEFAULT 0 COMMENT  'number of rows in the result';
```

src/qmeta/QMetaMysql.cc
```diff
@@ -287,6 +287,8 @@ QueryId QMetaMysql::registerQuery(QInfo const& qInfo, TableNames const& tables)
     query += msgTableName;
     query += ", ";
     query += resultLocation;
+    query += ", ";
+    query += boost::lexical_cast<string>(qInfo.chunkCount());
```
iagaponenko, line 291: No need for this in modern C++11 (or newer). There is the standard library function:
```
std::to_string(qInfo.chunkCount())
```
I suggest refactoring other uses of `boost::lexical_cast<string>()` in this code where it makes sense to do.

src/czar/MessageTable.cc
```diff
@@ -138,7 +138,8 @@ void MessageTable::_saveQueryMessages(ccontrol::UserQuery::Ptr const& userQuery)
     cMsg += multiErrStr;
     LOGS(_log, LOG_LVL_DEBUG, " MULTIERROR:" << cMsg);
     std::string summaryQ = (boost::format(::writeTmpl) % _tableName % "-1" % "-1" %
-                            _sqlConn->escapeString(cMsg) % severity % std::time(nullptr)).str();
+                            _sqlConn->escapeString(cMsg) % severity % std::time(nullptr))
+                                   .str();
```
iagaponenko, line 142: Putting this to the next line doesn't look right. Did you do this, or this was done automatically by the `clang-format` tool?

src/ccontrol/UserQuerySelect.cc
```diff
@@ -538,8 +538,11 @@ void UserQuerySelect::qMetaRegister(std::string const& resultLocation, std::stri
         // Special token #QID# is replaced with query ID later.
         _resultLoc = "table:result_#QID#";
     }
+
+    int chunkCount = _qSession->getChunksSize();
```
iagaponenko, line 542: It's safer to declare it as:
```
int const chunkCount = ...
```

src/ccontrol/UserQueryFactory.cc
```diff
@@ -120,7 +120,7 @@ UserQuery::Ptr UserQueryFactory::newUserQuery(std::string const& aQuery, std::st
     if (UserQueryType::isSubmit(query, stripped)) {
         // SUBMIT is only allowed with SELECT for now, complain if anything else is there
         if (!UserQueryType::isSelect(stripped)) {
-            auto uq = std::make_shared<UserQueryInvalid>("Invalid or unsupported query: " + query);
+            auto uq = std::make_shared<UserQueryInvalid>("SUBMIT only valid with SELECT queries: " + query);
```
iagaponenko, line 123: The temporary variable is not needed here. Could you just do this?
```
return std::make_shared<UserQueryInvalid>("SUBMIT only valid with SELECT queries: " + query);
```

src/ccontrol/UserQuerySelect.cc
```diff
@@ -607,6 +612,18 @@ void UserQuerySelect::_qMetaUpdateStatus(qmeta::QInfo::QStatus qStatus) {
...
 }
 
+void UserQuerySelect::_qMetaUpdateMessages() {
+    // message table
+
+    auto msgStore = getMessageStore();
+    try {
+        _queryMetadata->addQueryMessages(_qMetaQueryId, msgStore);
+    } catch (qmeta::SqlError const& ex) {
+        LOGS(_log, LOG_LVL_ERROR, "UserQuerySelect::_qMetaUpdateMessages failed " << ex.what());
```
iagaponenko, line 626: Should this be treated as `WARN` since the code is not re-throwing the exception or taking any other measures to address the problem?
In theory, if we can't update any table in the database this should be a reason to raise an alarm.

src/czar/MessageTable.cc
```diff
@@ -125,8 +125,7 @@ void MessageTable::_saveQueryMessages(ccontrol::UserQuery::Ptr const& userQuery)
 
         char const* severity = (qm.severity == MSG_INFO ? "INFO" : "ERROR");
         std::string query = (boost::format(::writeTmpl) % _tableName % qm.chunkId % qm.code %
-                             _sqlConn->escapeString(qm.description) % severity % qm.timestamp)
-                                    .str();
+                             _sqlConn->escapeString(qm.description) % severity % qm.timestamp).str();
```
iagaponenko: `boost::format` format is really ugly! We need to get rid of it ... eventually.

src/qmeta/QMeta.h
```diff
@@ -35,6 +35,11 @@
 #include "qmeta/QStats.h"
 #include "qmeta/types.h"
 
+namespace lsst::qserv::qdisp {
+    class MessageStore;
+    class QueryMessage;
```
iagaponenko: According to @n8pease, spaces aren't allowed in the forward declarations.

src/qmeta/QMetaMysql.cc
```diff
@@ -794,11 +797,59 @@ void QMetaMysql::saveResultQuery(QueryId queryId, std::string const& query) {
...
+    query += boost::lexical_cast<string>(queryId);
+    query += ", ";
+    query += boost::lexical_cast<string>(qMsg.chunkId);
+    query += ", ";
+    query += _conn->escapeString(severity);
+    query += ", ";
+    query += _conn->escapeString(qMsg.description);
+    query += ", ";
+    query += boost::lexical_cast<string>(qMsg.timestamp);
+    query += ")";
```
iagaponenko, line 859: This "staircase" could be written in one (or two, if needed to wrap after 110 characters) line:
```
string const query = "INSERT INTO QMessages (queryId, chunkId, code, message, severity, timestamp) VALUES (" + to_string(queryId) +  ", "to_string(qMsg.chunkId) + ", "
    + _conn->escapeString(severity) + ", " +  _conn->escapeString(qMsg.description) + ", " + to_string(qMsg.timestamp) + ")";
```

src/ccontrol/UserQueryDrop.cc
```diff
@@ -115,18 +115,18 @@ void UserQueryDrop::submit() {
         // Has it disappeared already?
         LOGS(_log, LOG_LVL_ERROR, "database disappeared from CSS");
         std::string message = "Unknown database " + _dbName;
-        _messageStore->addMessage(-1, 1051, message, MessageSeverity::MSG_ERROR);
+        _messageStore->addMessage(-1, "CSS", 1051, message, MessageSeverity::MSG_ERROR);
```
iagaponenko, line 118: I know you didn't add those, though, I'm curious, what's the meaning of these hardwired "codes" such as `1051`? I'm seeing those in the new `QMessages` table as well. What are they telling us, and what are we supposed to do about these numbers?

src/ccontrol/UserQueryFactory.cc
```diff
@@ -208,8 +208,8 @@ UserQuery::Ptr UserQueryFactory::newUserQuery(std::string const& aQuery, std::st
                 _userQuerySharedResources->queryStatsData, _userQuerySharedResources->semaMgrConnections,
                 _userQuerySharedResources->qMetaCzarId, errorExtra, async, resultDb);
         if (sessionValid) {
-            uq->qMetaRegister(resultLocation, msgTableName);
             uq->setupChunking();
+            uq->qMetaRegister(resultLocation, msgTableName);
```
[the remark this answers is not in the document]
jgates108: The first one figures out how many chunks are in the query so that can be added when the query is registered in `qMetaRegister`.

src/qmeta/QMetaMysql.cc
jgates108: This was more of a "when in rome" type thing. Using to_string is fine with me.

src/czar/MessageTable.cc
jgates108: That's the formatter, not me.

src/qmeta/QMeta.h
jgates108: auto formatter

src/ccontrol/UserQueryDrop.cc
jgates108: My guess is that it's a mysql error code. mysql error 1051 is unknown table, which seems appropriate.

src/qmeta/QMetaMysql.cc
jgates108: I found breaking it up easier to read, and to ensure escaped quotes matched.

src/ccontrol/UserQueryFactory.cc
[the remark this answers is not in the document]
iagaponenko: Based on this explanation I see the ordering dependency between these calls, which is unsafe. I have also analyzed all uses of these two methods and found that `setupChunking` is used only once at the code location shown above. The other method `qMetaRegister` is used twice within the same context (class `UserQueryFactory`). The first time `qMetaRegister` may get be called after instantiating the optimized version of the unconstrained `COUNT(*)` (class `UserQuerySelectCountStar`) if that query could be answered from the special row counters table w/o sending chunk queries to workers. And the second time, `qMetaRegister` is called right after `setupChunking` (class `UserQuerySelect`).

Based on these, the method `setupChunking` needs to become the private method of class `UserQuerySelect` and be called internally as the very first step by the method `UserQuerySelect::qMetaRegister`. Note that `UserQuerySelectCountStar::qMetaRegister()` and `UserQuerySelect::qMetaRegister` are completely independent of each other (no sharing via a common base class, etc.).

src/ccontrol/UserQuerySelect.cc
iagaponenko: I didn't get you. My suggestion was to do this:
```
int const chunkCount = _qSession->getChunksSize();
```
Where:
```
class QuerySession {
public:
    int getChunksSize() const { return _chunks.size(); }
};
```
I don't think this requires any copy constructor because `int` is the built-in primitive data type.

src/qmeta/QMetaMysql.cc
iagaponenko: That's fine. Please, replace `boost::lexical_cast<string>` with `std::to_string`.

src/ccontrol/UserQueryDrop.cc
iagaponenko: Using this code here is incorrect for two reasons. First of all, the previous line is complaining about a database:
```
std::string message = "Unknown database " + _dbName;
```
 Meanwhile, the MySQL error code 1051 corresponds to the macro definition for `ER_BAD_TABLE_ERROR`, which is posted when the table is not found. See: https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html#error_er_bad_table_error

The second comment is our code should not use hard coded MySQL error codes. If one needs to report MySQL-specific code then the different code `ER_BAD_DB_ERROR` should be used:
```
// Third party headers
#include <mysql/mysqld_error.h>
#include <mysql/errmsg.h>

std::string message = "Unknown database " + _dbName;
_messageStore->addMessage(-1, "CSS", ER_BAD_DB_ERROR, message, MessageSeverity::MSG_ERROR);
```

src/qmeta/schema/migrate-4-to-5.sql
```diff
@@ -26,7 +26,7 @@ COMMENT = 'Table of messages generated during queries.';
 -- Add the chunkCount column to QInfo.
 --  
 ALTER TABLE `QInfo` 
-  ADD `chunkCount` INT,
-  ADD `resultBytes` INT DEFAULT 0,
-  ADD `resultRows` INT DEFAULT 0;
+  ADD `chunkCount` INT NOT NULL COMMENT 'number of chunks needed by the query',
+  ADD `resultBytes` INT DEFAULT 0 COMMENT 'number of bytes in the result',
+  ADD `resultRows` INT DEFAULT 0 COMMENT  'number of rows in the result',;
```
iagaponenko, line 31: This has a syntax error because the line ends with the symbol `,` followed by the symbol `;`. MySQL will choke on this statement.

iagaponenko (approved):
Looks good to me.

PR #709, opened 2022-05-10
DM-34747: Add support to Qserv for standalone boolean value expressions in WHERE clauses

n8pease:


itest_src/datasets/case02/queries/1004.2_varObjects.sql
```diff
@@ -3,7 +3,6 @@
 
 SELECT objectId
 FROM   Object
-WHERE  scisql_s2PtInCircle(ra_PS, decl_PS, 1.5, 3, 1) = 1 -- noQserv
--- withQserv WHERE qserv_areaspec_circle(1.5, 3, 1)
+WHERE  scisql_s2PtInCircle(ra_PS, decl_PS, 1.5, 3, 1)
```
fritzm, line 6: I think it would probably be good to have test cases for both with _and_ without `=1` versions of this now non `qserv_areaspec_circle` query.

Also, you do want the `AND variability ...` clause here; now that the conditionals are gone, should you remove the `--`?

fritzm:
Could you also please change the commit message 'add support for CONTAINS without "=1"' to something more descriptive like 'add  ValueExprePredicate'?

fritzm (approved):
Looks good to me -- nice, general, fix on a short timeline.  A couple minor suggestions/asks left in comments.  Thanks!

itest_src/datasets/case02/queries/1004.2_varObjects.sql
iagaponenko: We do need a test with a more complex expression as this was discussed earlier in the team's Slack channel.

src/query/ValueExprPredicate.h
```diff
@@ -0,0 +1,83 @@
...
+
+// System headers
+#include <memory>
+#include <string>
+
+// Local headers
+#include "query/Predicate.h"
+
+namespace lsst::qserv::query {
+class ValueExpr;
```
iagaponenko, line 35: Very minor comment. Usually, this needs to be indented by 4 spaces. Was the indentation removed by `clang-format`?

iagaponenko (approved):
Looks good to me.

PR #713, opened 2022-05-24
Move partition submodule into Qserv

fritzm:
Much of this shows up as "new files", since the PR doesn't cross into the (former) submodule.  The only change that was required to the `partition` sources themselves was changing occurrences of `#include "lsst/partition/..."` to `#include "partition/..."` throughout.

(Then, of course, `clang-format`.  And then, of course, fixing the latent compiler warns...  I've organized these onto their own commits.)


fritzm:
Okay, done fiddling with this now.  Probably best to take this one by commit, because the `clang-format` and `clang-tidy` commits pee all over everything.

@iagaponenko I think I may have overwritten some of your previous comments when I force pushed?  If you have small cleanups that you spy that you really think we should do now, I'm happy to add them to another commit here.  Otherwise, we could move things forward after this is merged on another "clean-up" ticket.

PR #719, opened 2022-06-23
Added support for copying the query text

iagaponenko:
Refactored tables on the query monitoring page to allow copying queries
to the clipboard (a fallback solution is offered for cases
when the clipboard is not writeable). Separated query formatting controls
from browsing the table rows in order to allow additional operations
over queries (the first one was mentoned earlier).


fritzm (approved):
LGTM, thanks!

PR #722, opened 2022-06-28
Tickets/dm 35189

iagaponenko:


src/replica/HttpProcessor.cc
```diff
@@ -245,6 +245,12 @@ void HttpProcessor::registerServices() {
                                                                  self->_processorConfig, req, resp,
                                                                  "SELECT-QUERY-BY-ID");
                              });
+    httpServer()->addHandler("GET", "/replication/qserv/css/shared-scan",
+                             [self](qhttp::Request::Ptr const& req, qhttp::Response::Ptr const& resp) {
```
fritzm, line 249: For your consideration: if you capture request and response smart pointers by _value_ in this lambda, the reference counts are incremented, and the lambda is guaranteed to have a live object when it is called; when the last outstanding lambda runs or is canceled, the ref count goes to zero and the object is freed.  This is the design intention of using smart pointers for these arguments.

If you capture by reference, how are you certain that the request object is still alive when the asynchronous handler is called at some indefinite point in the future, or in such a circumstance where e.g. the io-context is being destroyed or its thread has been asked to exit?

I know we have had this discussion in the past, and you are hard to convince on this point.  Just wanted to explain what the _intent_ was here...

fritzm (approved):
LGTM, but one comment I would ask you to at least consider, if not adopt.

src/replica/HttpProcessor.cc
iagaponenko: In principle, the lifetime of the request/response objects is guaranteed by `self` which is the shared pointer:
```
 auto const self = shared_from_base<HttpProcessor>();
```
And this pointer is passed into all lambdas in this context by value. The class `HttpProcessor` has its very base class defined as:
```
class HttpProcessor : public HttpSvc, public EventLogger { ... }
class HttpSvc : public std::enable_shared_from_this<HttpSvc> { ... }
```
An instance of `qhttp::Server` lives within the class `HttpSvc` (the base class of `= `HttpProcessor`):
```
class HttpSvc : public std::enable_shared_from_this<HttpSvc> {
...
private:
    qhttp::Server::Ptr _httpServer;
};
```
Therefore (I think) the lifetime of the request/responses is no shorter than the lifetime of that `self` (`HttpProcessor`). The only scenario where things may theoretically go wrong here would be if the `qhttp` server was asked to `stop` (there is the corresponding method for that). I don't know what's going to happen to the request/response objects (as well as to any other outstanding operations witin `qhttp`) in this case.

Though, in general, I would agree with you that adding an extra "line of defense" wouldn't hurt here. I will migrate this implementation to reinforce it.

PR #724, opened 2022-07-13
Tickets/dm 35456

iagaponenko:


src/replica/HttpModuleBase.cc
```diff
@@ -63,10 +64,51 @@ void HttpModuleBase::execute(string const& subModuleName, HttpAuthType const aut
...
+            }
+            version = body().required<unsigned int>(versionAttrName);
+        }
+    } catch (...) {
+        throw HttpError(func, "The required parameter " + versionAttrName + " is not a number.", errorEx);
+    }
+    if (!(minVersion <= version && version <= maxVersion)) {
+        throw HttpError(func,
+                        "The requested version " + to_string(version) +
+                                " of the API is not in the range supported by the service.",
```
fritzm, line 101: Maybe return the supported range in the error message here too?

fritzm:
Should there be some test coverage for the new version-checking feature?

fritzm (approved):
Looks good.  Minor comment on one error message.  Should there also be some test coverage for out-of-range, unsupplied, and/or "0" versions?

src/replica/HttpModuleBase.cc
iagaponenko: Actually, the code already does this. In the expanded version of the method you would see:
```
json const errorEx = json::object({{"min_version", minVersion}, {"max_version", maxVersion}});
try {
    ...
} catch (...) {
    throw HttpError(func, "The required parameter " + versionAttrName + " is not a number.", errorEx);
}
if (!(minVersion <= version && version <= maxVersion)) {
    throw HttpError(func,
                    "The requested version " + to_string(version) +
                            " of the API is not in the range supported by the service.",
                    errorEx);
}
```
The exception class has the third parameter that accepts an optional dictionary that depends on an error.

iagaponenko:
> Looks good. Minor comment on one error message. Should there also be some test coverage for out-of-range, unsupplied, and/or "0" versions?

This is a good idea. Unfortunately, it would be hard to implement it for the version testing alone before I have mock versions of the REST modules that could be tested w/o setting up and running the real HTTP server. Most services have side effects, or, at least, they require some client connected to the service.

An alternative approach could be to extend the current integration test to expand coverage for the Replication/Ingest system.

I need to think more about both scenarios. In any case, this should be a dedicated effort since the amount of work for creating such infrastructure would exceed the amount of work on the improvement state in the ticket's description. I have registered the following improvement https://confluence.lsstcorp.org/display/DM/Testing+framework+for+the+REST+API in my TODO list at https://confluence.lsstcorp.org/display/DM/Completed%2C+on-going+and+proposed+improvements (see section/table 2 of the document)

PR #725, opened 2022-07-23
Tickets/dm 35696

iagaponenko:


src/replica/ReplicationTask.h
```diff
@@ -50,54 +49,40 @@ class ReplicationTask : public Task {
...
+     * @param numReplicas The desired number of replicas.
+     * @param purge Purge excess replicas if 'true'.
+     * @return The smart pointer to a new object
      */
     static Ptr create(Controller::Ptr const& controller,
                       Task::AbnormalTerminationCallbackType const& onTerminated,
-                      unsigned int qservSyncTimeoutSec, unsigned int replicationIntervalSec,
-                      unsigned int numReplicas, bool purge);
+                      unsigned int qservSyncTimeoutSec, bool forceQservSync,
+                      unsigned int replicationIntervalSec, unsigned int numReplicas, bool purge);
```
andy-slac, line 67: Not sure if I asked you about this already - maybe it's worth switching to `std::chrono` types for timeouts, for better type safety?

src/replica/Task.h
```diff
@@ -200,46 +179,37 @@ class Task : public EventLogger, public std::enable_shared_from_this<Task> {
...
 
     /**
-     * Log a message into the Logger's LOG_LVL_INFO stream
-     *
-     * @param msg
-     *   a message to be logged
+     * Log a message into the Logger's LOG_LVL_INFO stream.
+     * @param msg A message to be logged
      */
     void info(std::string const& msg) { LOGS(_log, LOG_LVL_INFO, context() << msg); }
```
andy-slac, line 192: This is an unconventional use of logging, I think you are supposed to use `LOGS` macro in the code directly, otherwise the context information in the log message is not very useful.

src/wpublish/SetChunkListQservRequest.h
```diff
@@ -45,7 +45,7 @@ class SetChunkListQservRequest : public QservRequest {
         SUCCESS,  // successful completion of a request
         INVALID,  // invalid parameters of the request
         IN_USE,   // request is rejected because one of the chunks is in use
-        ERROR     // an error occured during command execution
+        ERROR     // an error ocurred during command execution
```
andy-slac, line 48: `occurred` is the word you are looking for 🙂

andy-slac (approved):
Looks good, few minor comments.

src/replica/ReplicationTask.h
iagaponenko: This is a good idea. It was already discussed in the past. Iy's in my refactoring list for the R-I system. The `unsigned int` type is used in hundreds of locations across the code. Originally, this was used because the code is based on Boost ASIO which didn't accept the `std::chrono` at the time when this code was written. Boost had (and AFAIK still has its own type `boost::asio::chrono` for that). Besides, in some cases, I need to pass timeouts from the Controllers to the R-I workers via Protobuf. So, I thought that I wouldn't gain much due to these "border conditions" (type transformations between `std::chrono`/`boost::asio::chrono`/Protobuf. I should revisit the status of Boots ASIO to see if it now accepts `std::chrono` natutrally. I'm not sure how convincing this story sounds.
However, thank you for bringing this up!

src/replica/Task.h
iagaponenko: I will think about this. The method was introduced to reduce the noise in the code by calling:
```
info("Something happened");
```
vs:
```
LOGS(_log, LOG_LVL_INFO, context() << "Something happened");
```
I will investigate an option to replace `info()` with the macro-definition to benefit from both.
This pattern is presently used in all modules of the R-I system's REST API, not just here. So, for the sake of code consistency, I'd rather do this as a separate refactoring effort rather than fixing this particular usage.
(I maintain a list of things to refactor in Confluence which is presently down).

PR #727, opened 2022-07-27
Increased key limits in the configurations of the MariaDB services

iagaponenko:
The previous limits were too low for production instances


fritzm (approved):
LGTM, thanks!

PR #729, opened 2022-07-28
DM-35744: Add some additional trace logging to RelationGraph

fritzm:


src/qana/RelationGraph.cc
```diff
@@ -878,6 +882,27 @@ void computeMinimumOverlap(Vertex& vtx) {
...
                 u->overlap = requiredOverlap;
                 q.enqueue(u);
+                LOGS(_log, LOG_LVL_TRACE, " (updated and queued)");
+            }
+        }
+    }
+}
+
+/// dump graph
+void dumpGraph(std::list<Vertex> const& vertices) {
```
iagaponenko, line 892: Shouldn't this free-standing function be put into the anonymous namespace at the beginning of this compilation unit?

iagaponenko: I don't see any harm to have it in the `lsst::qserv::qana` namespace. I'm just curious what's the right code design rule here.

iagaponenko (approved):
Looks good to me

src/qana/RelationGraph.cc
fritzm: Oh, I though I *had* put in in the anon namespace...  Will fix, thx

PR #733, opened 2022-08-19
Tickets/dm 35578

iagaponenko:


src/replica/ConfigTestApp.cc
```diff
@@ -881,131 +875,92 @@ bool ConfigTestApp::_testTables() {
...
+    table22.directorTable = "director-2";
+    table22.directorTableKey = "director_id_key";
+    table22.latitudeColName = "decl";
+    table22.longitudeColName = "ra";
+    table22.columns.emplace_back(table22.directorTableKey, "INT UNSIGNED");
+    table22.columns.emplace_back(table22.latitudeColName, "DOUBLE");
+    table22.columns.emplace_back(table22.longitudeColName, "DOUBLE");
+
+    success = success && addTable(table22);
+    {
```
jgates108, line 960: table1of2, table2of2, or table1st, tabl2nd would probably be slightly better names. Just looking at the names I expect 20 more tables.

iagaponenko: Good point! I will fix that.

src/replica/HttpConfigurationModule.cc
```diff
@@ -275,17 +275,22 @@ json HttpConfigurationModule::_deleteTable() {
 json HttpConfigurationModule::_addTable() {
     debug(__func__);
 
-    auto const database = body().required<string>("database");
-    auto const table = body().required<string>("name");
-    auto const isPartitioned = body().required<int>("is_partitioned") != 0;
+    // FIXME: extend the service to accept all attributes of the table.
+    // Create a method in the base class to extract standard attributes from
+    // from the requests's body. Use this method here and in the method HttpIngestModule::_addTable
```
jgates108, line 280: IS there a Jira ticket for this FIXME?

src/replica/HttpIngestModule.cc
```diff
@@ -684,18 +685,18 @@ json HttpIngestModule::_deleteTable() {
...
+                          conn->sqlId("qservMeta", ::directorIndexTable(database.name, table.name)));
         }
         // Remove the row counters table (if any)
         try {
             conn->execute("DROP TABLE IF EXISTS " +
-                          conn->sqlId("qservMeta", ::rowCountersTable(databaseInfo.name, table)));
+                          conn->sqlId("qservMeta", ::rowCountersTable(database.name, table.name)));
         } catch (invalid_argument const&) {
             // This exception may be thrown by the table name generator if
             // it couldn't build a correct name due to MySQL limitations.
```
jgates108, line 702: Is it worth logining the exceptions as warnings?

src/replica/testConfiguration.cc
```diff
@@ -95,8 +95,11 @@ BOOST_AUTO_TEST_CASE(ConfigurationTestStaticParameters) {
 
 BOOST_AUTO_TEST_CASE(ConfigurationInitTestJSON) {
     LOGS_INFO("Testing JSON initialization");
-
+#if 0
     BOOST_REQUIRE_NO_THROW(config = Configuration::load(ConfigTestData::data()));
+#else
+    config = Configuration::load(ConfigTestData::data());
+#endif
```
jgates108, line 102: Is is useful to hold onto the BOOST_RQUIRE_NO_THROW line?

iagaponenko: That was a leftover from some debugging effort. I will eliminate the second branch of the condition. Thank you for noticing it!

src/replica/HttpIngestModule.cc
iagaponenko: You're right - this has to be logged as `ERROR`. I will fix it.

jgates108 (approved):
It looks good, just minor comments.

admin/local/docker/compose/log/log-repl-controller.cnf
```diff
@@ -8,7 +8,7 @@ log4j.appender.CONSOLE.layout.ConversionPattern=[%d{yyyy-MM-ddTHH:mm:ss.SSSZ}] [
 # Tune log at the module level
 #log4j.logger.lsst.qserv.replica=DEBUG
 #log4j.logger.lsst.qserv.replica.DatabaseMySQL=DEBUG
-#log4j.logger.lsst.qserv.replica.MessengerConnector=DEBUG
+log4j.logger.lsst.qserv.replica.MessengerConnector=TRACE
```
andy-slac, line 11: Do you always want TRACE logging, or was it for testing only?

andy-slac (approved):
Looks good, small bunch of comments. I mostly looked at Python and some C++ code, skipped test data completely. Lots of C++ changes look like they are just renaming of variables, I did not look at those closely.

admin/local/docker/compose/log/log-repl-controller.cnf
iagaponenko: This particular config is used only by the integration test. I would like to keep the `TRACE` level on for now. This configuration won't "contaminate" any production deployments.

src/replica/schema/migrate-None-to-11.sql
```diff
@@ -66,7 +66,13 @@ CREATE TABLE IF NOT EXISTS `config_database_table` (
   `table`           VARCHAR(255)  NOT NULL ,
   `is_partitioned`  BOOLEAN NOT NULL ,
   `director_table`  VARCHAR(255) DEFAULT "" , -- The name of the corresponding 'director' table (if any)
+                                              -- or the first referenced 'director' table (if the current table is RefMatch)
   `director_key`    VARCHAR(255) DEFAULT "" , -- The name of the table's FK key linking to the PK of the `director` table (if any)
+                                              -- or the first referenced 'director' table's FK key (if the current table is RefMatch)
+  `director_table2` VARCHAR(255) DEFAULT "" , -- The name of the second referenced 'director' table (if the current table is RefMatch)
+  `director_key2`   VARCHAR(255) DEFAULT "" , -- The name of the second referenced 'director' table's FK key (if the current table is RefMatch)
+  `flag`            VARCHAR(255) DEFAULT "" , -- The name of a column for flags (if the current table is RefMatch)
+  `ang_sep`         DOUBLE       DEFAULT 0 ,  -- The angular separation parameter (if the current table is RefMatch)
```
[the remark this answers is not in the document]
iagaponenko: As far as I know, this parameter is closely related to `overlap`, and it's expressed in the same units. I think it's in degrees....may be. I couldn't find any specific reference for the units in the class `lsst::qserv::partition::Chunker`.

src/partition/sph-partition-matches.cc
```diff
@@ -71,6 +71,13 @@
 namespace fs = boost::filesystem;
 namespace po = boost::program_options;
 
+namespace {
+std::string coords2str(std::pair<double, double> const& loc, int chunkId) {
+    return "(lon:" + std::to_string(loc.first) + ",lat:" + std::to_string(loc.second) +
```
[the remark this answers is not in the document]
iagaponenko: The partitioner library and tools are consistently using `longitude` and `latitude`. That's why I used those. On the other hand, all table schemas I'm seen so far use the naming convention based on `ra` and `dec`. It's a bit confusing. I wish we had the common convention everywhere.

src/admin/python/lsst/qserv/admin/itest.py
```diff
@@ -376,8 +376,8 @@ def run(self) -> None:
         """
         _log.info("Running %s", self.case_id)
         for query_file in os.listdir(self.queries_dir):
-            if "FIXME" in query_file:
-                _log.info("Skipping FIXME query %s", os.path.basename(query_file))
+            if not query_file.endswith(".sql"):
```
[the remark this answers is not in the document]
iagaponenko: What's wrong with using `str.endswith`? Is there any performance penalty for calling this function versus what you've suggested to do instead? Just curious.
Note that my intent here is to select files that end with `.sql`,

src/admin/python/lsst/qserv/admin/itest_table.py
```diff
@@ -0,0 +1,88 @@
...
+    def fields_escaped_by(self) -> str:
+        return self._csv_dialect_attr("fields_escaped_by", "\\\\")
+
+    @property
+    def lines_terminated_by(self) -> str:
+        return self._csv_dialect_attr("lines_terminated_by", "\\n")
+
+    def _csv_dialect_attr(self, attr: str, default_val: str = "") -> str:
+        if (attr in self.ingest_config) and (self.ingest_config[attr] != ""):
+            return self.ingest_config[attr]
```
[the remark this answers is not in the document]
iagaponenko: 👍

src/admin/python/lsst/qserv/admin/itest_table.py
```diff
@@ -0,0 +1,88 @@
...
+    ref_db_table_schema_file: str
+
+    @property
+    def is_partitioned(self) -> bool:
+        return bool(self.ingest_config["is_partitioned"])
+
+    @property
+    def is_match(self) -> bool:
+        attr = "director_table2"
+        return (attr in self.ingest_config) and (self.ingest_config[attr] != "")
```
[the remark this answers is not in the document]
iagaponenko: 👍

src/admin/python/lsst/qserv/admin/itest_load.py
```diff
@@ -159,57 +159,22 @@ def _load_ref_data(
...
-        The data separator used in the data_file.
+    table : `LoadTable`
+        The descriptor of the table to load data into.
     """
-    sql = f"LOAD DATA LOCAL INFILE '{data_file}' INTO TABLE {table} FIELDS TERMINATED BY '{field_sep}'"
+    sql = f"LOAD DATA LOCAL INFILE '{data_file}' INTO TABLE {table.table_name}"
+    sql = sql + f" FIELDS TERMINATED BY '{table.fields_terminated_by}'"
+    sql = sql + f" ENCLOSED BY '{table.fields_enclosed_by}'"
+    sql = sql + f" ESCAPED BY '{table.fields_escaped_by}'"
+    sql = sql + f" LINES TERMINATED BY '{table.lines_terminated_by}'"
```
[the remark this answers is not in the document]
iagaponenko: I will look at this option.

src/partition/sph-partition-matches.cc
```diff
@@ -296,6 +305,10 @@ void Worker::defineOptions(po::options_description& opts) {
                        "the matched identifiers aren't the same. Otherwise just complain "
                        "and assume that the matched objects are still within the specified "
                        "overlap radius.");
+    part.add_options()("part.position-mismatch-ignore", po::bool_switch()->default_value(false),
+                       "Do not abort processing if positions of the matched objects aren't "
```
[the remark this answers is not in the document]
iagaponenko: Goof point.

src/replica/ConfigTable.h
```diff
@@ -0,0 +1,173 @@
...
+class TableInfo {
+public:
+    std::string name;      ///< The name of the table.
+    std::string database;  ///< The name of the parent database.
+
+    // The publishing status and major events in the table's history.
+
+    bool isPublished = false;
+    uint64_t createTime = 0;
+    uint64_t publishTime = 0;
```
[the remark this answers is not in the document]
iagaponenko: This suggestion has been registered at my TODO list: https://confluence.lsstcorp.org/display/DM/Completed%2C+on-going+and+proposed+improvements.
I will do this in the next refactoring as there are many uses of this plain integer type instead of `<crono>`.

src/replica/ConfigTable.h
```diff
@@ -0,0 +1,173 @@
...
+     * as for the firts director attribute \name directorTable.
+     */
+    std::string directorTable2;
+
+    /**
+     * @brief The name of a column representing object identifiers.
+     *
+     * All partitioned tables (both "director" and the "dependent" tables, including
+     * the "RefMatch" ones) will have a non-empty value here.
+     * In the "director" table this is the unique PK identifying table rows.
```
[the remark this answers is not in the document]
iagaponenko: It's the name of a column that makes the PK. I will fix the comment to make it clear.

src/replica/ConfigTable.h
```diff
@@ -0,0 +1,173 @@
...
+    std::string directorTableKey;
+
+    /// The name of the "RefMatch" table's key representing object (RefMatch tables only)
+    std::string directorTableKey2;
+
+    /// The name of a column that's used to store flags  (RefMatch tables only)
+    std::string flagColName;
+
+    /// The angular separation parameter (RefMatch tables only)
+    double angSep = 0.0;
```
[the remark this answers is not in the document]
iagaponenko: I don't know. Interesting enough, even `CSS` doesn't know that. This parameter is passed into class:
https://github.com/lsst/qserv/blob/main/src/css/MatchTableParams.h

where corresponding data member is declared as:
```
/// A container for match-table metadata.
struct MatchTableParams {
    MatchTableParams() : angSep(0.0) {}
    MatchTableParams(std::string const& dirTable1_, std::string const& dirColName1_,
                     std::string const& dirTable2_, std::string const& dirColName2_,
                     std::string const& flagColName_, double angSep_)
            : dirTable1(dirTable1_),
              dirColName1(dirColName1_),
              dirTable2(dirTable2_),
              dirColName2(dirColName2_),
              flagColName(flagColName_),
              angSep(angSep_) {}

    std::string dirTable1;    ///< First director-table involved in match.
    std::string dirColName1;  ///< The column used to join with dirTable1.
    std::string dirTable2;    ///< Second director-table involved in match.
    std::string dirColName2;  ///< The column used to join with dirTable2.
    std::string flagColName;  ///< Match-flags column name.
    double angSep;            ///< Max angular separation between objects in director tables.

    /// Returns true if table is a match table
    bool isMatchTable() const { return not dirTable1.empty(); }

```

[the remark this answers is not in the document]
iagaponenko: I would expect (as minimum) the name of the parameter to carry the units, like in:
```
double angSepDeg;
```
or (as maximum) using a proper C++ abstraction in the true C++ way:
```
class AngSepDeg {
public:
    /// @throws if val is not strictly greater than 0 and less than some reasonable max
    AngSepDeg(double val);
};
```
That would be much safer. Especially, if this class was used everywhere in Qserv code.

src/admin/python/lsst/qserv/admin/itest.py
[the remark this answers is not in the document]
andy-slac: I saw you using `os.path.splitext` in other places, so maybe for consistency reason? `splitext` is supposed to handle some corner cases better, but probably does not matter in your case.

src/replica/HttpConfigurationModule.cc
iagaponenko: I don't have (yet) any JIRA ticket for that. I will have the one later. I have registered this concern in the refactoring & code improvement list for the Replication/Ingest system at: https://confluence.lsstcorp.org/display/DM/Completed%2C+on-going+and+proposed+improvements

src/replica/ConfigTable.h
```diff
@@ -0,0 +1,173 @@
...
+     * Unlike other tables, the matched director table is also allowed to include
+     * the name of a database where the referenced director is residing (in case if
+     * this is not the same database as the one wher the "RefMatch" itself is residing.
+     * In this case  a value stored in the attribute may look like:
+     * @code
+     * [<database>.]<table>
+     * @endcode
+     * @note Neither name (thw database or the table) above should have any quotes.
+     * The name of the databasde (if any) would be retreived using the convenience
+     * method \name databaseName().
```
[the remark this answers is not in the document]
iagaponenko: How about introducing the following class to represent table dependencies:
```
struct DirectorTableRef {
    std::string directorDatabase;  // optional
    std::string directorTable;
    std::string directorTableKey;
};
```
Note that the current version of the new class `TableInfo` has two optional links to the directors that are needed for the child or the `RefMatch` tables:
```
class TableInfo {
public:
    ..
    std::string directorTable;
    std::string directorTableKey;
    std::string directorTable2;
    std::string directorTableKey2;
    ..
};
```
If I encapsulate the relevant parameters into `DirectorTableRef` then I would have:
```
class TableInfo {
public:
    struct DirectorTableRef { ... };
    DirectorTableRef directorTable;
    DirectorTableRef directorTable2;
    ...
};
```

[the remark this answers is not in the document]
andy-slac: 👍 Looks reasonable. `primaryKeyName` is still a bit confusing, can it be called `primaryKeyColumn`? And you may want few convenience methods in that class, e.g. constructor from dotted name, formatting, etc., up to you of course.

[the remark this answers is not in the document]
iagaponenko: Surely, I will add those methods as needed. Naming the column `primaryKeyColumn` is a good idea. Thank you!

src/partition/sph-partition-matches.cc
```diff
@@ -293,9 +302,13 @@ void Worker::defineOptions(po::options_description& opts) {
                        "chunk info is available for the identifier.");
     part.add_options()("part.chunk-mismatch-abort", po::bool_switch()->default_value(false),
                        "Abort processing if the chunk numbers found in the index maps for "
-                       "the matched identifiers aren't the same. Otherwise just complain "
+                       "the matched identifiers are not the same. Otherwise just complain "
                        "and assume that the matched objects are still within the specified "
                        "overlap radius.");
+    part.add_options()("part.position-mismatch-ignore", po::bool_switch()->default_value(false),
+                       "Do not abort processing if positions of the matched objects are not "
+                       " within the specified overlap overlap radius. Just issue the warning "
```
fritzm, line 310: typo "overlap overlap"

src/partition/sph-partition-matches.cc
```diff
@@ -293,9 +302,13 @@ void Worker::defineOptions(po::options_description& opts) {
                        "chunk info is available for the identifier.");
     part.add_options()("part.chunk-mismatch-abort", po::bool_switch()->default_value(false),
```
fritzm, line 303: I've been curious about this option -- do you think it is useful to have?  Isn't the whole _point_ of the match partitioner to deal with this case?  Why would we want to abort?

fritzm:
In commit https://github.com/lsst/qserv/pull/733/commits/d6ee485fb44e04cbb7e21f4329e740ae56f48f66, your commit comment reads:

> In the new version of the tool, there is a flag named --part.position-mismatch-ignore allowing to report the mismatch and continue processing. In this case, the non-matched entries will be registered with the flag set to 3. The flag is meant to be used when debugging the partitioned data.In the new version of the tool, there is a flag named --part.position-mismatch ignore allowing to report the mismatch and continue processing. In this case, the non-matched entries will be registered with the flag set to 3. The flag is meant to be used when debugging the partitioned data.

Is this part about "3" still accurate?  I did not see that in the diffs.  (Also, per discussion about flag bits, if it _were_ true, I think it would be incorrect -- "3" in flags already has a meaning, which is both ends of match row within one chunk.)

itest_src/datasets/case01/queries/1084_not_like.sql
```diff
@@ -1,7 +1,3 @@
 -- test for NOT LIKE
 
-
--- pragma sortresult
```
fritzm, line 4: Are you sure these `--pragma sortresults` comments are not needed?  My understanding was they informed the integration test to sort results before comparing qserv result to ref db results, because order of row return is not guaranteed?

fritzm (approved):
Looks good to me -- left a few comments -- thanks!

src/partition/sph-partition-matches.cc
iagaponenko: Will fix it now.

src/partition/sph-partition-matches.cc
iagaponenko: I added that option when debugging the input for the table `MatchesTrurth` in `DP02`. That was done (I think, on a different ticket). Perhaps, it's no longer needed. If you think the option is no longer needed then I can eliminate it.

itest_src/datasets/case01/queries/1084_not_like.sql
iagaponenko: I have checked the code of Qserv (C++ code) and the integration tests, and I'm not seeing anything that might refer to this. This also doesn't seem like a valid MySQL syntax.

iagaponenko:
@fritzm to your question:

> In commit https://github.com/lsst/qserv/commit/d6ee485fb44e04cbb7e21f4329e740ae56f48f66, your commit comment reads:

>> In the new version of the tool, there is a flag named --part.position-mismatch-ignore allowing to report the mismatch and continue processing. In this case, the non-matched entries will be registered with the flag set to 3. The flag is meant to be used when debugging the partitioned data.

> Is this part about "3" still accurate? I did not see that in the diffs. (Also, per discussion about flag bits, if it were true, I think i> t would be incorrect -- "3" in flags already has a meaning, which is both ends of match row within one chunk.)

The comment for the commit is obsolete. I had that flag on the original version of the commit. Later, I realized that I don't need this flag. I'm going to fix the commit message to stop confusing others ... and myself. Thank you for noticing this!

src/partition/sph-partition-matches.cc
iagaponenko: And, by the way, I'm going to add the following to the documentation of the parameter:
```
WARNING: do not use this option in production as it is meant to be used for debugging the incorrect input files.
```
A similar warning will be added to the parameter:
```
--part.position-mismatch-ignore
```
The original version of the partitioner was way too strict in this area. I added these parameters to allow debugging the inputs if it has problems. W/o these options the partitioner would choke on the very first incorrect line and won't let me see all incorrect lines in the input file(s).

itest_src/datasets/case01/queries/1084_not_like.sql
andy-slac: It's definitely not mysql syntax, as it is commented out. "pragma" comment is handled by the integration test script,: https://github.com/lsst/qserv/blob/main/src/admin/python/lsst/qserv/admin/itest.py#L78

iagaponenko: Good catch! Though, it looks like the test is sorting results in the text files 🐱 ```
        if outfile is None:
            return _do(subprocess.PIPE)
        else:
            os.makedirs(os.path.dirname(outfile), exist_ok=True)
            with open(outfile, "w") as f:
                _do(f)
                if self.sort_result:
                    self._sort(outfile)
```
I don't think this is the right approach as we want to be sure Qserv did the right thing.

@fritzm what do you think on this?

src/partition/sph-partition-matches.cc
iagaponenko: This option has been removed as it's no longer needed and to prevent confusing others.

itest_src/datasets/case01/queries/1084_not_like.sql
andy-slac: Why is sorting not "the right thing"? How do you compare result rows that appear in random order?

PR #737, opened 2022-09-01
Fix entrypoint defaults for czar and proxy configs

fritzm:


fjammes:
Hi @fritzm LGTM. Thanks for this quick fix!

PR #741, opened 2022-09-15
Fix 'qserv prepare-data' option

fjammes:


src/admin/python/lsst/qserv/admin/qservCli/launch.py
```diff
@@ -1189,6 +1191,8 @@ def prepare_data(
         f"src={itest_file},dst=/usr/local/etc/integration_tests.yaml,type=bind",
         "--mount",
         f"src={os.path.join(qserv_root, testdata_subdir)},dst={tests_data['qserv-testdata-dir']},type=bind",
+        "--mount",
+        f"src={outdir},dst=/tmp/data,type=bind",
```
andy-slac, line 1195: I guess `/tmp/data` here must match `tmp_data_dir` defined in `itest_load`? Maybe make it explicit by doing 

    f"src={outdir},dst={itest_load.tmp_data_dir},type=bind"

andy-slac (approved):
Looks good.

src/admin/python/lsst/qserv/admin/qservCli/launch.py
fjammes: :+1:

PR #748, opened 2022-10-20
Tickets/dm 36362

jgates108:


src/util/FileMonitor.h
```diff
@@ -37,38 +36,36 @@ namespace lsst::qserv::util {
...
 
-	/// start running `_thrd` containing `checkLoop()`.
-	void run();
+    /// start running `_thrd` containing `checkLoop()`.
+    void run();
 
-	/// Stop the thread `_thrd`.
-	void stop() { _loop = false; }
+    /// Stop the thread `_thrd`.
+    void stop() { _loop = false; }
```
iagaponenko, line 54: You don't need to have this method because it has no users but the class's destructor.

src/util/FileMonitor.h
```diff
@@ -37,38 +36,36 @@ namespace lsst::qserv::util {
...
 
-	/// Join to `_thrd`, if it is joinable.
-	void join();
+    /// Join to `_thrd`, if it is joinable.
+    void join();
```
iagaponenko, line 57: You don't need to have this method because it has no users but the class's destructor.

src/util/FileMonitor.h
```diff
@@ -37,38 +36,36 @@ namespace lsst::qserv::util {
 /// events wait for.
 class FileMonitor {
 public:
-	using Ptr = std::shared_ptr<FileMonitor>;
+    using Ptr = std::shared_ptr<FileMonitor>;
```
iagaponenko, line 39: This declaration is useless since the rest of the code is using:
```
std::shared_ptr<FileMonitor>
```
We have this historical artifact in many places of the Qserv code. We may need to get rid of it (eventually).

src/util/FileMonitor.h
```diff
@@ -37,38 +36,36 @@ namespace lsst::qserv::util {
...
 private:
-	void _setup(); ///< Do the class initialization.
-	void _checkLoop(); ///< Check for changes to the file in the thread `_thrd`.
+    void _setup();      ///< Do the class initialization.
+    void _checkLoop();  ///< Check for changes to the file in the thread `_thrd`.
 
-	std::string _fileName; ///< name of the file being watched, including path.
-	std::thread _thrd; ///< Thread monitoring file changes
-	std::atomic<bool> _loop{true}; ///< setting to false will end the loop
+    std::string _fileName;          ///< name of the file being watched, including path.
```
iagaponenko, line 63: Since the name is set only once by the only constructor of the class then it has to be declared as:
```
std::string const _fileName;
```

iagaponenko (changes requested):

src/czar/Czar.cc
```diff
@@ -134,6 +137,18 @@ Czar::Czar(string const& configPath, string const& czarName)
...
     LOGS(_log, LOG_LVL_INFO, "Czar config: " << _czarConfig);
+
+    // Watch to see if the log configuration is changed.
+    // If LSST_LOG_CONFIG is not defined, there's no good way to know what log
+    // configuration file is in use.
+    string logConfigFile = std::getenv("LSST_LOG_CONFIG");
+    if (logConfigFile == "") {
+    	LOGS(_log, LOG_LVL_ERROR, "FileMonitor LSST_LOG_CONFIG was blank, no log configuration file to watch.");
+    } else {
+        LOGS(_log, LOG_LVL_ERROR, "logConfigFile=" << logConfigFile);
```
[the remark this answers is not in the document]
jgates108: WARN or INFO is the question. There should never be a lot of these messages in the log file, but it's fairly easy to load a new configuration and suddenly have a blank log file. It's nice to have evidence for why that may have happened.

src/util/FileMonitor.h
jgates108: I disagree. I find it is convenient shorthand, but in many cases it can't be used because the header file has not been included, usually in other header files.

src/util/FileMonitor.h
jgates108: Just because it isn't really needed, doesn't mean that it shouldn't be there. It helps make the code more readable to have these functions for basic tasks separate.

admin/local/docker/compose/log/log.cnf
```diff
@@ -1,6 +1,9 @@
-log4j.rootLogger=DEBUG, CONSOLE
+#log4j.rootLogger=DEBUG, CONSOLE
+log4j.rootLogger=E, CONSOLE
```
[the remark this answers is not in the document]
iagaponenko: Got it!

src/util/FileMonitor.h
iagaponenko: My point was that there is no reason to have this method in the `public:` area of the class. Seeing a class with to many publicly visible external controls is confusing. And it complicates the interface of the class. It also makes the design fragile as these methods could be called not when the original developer expects it.

iagaponenko:
After looking through the design of the new class one more time, and analyzing use cases for the class in Qserv, I would argue that the interface of the class should have only two public methods:
* constructor
* destructor

The constructor should cal the private methods `_setup()` and `_run()`. The destructor would do oposite operations `_stop()` and `_join()`.

I see no reason to have any other publicly visible methods in this class. This is a trivial utility class with no persistent side effects. Therefore it may begin doing what it's meant to do (tracking changes in the config file) at a point of construction and finish its work at a point of destruction.

src/util/FileMonitor.h
iagaponenko: I may leave this definition. My point was that it's not used. No reason to have elements of an interface that aren't used. These elements complicate the interface w/o providing any benefits to its users.

iagaponenko (approved):
LGTM

PR #753, opened 2022-11-23
Tickets/dm 36965

jgates108:


src/wsched/ScanScheduler.h
```diff
@@ -80,6 +80,8 @@ class ScanScheduler : public SchedulerBase {
     double getMaxTimeMinutes() const { return _maxTimeMinutes; }
     bool removeTask(wbase::Task::Ptr const& task, bool removeRunning) override;
 
+    nlohmann::json getJson() const;  ///< Return a json version of this object.
```
iagaponenko, line 83: I'm not seeing any implementation of this method in the CC file of the class.

src/wbase/Task.cc
```diff
@@ -135,6 +147,23 @@ Task::Task(TaskMsgPtr const& t, std::string const& query, int fragmentNumber,
     _scanInfo.scanRating = msg->scanpriority();
     _scanInfo.sortTablesSlowestFirst();
     _scanInteractive = msg->scaninteractive();
+
+    /* &&&
+    /// For all of the histograms, all entries should be kept at least until the work is finished.
+    std::chrono::milliseconds maxAge(48h);
+    histSizePerChunk = util::Histogram::Ptr(new util::Histogram(
+            string("SizePerChunk") + _idStr, {1'000, 10'0000, 1'000'000, 10'000'000, 100'000'000}, maxAge,
```
iagaponenko, line 155: The binning of the histogram is not intuitively clear. here is what you have:
```
1'000
10'0000    <-- this is probably a mistake. Should it be 10'000 instead?
1'000'000
10'000'000
100'000'000
```
I would make the histogram binning to be strictly exponential, like:
```
1'000
10'000
100'000
1'000'000
10'000'000
100'000'000
```

src/replica/CMakeLists.txt
```diff
@@ -420,6 +420,8 @@ target_sources(replica PRIVATE
     TransactionsApp.h
     Url.cc
     Url.h
+    UserQueryWInfo.cc
+    UserQueryWInfo.h
     VerifyApp.cc
```
iagaponenko, line 423: These entries shouldn't be in the module `replica`.

src/util/Histogram.h
```diff
@@ -0,0 +1,163 @@
...
+    Histogram(Histogram const&) = delete;
+    Histogram& operator=(Histogram const&) = delete;
+
+    /// Add a time and value to the histogram.
+    std::string addEntry(TIMEPOINT stamp, double val, std::string const& note = "");
+
+    /// Add a time entry using CLOCK::now() as the time stamp.
+    std::string addEntry(double val, std::string const& note = "");
+
+    void setMaxSize(size_t maxSize);                   ///< Set the maximum number of entries to keep.
```
iagaponenko, line 114: I'm not seeing any client of the method in the PR. Is it really needed?

src/util/Histogram.h
```diff
@@ -0,0 +1,163 @@
...
+              std::chrono::milliseconds maxAge, size_t maxSize = 1000);
+    Histogram() = delete;
+    Histogram(Histogram const&) = delete;
+    Histogram& operator=(Histogram const&) = delete;
+
+    /// Add a time and value to the histogram.
+    std::string addEntry(TIMEPOINT stamp, double val, std::string const& note = "");
+
+    /// Add a time entry using CLOCK::now() as the time stamp.
+    std::string addEntry(double val, std::string const& note = "");
```
iagaponenko, line 112: The first comment on the method is about its return value. If the return value is still needed by the rest of the code (in this PR) then it's better to be documented here (in the header).

Also, the default value for the string should be `std::string()`:
```
    std::string addEntry(double val, std::string const& note = std::string());
```

src/util/Histogram.h
```diff
@@ -0,0 +1,163 @@
...
+    /// @param maxAge - maximum age in milliseconds, such that entries older than
+    ///                       it will be removed from the histogram. Default is 1 hour.
+    Histogram(std::string const& label, std::vector<double> const& bucketVals,
+              std::chrono::milliseconds maxAge, size_t maxSize = 1000);
+    Histogram() = delete;
+    Histogram(Histogram const&) = delete;
+    Histogram& operator=(Histogram const&) = delete;
+
+    /// Add a time and value to the histogram.
+    std::string addEntry(TIMEPOINT stamp, double val, std::string const& note = "");
```
iagaponenko, line 109: The first comment on the method is about its return value. If the return value is still needed by the rest of the code then it's better to be documented here (in the header). I have looked at the PR, and I'm not seeing any use for the method's return, not even in the unit test.

The second comment is that the default value for the string should be `std::string()`:
```
    std::string addEntry(TIMEPOINT stamp, double val, std::string const& note = std::string());
```

src/util/Histogram.h
```diff
@@ -0,0 +1,163 @@
...
+    /// Normal values of `incr` should be -1 or 1.
+    void _changeCountsBy(double val, int incr);
+
+    /// Remove old entries, _mtx must be held when calling.
+    void _checkEntries();
+
+    std::string _label;
+    mutable std::mutex _mtx;
+    std::vector<Bucket> _buckets;
+    int64_t _overMaxCount = 0;
```
iagaponenko, line 151: Why the signed integer type is here? does the Histogram design allow negative counters?

src/util/Histogram.h
```diff
@@ -0,0 +1,163 @@
...
+    double _total = 0.0;
+    int64_t _totalCount = 0;
```
iagaponenko, line 153: Why the signed integer type is here? does the Histogram design allow negative counters?

iagaponenko (commented):
My understanding was that this PR is still in progress. I'm still not seeing how it's going to be integrated with the monitoring probes of the Replication system. The Replication system's monitoring probes would need to find a route to the query tasks in order to obtain the JSON objects representing the corresponding histograms.

I have also made a number of general comments on the code itself.

iagaponenko:
Another comment I have is that instead of introducing the custom class `Histogram` you might look at the existing solutions like the one offered by the BOOST library: https://www.boost.org/doc/libs/1_70_0/libs/histogram/doc/html/index.html 
This would save you from implementing another histogramming class, writing a unit test to prove that it's implemented correctly, etc.

In your code, you may always add extra functionality on top of the BOOST's histograms if needed.

src/util/Histogram.cc
```diff
@@ -0,0 +1,229 @@
...
+
+std::string Histogram::getString(std::string const& note) {
+    std::lock_guard<std::mutex> lock(_mtx);
+    return _getString(note);
+}
+
+/// _mtx must be locked before calling this function.
+///
+string Histogram::_getString(std::string const& note) {
+    stringstream os;
```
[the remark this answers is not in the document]
jgates108: I don't like that for a few reasons.

1. You care about what mutex is locked, and the lock guard could easily have the wrong mutex. There needs to be comment with the lock_guard saying what mutex should be locked, or include the mutex in the name of the lock_guard parameter.
2. If someone does pass the wrong mutex in the lock_guard, it's a placebo. They see that and think there ok because there's a lock_guard there.
3. A lock_guard wont work in a unique_lock is needed.
4. There are ways of doing this that actually check that the correct mutex is locked and don't care about lock_guard or unique_lock. They require some effort, but if the  correct mutex is not locked by this thread, it will always trigger. It requires a specially written mutex, but it's compatible with everything we do.

src/util/Histogram.h
```diff
@@ -0,0 +1,163 @@
...
+
+    /// Return the _maxVal for the bucket at `index`.
+    double getBucketMaxVal(size_t index) const;
+
+    double getAvg() const;                        ///< Return the average value of all current entries.
+    double getTotal() const;                      ///< Return the total value of all entries.
+    double getSize() const;                       ///< Return the number of entries.
+    std::chrono::milliseconds getMaxAge() const;  ///< Return the maximum age allowed.
+    size_t getMaxSize() const;                    ///< Return the maximum number of entries allowed.
+    void checkEntries();                          ///< Remove old entries or entries beyond maxSize.
```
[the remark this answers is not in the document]
jgates108: checkEntries() only checks the first entry, which should be the oldest. If that is too old, it deletes it, and checks the next entry, etc.

src/util/Histogram.h
```diff
@@ -0,0 +1,163 @@
...
+    /// This keeps track of the count of entries with values greater than
+    /// the previous bucket's _maxVal and less than this bucket's _maxVal.
+    class Bucket {
+    public:
+        Bucket(double maxV) : _maxVal(maxV) {}
+        Bucket() = delete;
+        Bucket(Bucket const&) = default;
+
+        double getMaxVal() const { return _maxVal; }
+        int64_t count = 0;
```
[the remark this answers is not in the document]
jgates108: It's used with subtraction. If there's a logic error and it goes negative, it's much less likely to crash the system than an unsigned. the maximum value of an int64_t is also far higher than ever expected.

src/util/Histogram.cc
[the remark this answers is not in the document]
iagaponenko: For number 1, you may name the parameter after the name of the corresponding mutex to be locked by the lock guard:
```
class Histogram {
public:
  std::string _getString(std::lock_guard<std::mutex> const& lockOnMtx1,
                         std::string const& note);
};
```
This won't guarantee anything (beyond a requirement to lock some mutex). However, seeing this signature of the method in the class's contract (the header file) is hopefully better than having a hidden comment in the CC file like:
```
/// _mtx must be locked before calling this function.
///
string Histogram::_getString(std::string const& note) {
...
```
For number 2, it's true. I see no way to prevent this kind of mistake. My solution is not the panacea for all possible problems that might exist in the multithreaded code. This is just a step towards a better-structured code that helps a developer to avoid some traps (of forgetting to lock a mutex). Passing the lock as a parameter of a method would at least force the developer to think about the synchronization requirements of the called method. It's much easier to overlook the hidden comment in the CC file. Besides, the contemporary IDEs (VS Code, etc.) would happily pop up info on the method from the header file.

For 3, yes, I'm aware of this. That's why I never claimed that this is an ideal solution. The solution was mainly designed for implementing simple synchronized methods like the one you have here.

For 4, I myself would be very interested to learn about these techniques. Those would be useful for cases that go beyond solving the problem of implementing trivially synchronized methods.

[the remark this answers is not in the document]
jgates108: This post https://stackoverflow.com/questions/21892934/how-to-assert-if-a-stdmutex-is-locked . The answer where they define a new mutex that stores the thread id of the locking thread and has a function to verify that the current thread. If you want I could set that up.

src/util/Histogram.cc
```diff
@@ -0,0 +1,229 @@
...
+
+nlohmann::json Histogram::getJson() const {
+    std::lock_guard<std::mutex> lock(_mtx);
+    nlohmann::json rJson = {
+            {"HistogramId", _label}, {"avg", _getAvg()}, {"size", _entries.size()}, {"total", _total}};
+    for (size_t j = 0; j < _buckets.size(); ++j) {
+        auto const& bk = _buckets[j];
+        rJson["buckets"][j] = {{"index", j}, {"maxVal", bk.getMaxVal()}, {"count", bk.count}};
+    }
+
```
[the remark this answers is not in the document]
jgates108: Code was changed to use nlohmann::json::array.

src/util/Histogram.cc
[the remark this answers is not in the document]
jgates108: Looking at the code, there's a util::Mutex class that does this, but it seems to go further than necessary by keeping a list of locked mutexes. I don't quite understand the purpose of the list of locked mutex. It makes locking much more expensive, and even with the list, it would still be very difficult to figure out what went wrong. `assert(_utilMtx.lockedByCaller());` would be as effective, show where the error occured, and be much faster.  Writing our own macros would probably be worth it as well, or do something like `_utilMtx::lockedByCaller(ERR_LOC)`, that way error messages could be logged or throw.

src/util/Histogram.cc
```diff
@@ -0,0 +1,229 @@
...
+Histogram::Histogram(string const& label, std::vector<double> const& bucketVals,
+                     std::chrono::milliseconds maxAgeMillis, size_t maxSize)
+        : _label(label), _maxSize(maxSize), _maxAge(maxAgeMillis) {
+    // sort vector and remove duplicates.
+    std::set<double> valSet;
+    for (auto&& v : bucketVals) {
+        valSet.insert(v);
+    }
+    for (auto&& v : valSet) {
+        _buckets.emplace_back(Bucket(v));
```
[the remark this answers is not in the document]
jgates108: Fixed

src/util/Histogram.cc
```diff
@@ -0,0 +1,229 @@
...
+}
+
+/// _mtx must be locked before calling this function.
+///
+string Histogram::_getString(std::string const& note) {
+    stringstream os;
+
+    os << _label << " " << note << " size=" << _entries.size() << " total=" << _total << " avg=" << _getAvg()
+       << " ";
+    double maxB = -DBL_MAX;
```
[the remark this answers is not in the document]
jgates108: Using std::numeric_limits<double> now. The text translation of inifinity() was 'null', so the final bucket max value is set to "infinite" in the code.

src/util/Histogram.h
jgates108: Fixed

src/util/Histogram.h
jgates108: If there's a logic error with subtraction, massive positive values will be much more harmful than small negative numbers. It will also be nearly impossible to reach the positive maximum value of an int64_t.

iagaponenko (approved):
Look good to me. The next step whould be to inject the histograms into the Replication system's monitoring stream at the worker. This should be done in a separate ticket.

PR #754, opened 2022-12-01
Added logging to help locate lockup.

jgates108:
Looking through the log for discrepancies among the number of different instances could help determine what is going wrong.


fritzm (approved):
LGTM -- thanks!

PR #759, opened 2023-01-13
DM-37334: Worker lock-ups

jgates108:


fritzm:
Closing this stale PR for now.  The Jira issue has gone over to Igor as designated custodian for XRootD Qserv.  Igor will try and repro, and if still necessary will put up a new PR against the current latest XRootD version.

PR #764, opened 2023-01-28
mypy fixes for recent changes to mysql.connector package

fritzm:
Latest Qserv release build attempt pulled in a new version of `mysql-connector-python`, with type hint changes.

The main issue is that `mysql.connector.connect()` is now hinted as `Union[PooledMySQLConnection, MySQLConnection, CMySQLConnection]`.  You only get a pooled connection back if you pass pool kwargs to `connect`.  The pooled connection is meagre compared to the other two, and doesn't support all the calls we make in our code.  We don't pass pool kwargs anywhere to `connect`, and aren't expecting any pooled connections.

The other two varieties (and their associated cursor classes) are much more congruent, and share ABCs `MySQLConnectionAbstract` and `MySQLCursorAbstract` which have (almost) everything we need/use.  The `MySQL...` versions are pure Python, while the `CMySQL...` versions leverage a C extension for greater performance.  Which flavor you get back from `connect` by default depends on platform and whether or not the C extension is available/installed, but defaults are now trending toward the C-accelerated versions.

Mostly here I've opted for `typing.cast` to the ABCs, because we can know from the immediately surrounding code that we aren't passing pool kwargs, and it makes us insensitive to whether we get a pure Python or a C-accelerated connection.  Some alternatives I could think of:

* Use runtime `assert isinstance` instead of `cast`, for future-proofing against code changes?
* Say we don't care about performance for this application and explicitly construct tightly typed pure Python `MySQLConnector` by calling the `MySQLConnector` ctor instead of the generic `connector.connect` (or the other way around, and make the C extension a prerequisite for Qserv; perhaps that's even better.)

Opinions?  Thanks.  This is holding up the next Qserv release, so we should try to pick something soon...

### DIFF
--- a/src/admin/python/lsst/qserv/admin/itest_load.py
+++ b/src/admin/python/lsst/qserv/admin/itest_load.py
@@ -28,12 +28,12 @@ import gzip
 import json
 import logging
 import mysql.connector
-from mysql.connector.cursor import MySQLCursor
+from mysql.connector.abstracts import MySQLConnectionAbstract, MySQLCursorAbstract
 import os
 import shutil
 import subprocess
 from tempfile import TemporaryDirectory
-from typing import Any, Dict, Generator, List, NamedTuple, Optional, Tuple
+from typing import Any, Dict, Generator, List, NamedTuple, Optional, Tuple, Union, cast
 import yaml
 
 from .constants import tmp_data_dir
@@ -57,7 +57,7 @@ def unzip(source: str, destination: str) -> None:
         shutil.copyfileobj(_source, _target)
 
 
-def execute(cursor: MySQLCursor, stmt: str, multi: bool = False) -> None:
+def execute(cursor: MySQLCursorAbstract, stmt: str, multi: bool = False) -> None:
     """Execute a statement in a cursor, and clean up the cursor so it can be
     closed without causing warnings on the server."""
     warnings = []
@@ -73,6 +73,11 @@ def execute(cursor: MySQLCursor, stmt: str, multi: bool = False) -> None:
         cursors = [cursor]
 
     for cursor in cursors:
+        # Cast here because MySQLCursorAbtract does not have with_rows for some reason, even though both of
+        # its subclasses do...
+        cursor = cast(
+            Union[mysql.connector.cursor.MySQLCursor, mysql.connector.cursor_cext.CMySQLCursor], cursor
+        )
         if cursor.with_rows:
             for row in cursor.fetchall():
                 results.append(row)
@@ -121,12 +126,12 @@ def _create_ref_db(ref_db_admin: str, name: str) -> None:
                 execute(cursor, stmt)
 
 
-def _create_ref_table(cnx: mysql.connector.connection, db: str, schema_file: str) -> None:
+def _create_ref_table(cnx: MySQLConnectionAbstract, db: str, schema_file: str) -> None:
     """Create a table in the mysql database used for integration test reference.
 
     Parameters
     ----------
-    cnx : `mysql.connector.connection`
+    cnx : `mysql.connector.abstracts.MySQLConnectionAbstract`
         Connection to the reference database that has permission to create a
         table.
     db : `str`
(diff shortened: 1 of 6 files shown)


fritzm:
Note that the CI integration test fails here seem to be due to a FK constraint violation when ingesting one of the test cases; this is probably also exposed by the updated connector library?  This will also need to be addressed before release, but I don't believe it is related to the type hinting changes here, which I would like to get reviewed and merged first.

fritzm:
Rebased on documentation fixes...

src/admin/python/lsst/qserv/admin/itest_load.py
```diff
@@ -28,12 +28,12 @@
 import json
 import logging
 import mysql.connector
-from mysql.connector.cursor import MySQLCursor
+from mysql.connector.abstracts import MySQLCursorAbstract
```
ktlim, line 31: Also `MySQLConnectionAbstract` to shorten some of the lines below?

ktlim (approved):

src/admin/python/lsst/qserv/admin/itest_load.py
fritzm: Yes, thanks!

src/admin/python/lsst/qserv/admin/itest_load.py
```diff
@@ -121,7 +126,7 @@ def _create_ref_db(ref_db_admin: str, name: str) -> None:
                 execute(cursor, stmt)
 
 
-def _create_ref_table(cnx: mysql.connector.connection, db: str, schema_file: str) -> None:
+def _create_ref_table(cnx: MySQLConnectionAbstract, db: str, schema_file: str) -> None:
```
andy-slac, line 129: Maybe also change docstring for `cnx` parameter below to specify the same new type?

src/admin/python/lsst/qserv/admin/itest_load.py
```diff
@@ -145,9 +150,7 @@ def _create_ref_table(cnx: mysql.connector.connection, db: str, schema_file: str
             execute(cursor, f.read(), multi=True)
 
 
-def _load_ref_data(
-    cnx: mysql.connector.connection, data_file: str, db: str, table: LoadTable
-) -> None:
+def _load_ref_data(cnx: MySQLConnectionAbstract, data_file: str, db: str, table: LoadTable) -> None:
```
andy-slac, line 153: Same thing with docstring here.

src/admin/python/lsst/qserv/admin/itest_load.py
```diff
@@ -73,6 +73,11 @@ def execute(cursor: MySQLCursor, stmt: str, multi: bool = False) -> None:
         cursors = [cursor]
 
     for cursor in cursors:
+        # Cast here because MySQLCursorAbtract does not have with_rows for some reason, even though both of
+        # its subclasses do...
+        cursor = cast(
+            Union[mysql.connector.cursor.MySQLCursor, mysql.connector.cursor_cext.CMySQLCursor], cursor
```
andy-slac, line 79: Are you still at Python3.9? With 3.10 you can replace `Union` with `|` which is shorter and looks more natural.

src/admin/python/lsst/qserv/admin/itest_load.py
```diff
@@ -73,6 +73,11 @@ def execute(cursor: MySQLCursor, stmt: str, multi: bool = False) -> None:
         cursors = [cursor]
 
     for cursor in cursors:
+        # Cast here because MySQLCursorAbtract does not have with_rows for some reason, even though both of
+        # its subclasses do...
```
andy-slac, line 77: Could be an oversight, you can probably complain to mysql folks about it.

andy-slac (approved):
Looks good, minor comments. I'm OK with `cast` vs `assert isinstance`, if that breaks then you'll probably get an exception elsewhere. And mysql connector probably needs an improvement too.

src/admin/python/lsst/qserv/admin/itest_load.py
fritzm: We're Python3.8, here, sadly!  (Qserv containers currently build off stock alma 8.)

src/admin/python/lsst/qserv/admin/itest_load.py
fritzm: I'll file an issue.

src/admin/python/lsst/qserv/admin/itest_load.py
fritzm: Yes, thanks for the catch...

PR #769, opened 2023-02-24
Configuration options to store query results at worker filesystem

iagaponenko:


jgates108 (approved):
It looks good to me. I do have a question about the file resources. xrootd keeps a flat list of all resources. If all workers are keeping the results in `/qserv/data/results/` how are the different workers differentiated from each other? Or how are the files named to differentiat them? Concievably, if 2 different workers generate complete results for the same chunk/subchunk for the same query Id, the results should be interchangeable.

iagaponenko:
> It looks good to me. I do have a question about the file resources. xrootd keeps a flat list of all resources. If all workers are keeping the results in `/qserv/data/results/` how are the different workers differentiated from each other? Or how are the files named to differentiat them? Concievably, if 2 different workers generate complete results for the same chunk/subchunk for the same query Id, the results should be interchangeable.

The final naming convention for file resources (the names of the files) is yet to be finalized. There are various options here. The most trivial one would be to generate some unique file name using the name generator (such as UUID) run at a worker. The name would be explicitly passed back to `czar` in the Protobuf response. The second option would be to come up with some deterministic scheme based on non-ambiguous combinations of the query properties, such as:
* query ID
* chunk ID
* attempt count

Either way should work.

fritzm:
Changes were merged, but PR was orphaned

PR #770, opened 2023-02-28
Moved timers and statistics functions to where they now need to be.

jgates108:


iagaponenko (commented):
The CI is failing at https://github.com/lsst/qserv/actions/runs/4288136941/jobs/7469803409

It looks like `wcontrol/CMakeList.txt` is looking for the non-existing file `WorkerStats.cc`. I'm not seeing this file in the PR.

iagaponenko:
Please, note what GitHub says on 3 files:
...
This branch has conflicts that must be resolved
Use the web editor or the command line  to resolve conflicts.
Conflicting files
```
src/wbase/SendChannelShared.cc
src/wcontrol/Foreman.cc
src/wdb/Query
```

src/global/constants.h
```diff
@@ -22,11 +22,15 @@
...
 #define LSST_QSERV_CONSTANTS_H
+
 /**
  * @brief  Global constants.
  *
  */
 
+// Sytem includes
+#include <chrono>
+
```
iagaponenko, line 33: Sorry, this is a bad idea to get this header included in this particular header file. All classes in Qserv will catch this include, even though only a few of those actually need this. Besides, we should not use our own headers as "packages" for including standard (or any other headers). Each header should define custom products only. 

The side effect of having the include here - is an extra dependency for all (a few hundred) Qserv classes and increased compilation time.

src/global/constants.h
```diff
@@ -68,5 +72,8 @@ const int NOTSET = -1;
  */
 enum MessageSeverity { MSG_INFO = 1, MSG_ERROR };
 
+using CLOCK = std::chrono::system_clock;
+using TIMEPOINT = std::chrono::time_point<CLOCK>;
```
iagaponenko, line 76: Please, remove these from the header. If you think you need to share these definitions by the dependent classes (my understanding is that there only a few of those) then you may add a specialized header file that would have:
```
/// File: global/clock_defs.h

// System includes
#include <chrono>

using CLOCK = std::chrono::system_clock;
using TIMEPOINT = std::chrono::time_point<CLOCK>;
```

src/global/constants.h
```diff
@@ -22,11 +22,15 @@
  */
 #ifndef LSST_QSERV_CONSTANTS_H
 #define LSST_QSERV_CONSTANTS_H
+
 /**
  * @brief  Global constants.
  *
  */
 
+// Sytem includes
```
iagaponenko, line 31: The typo in "Sytem". Must be "System".

src/qdisp/CzarStats.cc
```diff
@@ -58,15 +57,18 @@ void CzarStats::setup(qdisp::QdispPool::Ptr const& qdispPool) {
...
 
 CzarStats::CzarStats(qdisp::QdispPool::Ptr const& qdispPool) : _qdispPool(qdispPool) {
-    _histTrmitRecvRate = util::HistogramRolling::Ptr(new util::HistogramRolling(
-            "TransmitRecvRateBytesPerSec", {1'000, 1'000'000, 500'000'000, 1'000'000'000}, 1h, 10000));
-    auto bucketVals = {0.1, 1.0, 10.0, 100.0, 1000.0};
-    _histRespSetup =
-            util::HistogramRolling::Ptr(new util::HistogramRolling("RespSetupTime", bucketVals, 1h, 10000));
-    _histRespWait =
-            util::HistogramRolling::Ptr(new util::HistogramRolling("RespWaitTime", bucketVals, 1h, 10000));
+    auto bucketValsRates = {1'000.0, 1'000'000.0, 500'000'000.0, 1'000'000'000.0};
```
iagaponenko, line 60: Since you're not going to modify this object then do:
```
auto const ...
```

src/qdisp/CzarStats.h
```diff
@@ -123,15 +127,14 @@ class TimeCountTracker {
 public:
```
iagaponenko, line 127: This utility class doesn't really belong to this header (`qdisp/CzarStats.h`). I'd rather put it into the same header where you define `TIMEPOINT` and `CLOCK`. And my suggestion is to put `TIMEPOINT`, `CLOCK`, and `TimeCountTracker` into the dedicated header `global/clock_defs.h`.

In general, we should be aiming at having each namespace-scoped class in the dedicated header. Exceptions are - utility headers shared by many other dependent classes, nested classes defined within other classes, or implementation-specific classes in the *.CC files.

src/util/Histogram.h
```diff
@@ -39,6 +39,9 @@
 // Third party headers
 #include <nlohmann/json.hpp>
 
+// qserv headers
+#include "global/constants.h"
```
iagaponenko, line 43: Please, see my comments posted earlier for this header. Definitions that you are importing here should be placed into the dedicated header:
```
#include "global/clock_defs.h"
```
or something.

src/wbase/SendChannelShared.cc
```diff
@@ -247,37 +242,41 @@ bool SendChannelShared::_transmit(bool erred) {
...
+        // Only wait if this is really the last message to finish transmitting for this SendChannelShared.
+        // Note: If this does speed up transmission of data to the czar then it is probably worth it to
+        //       add code to wait if more than X buffers are in use at the same time, which will likely
+        //       make a big performance crater, but better than crashing due to out of memory.
+        //       It may be best to reduce the number of threads per scheduler to 1 until it clears up instead
+        //       of waiting for every send.
+        //       If this doesn't improve transmission performance, just change it back to waiting all of the
+        //       time.
+        if (last) {
+            streamBuf->waitForDoneWithThis();  // Block until this buffer has been sent.
```
iagaponenko, line 262: This is actually a very big change to the behavior of Qserv workers. It needs to be in a separate JIRA ticket (and PR) to track and isolate possible side effects of the change. My understanding is that the declared scope of the present PR is just to streamline statistic collection and Qserv monitoring as in:
```
Moved timers and statistics functions to where they now need to be
```

Please, either remove this from this PR or redefine the scope of the JIRA ticket and the PR to reflect actual changes in the code.

iagaponenko (changes requested):
I'm sorry, I can't approve this PR. Please, see comments posted on the PR.

src/wbase/SendChannelShared.cc
jgates108: This is the change that forced most of the other changes. Moving it to another ticket is pretty simple.

src/global/clock_defs.h
```diff
@@ -0,0 +1,88 @@
...
+    TimeCountTracker(TimeCountTracker const&) = delete;
+    TimeCountTracker& operator=(TimeCountTracker const&) = delete;
+
+    /// Constructor that includes the callback function that the destructor will call.
+    TimeCountTracker(CALLBACKFUNC callback) : _callback(callback) {
+        auto now = CLOCK::now();
+        _startTime = now;
+        _endTime = now;
+    }
+
```
iagaponenko, line 56: You may want to push the implementations into the CC file `clock_defs.cc`. This would simplify the interface and reduce the compilation time of the dependent modules from where this header is included. Normally, inline implementations are made to improve the run-time performance of the code if the functions are on the performance-critical path. Objects of the current class aren't created too often. Besides, the amount of time spent within these functions are negligible compared with one of the calling contexts (query processing). So, the compile-time dependency and compilation time would play a more important role here.

src/global/clock_defs.h
```diff
@@ -0,0 +1,88 @@
...
+    /// Call the callback function as the dying act.
+    ~TimeCountTracker() {
+        TType sum;
+        {
+            std::lock_guard lg(_mtx);
+            _endTime = CLOCK::now();
+            sum = _sum;
+        }
+        _callback(_startTime, _endTime, sum, _success);
+    }
```
iagaponenko, line 66: You may want to push the implementations into the CC file `clock_defs.cc`. Same comment as for the c-tor.

src/global/clock_defs.h
```diff
@@ -0,0 +1,88 @@
...
+
+    /// Add val to _sum
+    void addToValue(double val) {
+        std::lock_guard lg(_mtx);
+        _sum += val;
+    }
```
iagaponenko, line 72: You may want to push the implementations into the CC file `clock_defs.cc`. Same comment as for the c-tor.

iagaponenko (approved):
Looks good to me. Though, a couple of minor improvements in `global/clock_defs.h` posted in the comments would be welcome.

src/qdisp/CzarStats.cc
iagaponenko: The suggestion has been withdrawn as it won't work for the list-initializer which by itself is already `const`.

src/global/clock_defs.h
jgates108: 'double val' should be `TType val`

PR #773, opened 2023-03-24
Tickets/dm 38257

jgates108:


src/wbase/UserQueryInfo.cc
```diff
@@ -0,0 +1,91 @@
...
+ * the GNU General Public License along with this program.  If not,
+ * see <http://www.lsstcorp.org/LegalNotices/>.
+ */
+
+// Class header
+#include "wbase/UserQueryInfo.h"
+
+// System headers
+
+// Qserv headers
```
iagaponenko, line 27: This comment w/o any headers included after that is useless.

src/wbase/UserQueryInfo.cc
```diff
@@ -0,0 +1,91 @@
...
+
+// LSST headers
+#include "lsst/log/Log.h"
+
+
```
iagaponenko: The extra empty line to be removed.

iagaponenko (changes requested):
The implementation of `UserQueryInfo` is not thread-safe. Please, fix it.

src/wbase/UserQueryInfo.cc
jgates108: But there is a qserv header.

src/wbase/UserQueryInfo.h
```diff
@@ -0,0 +1,123 @@
...
+
+    UserQueryInfo(QueryId);
+    UserQueryInfo() = delete;
+    UserQueryInfo(UserQueryInfo const&) = delete;
+    UserQueryInfo& operator=(UserQueryInfo const&) = delete;
+
+    ~UserQueryInfo() = default;
+
+    /// &&& doc
+    size_t addTemplate(std::string const& templateStr);
```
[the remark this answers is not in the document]
jgates108: _uqMtx was supposed to be used to protect that. Good catch, I'm not sure what happened there.

src/wbase/UserQueryInfo.h
```diff
@@ -0,0 +1,123 @@
...
+    std::string& getTemplate(size_t id);
+
+private:
+    static Map _uqMap;
+    static std::mutex _uqMapMtx; ///< protects _uqMap
+
+    QueryId const _qId; ///< The User Query Id number.
+
+
+
```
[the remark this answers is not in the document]
jgates108: I thought single empty lines were allowed?  My eyes like occasional breaks when there is some context change in the member variables.

src/wbase/UserQueryInfo.cc
iagaponenko: I don't see any `#include` statements after that comment.

src/wbase/UserQueryInfo.h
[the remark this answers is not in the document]
iagaponenko: I suggest running `clang-format` before each committing changes. I think it should take care of empty lines, etc. Normally I do this at the same time when I'm compiling the code:
```
qserv build -j24 --no-build-image --clang-format=REFORMAT
```
You may also run the tool w/o compiling the code by:
```
qserv build -j24 --no-cmake --no-make --no-build-image --clang-format=REFORMAT
```
The second command is useful when you're modifying previous commits:
```
git rebase -i <old-commit>
<modify the code}
qserv build -j24 --no-cmake --no-make --no-build-image --clang-format=REFORMAT
git add <modified-file>
git commit --amend
git rebase --continue
```
In this scenario, you won't be re-compiling the code based on the older commit which could make your development cycle longer.

src/wbase/Task.cc
```diff
@@ -91,21 +77,15 @@ bool Task::ChunkEqual::operator()(Task::Ptr const& x, Task::Ptr const& y) {
...
 // Task::PtrChunkIdGreater functor
 bool Task::ChunkIdGreater::operator()(Task::Ptr const& x, Task::Ptr const& y) {
     if (!x || !y) {
         return false;
     }
-    if ((!x->msg) || (!y->msg)) {
-        return false;
-    }
-    return x->msg->chunkid() > y->msg->chunkid();
+    return x->_chunkId > y->_chunkId;
```
iagaponenko, line 88: Same comment as for `ChunkEqual` - the comparison should also involve the new attribute `_hasChunkId` which is relevant in this context. Something like this:
```
    if (!x->_hasChunkId || !y->_hasChunkId) {
        throw std::runtime_error("Task::ChunkIdGreater::" + std::string(__func__) + ": incompatible states to allow the comparison");
    }
    return x->_chunkId > y->_chunkId;
```

src/wbase/UserQueryInfo.cc
jgates108: #include "util/Bug.h"

src/wbase/Task.cc
```diff
@@ -125,39 +105,73 @@ std::atomic<uint32_t> taskSequence{0};
...
-        : msg(t),
+Task::Task(TaskMsgPtr const& t, int fragmentNumber, std::shared_ptr<UserQueryInfo> const& userQueryInfo,
+           size_t templateId, int subchunkId, std::shared_ptr<SendChannelShared> const& sc)
+        : _userQueryInfo(userQueryInfo),
           _sendChannel(sc),
           _tSeq(++taskSequence),
           _qId(t->queryid()),
+          _templateId(templateId),
+          _hasChunkId(t->has_chunkid()),
+          _chunkId(t->has_chunkid() ? t->chunkid() : -1),
```
iagaponenko, line 116: As a general comment, I think that negative numbers make no sense for in-chunk and sub-chunk identifiers which should be modeled as natural numbers: 0, 1, 2, etc. Chunks/sub-chunks aren't meant to be used in arithmetic operations (addition or substractions where the sign is needed). And yes, I'm aware that the current Protobuf definition for these attributes incorrectly defines those as signed integers. I think that was a mistake. 

Apart from that statement, my specific comment in this context is that you have already introduced the special flag `_hasChunkId` that allows tracking the state of the `chunkId` attribute.  Similar actions should be also applied to other optional attributes that you're fetching from the Protobuf messages and storing within tasks. If you do this for all optional attributes then your code would be more consistent and safer.

Besides, you could use those `has` states to validate attributes in the comparison operations. See my comments posted on this subject for operations `ChunkIdGreater` and `ChunkIdEqual`.

src/wbase/Task.cc
```diff
@@ -125,39 +105,73 @@ std::atomic<uint32_t> taskSequence{0};
...
+
+    // Create sets and vectors for 'aquiring' subchunk temporary tables.
+    proto::TaskMsg_Fragment const& fragment(t->fragment(_queryFragmentNum));
+    if (!_fragmentHasSubchunks) {
+        /// &&& ??? Why aquire anything if there are no subchunks in the fragment?
+        /// &&& ??? All tasks for this user query probably use the same tables
+        for (auto const& scanTbl : t->scantable()) {
+            _dbTbls.emplace(scanTbl.db(), scanTbl.table());
+        }
+        assert(t->has_db());
```
iagaponenko, line 158: I think this assertion has to be done much earlier in this constructor regardless of the state of `_fragmentHasSubchunks`.

src/wbase/Task.cc
```diff
@@ -206,6 +217,13 @@ std::vector<Task::Ptr> Task::createTasks(std::shared_ptr<proto::TaskMsg> const&
     return vect;
 }
 
+string Task::getQueryString() const {
+    string qs = _userQueryInfo->getTemplate(_templateId);
+    boost::algorithm::replace_all(qs, CHUNK_TAG, to_string(_chunkId));
+    boost::algorithm::replace_all(qs, SUBCHUNK_TAG, to_string(_subchunkId));
+    return qs;
```
iagaponenko, line 232: Note that your current design of the class `Task` allows `_chunkId` and `_subchunkId` to be negative numbers. What kind of query those (negative) values would produce here? Perhaps you need to check if the task instance really has valid non-negative attributes here and throw an exception?

src/wbase/Task.cc
```diff
@@ -217,16 +235,11 @@ wpublish::QueryStatistics::Ptr Task::getQueryStats() const {
...
-        return msg->chunkid();
-    }
-    return -1;
-}
+int Task::getChunkId() const { return _chunkId; }
 
 /// Flag the Task as cancelled, try to stop the SQL query, and try to remove it from the schedule.
 void Task::cancel() {
-    LOGS(_log, LOG_LVL_INFO, "Task::cancel " << _idStr);
+    LOGS(_log, LOG_LVL_INFO, "Task::cancel " << getIdStr());
```
iagaponenko, line 250: Perhaps, `LOG_LVL_WARN` would be better in this context. Task cancellation is usually an abnormal, though, not a catastrophic event that might require a bit more attention when inspecting log streams.

src/wbase/UserQueryInfo.h
```diff
@@ -0,0 +1,123 @@
...
+// Qserv headers
+#include "global/intTypes.h"
+
+
+// This header declarations
+namespace lsst::qserv::wbase {
+
+/// &&& doc
+class UserQueryInfo {
```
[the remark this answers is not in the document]
jgates108: There are two collection objects that are important (both have changed to make them safe):

The first is the static map, `_uqMap`,  which contains weak pointers to all UserQueryInfo instances and is ordered by `QueryId`. `QueryId` is the same for all `Task` and `Job` objects in the UserQuery. `_uqMapMtx` is used to protect `_uqMap` from race conditions. Before, it used normal shared pointers and wouldn't delete the map entry when the UserQueryInfo was no longer needed. That should be fixed now.

The second is the vector of template strings, `_templates`. The index into the vector is not a pointer, it's just an unsigned integer index pointing to a std::string. Even if the vector is moved due to needing more memory, the index to a string should still be valid. New strings are always added to the end, so the previous indexes will not change. `_uqMtx` is now being used to prevent race conditions.

`_templates` is kind of annoying as it seems like it would be really nice to use the template string as the key so it can be found quickly when searching. However, using the template as the key completely defeats the purpose of saving space, and as you pointed out, putting them in a `set` or `map` could cause them to move and invalidate the indexes. Luckily, there are only a small number of templates  for any UserQuery (I think 2 max, but I'm not certain), so searching a `vector` for a string match is not much worse than a map or set.

src/wbase/Task.h
```diff
@@ -220,16 +220,41 @@ class Task : public util::CommandForThreadPool {
...
     int const _attemptCount = 0;  // attemptCount from czar
-    /// _idStr for logging only.
-    std::string const _idStr = makeIdStr(true);
-    std::string _queryString;   ///< The query this task will run.
-    int _queryFragmentNum = 0;  ///< The fragment number of the query in the task message.
+    int const _queryFragmentNum;  ///< The fragment number of the query in the task message. // &&& delete???
+    bool const _fragmentHasSubchunks;  ///< True if the fragment in this query has subchunks.
+    int64_t const _session;            ///< XrdSsi session.
+    bool const _hasDb;                 ///< true if db was in message from czar.
+    std::string _db;                   ///< Task database (&&& move to UserQueryInfo ???)
```
iagaponenko: Why this is not declared as `const`? Is the value of this attribute going to be modified later?

iagaponenko (commented):
My general comment on the new attributes that are pulled from the Protobuf message into the class `Task` is that each optional attribute should have the pairing "has" flag. The flag would indicate if the corresponding attribute was actually provided in the Protobuf method. This would also allow reinforcing implementations of the methods relying on values of the attroibutes (such as comparison operator and the query generators).

src/wbase/UserQueryInfo.h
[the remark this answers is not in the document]
jgates108: I have been running `qserv build --clang-format REFORMAT  --cmake -j12` to build and it has been shifting text around. 

`qserv build --clang-format=REFORMAT  --cmake -j12` didn't really do anything different.

src/wbase/UserQueryInfo.h
```diff
@@ -0,0 +1,123 @@
...
+    UserQueryInfo(UserQueryInfo const&) = delete;
+    UserQueryInfo& operator=(UserQueryInfo const&) = delete;
+
+    ~UserQueryInfo() = default;
+
+    /// &&& doc
+    size_t addTemplate(std::string const& templateStr);
+
+    /// &&& doc
+    std::string& getTemplate(size_t id);
```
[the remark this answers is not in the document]
jgates108: Fixed

src/wbase/Task.cc
```diff
@@ -91,21 +77,15 @@ bool Task::ChunkEqual::operator()(Task::Ptr const& x, Task::Ptr const& y) {
     if (!x || !y) {
         return false;
     }
-    if ((!x->msg) || (!y->msg)) {
-        return false;
-    }
-    return x->msg->has_chunkid() && y->msg->has_chunkid() && x->msg->chunkid() == y->msg->chunkid();
+    return x->_chunkId == y->_chunkId;
```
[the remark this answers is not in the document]
jgates108: If `has_chunkId` is false, _chunkID is `-1`, so it will always be consistent.

src/wbase/Task.cc
jgates108: The assert is a holdover from the previous code and is probably a really bad way to go about that as it will kill the worker and cause massive trouble. If `has_db()` is false, is will cause `QueryRunner` problems that should get an error sent back to the czar and maybe give the user a decent error message.

src/wbase/Task.cc
jgates108: It would produce an invalid query, if it ever got that far, and would then trip a user error. The chunk/subchunk numbers tend to show up in the middle of table names  "Object_-1sub" will kick back a SQL error, which is better than "Object_0sub", which could result in invalid rows being added.

I think it's better in this case to generate the SQL error as this should not happen. CHUNK_TAG and SUBCHUNK_TAG should never be found in a query that does not have chunk or subchunk numbers in them.

src/wbase/Task.cc
jgates108: Maybe, the problem is when one `Task` is cancelled thousands of `Task`s tend to get cancelled and it just fills up the log, quite possibly burrying the thing that matters.

src/wbase/Task.h
```diff
@@ -185,7 +185,7 @@ class Task : public util::CommandForThreadPool {
     void setSafeToMoveRunning(bool val) { _safeToMoveRunning = val; }  ///< For testing only.
 
     static IdSet allIds;  // set of all task jobId numbers that are not complete.
-    std::string getIdStr() const { return _idStr; }
+    std::string getIdStr() const { return makeIdStr(); }
```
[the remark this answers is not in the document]
jgates108: I've added a comment saying the returned string is for logging purposes only.

`_idStr` is difficult to set consistently as parts of the value of the query id (which is really important) can't be known until the `TaskMsg` is read. An invalid `TaskMsg` causes all kinds of issues. Using `makeIdStr()` gives the most accurate identifier possible at any given time.

src/wbase/Task.h
```diff
@@ -220,16 +220,41 @@ class Task : public util::CommandForThreadPool {
...
+
+    /// Do not alter the returned value. &&& ??? just return a copy?
+    DbTableSet& getDbTbls() { return _dbTbls; }
+
+    /// Do not alter the returned value. &&& ??? just return a copy?
+    IntVector& getSubchunksVect() { return _subchunksVect; }
+
 private:
+    std::shared_ptr<UserQueryInfo> _userQueryInfo;  ///< Details common to Tasks in this UserQuery.
     std::shared_ptr<SendChannelShared> _sendChannel;
```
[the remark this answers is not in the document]
jgates108: `_sendChannel` needs to be reset to free resources well before `Task` is deleted.

src/wbase/UserQueryInfo.h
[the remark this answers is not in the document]
jgates108: Making it public makes it a lot easier to add new templates.

src/wbase/Task.cc
jgates108: I disagree as an invalid chunkId or subchunkId in the middle of a table or column name should never happen, but if it does '-1' will cause a SQL parse error, where '0' would could go through in some cases. I believe '0' is a valid id.

src/wbase/Task.h
```diff
@@ -227,11 +227,14 @@ class Task : public util::CommandForThreadPool {
     bool getFragmentHasSubchunks() const { return _fragmentHasSubchunks; }
     int getSubchunkId() const { return _subchunkId; }
 
-    /// Do not alter the returned value. &&& ??? just return a copy?
+    /// Do not alter the returned value, returns a set of database tables.
+    /// Note: Copying would be safer, but it could be an expensive thing to copy.
+    ///       This only used by ChunkResourceRequest acquire. The set is usually
+    ///       short, but this is called fairly frequently.
     DbTableSet& getDbTbls() { return _dbTbls; }
```
iagaponenko, line 235: Why return the non-`const` reference? It's highly unsafe as it allows a caller to modify a private data member. You comment "Do not alter" is not a protection here 🐱

src/wbase/Task.h
```diff
@@ -227,11 +227,14 @@ class Task : public util::CommandForThreadPool {
...
 
-    /// Do not alter the returned value. &&& ??? just return a copy?
-    IntVector& getSubchunksVect() { return _subchunksVect; }
+    /// Returns a copy of the list of subchunk ids.
+    IntVector getSubchunksVect() { return _subchunksVect; }
```
iagaponenko, line 238: What's the point in returning a copy vs a `const` reference to the vector? Is there any concern with the life expectancy of the vector stored withing `Task`?

src/wbase/UserQueryInfo.h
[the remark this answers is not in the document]
iagaponenko: My main concern was that `_templates` can be populated in 2 ways - first by calling `addTemplate` which is actually fine to me because the method would extend the collection and return a stable (provided you're not going to introduce operations for deleting elements) index of the added templated within the collection. However, you also have an alternative way of populating `_templates`  via the static method `uqMapInsert` where you're sending in a collection of templates for which you don't have any clear semantic of "template id". I find this dangerous and confusing. Do you really want  that`uqMapInsert` to be present in the API?

src/wbase/UserQueryInfo.cc
iagaponenko: I'm confused since I'm not seeing that include. If I open the file viewer to see the complete file then all I see is:
```
// Class header
#include "wbase/UserQueryInfo.h"

// System headers

// Qserv headers

// LSST headers
#include "lsst/log/Log.h"
```
Are you sure you have committed & pushed everything from your local GIt repo to this PR in GitHub?

src/wbase/Task.cc
iagaponenko: It's better to throw an exception in the ctor of the class `Task` and not wait (potentially for hours) before the query generator will be called when the task would be allowed to run for real. The code that does an early parameter validation in general is much safer than this one.

src/wbase/Task.h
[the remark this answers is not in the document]
iagaponenko: That still leaves a question why not call `makeIdStr()` directly for logging purposes if needed? Why have another "proxy" method `getIdStr()` for that? Would it be bad for some reason?

If the only reason was that there was some existing code relying on `getIdStr()` then it's probably fine to leave that method.

src/wbase/Task.cc
jgates108: The test is already there. Why should I test for the same thing again? 
From `void SsiRequest::execute(XrdSsiRequest& req)` before it tries making Tasks.
```
          if (!taskMsg->has_db() || !taskMsg->has_chunkid() || (ru.db() != taskMsg->db()) ||
                (ru.chunk() != taskMsg->chunkid())) {
                reportError("Mismatched db/chunk in TaskMsg on resource db=" + ru.db() +
                            " chunkId=" + std::to_string(ru.chunk()));
                return;
            }
```

There is no easy test for a valid subchunkId or if a query should even have one, but those checks are being made.

src/wbase/Task.h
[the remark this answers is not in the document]
jgates108: Good point. `makeIdStr()` has the wrong connotation behind it and fewer occurrences, so it's being replaced with `getIdStr()`

src/wbase/Task.h
jgates108: The note basically says it all. I was largely looking for a second opinion. Second opinion says no reference, so make a copy.

src/wbase/Task.h
jgates108: Copying a vector of int is always fairly cheap, so no reason not to be safe.

src/wbase/UserQueryInfo.cc
jgates108: Line 25, on this page `// Qserv headers`
Line 26, on this page, directly below the comment I am typing now, `#include "util/Bug.h"`

src/wbase/UserQueryInfo.h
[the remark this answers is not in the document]
jgates108: Now I am confused. The only way of adding templates to `_templates` is to call `addTemplate`.

`uqMapInsert` adds a new `UserQueryInfo` object to the static `_uqMap` with key `QueryId`. Each `UserQueryInfo` instance has its own `_templates` vector, which needs to be filled by calling `addTemplate`.

src/wbase/Task.h
jgates108: Actually, returning a constant reference for both of these should be fine. The final `acquire` object makes its own copies.

src/wbase/Task.h
iagaponenko: I disagree with the statement that:
```
Copying a vector of int is always fairly cheap, ...
```
It depends on a size of a vector. This requires 1) calling (executing) a copy c-tor, 2) allocating memory for the vector itself, 3) allocating a block of memory where the vector elements will be stored. This operation will be more expensive than returning a reference to the vector (64-bit address). The latter won't require any memory allocation or copy constructor, etc.

As far as my other comment (regarding safety) was concerned then what you say here:
```
...so no reason not to be safe.
```
is not what I was asking for. My question was about if there would be thread safety or object lifecycle concerns causing this method not to return a reference to an internal data member (the vector in this case), especially if this member is mutable. In this case, a copy of the vector would need to be returned under the protection of a lock on a mutex. Something like this:
```
class Task {
public:
    void modifySubchunksVect() {
        std::lock_guard<std::mutex> const lock(_mtx);
        // Now it's safe to modify _subchunksVect
    }
    IntVector getSubchunksVect() {
        std::lock_guard<std::mutex> const lock(_mtx);
        // Now it's safe to call a copy c-tor of the vector to make a copy
        return _subchunksVect;
    }
```
What's shown above is one of two cases when you want to return a copy of a vector rather than a reference to the vector. The second case when you don't want to return the reference would be if the scope (the Task) would get destroyed while you had a reference in the caller's context:
```
Task* task = new Task(...);
Task::IntVector const& vec = task->getSubchunksVect();
delete task;
// Now you can't use the reference to the vector
```

In the end, your options here would (depending on a specific thread-safety or life expectancy requirements) be either of those:
```
    // Absolutely safe for threads and life expectancy.
   / / Very inefficient
    IntVector getSubchunksVect() {
        std::lock_guard<std::mutex> const lock(_mtx);
        // Now it's safe to call a copy c-tor of the vector to make a copy
        return _subchunksVect;
    }

    // Unsafe for threads and life expectancy, or vector mutability.
    // Extremely efficient.
    IntVector const& getSubchunksVect() { eturn _subchunksVect; }
```
What you have in your code is neither safe nor efficient.

src/wbase/Task.cc
iagaponenko: If you don't want to introduce the new attribute `_hasChunkId` and still want to use `-1` then you still need to throw an exception if the states are not comparable. The latter means if either or both objects store `-1` as the chunk identifier. This reinforcement (attribute validation) would lead to a safer code which would allow for detecting issues with chunks at an earlier stage.

src/wbase/Task.cc
iagaponenko: That's good.

iagaponenko (approved):
Looks fine. Please, consider my comments. I would still like to get this PR tested against some large catalog to see how much memory we would safe with this improvement.

src/wbase/Task.h
jgates108: You don't seem to be looking at the most recent commit, which is in the pull request. The current code doesn't match what it is in your comments. The reason it is safe is that both are set only in the constructor, making reading safe without locking at any point thereafter. Defining them as constants would be nice, but a bit tricky to actually get them defined. The only thing to worry about is the caller changing the contents of the data structure. I can add a comment to the definitions (my preference) or add locks and make copies, which are then copied again. Please note, these do get called fairly frequently.


    /// Returns a reference to _dbTbls.
    const DbTableSet& getDbTbls() const { return _dbTbls; }

    /// Return a reference to the list of subchunk ids.
    const IntVector& getSubchunksVect() const { return _subchunksVect; }

src/wbase/Task.cc
jgates108: This is a comparison that above all else, needs to be consistent so items can be sorted. Having it throw an exception would be disaterous to any form of sorting.

What these comparison functions do is sort the items of the lists so they can be compared consistently. 
As long as all tasks without a valid chunk id have a value of -1, the sorting will always be consistent and useful, even though it probably never happens.

src/wbase/Task.cc
```diff
@@ -217,20 +248,16 @@ wpublish::QueryStatistics::Ptr Task::getQueryStats() const {
 }
 
 /// @return the chunkId for this task. If the task has no chunkId, return -1.
-int Task::getChunkId() const {
-    if (msg->has_chunkid()) {
-        return msg->chunkid();
-    }
-    return -1;
-}
+int Task::getChunkId() const { return _chunkId; }
```
jgates108, line 251: Note the original behavior was to return -1 when there was no chunk id.

PR #775, opened 2023-04-08
DM-38621: Restrict director index restrictions to `=` and `IN`

fritzm:
Director index queries currently take place on the query submission thread, and thus can block the single Lua thread in mysql-proxy if not completed expeditiously.

We have seen instances of this occurring "in the wild" on IDF, where queries are sometimes submitted resulting in `objectId >= 0` predicates in the `WHERE` clause, causing millions-of-rows queries to be dispatched to the `Object` director index during query submission, holding up the proxy for minutes at a time.

This ticket is for seat-belt work to confine director index restrictor generation to cases of `=` and `IN` predicates only, which can generally be assumed performant. Other predicate restrictions on director index columns will default through to full scans.

### DIFF
--- a/src/qana/QservRestrictorPlugin.cc
+++ b/src/qana/QservRestrictorPlugin.cc
@@ -337,7 +337,7 @@ std::shared_ptr<query::SecIdxInRestrictor> makeSecondaryIndexRestrictor(query::I
 }
 
 /**
- * @brief Make a Secondary Index comparison restrictor for the given between predicate, if one of the
+ * @brief Make a Secondary Index 'between' restrictor for the given between predicate, if one of the
  *      columns in the predicate is a director column.
  *
  * @param betweenPredicate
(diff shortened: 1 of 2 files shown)


src/qana/QservRestrictorPlugin.cc
```diff
@@ -337,15 +337,15 @@ std::shared_ptr<query::SecIdxInRestrictor> makeSecondaryIndexRestrictor(query::I
...
+ * @brief Make a Secondary Index 'between' restrictor for the given between predicate, if one of the
  *      columns in the predicate is a director column.
  *
  * @param betweenPredicate
  * @param context
  * @return std::shared_ptr<query::SIBetweenRestr> The restrictor that corresponds to the given predicate if
  * one of the columns is a director column, otherwise nullptr.
  */
-std::shared_ptr<query::SecIdxBetweenRestrictor> makeSecondaryIndexRestrictor(
+__attribute__((unused)) std::shared_ptr<query::SecIdxBetweenRestrictor> makeSecondaryIndexRestrictor(
```
iagaponenko, line 348: Since Qserv is officially at C++17 then you might consider a more portable syntax for this:
```
[[maybe_unused]]
```
see: https://en.cppreference.com/w/cpp/language/attributes/maybe_unused

iagaponenko (approved):
LGTM

PR #776, opened 2023-04-08
Add make target for running clang-format

fritzm:
Using the `qserv` CLI to run `clang-format` isn't exactly convenient from inside a long-running build container.  This PR adds a few lines of cmake config so `make clang-format` does the desired thing from within the build container.


src/CMakeLists.txt
```diff
@@ -30,6 +30,11 @@ include_directories(
     ${CMAKE_CURRENT_BINARY_DIR}
 )
 
+file(GLOB_RECURSE CLANG_FILES *.cc *.h)
+add_custom_target(clang-format
+    COMMAND /usr/bin/clang-format -style=file --verbose -i ${CLANG_FILES}
+)
+
```
iagaponenko, line 37: It would be also nice to have another target to check the formatting. Something like this:
```
file(GLOB_RECURSE CLANG_FILES *.cc *.h)
add_custom_target(clang-format-check
    COMMAND /usr/bin/clang-format -style=file --dry-run -Werror --verbose -i ${CLANG_FILES}
)
```
A similar "check" option exists in the command line interface for compiling Qserv:
```
qserv build --clang-format [CHECK|REFORMAT|OFF]
```

iagaponenko (approved):
Looks good to me.

PR #778, opened 2023-04-21
Tickets/dm 38730

jgates108:


src/xrdsvc/SsiRequest.cc
```diff
@@ -151,6 +151,10 @@ void SsiRequest::execute(XrdSsiRequest& req) {
                     wbase::SendChannelShared::create(sendChannelBase, _transmitMgr, taskMsg->czarid());
             auto tasks = wbase::Task::createTasks(taskMsg, sendChannel);
 
+            for (auto const& task : tasks) {
+                _tasks.emplace_back(task);
```
iagaponenko, line 155: Wrong use of `std::vector<T>::emplace_back(...)`. The method is designed for CONSTRUCTING a new element and placing it into a collection. For example, if you had this class:
```
struct Foo {
   Foo(int a, float b, char c);
};
```
Then the following code would be the right use for `emplace_back`:
```
std::vector<Foo> coll;
coll.emplace_back(1, 3.14, 'a');
```
This code is _functionally_ equivalent to the less efficient sequence which _may_ require calling a copy c-tor by `push_back`:
```
Foo foo(1, 3.14, 'a');
std::vector<Foo> coll;
coll.push_back(foo);
```
The bottom line: for shared pointers use `push_back`.

For further details, see: https://en.cppreference.com/w/cpp/container/vector/emplace_back

iagaponenko (approved):
Looks good to me.

PR #780, opened 2023-04-22
DM-38833: update submodules

fritzm:
This just updates to latest `log` and `sphgeom` so we don't fall too far behind.


iagaponenko (approved):
Looks good to me. Though, it's not clear to me how we're tracking versions of the submodules.

PR #783, opened 2023-05-03
Tickets/dm 38909

jgates108:


src/xrdsvc/StreamBuffer.cc
```diff
@@ -154,6 +154,10 @@ void StreamBuffer::cancel() {
     // better a possible leak than corrupted memory or a permanently wedged
     // thread in a limited pool.
     // In any case, this code having an effect should be extremely rare.
+    // FUTURE: It would be nice to eliminate this possible memory leak.
+    //       Possible fix, atomic<bool> _recycleCalled, create thread
+    //       to check if _recycleCalled == true. If true or 24 hours pass
+    //       use `Ptr keepAlive = std::move(_selfKeepAlive);` to kill the object.
     {
```
iagaponenko, line 161: Is this really the "leak" or an extended lifespan of an object? If this is the former then it has to be investigated and fixed right away. If this is the latter then the proposed time-based cleanup may be an option before we have an extended Czar-to-workers communication protocol for sending direct query cancellation/completion signals to workers.

iagaponenko (commented):
I have noticed that objects of the new class`util::HoldTrack` have been placed in many lines of the code. My understanding is that the class was made to investigate the last worker lockup.  And, it seems to me, the tool has played its role. However, I don't think any use of the tool should be left in the production code due to the potential performance implications of using the class. This class has mutex, atomics, and synchronized methods that would be called from multiple threads. Some methods of the class are internally relying on serializing object states into `std::ostringstream`, which by itself is another performance hit. So, If you still want to retain this instrument (the new class) for investigating more race conditions in the future then it would be fine to keep it in the module `util::`. However, all uses of this class in the production code are better to be eliminated or commented out. You will also be able to resume using class later if needed.

src/util/HoldTrack.h
```diff
@@ -0,0 +1,120 @@
...
+
+    HoldTrack() = delete;
+    HoldTrack(HoldTrack const&) = delete;
+    HoldTrack& operator=(HoldTrack const&) = delete;
+
+    ~HoldTrack() = default;
+
+    /// Setup the global instance for storing hold information.
+    /// Items that have existed for longer than `durationLimit` will be logged.
+    static void setup(double durationLimitSeconds);
```
iagaponenko, line 64: This method should use the standard data type:
```
std::chrono::milliseconds
```
If you still like using `double` for some reason within the implementation then this is fine.

src/util/HoldTrack.h
```diff
@@ -0,0 +1,120 @@
...
+            }
+        }
+
+    private:
+        bool _inserted = false;  ///< set to true when a mark is inserted into the set.
+        KeyType _key;            ///< that matches this Mark in the set.
+    };
+
+private:
+    HoldTrack(double durationLimitSeconds) : _durationLimitMillisec(durationLimitSeconds * 1000.0) {
```
iagaponenko, line 96: Should use:
```
std::chrono::milliseconds
```

src/util/HoldTrack.h
```diff
@@ -0,0 +1,120 @@
...
+        _enabled = true;
```
iagaponenko, line 97: This doesn't seem right to me to modify the static member from the c-tor of the instance

iagaponenko (commented):
More comments on the class `util::HoldTrack`. It has a very unusual and counterintuitive design. I propose changing it to have the following interface:
```
class HoldTrack {
public:
    static std::shared_ptr<HoldTrack> create(std::chrono::milliseconds);
    static std::shared_ptr<HoldTrack> instance();
   void addKey(KeyType const& key, Issue::Context const& ctx, string const& note);
   void removeKey();
   ...
private:
    static std::shared_ptr<HoldTrack> _ptr;
};
```
And move the internal class `Mark` into the namespace instead of having it the inner class:
```
class MarkHoldTrack {
public:
   ...
};
```
That class `MarkHoldTrack` would be using the public API of the  singleton of the class `HoldTrack` by calling:
```
auto ptr = HoldTrack::instance();
if (ptr == nullptr) { /* then no tracking */ }
```

src/util/HoldTrack.h
```diff
@@ -0,0 +1,120 @@
...
+    /// Add `key` to _keyMap with ctx prepended to note.
+    void _addKey(KeyType const& key, Issue::Context const& ctx, std::string const& note);
+
+    /// Remove `key` from _keyMap.
+    void _removeKey(KeyType const& key);
+
+    /// Return the next sequence number.
+    static uint64_t _getSeq() { return _seq++; }
+
+    static std::atomic<bool> _enabled;  ///< Set to true to enable tracking.
```
iagaponenko, line 109: This attribute's initialization and use are not consistent. It's initialized to be `false` during static initialization of the process. It's set to be `true` when an instance of the "global" class `HoldTrack` gets created and this private variable is used outside of the class `HoldTrack` by another class `Mark` (even though, it's "internal").

src/util/HoldTrack.h
jgates108: It's meant to be used as way to turn the feature on and off efficiently at run time, but there's no way of doing that at this time. I'll probably just remove _enabled for now.

src/xrdsvc/StreamBuffer.cc
jgates108: There's an ugly race condition where it's difficult to tell if xrootd is going to call Recycle() or not. I've been thinking about it. Waiting a very long time would work, but it's ugly. I'm also thinking that when the SsiRequest is getting deleted, it's probably safe to say that xrootd isn't going to be calling Recycle(), but I'm not 100% sure about that.

jgates108:
There are a couple of instances where HoldTrack::Mark have been left in. SsiRequest::Finish, as that 's is a short call and something bad has almost certainly happened if that shows up in the log. The efficiency is about 2*log(n) so having a small number of them should have negligable impact.

src/xrdsvc/StreamBuffer.cc
iagaponenko: I have filed the JIRA ticket for this race condition: https://jira.lsstcorp.org/browse/DM-39041

PR #784, opened 2023-05-08
Removed transmitCancelError.

jgates108:


iagaponenko (approved):
The code looks good to me. Though, I must admit I don't fully understand how the cancellation logic works. It's quite complex, and it involves many different classes. The cancellation sequence crosses many synchronization domains (classes, methods guarded mutexes, atomics). I have an impression this code requires serious refactoring and possibly reimplementation.

PR #789, opened 2023-05-27
Tickets/dm 39425

iagaponenko:


src/wpublish/QueriesAndChunks.h
```diff
@@ -50,14 +53,19 @@ class QueryStatistics {
...
+    /// The dictionary type for all tasks for this user query keyed by:
+    ///   [jobId][fragmentNum][templateId][subChunkId]
+    using JobIdKeyType = int;
+    using FragmentNumKeyType = int;
+    using TemplateIdKeyType = size_t;
+    using SubChunkIdKeyType = int;
+    using TaskMap =
+            std::map<JobIdKeyType,
+                     std::map<FragmentNumKeyType,
+                              std::map<TemplateIdKeyType, std::map<SubChunkIdKeyType, wbase::Task::Ptr>>>>;
```
jgates108, line 65: I'm not sure why you're going with a multilevel map instead of a std::pair or std::tuple key. The performance is going to be roughly the same for finding a unique element and accessing all the elements is a simple `for (auto&& taskElement : taskMap) { ... }`. This is probably better if stats for all subchunks in a single chunk are needed.
Also, I'd probaly put subchunk in front of templateId. Currently, I think it's a maximum of 2 templates per subchunk and the template queries tend to be very similar. Separating everything out by subchunkId is probably more useful than templateId.

jgates108 (approved):
Looks good, I'm just curious about one thing.

PR #791, opened 2023-06-05
Update qserv-ingest documentation

fjammes:


doc/admin/qserv-ingest/ingest.yaml
```diff
@@ -0,0 +1,69 @@
+version: 15
+ingest:
+    http:
+        # Optional
+        # Timeout for POST and PUT queries in seconds
+        # POST and PUT queries which might be very long like:
```
iagaponenko, line 6:
```
...might be very long for the following operations:
```

doc/admin/qserv-ingest/ingest.yaml
```diff
@@ -0,0 +1,69 @@
...
+        # - postint contribution
+        # - closing transaction
+        # - publishing database
+        # have no timeout at all
```
iagaponenko, line 10: After that, I would also add 2 more long-running operations:
```
# - building the "director" index
# - building table indexes at workers
```

doc/admin/qserv-ingest/ingest.yaml
```diff
@@ -0,0 +1,69 @@
+version: 15
+ingest:
+    http:
+        # Optional
```
iagaponenko, line 4: If you say `Optional' then you may also mention what is the value of the default timeout, if it's some pre-defined long value, or no timeout (waiting indefinitely).

Alternatively, you may want to explain this in some RST file instead of this YAML file.

doc/admin/qserv-ingest/ingest.yaml
```diff
@@ -0,0 +1,69 @@
...
+      # Optional
+      # Allow to customize metadata URL
+      # Default to "{servers[0]}{path}"
+      url: http://dataserver/datasets/DC2/
+
+    input:
+        # List of http servers providing input dataset
+        # The ingest process will load-balance the download of input files accross
+        # these servers.
+        # Use file:// as first element in list when using local data
```
iagaponenko, line 25: Would using `file://` screw your previous instructions for:
```
metadata:
      # Optional
      # Allow to customize metadata URL
      # Default to "{servers[0]}{path}"
```
In this case, the metadata operations would be attempted via the file service. Or, it still would work fine? Just curious.

iagaponenko (approved):
Looks good to me. Please, see my comments and questions.

doc/admin/k8s.rst
```diff
@@ -1,5 +1,25 @@
...
+
+Access to Qserv dashboard
+-------------------------
+
+This access is only available for users having access to the cluster with ``kubectl``.
+
+.. code:: bash
+
+    # Open a tunnel between local machine and Qserv dashboard
+    kubectl port-forward qserv-repl-ctl-0 8080 &
```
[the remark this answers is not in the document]
fjammes: :+1:

doc/admin/qserv-ingest/ingest.yaml
fjammes: This path is for retrieving data, the metadata option support the situation where the metadata file is in an other place than the data. It can be useful to test new metadata file version with existing data.

`file://` was implemented by @hsinfang and I did not test or use this feature. I do not know if it still works fine.

PR #793, opened 2023-06-20
Tickets/dm 39704

iagaponenko:


src/replica/DirectorIndexJob.cc
```diff
@@ -388,18 +386,19 @@ void DirectorIndexJob::_loadDataIntoTable() {
         if (request == nullptr) break;
 
         // Load request's data into the destination table.
+        bool const localFile = true;
         try {
             string const query = g.loadDataInfile(
                     request->responseData().fileName, directorIndexTableName(database(), directorTable()),
                     controller()->serviceProvider()->config()->get<string>("worker", "ingest-charset-name"),
-                    localFile());
+                    localFile);
```
jgates108, line 394: Would it be better to remove the parameter from `loadDataInfile`?

src/replica/DirectorIndexJob.cc
```diff
@@ -388,18 +386,19 @@ void DirectorIndexJob::_loadDataIntoTable() {
...
             h.conn->executeInOwnTransaction([&](auto conn) {
                 conn->execute(query);
                 // Loading operations based on this mechanism won't result in throwing exceptions in
                 // case of certain types of problems encountered during the loading, such as
                 // out-of-range data, duplicate keys, etc. These errors are reported as warnings
                 // which need to be retrieved using a special call to the database API.
-                if (localFile()) {
+                if (localFile) {
```
jgates108, line 401: Would it be better to remove the `if` statement?

jgates108 (approved):
Just a couple of very minor comments, looks good.

src/replica/DirectorIndexJob.cc
iagaponenko: This can't be done as this option is still used by the worker ingest service for loading chunk contributions. In this case, both worker Ingest service and worker MySQL share the folder as they are guaranteed to be run within the same Kubernetes pod (and on the same host in the Docker-based environment). The `local=0` mode provides a noticeable speedup in the performance of the bulk ingest operations. We just can't do this for constructing the "director" index.

src/replica/DirectorIndexJob.cc
iagaponenko: I too thought about removing the if. I will consider it.

PR #794, opened 2023-06-24
Reinforced the implementation to check if the desired folder exists

iagaponenko:
And if it exists then also test if the path is a valid folder before attempting to create the one. The unit test was improved to track this scenario as well.


fritzm (approved):
LGTM -- thanks!

PR #795, opened 2023-07-06
Added symbolic definitions for known HTTP status codes in qhttp

iagaponenko:
Also migrated clients of the module to use these definitions instead of numeric literals ("magic" constants).


fritzm (approved):
LGTM -- thanks!

PR #800, opened 2023-07-19
Fixed a bug in qhttp response handler

iagaponenko:
The older version of the code was throwing an exception when creating HTTP headers for file sizes exceeding signed 32-bit representation (~2 GB).


fritzm (approved):
LGTM -- thanks!

PR #801, opened 2023-07-20
Tickets/dm 39819

iagaponenko:


src/ccontrol/MergingHandler.h
```diff
@@ -100,6 +100,7 @@ class MergingHandler : public qdisp::ResponseHandler {
     void _setError(int code, std::string const& msg);    ///< Set error code and string
     bool _setResult(BufPtr const& bufPtr, int blen);     ///< Extract the result from the protobuffer.
     bool _verifyResult(BufPtr const& bufPtr, int blen);  ///< Check the result against hash in the header.
+    bool _noErrorsInResult();  ///< Check if the rsult message object has no errors, report the ones (if any).
```
jgates108, line 103: rsult -> result

src/czar/CzarConfig.h
```diff
@@ -44,15 +49,33 @@ namespace lsst::qserv::czar {
...
+     */
+    static std::shared_ptr<CzarConfig> create(std::string const& configFileName);
+
+    /**
+     * Get a pointer to an instance that was created by the last call to
+     * the method 'create'.
+     * @return the shared pointer to the configuration object
+     * @throws std::logic_error when attempting to call the bethod before creating an instance.
+     */
+    static std::shared_ptr<CzarConfig> instance();
```
jgates108, line 69: I think having `create()` get called more than once should probably be an error. Having 2 different instances of the configuration that do not match could be extremely confusing.

src/mysql/MySqlConnection.cc
```diff
@@ -155,7 +155,7 @@ bool MySqlConnection::queryUnbuffered(std::string const& query) {
 int MySqlConnection::cancel() {
     std::lock_guard<std::mutex> lock(_interruptMutex);
     int rc;
-    if (!_isExecuting || _interrupted) {
+    if (_interrupted) {
```
jgates108, line 158: Is this the bug with query cancelation you found?

jgates108 (approved):

src/mysql/MySqlConnection.cc
iagaponenko: Formally, I wouldn't really call this a bug. The code was designed on the assumption that the method could be called only for queries that are still in the "executing" state before the blocking MySQL's function `mysql_real_query` returns to a caller the query completion status. When `mysql_real_query` is finished, MySQL has a result set (or an iterator of such result set) that still needs to be extracted. And before the result retrieval is finished, MySQL has to maintain the state of the query and keep the relevant (substantial) temporary resources that are needed for that. We also know that (depending on the size of a result set and the back pressure from Qserv), the result extraction phase may be (and indeed, it is) much longer than the query execution phase. This is especially true when Qserv worker is streaming results back to `Czar` over the SSI protocol. And things are getting significantly worse if the result set is (very) large. So, my fix solves this problem by terminating queries at both the "execution" and  "result extraction" phases. In the latter case `wbd::QueryRunner` would just fail to retrieve the next batch of rows from the connection. This isn't a problem since the task and the MySQL connection would be already marked as canceled.

src/czar/CzarConfig.h
iagaponenko: Good point! I will fix that.

PR #804, opened 2023-08-24
DM-40506: Synchronization bug in the QMeta query runner class of Qserv

iagaponenko:
Minor refactoring in the class.


jgates108 (approved):
Looks good.

PR #807, opened 2023-09-07
Make director index column match case-insensistive

fritzm:
This ~1-liner was sitting around on my disk.  Figure we might as well get it reviewed and in...


iagaponenko (approved):
LGTM

PR #810, opened 2023-09-14
Tickets/dm 40787

iagaponenko:


src/wbase/Task.cc
```diff
@@ -156,13 +144,13 @@ Task::Task(TaskMsgPtr const& t, int fragmentNumber, std::shared_ptr<UserQueryInf
     if (resultDeliveryProtocol != wconfig::WorkerConfig::ResultDeliveryProtocol::SSI) {
         _resultFilePath = ::buildResultFilePath(t, workerConfig->resultsDirname());
         if (resultDeliveryProtocol == wconfig::WorkerConfig::ResultDeliveryProtocol::XROOT) {
-            // NOTE: one extra '/' after the <server>[:<port>] spec is required to make
+            // NOTE: one extra '/' after the <vnid>[:<port>] spec is required to make
             // a "valid" XROOTD url.
-            _resultFileXrootUrl = "xroot://" + ::get_hostname() + ":" +
-                                  to_string(workerConfig->resultsXrootdPort()) + "/" + _resultFilePath;
+            _resultFileXrootUrl = "xroot://" + vnid + ":" + to_string(workerConfig->resultsXrootdPort()) +
```
jgates108, line 149: There should probably be a comment explaining what a vnid is and why it's being used instead of the hostname.

jgates108 (approved):
Looks good.

src/wbase/Task.cc
iagaponenko: Good point!

iagaponenko: For the record, the `vnid` track has turned out to be absolutely wrong. Upon further testing, the file-based protocol is based on establishing direct TCP connections from `Czar` to workers using the host/port of the latter. `vnid` is not used in this communication path. So, I modified this commit and replaced `vnid` with the worker's `FQDN`.

PR #812, opened 2023-09-23
DM-40891: Migrate Qserv to almalinux 9

iagaponenko:


admin/tools/docker/base/Dockerfile
```diff
@@ -121,9 +119,9 @@ RUN cd /tmp \
     && rm -rf mysqlproxy
 
 RUN cd /tmp \
-    && git clone https://github.com/lsst/xrootd.git \
+    && git clone https://github.com/xrootd/xrootd.git \
     && cd xrootd \
-    && git checkout ssi-5.3.x \
+    && git checkout master \
```
fritzm, line 124: We should use a released tag of XRootD here, and not lastest `master`, otherwise this will float on each build.  Check with Andy H. on which version should be used?

admin/tools/docker/base/Dockerfile
```diff
@@ -202,23 +200,22 @@ RUN dnf install -y 'dnf-command(config-manager)' \
         openssl \
         procps-ng \
         protobuf \
-        python38 \
+        python3 \
+        pip \
         tree \
         vim \
-    && dnf module install -y mariadb/client \
+    && dnf install -y mariadb \
```
fritzm, line 207: Please check that this is client-only, to make sure we don't bloat the containers?

admin/tools/docker/base/Dockerfile
```diff
@@ -50,22 +50,20 @@ RUN dnf install -y 'dnf-command(config-manager)' \
...
 RUN dnf update -y \
     && dnf install -y \
        https://apache.jfrog.io/artifactory/arrow/almalinux/$(cut -d: -f5 /etc/system-release-cpe | cut -d. -f1)/apache-arrow-release-latest.rpm \
     && dnf config-manager --set-enabled epel \
-    && dnf config-manager --set-enabled powertools \
+    && dnf config-manager --set-enabled crb \
     && dnf install -y \
-      arrow-devel-$ARROW_VERSION \
-      parquet-devel-$ARROW_VERSION \
+      arrow-devel \
```
fritzm, line 65: You explicitly call out versions 13 below -- should these be explicit to match?

fritzm (approved):
Thanks, Igor!!  A few comments left.  Please consider change to specify XRootD version explicitly a hard requirement.

admin/tools/docker/base/Dockerfile
iagaponenko: The older version of the run container had these lines where the version was mentioned twice (in different form):
```
ENV ARROW_VERSION 11.0.0

dnf install -y arrow11-libs-$ARROW_VERSION \
dnf install -y parquet11-libs-$ARROW_VERSION \
```
And the build container in the older version had:
```
dnf install -y arrow-devel-$ARROW_VERSION
```
You may see the "assymetry". I have no clue why developers chose this naming convention. The only theory I have is that the packages are named after the major version, while leaving some degree of freedom for the middle an dminor version numbers.

AlmaLinux 9 comes with version 13.0.0 of the tools. So, we don't need to mention it in the build container. The run container still requires:
```
dnf install -y arrow13-libs
dnf install -y parquet13-libs
```

admin/tools/docker/base/Dockerfile
iagaponenko: Tou're right - `mariadb` would install the server as well. It looks like this is a super-package that includes `mariadb-server` and client tools. I will further investigate this. In theory, a special client-only package `mysql` could be used. 

I just couldn't get this to work in AlmaLinux 9:
```
dnf module install -y mariadb/client
```

admin/tools/docker/base/Dockerfile
iagaponenko: Migrated to the stable release tag `v5.6.2`. Verified that this release includes the desired functionality.

admin/tools/docker/base/Dockerfile
iagaponenko: For the record. Replaced `mariadb` with the client package `mysql`. Verified that Qserv works after the change.

fritzm (approved):
LGTM, thanks!

PR #813, opened 2023-10-19
Tickets/dm 40162

iagaponenko:


src/replica/HttpQservMonitorModule.h
```diff
@@ -54,11 +54,13 @@ class HttpQservMonitorModule : public HttpModule {
...
      * Supported values for parameter 'subModuleName':
      *
-     *   WORKERS    - get the status info of many workers
-     *   WORKER     - get the status info of a specific worker
-     *   WORKER-DB  - get the database status of a specific worker
-     *   CZAR       - get the status info of Czar
-     *   CZAR-DB    - get the database status of Czar
+     *   WORKERS       - get the status info of many workers
+     *   WORKER        - get the status info of a specific worker
+     *   WORKER-CONFIG - get onfiguration parameters of a specific worker
```
jgates108, line 59: onfiguration -> configuration

src/replica/HttpQservMonitorModule.h
```diff
@@ -54,11 +54,13 @@ class HttpQservMonitorModule : public HttpModule {
...
+     *   WORKER-DB     - get the database status of a specific worker
+     *   CZAR        - get the status info of Czar
+     *   CZAR-CONFIG - get configuration parameters of Czar
+     *   CZAR-DB     - get the database status of Czar
```
jgates108, line 63: Maybe add a space so czar descriptions line up with worker descriptions.

src/replica/QservMgtServices.cc
```diff
@@ -346,6 +346,40 @@ GetDbStatusQservMgtRequest::Ptr QservMgtServices::databaseStatus(
...
 }
 
+GetConfigQservMgtRequest::Ptr QservMgtServices::config(std::string const& worker, std::string const& jobId,
+                                                       GetConfigQservMgtRequest::CallbackType const& onFinish,
+                                                       unsigned int requestExpirationIvalSec) {
+    GetConfigQservMgtRequest::Ptr request;
+
+    // Make sure the XROOTD/SSI service is available before attempting
+    // any operations on requests
+
```
jgates108, line 356: I don't think this blank line helps anything.

src/util/ConfigStore.cc
```diff
@@ -113,25 +112,44 @@ int ConfigStore::getIntRequired(std::string const& key) const {
...
+    // Copy all parameters matching "<sectionName>.<param>" to new map (dropping "<sectionName>.")
+    std::string const section = sectionName + ".";
     std::map<std::string, std::string> sectionConfigMap;
     int len = section.length();
-    for (auto const& kv : _configMap) {
-        if (kv.first.compare(0, len, section) == 0) {
-            sectionConfigMap.insert(std::make_pair(std::string(kv.first, 4), kv.second));
+    for (auto&& [key, val] : _configMap) {
+        if (key.substr(0, len) == section) {
+            sectionConfigMap[key.substr(len)] = val;
```
jgates108, line 132: I like the change, but I think `auto const&` would be better than `auto&&` in this case, as `auto&&` could allow `val` to be altered.

src/xrdreq/GetConfigQservRequest.cc
```diff
@@ -0,0 +1,93 @@
+/*
+ * LSST Data Management System
+ * Copyright 2018 LSST Corporation.
```
jgates108, line 3: Need to use the right header message.

jgates108 (approved):

src/xrdreq/GetConfigQservRequest.cc
iagaponenko: Good suggestion! Many headers are obsolete.

src/util/ConfigStore.cc
iagaponenko: Well, `auto&&` in this context means "forwarding" of whatever types are returned by the range iterator. This is just the "lazy" way of writing `auto const&`. Neither `key` nor `val` could be modified in this context because a context (a method) where this iteration is happening is `const`.

src/replica/QservMgtServices.cc
iagaponenko: Good point!

src/replica/HttpQservMonitorModule.h
iagaponenko: Good suggestion!

src/replica/GetConfigQservMgtRequest.h
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * @return The info object returned back by the worker.
+     * @note The method will throw exception std::logic_error if called before
+     *   the request finishes or if it's finished with any status but SUCCESS.
+     */
+    nlohmann::json const& info() const;
+
+    /// @see QservMgtRequest::extendedPersistentState()
+    std::list<std::pair<std::string, std::string>> extendedPersistentState() const override;
```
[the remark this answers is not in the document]
iagaponenko: Good point. I could define this as:
```
virtual std::list<std::pair<std::string, std::string>> extendedPersistentState() const = default;
```

src/util/ConfigStore.cc
jgates108: No, auto&& preserves constantness. If it's a constant going in, it stays a constant. If it isn't a constant going in, then it isn't constant.
```
    std::map<string, int> numMap;
    numMap.insert(pair("one",1));
    numMap.insert(pair("two", 2));
    for (auto&& [key, val] : numMap) {
        val = 1234;
    }
```

iagaponenko: Your example is incorrect because the iterable collection isn't `const`. If you rewrite your code like this then it won't compile:
```
std::map<string, int> const numMap = {{"one",1}, {"two", 2}};
for (auto&& [key, val] : numMap) {
    val = 1234;   // <- compilation ERROR
}
```
In the above-shown scenario, both `key` and `val` look like this:
```
std::string const& key = ...
int const& val = ...
```
Back to the original comment. The simplified version of my code in this PR looks like this:
```
ConfigStore {
public:
    std::map<std::string, std::string> getSectionConfigMap(std::string sectionName) const {
        for (auto&& [key, val] : _configMap) {
            ...
        }
    }
private:
    std::map<std::string, std::string> const _configMap;
};
```
With such a definition, `_configMap` (as well as `key` and `val`) become immutable for two reasons:
- because of the `const` attribute of the method `getSectionConfigMap`
- and because the data member by itself is `const`
- Therefore type forwarding made by `auto&&` does the right thing (forwards both to be `std::string const&`).

PR #814, opened 2023-10-24
Tickets/dm 29447

iagaponenko:


src/replica/HttpIngestModule.cc
```diff
@@ -97,6 +98,23 @@ string jobCompletionErrorIfAny(SqlJob::Ptr const& job, string const& prefix) {
...
+ * Check if the provided type is not prohibited for using in the 'director' tables.
+ * @param func A context for error reporting.
+ * @param colName The name of a column (used for error reporting).
+ * @param colType The column type definition to be evaluated.
+ * @throw lsst::qserv::replica::HttpError if the type validation failed.
+ */
+void validateColumnType(string const& func, string const& colName, string const& colType) {
+    string const colTypeUpperCase = boost::algorithm::to_upper_copy(colType);
+    for (char const* prohibitedType : {"BLOB", "TEXT"}) {
+        if (colTypeUpperCase.find(prohibitedType) == string::npos) continue;
```
jgates108, line 111: I prefer `!= string::npos` instead of using continue, but it doesn't make much difference.

jgates108 (approved):
Looks good, just one really nitpicky comment.

PR #815, opened 2023-10-26
DM-41358: Fixed a bug in the task state transition

iagaponenko:
The change only affects the file-based result delivery protocol.


jgates108:
Looks good.

PR #816, opened 2023-10-27
DM-41477: Fixed a bug in the state transition logic of the worker tasks

iagaponenko:
In the previous code, the task were declared as "finished" after being "booted" from a scheduler. In reality, the tasks could be still in the active state processing the corresponding queries and consuming resources (memory, CPU, MySQL connections). This change made the task monitoring correct and allowed to uncover tasks that were previously seen as "invisible".

### DIFF
--- a/src/wdb/QueryRunner.cc
+++ b/src/wdb/QueryRunner.cc
@@ -35,6 +35,7 @@
 #include <cstddef>
 #include <iostream>
 #include <memory>
+#include <thread>
 
 // Third-party headers
 #include <google/protobuf/arena.h>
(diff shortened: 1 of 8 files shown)


src/wdb/QueryRunner.cc
```diff
@@ -333,6 +337,12 @@ bool QueryRunner::_dispatchChannel() {
             // metadata store of the worker, or keeping the state transisitons in a separate transient
             // store that won't be affected by the task destruction.
             _task->finished(std::chrono::system_clock::now());
+#endif
+            // DM-41358: Unlike the above-made call to _task->finished() that was disabled,
+            //           this one does the complete cleanup of metadata and stats associated
+            //           with the task completion. In the previous version of the code this
+            //           operation was initiated by BlendScheduler::commandFinish.
+            _queriesAndChunks->finishedTask(_task);
```
jgates108, line 345: This should probably go in the `class Release` defined in `runQuery(...)` so it's called even if there's an exception.
Also, `startedTask()` should probably be moved out of `BlendScheduler::commandStart()` and put just before `class Release` is instantiated so that it's very unlikely one gets called without the other.

jgates108 (approved):
Looks good.

PR #818, opened 2023-11-08
Web Dashboard: Fixed a bug on the query progress display

iagaponenko:
Now displaying the progess in the UTC timezone


fritzm (approved):
LGTM

PR #819, opened 2023-11-10
DM-41685: Update run base to arrow and parquet libs to version 14

fritzm:
This seems needed since the "default" versions picked up in the build container have now advanced to version 14.


iagaponenko (approved):
LGTM

PR #825, opened 2023-12-21
DM-42221: Pushed the Replication/Ingest system's classes into the dedicated subfolders

iagaponenko:
…folders

The classes are now clustered within subfolders based on class's role and functionality. CMake files were added to the relevant subfolders. Include headers were rearranged in alphabetical order.


fritzm (approved):
LGTM

PR #826, opened 2024-01-13
Fixed a bug in the Czar management protocol

iagaponenko:


jgates108 (approved):
Looks good. I'm not thrilled by the way responseCode() operates as this looks like a very easy mistake to make.

PR #827, opened 2024-01-19
DM-42562: Improved version of the response buffer management in qhttp

iagaponenko:
When sending the payload of the static files, the file content is split into records, where the maximum size of each record is limited by 2MB.

### DIFF
--- a/src/qhttp/Response.cc
+++ b/src/qhttp/Response.cc
@@ -24,6 +24,8 @@
 #include "qhttp/Response.h"
 
 // System headers
+#include <algorithm>
+#include <errno.h>
 #include <map>
 #include <memory>
 #include <string>
@@ -35,6 +37,7 @@
 #include "boost/asio.hpp"
 #include "boost/filesystem.hpp"
 #include "boost/filesystem/fstream.hpp"
+#include "boost/system/error_code.hpp"
 
 // Local headers
 #include "lsst/log/Log.h"
@@ -120,8 +123,11 @@ std::unordered_map<std::string, const std::string> contentTypesByExtension = {
 namespace lsst::qserv::qhttp {
 
 Response::Response(std::shared_ptr<Server> const server, std::shared_ptr<ip::tcp::socket> const socket,
-                   DoneCallback const& doneCallback)
-        : _server(std::move(server)), _socket(std::move(socket)), _doneCallback(doneCallback) {
+                   DoneCallback const& doneCallback, std::size_t const maxResponseBufSize)
+        : _server(std::move(server)),
+          _socket(std::move(socket)),
+          _doneCallback(doneCallback),
+          _maxResponseBufSize(maxResponseBufSize) {
     _transmissionStarted.clear();
 }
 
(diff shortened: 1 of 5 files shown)


src/qhttp/Response.cc
```diff
@@ -186,23 +209,49 @@ std::string Response::_headers() const {
...
                                   << logger(_socket) << "handler logic error: multiple responses sent");
-        return;
     }
+}
+
+void Response::_finishTransmit(boost::system::error_code const& ec, std::size_t sent) {
+    if (ec) {
+        LOGLS_ERROR(_log, logger(_server) << logger(_socket) << "write failed: " << ec.message());
+    }
+    _responseBuf.clear();
```
fritzm, line 223: Would an error/info log if we haven't sent the full file (e.g. `_bytesToRead != 0`) be a good idea?

src/qhttp/Response.cc
```diff
@@ -142,28 +147,46 @@ void Response::send(std::string const& content, std::string const& contentType)
...
+    _responseBuf = _headers() + "\r\n" + content;
+    _startTransmit();
+    asio::async_write(*_socket, asio::buffer(_responseBuf.data(), _responseBuf.size()),
+                      [self = shared_from_this()](boost::system::error_code const& ec, std::size_t sent) {
+                          self->_finishTransmit(ec, sent);
+                      });
 }
 
 void Response::sendFile(fs::path const& path) {
+    _bytesToRead = fs::file_size(path);
```
fritzm, line 159: `_bytesRemaining` might be slightly better name?  Or maybe `_bytesToRead` as the full file size (never changes) and `_bytesRemaining` to track how much remains to be sent?

fritzm (approved):
Looks good -- a couple minor and optional suggestions in comments.

src/qhttp/Response.cc
iagaponenko: Good point. Will fix it.

src/qhttp/Response.cc
iagaponenko: The method `_finishTransmit` gets called from 4 locations when `_bytesToRead == 0` or when `ec` is not a success. So, the extra check inside the method is not needed.

iagaponenko: These are invocations of the method mentioned above:
```
void Response::send(std::string const& content, std::string const& contentType) {
    ...
    asio::async_write(*_socket, asio::buffer(_responseBuf.data(), _responseBuf.size()),
                      [self = shared_from_this()](boost::system::error_code const& ec, std::size_t sent) {
                          self->_finishTransmit(ec, sent);
                      });
}
```
In that case `_bytesToRead` is always set to `0` in the constructor.

The following two cases are when something bad happens:
```
auto const ec = boost::system::system_error(errno, boost::system::generic_category());
_finishTransmit(ec.code(), _bytesSent);
```
```
if (ec) {
    self->_finishTransmit(ec, self->_bytesSent);
```

The last case is when the transfer is over:
```
if (self->_bytesRemaining == 0) {
    auto const ec = boost::system::errc::make_error_code(boost::system::errc::success);
    self->_finishTransmit(ec, self->_bytesSent);

```

fritzm (approved):

PR #830, opened 2024-02-02
DM-42770: Added "relaxed" memory management mode at Qserv worker

iagaponenko:
The memory management node is a variation of MemManNone except it won't report a failure to (memory) lock tables where such locking is required. With this mode, the scheduler would not enforce additional restrictions on the number of queries that are processed in parallel. In case of using the non-relaxed manager MemManNone the processing concurrency is forced to be 1 per queue regardles of what's specified in the worker's configuration or the number of the hardware threads on a machine.

### DIFF
--- a/src/xrdsvc/SsiService.cc
+++ b/src/xrdsvc/SsiService.cc
@@ -168,6 +168,9 @@ SsiService::SsiService(XrdSsiLogger* log) {
                 memman::MemMan::create(memManSize, workerConfig->getMemManLocation()));
     } else if (cfgMemMan == "MemManNone") {
         memMan = make_shared<memman::MemManNone>(1, false);
+    } else if (cfgMemMan == "MemManNoneRelaxed") {
+        bool const alwaysLock = true;
+        memMan = make_shared<memman::MemManNone>(1, alwaysLock);
     } else {
         LOGS(_log, LOG_LVL_ERROR, "Unrecognized memory manager " << cfgMemMan);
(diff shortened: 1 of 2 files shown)


src/xrdsvc/SsiService.cc
```diff
@@ -168,6 +168,9 @@ SsiService::SsiService(XrdSsiLogger* log) {
                 memman::MemMan::create(memManSize, workerConfig->getMemManLocation()));
     } else if (cfgMemMan == "MemManNone") {
         memMan = make_shared<memman::MemManNone>(1, false);
+    } else if (cfgMemMan == "MemManNoneRelaxed") {
+        bool const alwaysLock = true;
+        memMan = make_shared<memman::MemManNone>(1, alwaysLock);
```
jgates108, line 173: The naming here seems counter intuitive. I'm guessing alwaysLock = true, means consider it locked requardless of if it could be locked? I think some comments explaining what is going on with the different modes will be helpful.

jgates108 (approved):
Looks good, but I think some comments would be helpful.

src/xrdsvc/SsiService.cc
iagaponenko: Completely agree with you. However, the variable was named to match the name of the corresponding formal parameter of the constructor (see class `memman::MemManNone`). This parameter and the class existed before. So, I decided not to change anything.

PR #832, opened 2024-02-07
Added code to track booted queries.

jgates108:


src/wpublish/QueryStatistics.h
```diff
@@ -0,0 +1,173 @@
...
+    int _tasksCompleted = 0;
+    int _tasksRunning = 0;
+    int _tasksBooted = 0;                   ///< Number of Tasks booted for being too slow.
+    std::atomic<bool> _queryBooted{false};  ///< True when the entire query booted.
+
+    double _totalTimeMinutes = 0.0;
+
+    std::vector<wbase::Task::Ptr> _tasks;  ///< A collection of all tasks of the query
+
+    util::Histogram::Ptr _histTimeRunningPerTask;  ///< Histogram of SQL query run times.
```
iagaponenko, line 160: A few comments on the histogram data members. Firstly, I'm not seeing any _include_ statements in this header for those. Since you depend on the typedef for `::Ptr` then I would expect something like:
```
#include "util/Histogram.h"
```
A reason why this code might still compile could be purely accidental - because the header for `wpublish/QservStatistics.h` is used in some context where the `Histogram` is already included. We can't rely on this. Normally, all data types mentioned in a particular C++ header must be fully declared (if used by value, etc.) or forward-declared (if there are pointers or references to that type).

My second comment (suggestion) is directly related to the last statement. The `::Ptr` is "evil" in this case as it forces you to do the full include of the relevant header:
```
#include "util/Histogram.h"

```
Where you could simply do the light-weight forward declaration:
```
namespace lsst::qserv::util {
class Histogram;
}
```
And then do:
```
std::shared_ptr<util::Histogram> _histTimeRunningPerTask;
```
The lightweight forward declarations are helpful as they are known to speed up the compilation of a project.

It's okay to use `::Ptr` in the implementation (`*.cc`) files where the corresponding type is used.

src/wpublish/QueriesAndChunks.cc
```diff
@@ -240,6 +181,9 @@ void QueriesAndChunks::finishedTask(wbase::Task::Ptr const& task) {
             (*_newlyDeadQueries)[qId] = stats;
         }
     }
+    if (task->isBooted()) {
+        _bootedTaskTracker.removeTask(task);
```
iagaponenko, line 185: This is a bit confusing to me. If the task is "booted" then why should it be removed from the tracker? I would expect the opposite.

src/wbase/Task.h
```diff
@@ -295,6 +295,12 @@ class Task : public util::CommandForThreadPool {
     /// system in order to see what MySQL queries are being executed by tasks.
     void setMySqlThreadId(unsigned long id) { _mysqlThreadId.store(id); }
 
+    /// Return true if this task was already booted.
+    bool setBooted() { return _booted.exchange(true); }
+
+    /// Return true if the task was booted.
+    bool isBooted() { return _booted; }
+
```
iagaponenko, line 303: I'm not sure that simply adding yet another data member would be the right approach. I think that, the problem may need to be considered in a more general context of the class design. In particular, this class `wbase::Task` is a state machine that goes through a series of state transitions, depending on what's happening during query processing. The class has the following data members representing its state:
```
class Task {
private:
    std::atomic<TaskState> _state{TaskState::CREATED};
};
```
Where:
```
/**
 * The class State represents transient states of the worker-side tasks.
 * @note This class and the relevant functions are put into this header to
 * allow the compile-time (only) dependency onto this type from other modules
 * without needing to link against the current module's library.
 * Also note a choice of the underlying type which is meant to allow sending
 * values of the type as numeric attributes in the Protobuf messages w/o
 * introducing an additional (Protobuf) representation for those, or converting
 * the values to strings and vs.
 */
enum class TaskState : std::uint64_t {
    CREATED = 0,
    QUEUED,
    STARTED,
    EXECUTING_QUERY,
    READING_DATA,
    FINISHED
};
```
My impression is that "booting" is a significant event in the lifespan of the task, and as such it would need to be incorporated as a state. On the other hand, "booting" is not quite the state transition as it may mean one of 3 things:
1. merely detaching the task from a scheduler into the "free run" world
2. moving the task to some other low-priority scheduler ("snail", etc.)
3. or, canceling the task and (possibly queueing it into some low-priority) queue.

My understanding is that you're attempting to implement case number 2. Is this correct?

I also don't think you need to introduce these extra methods `setBooted()` and `isBooted()` since these will create redundancy and non-orthogonality. We already have the following methods which do effectively exactly what you wanted to achieve with the "boot" methods:
```
class Task {
public:
    void setTaskScheduler(TaskScheduler::Ptr const& scheduler) { _taskScheduler = scheduler; }
    TaskScheduler::Ptr getTaskScheduler() { return _taskScheduler.lock(); }
};
```
If you're implementing either scenario 1 (detaching from any scheduler) or 3 (canceling the task and re-queueing it) then you're going to reset the schedule to `nullptr`. If you're implementing scenario 2 (moving to "snail") then you are going to replace a value of the pointer with something else.

What I have not found in the new code is a timestamp when either of these significant events (1, 2, or 3) of making changes to the scheduler might happen. This could be done by modifying the implementation of the method `Task::setTaskScheduler()` to be more complex:
```
class Task {
public:
    void setTaskScheduler(TaskScheduler::Ptr const& scheduler) {
        _taskScheduler = scheduler;
       _schedChangeTime  =   std::chrono::system_clock::now();
    }

private:
    /// A timestamp when the last change of the scheduler was made
    std::chrono::system_clock::time_point _schedChangeTime;
};
```
This timestamp along with a value of the scheduler pointer is _exactly_ what is needed for the monitoring system to what's going on with the long-run tasks.

iagaponenko (commented):
Please, run Clang formatter on the code using:
```
qserv build -j8 --no-build-image --clang-format=REFORMAT
```
The current integration test has been broken sinc ethe code wasn't properly formatted before pushing it into GitHub.

src/wpublish/QueryStatistics.h
jgates108: The verbose explanation is a bit overkill. The the code compiles because "util/Histogram.h" is included in "wsched/Scheduler.h". I suppose it is mildly evil in that including the header will make a small increase in compile time, and it appears "wsched/Scheduler.h" doesn't really need it either. I can do something about that.

src/wbase/Task.h
jgates108: The problem is that "_booted" overlaps with other states, at least with EXECUTING_QUERY and possibly READING_DATA. Technically, what's happening is closest to your option 1, but the booted tasks should really be assigned to the snailScan so that snailScan is aware of their existance, and can then use them to make decisions about starting new Tasks from its queue. I don't think I'm going to go that far in this ticket, but it's probably what should happen.

I think keeping track of a task being booted is also important rguardless of the state of the task. What I mean is that as long as the system is holding onto the `Task`, we care about it being booted, even when it is `FINISHED`. When trying to figure out why a Task had issues, having been booted is something we probably want to know. Your proposed solution falls a little short in that every time a task get scheduled, it wipes out the previous timestamp, it also means there needs to be a separate timestamp for each TaskState change.
I think a better stratedgy would be to make a list of optional events for the task with timestamps. This would allow multiple scheduler changes to be recorded, as well multiple boot events. In most cases, this list would be empty. If we do decide to go with @fritzm 's idea of waterfall booting of Tasks, this would work reasonably well.

src/wpublish/QueriesAndChunks.cc
jgates108: The tracker only tracks "booted" tasks that are actively running as it's real purpose is to estimate how much of a drain on the system they are. Once the booted tasks finish, they are no longer a drain on the system.

src/wpublish/QueriesAndChunks.cc
```diff
@@ -381,31 +357,57 @@ void QueriesAndChunks::examineAll() {
...
                 if (iterChunk != tblSums.chunkPercentages.end()) {
+                    LOGS(_log, LOG_LVL_DEBUG,
+                         "examineAll c " << slowestTable << " chunkId=" << task->getChunkId() << " "
+                                         << task->getIdStr());
                     // We can only make the check if there's data on past chunks/tables.
                     double percent = iterChunk->second.percent;
                     bool valid = iterChunk->second.valid;
-                    double maxTimeChunk = percent * schedMaxTime;
+                    // Allow for some variation in time, also need twice as much time for 2 replicas.
+                    double leeway = 2.25;
```
iagaponenko, line 387: What this magic number is about? And why `2.25`? What if there would be 3 replicas or 1?

src/wpublish/QueriesAndChunks.cc
```diff
@@ -381,31 +357,57 @@ void QueriesAndChunks::examineAll() {
...
                     bool booting = runTimeMinutes > maxTimeChunk && valid;
-                    auto lvl = booting ? LOG_LVL_INFO : LOG_LVL_DEBUG;
+                    auto lvl = booting ? LOG_LVL_WARN : LOG_LVL_DEBUG;
                     LOGS(_log, lvl,
                          "examineAll " << (booting ? "booting" : "keeping") << " task " << task->getIdStr()
                                        << "maxTimeChunk(" << maxTimeChunk << ")=percent(" << percent
                                        << ")*schedMaxTime(" << schedMaxTime << ")"
                                        << " runTimeMinutes=" << runTimeMinutes << " valid=" << valid);
                     if (booting && !task->atMaxThreadCount()) {
+                        LOGS(_log, LOG_LVL_INFO,
```
iagaponenko, line 399: Why not to post this at the same logging level as defined above in:
```
auto lvl = booting ? LOG_LVL_WARN : LOG_LVL_DEBUG;
```

src/wpublish/QueriesAndChunks.cc
```diff
@@ -489,105 +499,168 @@ QueriesAndChunks::ScanTableSumsMap QueriesAndChunks::_calcScanTableSums() {
...
-}
+    // Track the total number of Tasks that are running while booted
+    // and move the worst offenders to the snailScan. The worst offender is the UserQuery with
+    // the most `Task`s running that are taking too long.
+    // The following function only counts booted `Task`s that are still EXECUTING,
+    // and `countQIdVect` is order by worst offender to least worst offender.
+    auto [bootedTaskCount, countQIdVect] = _bootedTaskTracker.getTotalBootedTaskCount();
+    // Sum of Tasks in
+    int uqBootedTasksSum = 0;
+    int maxDarkQueries = 5 * _maxTasksBooted;
```
iagaponenko, line 564: Another magic number which hasn't been explained.

src/wpublish/QueriesAndChunks.cc
```diff
@@ -489,105 +499,168 @@ QueriesAndChunks::ScanTableSumsMap QueriesAndChunks::_calcScanTableSums() {
...
+            QueryStatistics::Ptr queryToBoot = getStats(qIdToBoot);
+            if (queryToBoot != nullptr) {
+                if (_bootUserQuery(queryToBoot, bSched)) {
+                    uqBootedTasksSum += countQId.count;
+                    LOGS(_log, LOG_LVL_DEBUG,
+                         __func__ << " a booted uq=" << countQId.qId << ":count=" << countQId.count
+                                  << " uqBootedTasksSum=" << uqBootedTasksSum);
+                }
+            } else {
+                // This really should never happen.
```
iagaponenko, line 590: Then the code should throw:
```
util::Bug(ERR_LOC, "This really should never happen");
```

src/wpublish/QueriesAndChunks.cc
```diff
@@ -489,105 +499,168 @@ QueriesAndChunks::ScanTableSumsMap QueriesAndChunks::_calcScanTableSums() {
...
+    // Last check, if any of the queries have several booted `Task`s but have not been booted,
+    // they should be booted now.
+    for (auto const& countQId : countQIdVect) {
+        QueryId qIdToCheck = countQId.qId;
+        // Get query info
+        QueryStatistics::Ptr queryToCheck = getStats(qIdToCheck);
+        if (queryToCheck == nullptr) {
+            LOGS(_log, LOG_LVL_ERROR, __func__ << " Couldn't locate qIdToCheck=" << qIdToCheck);
+            // This really should never happen, but try to boot the next query in the list.
+            continue;
```
iagaponenko, line 605: Then the code should throw an exception to indicate a bug:
```
util::Bug(ERR_LOC, ...);
```

src/wpublish/QueriesAndChunks.cc
```diff
@@ -489,105 +499,168 @@ QueriesAndChunks::ScanTableSumsMap QueriesAndChunks::_calcScanTableSums() {
...
 
-/// @return true if all Tasks for this query are complete.
-/// Precondition, _qStatsMtx must be locked.
-bool QueryStatistics::_isMostlyDead() const { return _tasksCompleted >= _size; }
+bool QueriesAndChunks::_bootUserQuery(QueryStatistics::Ptr queryToBoot,
+                                      wsched::BlendScheduler::Ptr const& bSched) {
+    // Get query info
+    if (queryToBoot == nullptr) {
+        LOGS(_log, LOG_LVL_ERROR, __func__ << " queryToBoot is nullptr");
+        // This really should never happen.
```
iagaponenko, line 633: Another confusing statement. Please, consdier throwing exceptions for all those "should never happen".

src/wpublish/QueriesAndChunks.cc
```diff
@@ -489,105 +499,168 @@ QueriesAndChunks::ScanTableSumsMap QueriesAndChunks::_calcScanTableSums() {
...
+    }
 
-void QueryStatistics::addTaskTransmit(double timeSeconds, int64_t bytesTransmitted, int64_t rowsTransmitted,
-                                      double bufferFillSecs) {
-    _histTimeTransmittingPerTask->addEntry(timeSeconds);
-    _histRowsPerTask->addEntry(rowsTransmitted);
-    _histSizePerTask->addEntry(bytesTransmitted);
-    _histTimeBufferFillPerTask->addEntry(bufferFillSecs);
-}
+    bool queryWasBooted = false;
```
iagaponenko, line 642: The logic flow above is a bit hard to read. Why not to do something like this instead:
```
bool queryWasBooted = queryToBoot->_queryBooted;
if (queryWasBooted) return true;
```

src/wpublish/QueriesAndChunks.cc
```diff
@@ -489,105 +499,168 @@ QueriesAndChunks::ScanTableSumsMap QueriesAndChunks::_calcScanTableSums() {
...
+            LOGS(_log, LOG_LVL_DEBUG, __func__ << " schedTarg was snailScan for key=" << key);
+            continue;
+        }
 
-void QueryStatistics::addTaskRunQuery(double runTimeSeconds, double subchunkRunTimeSeconds) {
-    _histTimeRunningPerTask->addEntry(runTimeSeconds);
-    _histTimeSubchunkPerTask->addEntry(subchunkRunTimeSeconds);
+        LOGS(_log, LOG_LVL_INFO, "Booting uq queryId=" << queryToBoot->queryId);
+        queryToBoot->_queryBooted = true;
+        queryToBoot->_queryBootedTime = CLOCK::now();
```
iagaponenko, line 658: Is the current context modifying private data members of some other object? Is another object aware that its state could be modified externally w/o calling any public methods? This doesn't look safe.

src/wpublish/QueriesAndChunks.cc
```diff
@@ -601,21 +674,21 @@ vector<wbase::Task::Ptr> QueriesAndChunks::removeQueryFrom(QueryId const& qId,
...
-    lock.unlock();
 
     // Remove Tasks from their scheduler put them on 'removedList', but only if their Scheduler is the same
     // as 'sched' or if sched == nullptr.
     vector<wbase::Task::Ptr> taskList;
     {
-        lock_guard<mutex> taskLock(query->second->_qStatsMtx);
-        taskList = query->second->_tasks;
+        lock_guard<mutex> taskLock(query->_qStatsMtx);
+        taskList = query->_tasks;
```
iagaponenko, line 688: This coding pattern is dangerous since the calling context is accessing and using private data members of another object. Why not to redesign that `query` object by adding the following method:
```
class Query {
public:
    <task-list> getTasks() const {
        lock_guard<mutex> taskLock(_qStatsMtx);
        return_tasks;
   }
```
After that, you may simply call that method:
```
     taskList = query->getTasks();
```

src/wpublish/QueriesAndChunks.h
```diff
@@ -305,11 +301,15 @@ class QueriesAndChunks {
     std::chrono::seconds _examineAfter = std::chrono::minutes(5);
 
     /// Maximum number of tasks that can be booted until entire UserQuery is put on snailScan.
-    int _maxTasksBooted = 25;
+    int _maxTasksBooted = 15;
```
iagaponenko, line 304: Another magic number w/o a proper explanation.

src/wpublish/QueriesAndChunks.h
```diff
@@ -305,11 +301,15 @@ class QueriesAndChunks {
...
 
     /// Number of completed Tasks needed before ChunkTableStats::_avgCompletionTime can be
     /// considered valid enough to boot a Task.
     unsigned int _requiredTasksCompleted = 50;
```
iagaponenko, line 308: Another magic number w/o a proper explanation.

src/wbase/Task.cc
```diff
@@ -263,34 +266,51 @@ vector<Task::Ptr> Task::createTasks(shared_ptr<proto::TaskMsg> const& taskMsg,
...
     }
     sendChannel->setTaskCount(vect.size());
 
     return vect;
 }
 
+void Task::action(util::CmdData* data) {
+    string tIdStr = getIdStr();
+    if (_queryStarted.exchange(true)) {
+        LOGS(_log, LOG_LVL_INFO, "task was already started " << tIdStr);
```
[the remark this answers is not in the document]
jgates108: Done

src/util/ThreadPool.cc
```diff
@@ -117,6 +117,10 @@ void PoolEventThread::finishup() {
     }
 }
 
+CommandForThreadPool::~CommandForThreadPool() {
+    LOGS(_log, LOG_LVL_DEBUG, "~CommandForThreadPool::CommandForThreadPool()");
```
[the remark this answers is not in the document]
jgates108: Done

src/wbase/Task.cc
```diff
@@ -263,34 +266,51 @@ vector<Task::Ptr> Task::createTasks(shared_ptr<proto::TaskMsg> const& taskMsg,
...
+        return;
+    }
+
+    if (_unitTest) {
+        LOGS(_log, LOG_LVL_ERROR,
```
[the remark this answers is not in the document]
jgates108: There are a few unit tests that use this. The other option is to make a mock child of Task for those tests, which is a bit envolved. Its also worth noting that no queries will complete if _unitTest has been set.

src/wbase/Task.cc
```diff
@@ -263,34 +266,51 @@ vector<Task::Ptr> Task::createTasks(shared_ptr<proto::TaskMsg> const& taskMsg,
...
+    bool success = false;
+    try {
+        success = qr->runQuery();
+    } catch (UnsupportedError const& e) {
+        LOGS(_log, LOG_LVL_ERROR, __func__ << " runQuery threw UnsupportedError " << e.what() << tIdStr);
+    }
+    if (not success) {
+        LOGS(_log, LOG_LVL_ERROR, "runQuery failed " << tIdStr);
+        if (not getSendChannel()->kill("Foreman::_setRunFunc")) {
+            LOGS(_log, LOG_LVL_WARN, "runQuery sendChannel killed " << tIdStr);
```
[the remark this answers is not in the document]
jgates108: "sendChannel already killed"

src/wbase/Task.h
```diff
@@ -293,6 +299,19 @@ class Task : public util::CommandForThreadPool {
...
+    bool setBooted();
+
+    /// Return true if the task was booted.
+    bool isBooted() { return _booted; }
+
+    /// Only to be used in unit tests, use this to set a lambda function
+    /// to use in a unit test.
+    void setUnitTest(std::function<void(util::CmdData*)> func) {
+        _unitTest = true;
+        setFunc(func);
```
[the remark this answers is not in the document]
jgates108: The unit tests in question deal with the scheduler and booting tasks. The `func` provided by the unit test sleep for short or long periods to trip the booting logic.

src/wdb/testQueryRunner.cc
```diff
@@ -93,7 +93,9 @@ struct Fixture {
     }
     shared_ptr<QueriesAndChunks> queriesAndChunks() {
         bool resetForTesting = true;
-        return QueriesAndChunks::setupGlobal(chrono::seconds(1), chrono::seconds(300), 5, resetForTesting);
+        int maxTasksBooted = 5;
+        return QueriesAndChunks::setupGlobal(chrono::seconds(1), chrono::seconds(120), maxTasksBooted,
```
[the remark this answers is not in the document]
jgates108: Also just for the unit tests.

src/wpublish/QueriesAndChunks.cc
```diff
@@ -207,13 +148,16 @@ void QueriesAndChunks::queuedTask(wbase::Task::Ptr const& task) {
 /// Update statistics for the Task that just started.
 void QueriesAndChunks::startedTask(wbase::Task::Ptr const& task) {
     auto now = chrono::system_clock::now();
+    // TODO: Task::started() should probably be calling this function.
     task->started(now);
```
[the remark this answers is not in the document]
jgates108: Agreed, but that is also kind of a TODO maybe, and there's already enough being done in this ticket. I just removed the comment.

src/wpublish/QueriesAndChunks.cc
```diff
@@ -358,9 +331,12 @@ void QueriesAndChunks::examineAll() {
             lock_guard<mutex> lock(uq->_qStatsMtx);
             for (auto&& task : uq->_tasks) {
                 if (task->isRunning()) {
-                    auto const& sched = dynamic_pointer_cast<wsched::ScanScheduler>(task->getTaskScheduler());
-                    if (sched != nullptr) {
+                    auto const& tSched =
```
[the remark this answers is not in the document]
jgates108: The schedulers all have lifetimes as long as the worker, so there's no need to make a copy of the shared pointer. If `tSched` changes, the worker has crashed or is about to. `examineAll` is also a fairly expensive function and I'm also trying to keep the costs down.

src/wpublish/QueriesAndChunks.cc
jgates108: In the near future, the actual number of chunks in the full table scan, all the sizes of all the tables, and the number of replicas will be available, and the code will use those. This magic number will go away, and hopefully be an unhappy memory. 

`leeway` is meant to be large enough to keep from kicking innocent queries. As the TODO statement above tries to explain, there is no good value for this. The largest chunks are a lot larger than the smallest chunks, and there's no way for the worker to know how many chunks are in a real full table scan, so it's guessing, and it's probably not a even a very good guess. It really needs a reliable source of truth, which should be available soon. Even on the same system, the chunks used on a full tables scan may change at the whim of xrootd, and the worker has no way to know.

See [DM-43131](https://rubinobs.atlassian.net/browse/DM-43131) A reliable source for the details makes this problem a lot more tractable.

src/wpublish/QueriesAndChunks.cc
jgates108: That one is likely to stay and should be in a config file.

src/wpublish/QueriesAndChunks.cc
jgates108: It shouldn't happen, but if it does, it probably shouldn't crash the worker. I could see this possibly happening if a query got cancelled, but had long running tasks. Not being able to boot a cancelled task should not have any effect on performance.

src/wpublish/QueriesAndChunks.cc
jgates108: I can only see this possibly happening if something is cancelled, and in that case, it shouldn't crash the worker.

src/wpublish/QueriesAndChunks.cc
jgates108: I'll change the comments.

src/wpublish/QueriesAndChunks.cc
jgates108: Agreed. Originally QueryStatistics was a small struct, and this is leftover from that time. QueryStatistics is now pretty big and should no longer be a friend to QueriesAndChunks.

src/wpublish/QueriesAndChunks.h
jgates108: This is set by a configuration file later, so this value is pointless. Adding info to the comment.

src/wdb/testQueryRunner.cc
```diff
@@ -93,7 +93,9 @@ struct Fixture {
     }
     shared_ptr<QueriesAndChunks> queriesAndChunks() {
         bool resetForTesting = true;
-        return QueriesAndChunks::setupGlobal(chrono::seconds(1), chrono::seconds(300), 5, resetForTesting);
+        int maxTasksBooted = 5;
```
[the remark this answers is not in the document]
iagaponenko: Oh, right! I missed that!

src/wdb/testQueryRunner.cc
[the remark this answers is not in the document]
iagaponenko: True

src/wpublish/QueriesAndChunks.cc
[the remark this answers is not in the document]
iagaponenko: A problem with the left value being a reference to the shared pointer object is that `std::dynamic_pointer_cast` returns a value, not a reference to the pointer. So, in reality, the code is capturing a reference to some temporary shared pointer object. The life expectancy of that temporary object is disconnected from the reference. Please, read the documentation on the return type At: https://en.cppreference.com/w/cpp/memory/shared_ptr/pointer_cast

src/wpublish/QueriesAndChunks.cc
iagaponenko: > In the near future, the actual number of chunks in the full table scan, all the sizes of all the tables, and the number of replicas will be available, and the code will use those. This magic number will go away, and hopefully be an unhappy memory.

This is a wrong design assumption since workers should never have access to the number of replicas. That would require workers to effectively know what chunks other workers have. Implementing that would be a total mess for the Replication system due to complex state transitions. The Replication system would need to be made significantly more complex to redistribute all that information across workers. Besides, in the "shared nothing" architecture of Qserv workers are supported to act independently of each other.

If it's one of the key elements of the "booting" algorithm then, please, reconsider it.

The only option that I can see here would be if you implemented your own Czar-workers protocol where you could send this information from Czar to workers in each query as needed. Knowing table sizes is still fine.

jgates108: What the worker really needs to make a reasonable estimate is the list of chunks it will be expectected to handle in a full tables scan and the size of the tables in those chunks. This will probably come from the czar. 

Replication factor is only useful if the chunks in a full table scan is not available, as a replication factor of 2 would have half the chunks on the worker in a full table scan, a replication factor of 3 would use a third, etc. But, it's much better to just get the actual list of chunks from the czar.

PR #833, opened 2024-02-15
Tickets/dm 42617

iagaponenko:


src/czar/Czar.cc
```diff
@@ -538,6 +539,86 @@ void Czar::removeOldResultTables() {
...
+            result.completedChunks = stoi(colCompletedChunks[0]);
+            result.queryBeginEpoch = stoi(colQueryBeginEpoch[0]);
+            result.lastUpdateEpoch = stoi(colLastUpdateEpoch[0]);
+            break;
+        default:
+            // Should never be here.
+            string const msg = context +
+                               "Inconsistent stats returned for the user query, err=" + err.printErrMsg() +
+                               ", sql=" + sql;
+            throw runtime_error(msg);
```
jgates108, line 624: This is not something to be answered in this ticket, but maybe we should make a QservRuntimeException class based off utill::Issue and use that instead of runtime_error. That makes it easy to catch qserv related errors and also includes the source code location of the throw if it is caught.

src/czar/SubmitResult.h
```diff
@@ -25,25 +25,32 @@
...
 namespace lsst::qserv::czar {
 
 /// @addtogroup czar
 
 /**
  *  @ingroup czar
- *
- *  @brief Structure used for returning result from submitQuery.
+ *  @brief Structure used for returning result from Czar::submitQuery and from subsequent
+ *   calls to Czar::getQueryInfo
```
jgates108, line 38: Completely optional. The following has the same effect as `@brief`.

```
/** Structure used for returning result from Czar::submitQuery and from subsequent
 *   calls to Czar::getQueryInfo.
 ```

File: src/czar/Czar.cc [hunk 3]
```diff
@@ -145,27 +146,30 @@ Czar::Czar(string const& configFilePath, string const& czarName)
     const int year = 60 * 60 * 24 * 365;
     _idCounter = uint64_t(tv.tv_sec % year) * 1000 + tv.tv_usec / 1000;
 
+    auto databaseModels = qproc::DatabaseModels::create(_czarConfig->getCssConfigMap(),
+                                                        _czarConfig->getMySqlResultConfig());
+
+    // Need to be done first as it adds logging context for new threads
```
jgates108, line 153: `Needs` or `This Needs`

src/czar/Czar.cc
```diff
@@ -145,27 +146,30 @@ Czar::Czar(string const& configFilePath, string const& czarName)
...
+    _uqFactory.reset(new ccontrol::UserQueryFactory(databaseModels, _czarName));
+
     // Tell workers to cancel any queries that were submitted before this restart of Czar.
     // Figure out which query (if any) was recorded in Czar database before the restart.
     // The id will be used as the high-watermark for queries that need to be cancelled.
     // All queries that have identifiers that are strictly less than this one will
     // be affected by the operation.
+    //
+    // NOTE: This steps should be done after copnstructing the query factory where
```
jgates108: `This step should be done after constructing`

src/czar/Czar.cc
```diff
@@ -145,27 +146,30 @@ Czar::Czar(string const& configFilePath, string const& czarName)
...
+    //       the name of the Czar gets translated into a numeric icdentifier.
```
jgates108: icdentifier -> identifier

jgates108 (approved):
Looks good, just very minor comments.

src/czar/Czar.cc
iagaponenko: Good idea! At some point, we would need to review error reporting at Czar as many of those can be reported to Qserv users. Some messages may be quite confusing to users.

PR #834, opened 2024-03-06
DM-42638: The temporary message tables created by Qserv Czar are never deleted for the SQL COMMIT user query  

iagaponenko:


fritzm (approved):
LGTM, thanks!

PR #837, opened 2024-03-08
Update submodules

fritzm:
Updated `log` submodule moves us up to log4cxx v1.1.0, which required a couple minor tweaks to the base containers and the Qserv sources.

Don't forget to `git submodule update` for this one!


iagaponenko (approved):
Looks good to me.

PR #843, opened 2024-03-22
DM-43465: Allow default database context in query requests submitted via the Qserv HTTP frontend

iagaponenko:


fritzm (approved):
LGTM -- thanks!

PR #844, opened 2024-03-26
DM-43497: Handling binary column types in the HTTP frontend of Qserv

iagaponenko:


src/util/String.h
```diff
@@ -97,6 +97,16 @@ class String {
...
     }
+
+    /**
+     * Encode the input sequence of bytes into the hexadecimal representation packaged
+     * into a string.
+     * @param ptr A pointer to the byte sequence.
+     * @param length The number of bytes to translate.
+     * @return The encoded sequence of bytes or the empty string if the length=0.
+     * @throw std::invalid_argument If the pointer is nullptr.
+     */
```
jgates108, line 108: I think an example of input to output would help make it clearer what this is doing. From the jira, it's converting `[10,17,255,210]` to `"0A11FFD2"`

src/util/testString.cc
```diff
@@ -262,4 +263,56 @@ BOOST_AUTO_TEST_CASE(ToStringTest) {
...
+            "E0", "E1", "E2", "E3", "E4", "E5", "E6", "E7", "E8", "E9", "EA", "EB", "EC", "ED", "EE", "EF",
+            "F0", "F1", "F2", "F3", "F4", "F5", "F6", "F7", "F8", "F9", "FA", "FB", "FC", "FD", "FE", "FF"};
+
+    for (int i = 0; i < 256; ++i) {
+        char buf[1];
+        buf[0] = static_cast<char>(i);
+        BOOST_CHECK_EQUAL(util::String::toHex(buf, 1), char2hex[i]);
+    }
+
+    // Translate the long string that mqde of a monotonic sequence of all 256 characters.
```
jgates108, line 305: mqde?

jgates108 (approved):

src/util/String.h
iagaponenko: Good suggestion!

PR #845, opened 2024-03-27
DM-42810: Extend Qserv integration test to support the HTTP-based front-end

iagaponenko:


admin/local/docker/compose/docker-compose.yml
```diff
@@ -41,30 +41,37 @@ x-worker-repl:
...
-    czar_0_data:
-    czar_0_cfg:
-    czar_0_home:
-    czar_0_mariadb_lib:
-    czar_0_mariadb_log:
-    czar_0_mariadb_run:
-    repl_mgr_data:
-    repl_mgr_home:
-    repl_mgr_mariadb_lib:
+    volume_czar_xrootd:
```
fritzm, line 44: Is `volume_` prefix here redundant, given that inspection or specification contexts where these would be used are all exclusive to volumes?

If you think having it there makes things more understandable generally, then I am fine with it.  Understandability is "king".  Just asking!

src/admin/templates/http/etc/qserv-czar.cnf.jinja
```diff
@@ -0,0 +1,167 @@
...
+user = qsmaster
+port = {{ czar_db_port }}
+maxMsgSourceStore = 5
+
+# database connection for QStatusData
+# Using qmeta for now. May move this qservResult or make its own instance
+# databse based on what happens with multiple czars.
+[qstatus]
+passwd =
+#db = qservStatusData
```
fritzm, line 51: Remove dead wood comment?

src/admin/python/lsst/qserv/admin/qservCli/launch.py
```diff
@@ -32,11 +32,11 @@
 from typing import Dict, List, Optional
 
 from .opt import (
-    dashboard_port_ev,
-    mariadb_image_ev,
-    qserv_image_ev,
-    dh_user_ev,
-    dh_token_ev,
+    envval_dashboard_port,
```
fritzm, line 35: Slight preference for the slightly shorter `env_` prefix here throughout, but leave to your discretion :-)

iagaponenko: Good point! Will fix that.

src/admin/templates/http/etc/qserv-czar.cnf.jinja
iagaponenko: Will do it.

admin/local/docker/compose/docker-compose.yml
iagaponenko: Let me explain my motivation by using `czar_xrootd` as an example. If you browse the original (before my modifications) code for all occurrences of the name you would find 14 of those in 4 different contexts.

The first context is the name of a volume defined as:
```
volumes:
    czar_xrootd:
```
The identifier is referenced at (and other locations):
```
    czar_cmsd:
        ...
        volumes:
            - type: volume
              source: volume_czar_xrootd
```

The second context is the base name of a container:
```
    czar_xrootd:
        image: "${QSERV_IMAGE:?err}"
        init: true
        command: >
        ...
```
The third context is the virtual hostname (which is not the same as the name of a container):
```
    czar_xrootd:
        image: "${QSERV_IMAGE:?err}"
        ...
        hostname: czar_xrootd
```
The fourth context is the network mode (which could be related to the hostname?):
```
    czar_cmsd:
        image: "${QSERV_IMAGE:?err}"
        ...
        network_mode: "service:czar_xrootd"
```
So, the name is clearly overloaded. In my naive understanding, such overloading is probably normal for YAML, where the same name is treated differently depending on a context (where it's defined, and where it's referenced). However, I find it confusing and counterintuitive. Hence was my attempt to modify the configuration to prefix the class name (where class = "volume") in the name of an object.

BTW I run into a similar (name overloading) problem in the Python code.

Perhaps, for the "seasoned" YAML developer the overloading of names is not a problem.

fritzm: I see -- seems reasonable!

fritzm (approved):
LGTM -- thanks!

PR #846, opened 2024-03-27
tickets/DM-43384

jgates108:


src/czar/CzarChunkMap.h
```diff
@@ -0,0 +1,149 @@
...
+        /// Map of workers that have this chunk
+        std::map<std::string, std::weak_ptr<WorkerChunksData>> _workerHasThisMap;
+
+        std::string dump() const;
+    };
+
+    class WorkerChunksData {
+    public:
+        using Ptr = std::shared_ptr<WorkerChunksData>;
+        WorkerChunksData(std::string const& wId) : _workerId(wId) {}
```
iagaponenko, line 118: This isn't very important. However, it's always easier to read a code where there is a direct mapping of the construction parameters to the corresponding members like in:
```
 WorkerChunksData(std::string const& workerId) : _workerId(workerId) {}
```
...unless you're going to have the public/protected method like:
```
std::string const& workerId() const { return _workerId; }
```
In the latter case, the following usually works well:
```
WorkerChunksData(std::string const& workerId_) : _workerId(workerId_) {}
```

src/czar/CzarChunkMap.h
```diff
@@ -0,0 +1,149 @@
...
+        std::string dump() const;
+
+        std::string const _workerId;
+
+        /// Map of all chunks found on the worker
+        std::map<int, ChunkData::Ptr> _chunkDataMap; ///< key is chunkId // &&& needed?
+
+        /// Map of chunks this worker will handle during shared scans.
+        std::map<int, ChunkData::Ptr> _sharedScanChunkMap;
+        int64_t _sharedScanTotalSize = 0;
```
iagaponenko, line 150: Size is usually expressed using unsigned integer types.

src/czar/CzarChunkMap.h
```diff
@@ -0,0 +1,149 @@
...
+
+    CzarChunkMap() = delete;
+
+    static Ptr create(std::shared_ptr<qmeta::QMeta> const& qmeta) {
+        return Ptr(new CzarChunkMap(qmeta));
+    }
+
+    class WorkerChunksData;
+
+    class ChunkData {
```
iagaponenko, line 80: Having a public nested class that has all its members public may be a big problem. The same comment applies to the relevant class `WorkerChunksData`.  You allow any user of the class `CzarChunkMap` to mess with the internal state of `CzarChunkMap` as well as with internal states of `ChunkData` and `WorkerChunksData`. This design violates many key principles of the Object-oriented programming.

src/czar/CzarChunkMap.cc
```diff
@@ -63,23 +43,52 @@ LOG_LOGGER _log = LOG_GET("lsst.qserv.czar.CzarChunkMap");
...
+    qmeta::QMeta::ChunkMap qChunkMap = _qmeta->getChunkMap();
+    if (_lastUpdateTime >= qChunkMap.updateTime) {
+        LOGS(_log, LOG_LVL_DEBUG,
+             __func__ << " CzarChunkMap no need to read "
+                      << util::TimeUtils::timePointToDateTimeString(_lastUpdateTime)
+                      << " db=" << util::TimeUtils::timePointToDateTimeString(qChunkMap.updateTime));
+        return false;
+    }
+
+    auto const& jsChunks = qChunkMap.chunks;
```
iagaponenko: I thought the chunk map was migrated away from the JSON library.

src/czar/CzarChunkMap.h
```diff
@@ -0,0 +1,149 @@
...
+}
+
+namespace lsst::qserv::czar {
+
+class ChunkMapException : public util::Issue {
+public:
+    ChunkMapException(Context const& ctx, std::string const& msg) : util::Issue(ctx, msg) {}
+};
+
+class CzarChunkMap {
```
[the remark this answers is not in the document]
jgates108: That is probably not the most recent commit that you are commenting on. 10+ lines of documentation.

src/czar/CzarChunkMap.h
jgates108: All of the members are private with no method to change their values. Aside from calcTotalBytes, which I'm making private now. 

CzarChunkMap can only set the values because it a friend to this class. Nothing else should be able to alter these. This applies to all nested classes in CzarChunkMaps.

The underlying issue is that these maps will be used heavily. Making copies or locking mutexes will be expensive, so the nested classes need to be exposed for rapid cheap access, but need to be read only for safety. The friend class seemed to be the best way to do that. const isn't really an option for _totalBytes or _primaryScanWorker.

src/czar/CzarChunkMap.h
```diff
@@ -0,0 +1,149 @@
...
+        std::weak_ptr<WorkerChunksData> _primaryScanWorker;
+
+        /// &&& doc
+        void calcTotalBytes();
+
+        /// &&& doc
+        void addToWorkerHasThis(std::shared_ptr<WorkerChunksData> const& worker);
+
+        /// Key is databaseName+tableName, value is size in bytes.
+        std::map<std::pair<std::string, std::string>, int64_t> _dbTableMap;
```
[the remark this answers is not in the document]
jgates108: std::pair works as a key in a map as long as both items are comparable. Struct can be done, but the comparison needs to be defined. I'm not sure it merits the effort here as database name + table name is a normal pairing for database tables.

src/czar/CzarChunkMap.h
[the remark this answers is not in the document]
iagaponenko: I simply followed a link to the PR from the ticket. At a time when I was reviewing the PR I didn't see any other commits. Did you do `git push -f` on your recent mods?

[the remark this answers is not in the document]
jgates108: I could see the comments on the PR when I entered the previous message.

src/czar/CzarChunkMap.h
iagaponenko: Yes, I can see it now after reviewing other changes made to these classes in the second commit:
```
Added unit test.
```
So, please, disregard my comments.

iagaponenko (approved):
Looks good

src/czar/CzarChunkMap.h
jgates108: Changed name to workerId.

src/czar/CzarChunkMap.h
jgates108: Changed

jgates108:
This PR was made before the tickets/DM-43715 was created. Since that branch is still the same as main, the changes in the PR are the same, but this seems to have other effects on the PR.

PR #848, opened 2024-04-02
DM-43602: Rolled back schema and the relevant changes made in DM-43291

iagaponenko:


src/qmeta/QMetaMysql.cc
```diff
@@ -48,7 +47,7 @@ using namespace std;
 namespace {
 
 // Current version of QMeta schema
-char const VERSION_STR[] = "10";
+char const VERSION_STR[] = "9";
```
jgates108, line 50: slac6dev was advanced to version 10. That will need to be undone at some point.

jgates108 (approved):
Looks good.

PR #849, opened 2024-04-04
DM-43625: Extend the MySQL API of the Qserv Replication Framework

iagaponenko:
The new extension allows to generate collections of multi-row INSERT queries where the maximum length of each query is constrained by the specified limit.

### DIFF
--- a/src/replica/mysql/DatabaseMySQLGenerator.h
+++ b/src/replica/mysql/DatabaseMySQLGenerator.h
@@ -22,6 +22,7 @@
 #define LSST_QSERV_REPLICA_DATABASEMYSQLGENERATOR_H
 
 // System headers
+#include <list>
 #include <memory>
 #include <string>
 #include <tuple>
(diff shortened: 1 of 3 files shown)


src/replica/mysql/DatabaseMySQLGenerator.h
```diff
@@ -1083,6 +1111,14 @@ class QueryGenerator {
     std::string call(DoNotProcess const& packedProcAndArgs) const;
 
 private:
+    /**
+     * Check if the specified collection i snot empty.
```
jgates108, line 1115: i snot -> is not

src/replica/mysql/DatabaseMySQLGenerator.h
```diff
@@ -1083,6 +1111,14 @@ class QueryGenerator {
...
+     * @param func A scope from which the check was requested.
+     * @param coll A cvollection to be evaluated.
```
jgates108, line 1117: cvollection -> collection

src/replica/mysql/DatabaseMySQLGenerator.h
```diff
@@ -402,6 +403,33 @@ class QueryGenerator {
     std::string insertPacked(std::string const& tableName, std::string const& packedColumns,
                              std::vector<std::string> const& packedValues) const;
 
+    /**
+     * Generate a collection of complete INSERT statements for the given input, where
+     * the maximum size of each query string is determined by a value of the parameter
+     * \param maxQueryLength.
```
jgates108, line 409: Is that supposed to be @param? "the parmaeter \param maxquerylength." I'm don't think that's what you wanted.

jgates108 (approved):

src/replica/mysql/DatabaseMySQLGenerator.h
iagaponenko: My bad. I assumed I could use this syntax to generate a reference to the parameter. After re-reading the documentation on Doxygen that's not possible...and not needed here. Thank you for noticing this!

PR #850, opened 2024-04-05
DM-43603: Scalable implementation of the persistent chunk map in the Qserv Czar database

iagaponenko:


src/qmeta/QMeta.h
```diff
@@ -58,6 +58,45 @@ class QMeta {
...
+        /// @return 'true' if the map is empty (or constructed using the default constructor)
+        bool empty() const {
+            return workers.empty() || (std::chrono::time_point<std::chrono::system_clock>() == updateTime);
+        }
+
+        // NOTE: Separate types were added here for the sake of clarity to avoid
+        // a definition of the unreadable nested map.
+
+        typedef std::pair<unsigned int, size_t> Chunk;     ///< Chunk number and its size (bytes)
+        typedef std::vector<Chunk> Chunks;                 ///< Collection of chunks
```
jgates108, line 88: To clarify, Chunk<unsigned int = chunk id number, size_t = the size of the database.table in that chunk>. Correct?

src/qmeta/schema/migrate-None-to-10.sql.jinja
```diff
@@ -207,6 +207,28 @@ CREATE TABLE IF NOT EXISTS `QMessages` (
 ENGINE = InnoDB
 COMMENT = 'Table of messages generated during queries.';
 
+-- -----------------------------------------------------
+-- Table `chunkMap`
+-- -----------------------------------------------------
+
+CREATE TABLE IF NOT EXISTS `chunkMap` (
+  `worker` VARCHAR(256) NOT NULL COMMENT 'A unique identifier of a worker hosting the chuhnk replica',
```
jgates108, line 215: chuhnk -> chunk

src/sql/SqlResults.h
```diff
@@ -96,6 +96,8 @@ class SqlResults : boost::noncopyable {
                               std::vector<std::string>&, std::vector<std::string>&, SqlErrorObject&);
     bool extractFirst4Columns(std::vector<std::string>&, std::vector<std::string>&, std::vector<std::string>&,
                               std::vector<std::string>&, SqlErrorObject&);
+    bool extractFirst5Columns(std::vector<std::string>&, std::vector<std::string>&, std::vector<std::string>&,
+                              std::vector<std::string>&, std::vector<std::string>&, SqlErrorObject&);
```
jgates108, line 100: I don't think this should be changed in this ticket, but this is odd. `bool extractFirstXColumns(int numOfCols, vector<vector<string>>&, SqlErrorObject&);` with `string& worker = vect[i][0];` `string& database = vect[i][1];`, etc.

jgates108 (approved):

src/sql/SqlResults.h
iagaponenko: Do you mean missing parameter names in the method signatures, or something else? The following is perfectly legal syntax in C and C++:
```
void foo(T&)
```
However, I completely agree with you that it's a bit unusual in this context. I will fix a signature of this and similar methods of the class.

src/qmeta/QMeta.h
iagaponenko: Yes, that's correct. And since you asked this question (which means the interface is not self-explanatory) I will replace `std::pair` with the following struct:
```
struct ChunkInfo {
    unsigned int chunk = 0;
    size_t size = 0;
};
```

src/sql/SqlResults.h
jgates108: The syntax is fine. It's just 5 different functions when one would do.

iagaponenko: Yep, I agree. I will introduce another method to fill up a 2D vector `vector<vector<string>>` as per your suggestion.

PR #853, opened 2024-05-01
DM-43282: HTTP-based simple ingest service for fully-replicated tables in Qserv 

iagaponenko:


src/util/String.cc
```diff
@@ -110,18 +111,62 @@ vector<uint64_t> String::parseToVectUInt64(string const& str, string const& deli
                                              throwOnError, defaultVal);
 }
 
-string String::toHex(char const* ptr, size_t length) {
+string String::toHex(char const* ptr, size_t length, string const& prefix, bool lowerCase) {
```
fritzm, line 116: Should we consider using base64 encoding instead of hex?  Potentially better use of bandwidth, plus standard encoders/decoders available on Python side...

src/util/String.cc
```diff
@@ -171,4 +173,20 @@ string String::fromHex(string const& hex, string const& prefix) {
...
+}
+
+string String::toUpper(string const& str) {
+    string result = str;
+    transform(result.begin(), result.end(), result.begin(), ::toupper);
+    return result;
+}
+
+bool String::equalIgnoreCase(std::string const& str1, std::string const& str2) {
+    return toLower(str1) == toLower(str2);
```
fritzm, line 189: How about `std::equal(str1.begin(), str1.end(), str2.begin(), str2.end(), ichar_equals);`?

src/util/String.cc
iagaponenko: Adding base64 is a good idea. The "hex" encoding is an important case since we already depend on it in the Qserv integration tests for processing queries involving binary column types.

iagaponenko: Eventually, when `qhttp` will be ready to parse the multipart file attachments I will extend the ingest API to accept CSV-encoded payloads in the request body. This would be the most efficient way to send data & schema to the REST API. The present JSON-based packaging was just a good (and simple) way to start based on existing functionality and interfaces (`qhttp`).

src/util/String.cc
iagaponenko: That implies adding the comparator function, something along these lines:
```
bool ichar_equals(char a, char b)
{
    return std::tolower(static_cast<unsigned char>(a)) ==
           std::tolower(static_cast<unsigned char>(b));
}
```
Which is a reasonable option too.

fritzm (approved):
Looks good -- a couple small comments/qs left for your consideration

PR #854, opened 2024-05-12
DM-44281: Add support for base64 encoding for the binary column data in the HTTP frontend of Qserv

iagaponenko:


fritzm (approved):
LGTM -- thanks!

PR #859, opened 2024-07-02
DM-45049: Add support for SSL/TLS in the Qserv Czar frontend

iagaponenko:
This effort is about replacing the non-encrypted HTTP protocol with HTTPS in the Qserv Czar front end: https://confluence.lsstcorp.org/display/DM/HTTP+frontend+of+Qserv

The new implementation of the front end’s REST API will be based on the built-in HTTP server https://github.com/yhirose/cpp-httplib added to Qserv in DM-44780. 

No changes to the functionality of the front end are expected to be made. No modifications to the REST API version will be made either. The user documentation on the REST API will be updated accordingly.

The implementation provides two options for the SSL certificates:
- The certificates can be injected into the service’s container via a mount point at the default location expected by the front end. An alternative location of the certificate files could be specified via the optional parameters of the container’s entry point
-  If no certificates are found by the container's entry point at the expected (or overridden) location then the entry point will automatically generate self-signed certificates at the same location area.


fritzm (approved):
Nice! LGTM

PR #860, opened 2024-07-18
DM-45311: Upgraded the Apache Arrow and Parquet libraries to version 17

iagaponenko:
Note a change in the library version naming. It used to be "15", "16". As of the current version, it's "1700".


fritzm (approved):
LGTM -- thanks

PR #868, opened 2024-09-18
DM-46320: Supporting the multipart/form-data format in the Qserv's push-mode ingest protocol

iagaponenko:


src/http/FileUploadModule.cc
```diff
@@ -0,0 +1,112 @@
...
+            if (currentFile != nullptr) {
+                if (!currentFile->filename.empty()) {
+                    onEndOfFile();
+                } else {
+                    body().objJson[currentFile->name] = currentFile->content;
+                }
+            }
+        };
+        _contentReader(
+                [&](httplib::MultipartFormData const& file) -> bool {
```
fritzm, line 62: Do you really want `[&]` here?  Arg is already a const ref...

fritzm (approved):
LGTM -- thanks!  `httplib` seems pretty nice...

src/http/FileUploadModule.cc
iagaponenko: The problem is that this lambda calls 3 methods of the current class. So, I would need at least ``this`` to make the compiler happy. It also modifies a value of ``currentFile``. It's easier to pass ``&``.

PR #869, opened 2024-09-20
DM-46397: Extend Qserv integration test to use HTTP-based push mode catalog ingest

iagaponenko:


fritzm (commented):
This PR looks like it contains a lot of the same deltas as the previous one -- now that the previous is approved, can you merge it then rebase this before I review?

fritzm (approved):
LGTM -- thanks!

PR #875, opened 2024-11-02
DM-47345: Upgrade run-base container of Qserv to Apache Arrow and Parquet version 18

iagaponenko:
…on 18


fritzm (approved):
LGTM -- thanks!

PR #876, opened 2024-11-05
DM-46919: Extend the http frontend of Qserv to support pushing CSV files in the multipart/form-data body

iagaponenko:


src/czar/HttpCzarIngestModuleBase.cc
```diff
@@ -0,0 +1,374 @@
...
+    string const userDatabaseNamesPrefix = "user_";
+    if ((databaseName.size() <= userDatabaseNamesPrefix.size()) ||
+        !boost::iequals(databaseName.substr(0, userDatabaseNamesPrefix.size()), userDatabaseNamesPrefix)) {
+        auto err = "database name doesn't start with the prefix: " + userDatabaseNamesPrefix;
+        throw http::Error(func, err);
+    }
+}
+
+void HttpCzarIngestModuleBase::verifyUserTableName(string const& func, string const& tableName) const {
+    string const qservTableNamesPrefix = "qserv";
```
jgates108, line 129: The documentation said "qserv_" was reserved, although having "qserv" be reserved probably makes more sense.

src/czar/HttpCzarIngestModuleBase.h
```diff
@@ -0,0 +1,313 @@
...
+            std::function<std::map<std::string, std::string>(uint32_t)> const& submitRequestsToWorkers);
+
+    /**
+     * Verify the user-provided database name to ensure the name starts with the reserved
+     * prefix "user_".
+     * @param func The name of the calling function.
+     * @param databaseName The name of the database to be verified.
+     * @throw http::Error if the name is too short, or it doesn't start with the required prefix.
+     */
+    void verifyUserDatabaseName(std::string const& func, std::string const& databaseName) const;
```
jgates108, line 97: I think this could be static.

src/czar/HttpCzarIngestModuleBase.h
```diff
@@ -0,0 +1,313 @@
...
+
+    /**
+     * Verify the user-provided table name to ensure the name doesn't start with the reserved
+     * prefix "qserv_".
+     * @param func The name of the calling function.
+     * @param tableName The name of the table to be verified.
+     * @throw http::Error if the name is too short, or it starts with the reserved prefix.
+     */
+    void verifyUserTableName(std::string const& func, std::string const& tableName) const;
```
jgates108, line 106: I think this could be static.

src/czar/HttpCzarIngestModuleBase.cc
```diff
@@ -0,0 +1,374 @@
...
+    return numDirectors;
+}
+
+// These parameters correspond to the present partitioning model of 150k chunks.
+// In reality the parameters are not needed for ingesting regular tables which is
+// is the only kind of tables supported by the current implementation of the module.
+// Some value of the parameters are still required by the Replication/Ingest system's API.
+unsigned int defaultNumStripes = 340;
+unsigned int defaultNumSubStripes = 3;
+float defaultOverlap = 0.01667;
```
jgates108, line 70: These appear to be placeholder values that aren't really used? It would be nice to get rid of them to prevent future confusion.

src/czar/HttpCzarIngestModuleBase.cc
```diff
@@ -0,0 +1,374 @@
...
+    }
+    _createDatabase(databaseName);
+    _createDirectorTable(databaseName);
+}
+
+void HttpCzarIngestModuleBase::_createDatabase(string const& databaseName) {
+    json data = json::object({{"database", databaseName},
+                              {"num_stripes", ::defaultNumStripes},
+                              {"num_sub_stripes", ::defaultNumSubStripes},
+                              {"overlap", ::defaultOverlap}});
```
jgates108, line 180: The comment at definition of these values leads me to believe they have no real effect here. Maybe they should have have values that indicate that? A replicated table effectively has 0 stripes, 0 substripes, and 0 overlap. If it's a partioned tables, these should be real values.

src/czar/WorkerIngestProcessor.cc
```diff
@@ -0,0 +1,50 @@
...
+#include "czar/WorkerIngestProcessor.h"
+
+using namespace std;
+
+namespace lsst::qserv::czar::ingest {
+
+Request::Request(function<Result()> const& req, shared_ptr<ResultQueue> resultQueue)
+        : _req(req), _resultQueue(resultQueue) {}
+
+void Request::operator()() { _resultQueue->push(_req()); }
```
jgates108, line 32: I have a minor complaint about operator()(). It can be very difficult to find all instances where it is called when the ide isn't doing a good job. grep "()" generates vast numbers of false positives.

jgates108 (approved):

src/czar/WorkerIngestProcessor.cc
iagaponenko: Good point.

doc/user/http-frontend-ingest.rst
```diff
@@ -0,0 +1,577 @@
...
+
+  .. code-block::
+
+    user_<username>_
+
+  The rest of the name should include the name of a user or a role. For example:
+
+  .. code-block::
+
+    user_gapon
```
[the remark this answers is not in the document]
iagaponenko: This is an example of the name of a database. The (user) table ingest API requires the names of databases to follow the following convention:
```
user_<username>
```
Where `<username>` can be the account name of a user, the role name, or anything else. The most important thing is the prefix `user_` which allows the implementation of the Qserv Data Management system (Replication./Ingest) to treat these databases accordingly.

doc/ingest/api/concepts/publishing.rst
```diff
@@ -0,0 +1,68 @@
...
+Databases can also be unpublished to allow adding new tables, or for performaging significant changes to the table data or schema
+using the following service:
+
+- :ref:`ingest-db-table-management-unpublish-db` (REST)
+
+..  note::
+
+    The database unpublishing service does not affect the visibility of the database to Qserv users. All tables that existed
+    in the database before unpublishing can still be queried by Qserv users. The unpublishing operation is transparent to the users.
+
```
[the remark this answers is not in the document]
iagaponenko: The query will fail in an unpredictable way. The current integration model between Qserv query processor and the Qserv Data Management system is rather weak. That is by the original design to allow Qserv query processing to run w/o any DM system. The latter does not interfere with query processing during catalog ingests. No "clutch" exists yet. The database/table state transitions from unpublished to published, and back are mostly meant to ensure consistency of the ingested data during the data management phases. The only exception is when a new database (or a new table) is getting published. In such a case the transition is atomic. 
In general, catalog ingests are supposed (and, in most cases, will) happen when Qserv is not in use, during the DR ingesting periods, or the database maintenance periods.
When a table is deleted then all replicas are deleted as well regardless if there any outstanding queries.
Note that besides Qserv we also have the TAP layer above. Adding/removing/modifying databases/tables will also require making changes to the TAP schemas. So user queries will be blocked before they could reach Qserv during the large scale data ingests.

src/czar/HttpCzarIngestModuleBase.cc
iagaponenko: Empty (zero) values would work as well. This is related to your next comment.

src/czar/HttpCzarIngestModuleBase.cc
iagaponenko: Probably

PR #884, opened 2025-01-08
Tickets/dm 45548

jgates108:


src/ccontrol/UserQuerySelect.cc
```diff
@@ -233,6 +233,11 @@ std::string UserQuerySelect::getResultQuery() const {
 
 /// Begin running on all chunks added so far.
 void UserQuerySelect::submit() {
+    auto exec = _executive;
+    if (exec == nullptr) {
+        LOGS(_log, LOG_LVL_ERROR, "UserQuerySelect::submit() executive is null at start");
```
iagaponenko, line 238: This is a bit confusing. Shouldn't the user query object be constructed with a valid pointer to the "executive"? And if so then why is this test? What makes the pointer to suddenly disappear?
I'm also seeing the very same pattern of checking for the validity of a pointer all around the code of the PR. Perhaps there is a safer way of constructing objects to avoid putting the code with these tests?
My other worry is that if there is such an uncertainty (with the elusive pointer) then such code must be really hard to debug.

iagaponenko:
I have one general comment on redefining the types of the Standard Library. I think it is not a very good idea to introduce an extra layer of indirection for the pure (where no custom template parameters are involved) `std::` types. I would argue that this should to be done in the C++ headers due to expanded/exploded visibility of such definitions (It's probably fine to redefine types in a limited scope of the CC files. Though, this still should be done with care after assessing pros-vs-cons of such redefinition). Firstly, such a redirection complicates code analysis (by reviewers), Secondly, it blurs a separation between the Standard types and the custom types. Quite often it's hard to tell what type it is - if it's a user-defined type or the Standard one. So, the reviewer has to figure out where the type definition is made, go to the corresponding header, and memorize that type. My understanding is that a motivation for such a redefinition is probably to "save typing characters" or "see shorted lines" in the code. However, that benefit is rarely achieved when the headers are affected. Here is one example that I have found in the PR.

Consider redirections for `Clock` and `TimeType` made in `src/qmeta/JobStatus.h`:
```
class JobStatus {
public:
    typedef std::shared_ptr<JobStatus> Ptr;
    using Clock = std::chrono::system_clock;
    using TimeType = std::chrono::time_point<Clock>;
    ...
```
One can quickly see the first problem here - `TimeType` is not the `time point`. This changes expectations of the new type since `time type` is a very generic term (vs very specific `time point` of the Standard Library).

Now, let's look at where some of these redefined types are used in `src/ccontrol/UserQueryAsyncResult.cc`:
```
int64_t timestampMilli = ...;
qmeta::JobStatus::Clock::duration duration = std::chrono::milliseconds(timestampMilli);
qmeta::JobStatus::TimeType timestamp(duration);
```
This code is totally confusing. It causes a reviewer to look into the header where both types are defined. Compare this with the clean code w/o type redefinition:
```
int64_t timestampMilli = ...;
using namespace std;
chrono::system_clock::duration duration = chrono::milliseconds(timestampMilli);
chrono::time_point<chrono::system_clock> timestamp(duration);
```
That (second version of the) code is very easy to read (to someone who knows what the `std::chrono` is). That code doesn't create an extra a "mental barrier" for understanding the original intent of a developer ... which is to make a timestamp in the system clock.

Moreover, your next step in streamlining and improving the code would be to put those 2 lines into a function:
```
chrono::time_point<chrono::system_clock> systemTimeStamp(int64_t timestampMilli) {
    chrono::system_clock::duration duration = chrono::milliseconds(timestampMilli);
    return chrono::time_point<chrono::system_clock>(duration);
}
```
And since the function returns the final type then the recipient code would be reduced to use `auto` as shown below:
```
int64_t timestampMilli = ...;
auto timestamp = systemTimeStamp(timestampMilli);
```
That code is very easy to read and understand.

The bottom line is that when you redefine Standard types then you in essence introduce your own (non-standard) vocabulary. This adds extra complexity to the code on top of your application-specific complexity and vocabulary. It's better not to do this if possible.

src/ccontrol/UserQuerySelect.cc
jgates108: _executive being null at this point should really never happen, so that can be removed.

PR #885, opened 2025-01-24
Cleanup documentation

JeremyMcCormick:
Minor cleanup of the Qserv documentation:

- Removed the custom template for the landing page (index) and switched to using a standard `index.rst` file
- Removed the glossary, which only had two entries
- Removed the full table of contents page

Additionally, I had to add an ignore for the MySQL links to `conf.py`, as the build was showing a 403 response for these, indicating that they may be blacklisting all requests from GitHub Action runners.


fritzm (approved):
LGTM -- thanks!

PR #888, opened 2025-01-31
DM-48646: Reinforced the implementation of the FQDN resolver

iagaponenko:
The previous version was prone to crashes when DNS was not returning a well-formed result.


jgates108 (approved):
Looks good.

PR #900, opened 2025-03-31
DM-49795: Fixed a bug in the the worker configuration classes

iagaponenko:
Some data members were not initialized. This was causing occasionbal crashes of the application and possible memory corruption.

### DIFF
--- a/src/util/ConfigValMap.h
+++ b/src/util/ConfigValMap.h
@@ -162,11 +162,11 @@ public:
 
 protected:
     ConfigValT(std::string const& section, std::string const& name, bool required, T defVal, bool hidden)
-            : ConfigVal(section, name, required, hidden), _val(defVal) {}
+            : ConfigVal(section, name, required, hidden), _defVal(defVal), _val(_defVal) {}
 
 private:
-    T _val;     ///< Value for the item this class is storing.
-    T _defVal;  ///< Default value for the item this class is storing.
+    T const _defVal;  ///< Default value for the item this class is storing.
+    T _val;           ///< Value for the item this class is storing.
 };
 
 /// Bool is special case for json as the value should be "true" or "false" but
--- a/src/wconfig/WorkerConfig.h
+++ b/src/wconfig/WorkerConfig.h
@@ -88,9 +88,9 @@ public:
 private:
     ConfigValResultDeliveryProtocol(std::string const& section, std::string const& name, bool required,
                                     std::string const& defVal, bool hidden)
-            : ConfigVal(section, name, required, hidden), _val(parse(defVal)) {}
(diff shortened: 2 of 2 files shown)


src/wconfig/WorkerConfig.h
```diff
@@ -88,7 +88,11 @@ class ConfigValResultDeliveryProtocol : public util::ConfigVal {
 private:
     ConfigValResultDeliveryProtocol(std::string const& section, std::string const& name, bool required,
                                     std::string const& defVal, bool hidden)
-            : ConfigVal(section, name, required, hidden), _val(parse(defVal)) {}
+            : ConfigVal(section, name, required, hidden) {
+        _val = parse(defVal);
+        _defVal = _val;
+    }
+
```
jgates108, line 95: I ran ran into the same problem in the branch. My solution there is
```    
ConfigValResultDeliveryProtocol(std::string const& section, std::string const& name, bool required,
                                    std::string const& defVal, bool hidden)
            : ConfigVal(section, name, required, hidden), _defVal(parse(defVal)), _val(_defVal) {}
TEnum const _defVal;  ///< Default value for the item this class is storing.
TEnum _val;           ///< Value for the item this class is storing.
```

src/util/ConfigValMap.h
```diff
@@ -162,7 +162,10 @@ class ConfigValT : public ConfigVal {
 
 protected:
     ConfigValT(std::string const& section, std::string const& name, bool required, T defVal, bool hidden)
-            : ConfigVal(section, name, required, hidden), _val(defVal) {}
+            : ConfigVal(section, name, required, hidden) {
+        _val = defVal;
+        _defVal = defVal;
+    }
```
jgates108, line 168: I ran into the same problem in the branch. I changed the code to (also making the default const)
```
protected:
    ConfigValT(std::string const& section, std::string const& name, bool required, T defVal, bool hidden)
            : ConfigVal(section, name, required, hidden), _defVal(defVal), _val(_defVal) {}

private:
    T const _defVal;  ///< Default value for the item this class is storing.
    T _val;           ///< Value for the item this class is storing.
```

jgates108 (approved):
I already have a fix for this in the branch and I put the code in the review comments. The significant difference is _defVal is const in the branch now.

src/wconfig/WorkerConfig.h
iagaponenko: That's the same as I have.

src/util/ConfigValMap.h
iagaponenko: Good point! I will do the same in my PR before merging it into the main branch.

PR #902, opened 2025-04-01
DM-49803: Fixed a bug in the Qserv worker query runner

iagaponenko:
The older code was attempting to lock an uinitialized weak pointer, which happened to cause crashes of the process.
Also eliminated unused code from the same class.

### DIFF
--- a/src/wdb/QueryRunner.h
+++ b/src/wdb/QueryRunner.h
@@ -45,10 +45,6 @@
 #include "wbase/Task.h"
 #include "wdb/ChunkResource.h"
 
-namespace lsst::qserv::xrdsvc {
-class StreamBuffer;
-}  // namespace lsst::qserv::xrdsvc
-
 namespace lsst::qserv::wcontrol {
 class SqlConnMgr;
 }  // namespace lsst::qserv::wcontrol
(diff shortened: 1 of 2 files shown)


jgates108 (approved):
Looks good.

src/wdb/QueryRunner.h
```diff
@@ -91,24 +87,16 @@ class QueryRunner : public wbase::TaskQueryRunner, public std::enable_shared_fro
...
-
     wbase::Task::Ptr const _task;  ///< Actual task
 
-    qmeta::CzarId _czarId = 0;  ///< To be replaced with the czarId of the requesting czar.
-
     /// Resource reservation
     ChunkResourceMgr::Ptr _chunkResourceMgr;
-    std::string _dbName;
     std::atomic<bool> _cancelled{false};
-    std::weak_ptr<xrdsvc::StreamBuffer> _streamBuf;  ///< used release condition variable on cancel.
```
jgates108, line 110: As long as `this` is valid, locking `_streamBuf` should be valid. It lock should return nullptr if it hasn't been set. I suspect the problem was caused by the other memory issues we have seen.

iagaponenko: I agree with you. I too was surprised when I noticed this crash and a huge reference counter on the pointer. In any case, it was dead code that had to be eliminated.

PR #905, opened 2025-04-03
DM-49862: Switched to using a specific release of the cpp-httplib

iagaponenko:
Release v0.19.0 was made before a bug related to zstd was introduced in v0.20.0.


fritzm (approved):
LGTM; thanks!

PR #906, opened 2025-04-07
DM-49932: Create the prototype tables as the MyISAM tables to stay in synch with the data tables

iagaponenko:


jgates108 (approved):
Looks good.

PR #909, opened 2025-04-11
DM-50005: The temporary fix for the worker crash in the XROOTD/SSI stack

iagaponenko:
See XROOTD issue: https://github.com/xrootd/xrootd/issues/2469


fritzm (approved):
LGTM, thanks!

PR #910, opened 2025-04-11
DM-50057: Remove the XROOTD-based result delivery path from Qserv

iagaponenko:


src/proto/worker.proto
```diff
@@ -99,8 +99,7 @@ message ResponseSummary {
     optional uint32 rowcount = 6 [default = 0];
     optional uint64 transmitsize = 7 [default = 0];
     optional int32 attemptcount = 8 [default = 0];
-    optional string fileresource_xroot = 9 [default = ""];
-    optional string fileresource_http = 10 [default = ""];
+    required string fileresource_http = 11;
```
jgates108, line 102: I think that probably should have been left optional. A lot of people seem to have issues with required. It probably doesn't make much difference for us.

jgates108 (approved):
Looks good.

src/proto/worker.proto
iagaponenko: Good point! I will turn it back into `optional`. I also wanted to renumber the field to trigger the version checking during the message translation at Czar.

PR #911, opened 2025-04-11
Tickets/dm 49661

jgates108:
Changes made to simplify code, use jemalloc, remove library duplication, and help track memory usage.


src/CMakeLists.txt
```diff
@@ -21,7 +21,7 @@ find_package(Protobuf REQUIRED)
 find_package(pybind11 REQUIRED)
 find_package(Threads REQUIRED)
 find_package(XRootD REQUIRED)
-find_package(httplib REQUIRED)
+#&&& find_package(httplib REQUIRED)
```
jgates108, line 24: I believe this was a problem on main as well. I think and older version was used that had everything find_package was looking for?

iagaponenko: This works fine in the main branch. The problem was caused by the underlying shift in one of the dependencies of `httplib`. It was addressed in Qserv release: https://github.com/lsst/qserv/releases/tag/2025.4.1-rc2

The problem was analyzed and fixed in this ticket: https://rubinobs.atlassian.net/browse/DM-49862

iagaponenko: One way this can be addressed is by bringing the change into your development branch by:
```
git cherry-pick 96b68f41bbfcc02cdf78b14f8fa4ad1c7a4ea29f
```

iagaponenko (commented):
I posted a comment regarding `httplib` in `Dockerfile`

src/util/Error.h
```diff
@@ -74,8 +74,11 @@ struct ErrorCode {
  */
 class Error {
 public:
-    Error(int code = ErrorCode::NONE, std::string const& msg = "", int status = ErrorCode::NONE,
-          bool logLvLErr = true);
+    Error(int code, std::string const& msg = "", int status = ErrorCode::NONE, bool logLvLErr = true);
```
iagaponenko, line 77: You may want to add `explicit`:
```
  explicit Error(int code, std::string const& msg = "", int status = ErrorCode::NONE, bool logLvLErr = true);
```
Normally, you use `explicit` in all cases where you have a single parameter constructor because C++ may use the constructor as a type conversion operator. In this particular case, if you has some method accepting:
```
void foo(Error err) {
...
```
Then the conversion constructor would let you do:
```
foo(123);
```
Which would be the same as:
```
foo(Error(123));
```
You may not want this ... unless you rely on such a form of type conversion on purpose. These conversions may be a dangerous thing.

src/util/Error.h
```diff
@@ -101,9 +104,9 @@ class Error {
     bool isNone() { return (_code == util::ErrorCode::NONE); }
 
 private:
-    int _code;
-    std::string _msg;
-    int _status;
+    int _code = ErrorCode::NONE;
+    std::string _msg{""};
```
iagaponenko, line 108: You don't need this form of initialization because it's equivalent to:
```
std::string _msg;
```
C++ strings are constructed the valid empty strings.

src/util/InstanceCount.h
```diff
@@ -4,12 +4,28 @@
...
+
+class InstanceCountData {
+    InstanceCountData();
+    ~InstanceCountData();
+
+    friend InstanceCount;
+    friend std::ostream& operator<<(std::ostream& out, InstanceCount const& instanceCount);
+
+private:
+    std::map<std::string, int64_t> _instances;  //< Map of instances per class name.
```
iagaponenko, line 24: I think that the valid Doxygen comment is:
```
///< ...
```

src/util/InstanceCount.h
```diff
@@ -4,12 +4,28 @@
...
+#include <atomic>
 #include <map>
 #include <mutex>
 #include <string>
 
 namespace lsst::qserv::util {
 
+class InstanceCount;
+
+class InstanceCountData {
```
iagaponenko, line 16: If you don't have a use of this class in the public interface, then you may do:
```
class InstanceCountData;
```
Then push the class into the CC file to make the API of `InstanceCount` clean of the implementation-specific artifacts.

src/util/InstanceCount.h
```diff
@@ -26,9 +42,8 @@ class InstanceCount {
     friend std::ostream& operator<<(std::ostream& out, InstanceCount const& instanceCount);
 
 private:
-    std::string _className;                        //< Names of the of which this is a member.
-    static std::map<std::string, int> _instances;  //< Map of instances per class name.
-    static std::recursive_mutex _mx;               //< Protects _instances.
+    std::string _className;            ///< Name of instance being counted.
+    static InstanceCountData _icData;  ///< Map of counts and other data.
```
iagaponenko, line 46: Data members are normally declared after the methods, at the bottom of a class.

src/util/InstanceCount.cc
```diff
@@ -18,8 +21,17 @@ LOG_LOGGER _log = LOG_GET("lsst.qserv.util.InstanceCount");
 
 namespace lsst::qserv::util {
 
-std::map<std::string, int> InstanceCount::_instances;
-std::recursive_mutex InstanceCount::_mx;
+InstanceCountData InstanceCount::_icData;
+
+InstanceCountData::InstanceCountData() {
+    std::cout << "InstanceCountData " << " mx=" << (void*)(&_mx) << " _inst=" << (void*)(&_instances)
```
iagaponenko, line 27: Do you really want to post these messages to `std::cout` instead of the LSST Logger stream?

src/util/InstanceCount.cc
```diff
@@ -18,8 +21,17 @@ LOG_LOGGER _log = LOG_GET("lsst.qserv.util.InstanceCount");
...
+              << " t=" << (void*)(this) << endl;
+}
+
+InstanceCountData::~InstanceCountData() {
+    cout << "~InstanceCountData " << " mx=" << (void*)(&_mx) << " _inst=" << (void*)(&_instances)
```
iagaponenko, line 32: Sa ecomment for `std::cout` versus LSST Logger as above.

src/util/InstanceCount.cc
```diff
@@ -28,46 +40,73 @@ InstanceCount::InstanceCount(InstanceCount const& other) : _className{other._cla
 InstanceCount::InstanceCount(InstanceCount&& origin) : _className{origin._className} { _increment("mov"); }
 
 void InstanceCount::_increment(std::string const& source) {
-    std::lock_guard<std::recursive_mutex> lg(_mx);
+    std::lock_guard lg(_icData._mx);
```
iagaponenko, line 43: You don't have to use the prefix `std::` since you already did his earlier:
```
using namespace std;
```

src/util/InstanceCount.cc
```diff
@@ -28,46 +40,73 @@ InstanceCount::InstanceCount(InstanceCount const& other) : _className{other._cla
...
+    }
     std::pair<std::string const, int> entry(_className, 0);
-    auto ret = _instances.insert(entry);
+    auto ret = _icData._instances.insert(entry);
     auto iter = ret.first;
     iter->second += 1;
     LOGS(_log, LOG_LVL_TRACE, "InstanceCount " << source << " " << iter->first << "=" << iter->second);
+    if ((++(_icData._instanceLogLimiter)) % 10000 == 0) {
+        LOGS(_log, LOG_LVL_DEBUG, "InstanceCount brief " << *this << " icD=" << (void*)(&_icData));
+    }
```
iagaponenko, line 62: What is the purpose of all these changes in the method `InstanceCount::_increment()`?

src/util/MultiError.cc
```diff
@@ -54,6 +54,22 @@ std::string MultiError::toOneLineString() const {
...
+    return os.str();
+}
+
+util::Error MultiError::firstError() const {
+    Error err;
+    if (!empty()) {
+        err = _errorVector.front();
+    }
+    return err;
+}
```
iagaponenko, line 71: I would rewrite this method to be more concise:
```
util::Error MultiError::firstError() const {
    return empty() ? Error() : _errorVector.front();
}
```

src/util/MultiError.cc
```diff
@@ -54,6 +54,22 @@ std::string MultiError::toOneLineString() const {
 
 int MultiError::firstErrorCode() const { return empty() ? ErrorCode::NONE : _errorVector.front().getCode(); }
 
+string MultiError::firstErrorStr() const {
+    ostringstream os;
+    if (!empty()) {
+        os << _errorVector.front();
+    }
+    return os.str();
+}
```
iagaponenko, line 63: I would rewrite and speed up tfis metod as:
```
string MultiError::firstErrorStr() const {
    if (empty()) return string();
    stringstream os;
    os << _errorVector.front();
    return os.str();
}
```
This optimization will instantiate/constrict the stream object on the stack only if the collection is not empty. In most cases, it's empty.

src/util/ThreadPool.h
```diff
@@ -83,9 +83,9 @@ class CommandForThreadPool : public CommandTracked {
 public:
     using Ptr = std::shared_ptr<CommandForThreadPool>;
 
-    CommandForThreadPool() = default;
+    CommandForThreadPool() : CommandTracked() {}
```
iagaponenko, line 86: What's the reason for this change? It's the same as `= default`

src/wbase/Task.cc
```diff
@@ -427,13 +427,19 @@ wpublish::QueryStatistics::Ptr Task::getQueryStats() const {
...
         // Was already cancelled.
         return;
     }
 
-    LOGS(_log, LOG_LVL_DEBUG, "Task::cancel " << getIdStr());
+    if (logIt) {
+        if (!_ujData->getCancelled()) {
+            LOGS(_log, LOG_LVL_DEBUG, "Task::cancel " << getIdStr() << " UberJob still live.");
+        } else {
+            LOGS(_log, LOG_LVL_TRACE, "Task::cancel " << getIdStr());
```
iagaponenko, line 440: Just curious. What's the reason behind using different logging levels for checking the cancellation status of an object?

src/wbase/Task.cc
```diff
@@ -454,13 +460,12 @@ bool Task::checkCancelled() {
...
-/// @return true if task has already been cancelled.
-bool Task::setTaskQueryRunner(TaskQueryRunner::Ptr const& taskQueryRunner) {
+bool Task::setTaskQueryRunner(wdb::QueryRunner::Ptr const& taskQueryRunner) {
     _taskQueryRunner = taskQueryRunner;
     return checkCancelled();
 }
 
-void Task::freeTaskQueryRunner(TaskQueryRunner* tqr) {
+void Task::freeTaskQueryRunner(wdb::QueryRunner* tqr) {
     if (_taskQueryRunner.get() == tqr) {
```
iagaponenko, line 469: The row pointer-based comparison (as well as any uses of the C-style row pointers in C++) makes me feel uneasy 😄 I always find it safer to compare objects by their identities, like:
```
if (objA.id() == objB.id()) {
   ...
```

src/wbase/Task.h
```diff
@@ -204,17 +198,17 @@ class Task : public util::CommandForThreadPool {
...
     /// Check if this task should be cancelled and call cancel() as needed.
     /// @return true if this task was or needed to be cancelled.
     bool checkCancelled();
 
     TaskState state() const { return _state; }
     std::string getQueryString() const;
-    bool setTaskQueryRunner(
-            TaskQueryRunner::Ptr const& taskQueryRunner);  ///< return true if already cancelled.
-    void freeTaskQueryRunner(TaskQueryRunner* tqr);
+    /// Return true if already cancelled.
```
iagaponenko, line 209: To stay consistent in your coding style you may do:
```
    /// @return true if already cancelled.
```

src/wbase/UserQueryInfo.h
```diff
@@ -32,6 +32,7 @@
 
 // Qserv headers
 #include "global/intTypes.h"
+#include "util/InstanceCount.h"
```
iagaponenko, line 35: Why this change here? I have inspected the header of `UserQueryInfo` and I'm not seeing any uses of `util::InstanceCount` here.

iagaponenko (approved):
Looks fine to me. I have posted my comments on the code.

src/qdisp/Executive.cc
```diff
@@ -392,8 +390,14 @@ void Executive::markCompleted(JobId jobId, bool success) {
...
-        LOGS(_log, LOG_LVL_WARN,
-             "Executive: error executing " << err << " (status: " << err.getStatus() << ")");
+
+        {
+            lock_guard<mutex> lock(_errorsMutex);
+            errStr = _multiError.firstErrorStr();
+            err = _multiError.firstError();
+        }
+
+        LOGS(_log, LOG_LVL_DEBUG, "Executive: error executing " << err);
```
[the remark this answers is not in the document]
jgates108: `if (!success && !isRowLimitComplete()) {` means there was a problem. I am really hoping production systems have error logging set at INFO or WARN. Production absolutely should not be set DEBUG unless actively hunting down a problem. Also, `err` is exteremly likely to be true, so being empty could be useful.

src/qdisp/JobDescription.cc
```diff
@@ -48,19 +48,17 @@ LOG_LOGGER _log = LOG_GET("lsst.qserv.qdisp.JobDescription");
...
           _queryId(qId),
           _jobId(jobId),
           _qIdStr(QueryIdHelper::makeIdStr(_queryId, _jobId)),
           _resource(resource),
-          _respHandler(respHandler),
           _chunkQuerySpec(chunkQuerySpec),
-          _chunkResultName(chunkResultName),
           _mock(mock) {}
 
+JobDescription::~JobDescription() { LOGS(_log, LOG_LVL_TRACE, "~JobDescription()"); }
```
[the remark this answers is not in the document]
jgates108: Changing the log level to trace has a very similar effect, but getting rid of them is fine.

src/replica/apps/QservXrootdSsiApp.cc
```diff
@@ -0,0 +1,206 @@
+/*
```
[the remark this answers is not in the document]
jgates108: I suspect it will go when xrootd is eliminated. At this point, I know almost nothing about it, and I'm not sure where it came from. For the most part, I've been trying to avoid touching replica.

src/rproc/InfileMerger.h
```diff
@@ -227,6 +229,8 @@ class InfileMerger {
     bool _applySqlLocal(std::string const& sql, sql::SqlResults& results);
     bool _applySqlLocal(std::string const& sql, sql::SqlResults& results, sql::SqlErrorObject& errObj);
     bool _sqlConnect(sql::SqlErrorObject& errObj);
+
+    util::InstanceCount _icIm{"InfileMerger"};
```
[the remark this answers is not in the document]
jgates108: Done

src/util/Command.h
```diff
@@ -88,7 +88,7 @@ class Command {
 class CommandTracked : public Command, public Tracker {
 public:
     using Ptr = std::shared_ptr<CommandTracked>;
-    CommandTracked() = default;
+    CommandTracked() : Command(), Tracker() {}
```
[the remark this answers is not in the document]
jgates108: I saw some behavior that looked like a base class destructor wasn't being called for a different class. It was a pure virtual base class, and I may have been mistaken, but it make me nervous. So, verbose constructor for a class that is used a lot with multiple inheritance.

src/util/Error.h
jgates108: Good point.

src/util/InstanceCount.cc
jgates108: They need to go to `cout` as the instances are built statically and there's no guarantee that they will be called after `log` has been defined. Their only purpose is to detect that the object has been defined more than once, which was actually pretty useful as there wound up being multiple copies be made due to linker problems.

src/util/InstanceCount.cc
jgates108: Watching something like `Task` or `RowBuffer`, they can have hundreds of millions of instances. Generally, I've found I only care exactly how many instances of something exist if there are less than 5 of them, and I really care when they drop to 0. So, this takes a brief snap shot of everything being counted every 10k times an object being tracked is created. So when creating 25,000 Tasks for a userquery, you only get 2 log messages instead of 25000. Logging all 25,000 would be devestating to performance.

The destructor has code that limits logging to according to how many instances will be left, with 0 being really important. 

Without the limitations, you can't use `InstanceCount` with `Task` or `RowBuffer` as there's so much being logged with millions of them being created and destroyed that system performance tanks badly.

src/util/ThreadPool.h
jgates108: Same as the similar comment before.

src/wbase/Task.cc
jgates108: Mostly, it's trying to keep the log files under control. There can be tens of thousands of `Tasks` on a worker for a single User Query. If the user cancelled the query, or the Tasks are being cancelled because LIMIT has been reached, is really unlikely that the cancel log message will be useful and there will be a huge number of them.

On the other hand, `cancel()` being called for any other reason could be incredibly useful for debugging. Trying to keep the useful log messages without the junk messages is a bit of a trick.

src/wbase/Task.cc
jgates108: I'll add a comment. The pointer is just there to verify the correct thing is being deleted. If `tqr` doesn't match the pointer inside `_taskQueryRunner` nothing is deleted.
It's using the naked pointer as it's being stored inside a lambda function, and using a shared pointer there could cause lifetime issues.

src/wbase/Task.h
jgates108: I have found that putting @return on the first line eliminates the brief statement, which is annoying.

src/wbase/UserQueryInfo.h
jgates108: Actually, I want to track `UserQueryInfo` as there aren't that many of them and they have a relatively complicated life cycle. Good catch.

src/util/Command.h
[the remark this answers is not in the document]
iagaponenko: I don't think there is a reason to be concerned about the default construction. The C++ documentation on the `implicitly` defined (`=default` is the same) default constructors https://en.cppreference.com/w/cpp/language/default_constructor says the following:
```
An(until C++26)Otherwise, an(since C++26) implicitly-defined default constructor has the same effect as a user-defined constructor with empty body and empty initializer list. That is, it calls the default constructors of the bases and of the non-static members of this class. Class types with an empty user-provided constructor may get treated differently than those with an implicitly-defined default constructor during [value initialization](https://en.cppreference.com/w/cpp/language/value_initialization).
```
Based on this explanation, I would rather NOT define my own default constructor.

src/wbase/Task.h
iagaponenko: Interesting. I didn't know that. Did you see this on the Doxygen-generated documentation or in your IDE (Eclipse)?

PR #920, opened 2025-04-28
Added JobReadyMsg.

jgates108:
JobReadyMsg was created to move message creation and parsing to protojson. WorkerCzarComIssue was moved to its own files.


src/protojson/JobReadyMsg.h
```diff
@@ -0,0 +1,99 @@
...
+#include "global/clock_defs.h"
+#include "global/intTypes.h"
+#include "protojson/WorkerQueryStatusData.h"
+
+// This header declarations
+namespace lsst::qserv::protojson {
+
+/// This class handles the message used to inform the czar that a result file
+/// for an UberJob is ready.
+class JobReadyMsg {
```
iagaponenko, line 44: Do you still have separate classes `Job` and `UberJob` in the design? If so then a better name for class could be either `ResultReadyMsg` or `UberJobResultReady`.

src/protojson/JobReadyMsg.h
```diff
@@ -0,0 +1,99 @@
...
+public:
+    using Ptr = std::shared_ptr<JobReadyMsg>;
```
iagaponenko, line 46: My personal preference is to avoid using those `JobReadyMsg::Ptr` in favor of `std::shared_ptr<JobReadyMsg>`.

src/protojson/JobReadyMsg.cc
```diff
@@ -0,0 +1,116 @@
...
+LOG_LOGGER _log = LOG_GET("lsst.qserv.protojson.JobReadyMsg");
+}  // namespace
+
+namespace lsst::qserv::protojson {
+
+JobReadyMsg::Ptr JobReadyMsg::create(std::string const& replicationInstanceId,
+                                     std::string const& replicationAuthKey, std::string const& workerIdStr,
+                                     std::string const& czarName, CzarIdType czarId, QueryId queryId,
+                                     UberJobId uberJobId, std::string const& fileUrl, uint64_t rowCount,
+                                     uint64_t fileSize) {
```
iagaponenko, line 50: You may get rid of `std::` in the CC file since you already have:
```
line 37: using namespace std;
```

src/protojson/JobReadyMsg.cc
```diff
@@ -0,0 +1,116 @@
...
+    jrMsg->_czarId = czarId;
+    jrMsg->_queryId = queryId;
+    jrMsg->_uberJobId = uberJobId;
+    jrMsg->_fileUrl = fileUrl;
+    jrMsg->_rowCount = rowCount;
+    jrMsg->_fileSize = fileSize;
+    return jrMsg;
+}
+
+JobReadyMsg::Ptr JobReadyMsg::createFromJson(nlohmann::json const& jsWReq,
```
iagaponenko, line 63: You may get rid of `nlohmann::` here since you already have:
```
line 38: using namespace nlohmann;
```

src/protojson/JobReadyMsg.cc
```diff
@@ -0,0 +1,116 @@
...
+        jrMsg->_uberJobId = http::RequestBodyJSON::required<UberJobId>(jsWReq, "uberjobid");
+        jrMsg->_fileUrl = http::RequestBodyJSON::required<string>(jsWReq, "fileUrl");
+        jrMsg->_rowCount = http::RequestBodyJSON::required<uint64_t>(jsWReq, "rowCount");
+        jrMsg->_fileSize = http::RequestBodyJSON::required<uint64_t>(jsWReq, "fileSize");
+        return jrMsg;
+    } catch (invalid_argument const& exc) {
+        LOGS(_log, LOG_LVL_ERROR, string("JobReadyMsg::createJson invalid ") << exc.what());
+    }
+    return nullptr;
+}
```
iagaponenko, line 93: This is not a good code - the factory method should be simple. What you are doing here has to be done by another (overloaded) c-tor of the class. Something like:
```
JobReadyMsg::JobReadyMsg(nlohmann::json const& jsWReq, ... );
```
So your factory method would simply do:
```
JobReadyMsg::Ptr JobReadyMsg::createFromJson(nlohmann::json const& jsWReq, ...) {
    try {
       return shared_ptr<JobReadyMsg>(new JobReadyMsg(jsWReq, ...);
    } catch (invalid_argument const& exc) {
        LOGS(_log, LOG_LVL_ERROR, string("JobReadyMsg::createJson invalid ") << exc.what());
            return nullptr;
    }
}
```
The factory method should not be aware too much about the internals of the class construction and parameter validation which is normally a responsibility of constructors.

src/protojson/JobReadyMsg.cc
```diff
@@ -0,0 +1,116 @@
...
+
+json JobReadyMsg::serializeJson() {
```
iagaponenko, line 95: The name doesn't say what the method does - it converts the object into the `json` representation. Note that no serialization is happening here since the JSON object is the first-class citizen of the C++ type system. The serialization is happening when you are calling:
```
json::dump()
```
Hence a better name would be:
```
json JobReadyMsg::toJson()
```

src/protojson/JobReadyMsg.cc
```diff
@@ -0,0 +1,116 @@
...
+    shared_ptr<json> jsJrReqPtr = make_shared<json>();
+    json& jsJr = *jsJrReqPtr;
```
iagaponenko, line 97: I'm very confused about what you are doing in the last two lines, and why you need to make the shared pointer instead of having a simple construction on return:
```
return json::object({
    {"instance_id", _replicationInstanceId},
    {"auth_key", _replicationAuthKey},
...
}); 
```

src/protojson/JobReadyMsg.h
```diff
@@ -0,0 +1,99 @@
...
+
+    /// This function creates a JobReadyMsg object from the worker json `czarJson`, the
+    /// other parameters are used to verify the json message.
+    static Ptr createFromJson(nlohmann::json const& czarJson, std::string const& replicationInstanceId,
+                              std::string const& replicationAuthKey);
+
+    ~JobReadyMsg() = default;
+
+    /// Return a json object with data allowing collection of UberJob result file.
+    nlohmann::json serializeJson();
```
iagaponenko, line 73: A better name for the method (see earlier comment) would be `toJson` (like `std::to_string(<>)`, `std::strtoull(...)`, etc.).

The second comment is that the method should be `const` since it's not supposed to modify the state of the object.

src/protojson/CMakeLists.txt
```diff
@@ -1,8 +1,11 @@
 add_library(protojson SHARED)
 
 target_sources(protojson PRIVATE
+    JobErrorMsg.cc
+    JobReadyMsg.cc
 	ScanTableInfo.cc
     UberJobMsg.cc
+    WorkerCzarComIssue.cc
```
iagaponenko, line 8: Check the indentation of the lines in this section. Line number 6 (for `ScanTableInfo.cc`) has 4 more spaces .

src/protojson/JobErrorMsg.h
```diff
@@ -0,0 +1,96 @@
...
+#include "global/clock_defs.h"
+#include "global/intTypes.h"
+#include "protojson/WorkerQueryStatusData.h"
+
+// This header declarations
+namespace lsst::qserv::protojson {
+
+/// This class handles the message used to inform the czar that there has
+/// been a problem with an UberJob.
+class JobErrorMsg {
```
iagaponenko, line 44: Is this class for a "job" or for an "uber-job"? According to the comment, it's the latter. So, perhaps a better name would be `UberJobErrorMsg`?

src/protojson/JobErrorMsg.h
```diff
@@ -0,0 +1,96 @@
...
+    static Ptr createFromJson(nlohmann::json const& czarJson, std::string const& replicationInstanceId,
+                              std::string const& replicationAuthKey);
+
+    ~JobErrorMsg() = default;
+
+    /// Return a json object with data allowing collection of UberJob result file.
+    nlohmann::json serializeJson();
+
+    std::string getWorkerId() const { return _workerId; }
+    std::string getCzarName() const { return _czarName; }
```
iagaponenko, line 75: Same comments on the return types for strings, which could be:
```
std::string const&
```

src/protojson/JobErrorMsg.h
```diff
@@ -0,0 +1,96 @@
...
+
+    /// This function creates a JobErrorMsg object from the worker json `czarJson`, the
+    /// other parameters are used to verify the json message.
+    static Ptr createFromJson(nlohmann::json const& czarJson, std::string const& replicationInstanceId,
+                              std::string const& replicationAuthKey);
+
+    ~JobErrorMsg() = default;
+
+    /// Return a json object with data allowing collection of UberJob result file.
+    nlohmann::json serializeJson();
```
iagaponenko, line 72: The same comment on the name and qualifier of the method as for the class `JobReadyMsg` - it's better to be:
```
nlohmann::json toJson() const;
```

src/protojson/JobErrorMsg.cc
```diff
@@ -0,0 +1,111 @@
...
+LOG_LOGGER _log = LOG_GET("lsst.qserv.protojson.JobErrorMsg");
+}  // namespace
+
+namespace lsst::qserv::protojson {
+
+JobErrorMsg::Ptr JobErrorMsg::create(std::string const& replicationInstanceId,
+                                     std::string const& replicationAuthKey, std::string const& workerIdStr,
+                                     std::string const& czarName, CzarIdType czarId, QueryId queryId,
+                                     UberJobId uberJobId, int errorCode, std::string const& errorMsg) {
+    auto jrMsg = Ptr(new JobErrorMsg(replicationInstanceId, replicationAuthKey));
```
iagaponenko, line 50: You may eliminate `std::`

iagaponenko: Youd need to add another specialized constructor to parse `json` and construct from `json` as I explained for class `JobReadyMsg`.

src/protojson/JobErrorMsg.cc
```diff
@@ -0,0 +1,111 @@
...
+        return jrMsg;
+    } catch (invalid_argument const& exc) {
+        LOGS(_log, LOG_LVL_ERROR, string("JobErrorMsg::createJson invalid ") << exc.what());
+    }
+    return nullptr;
+}
+
+json JobErrorMsg::serializeJson() {
+    shared_ptr<json> jsJrReqPtr = make_shared<json>();
+    json& jsJr = *jsJrReqPtr;
```
iagaponenko, line 93: Same strange pattern of creating the pointer.

Do you have it in the rest of the new Qserv code? If so then please get rid of this pattern. Pointers should be avoided. There are very little benefits of using those in Qserv.

src/protojson/WorkerCzarComIssue.h
```diff
@@ -0,0 +1,131 @@
...
+/// the czar.
+/// Currently, this message is expected to only be needed rarely.
+class WorkerCzarComIssue {
+public:
+    using Ptr = std::shared_ptr<WorkerCzarComIssue>;
+
+    WorkerCzarComIssue() = delete;
+    ~WorkerCzarComIssue() = default;
+
+    std::string cName(const char* funcN) { return std::string("WorkerCzarComIssue") + funcN; }
```
iagaponenko, line 62: I already commented a similar function in the context of the class `JobReadyMsg`.
Firstly, I would hide the implementation of the method in the CC file.
Secondly, the name cName is meaningless (to me)
Thirdly, the raw pointers are unsafe. I would at least check if the pointer is not `nullptr` and throw an exception to reinforce the application and avoid SIGFAILT should the null be found here.

src/protojson/WorkerCzarComIssue.h
```diff
@@ -0,0 +1,131 @@
...
+        // TODO:UJ or list of failed transmits not empty.
+        return _thoughtCzarWasDead;
+    }
+
+    /// Set the contact information for the appropriate czar and worker.
+    void setContactInfo(WorkerContactInfo::Ptr const& wInfo_, CzarContactInfo::Ptr const& czInfo_) {
+        std::lock_guard lgWci(_wciMtx);
+        if (_wInfo == nullptr && wInfo_ != nullptr) _wInfo = wInfo_;
+        if (_czInfo == nullptr && czInfo_ != nullptr) _czInfo = czInfo_;
+    }
```
iagaponenko, line 90: Why have the online implementation of these methods in the header? This makes headers way too long and unreadable. Please note that in the modular programming paradigm (C++ and many other languages), _declarations_ is a way to define a contract of a class. Meanwhile, you would push all implementation details into the CC file. This makes your code readable and observable. Besides, it may as well speed up the compilation time of the project since headers may be included everywhere. More stuff you put into the header - longer the compilation could be.

src/protojson/WorkerCzarComIssue.h
```diff
@@ -0,0 +1,131 @@
...
+        std::lock_guard lgWci(_wciMtx);
+        return _wInfo;
+    }
+
+    /// Return a json version of the contents of this class.
+    std::shared_ptr<nlohmann::json> serializeJson();
+
+    /// Return a json object indicating the status of the message for the
+    /// original requester.
+    nlohmann::json serializeResponseJson();
```
iagaponenko, line 107: This needs to be:
```
nlohmann::json responseToJson() const;
```

src/protojson/WorkerCzarComIssue.h
```diff
@@ -0,0 +1,131 @@
...
+        return _czInfo;
+    }
+
+    WorkerContactInfo::Ptr getWorkerInfo() const {
+        std::lock_guard lgWci(_wciMtx);
+        return _wInfo;
+    }
+
+    /// Return a json version of the contents of this class.
+    std::shared_ptr<nlohmann::json> serializeJson();
```
iagaponenko, line 103: This has to be:
```
nlohmann::json toJson() const;
```

iagaponenko (approved):
Looks good. My main concerns are regarding coding style issues that I found in the PR. I think addressing those would make the code more maintainable in the long run.

src/protojson/JobReadyMsg.h
```diff
@@ -0,0 +1,99 @@
...
+public:
+    using Ptr = std::shared_ptr<JobReadyMsg>;
+
+    JobReadyMsg() = delete;
+    JobReadyMsg(std::string const& replicationInstanceId, std::string const& replicationAuthKey)
+            : _replicationInstanceId(replicationInstanceId), _replicationAuthKey(replicationAuthKey) {}
+    JobReadyMsg(JobReadyMsg const&) = delete;
+    JobReadyMsg& operator=(JobReadyMsg const&) = delete;
+
+    std::string cName(const char* fName) { return std::string("WorkerQueryStatusData::") + fName; }
```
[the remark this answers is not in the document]
jgates108: The point of this is it is a short function name to get useful and consistent logging information in a compact form. A long, descriptive member function name defeats that purpose. The normal argument is expected to be `__func__`.  Checking that against nullptr would be nice, but there's really no quick way to verify that `__func__` is giving you a valid string or that the pointer is at all reasonable. It would be great if there was a sting equivalent of `__func__`. Printing out internal information for logging is one of those things C++ doesn't do well. I'm still thinking about what the best course of action really is here.

src/protojson/WorkerCzarComIssue.h
jgates108: I agree with const and name change. However, it's using a pointer as the json gets passed around as a function argument. It's very easy to make sure that a relatively large object isn't being copied when you're using a pointer to that object.

src/protojson/JobReadyMsg.h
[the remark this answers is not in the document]
iagaponenko: Then (if the intent is to use `__func__`), what you might write here would be:
```
static
std::string context(std::string const& func) const {
    return "WorkerQueryStatusData::" + func;
}
```
Note `static`.

And if you're only going to use that "context" idea in the CC file then you can always use C macro:
```
#define CONTEXT_ ("WorkerQueryStatusData::" + std::string(__func__) + " ")
```
Then you could use this anywhere in the CC file. For example:
```
void WorkerQueryStatusData::foo() {
    throw std::runtime_error(CONTEXT_ + "something bad just has happened");
}
```
The right value of `__func__` would be deduced by the microprocessor/compiler.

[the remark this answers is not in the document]
iagaponenko: That `std::string cName(const char* fName)` is a bad (from my point of view) coding pattern introduced many years ago in Qserv. And it's surprisingly alive. I suspect it came here from XROOTD. At least I could see a lot of these things in the XROOTD code last time I was inspecting some crashes in there. XROOTD was written basically in C (not even in C++).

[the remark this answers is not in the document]
iagaponenko: One more comment - methods like `cName` should not be in the public API of the class as they are meant to be used for internal logging (throwing exceptions). The implementation generates a string that's tied to a specific class. This is another argument in favor of the macro definition `CONTEXT_` in the CC file rather than having this method in the public API of a class. The public API is for other classes and for code readers.

src/protojson/WorkerCzarComIssue.h
iagaponenko: Agreed. Then it makes sense.

src/protojson/JobReadyMsg.cc
jgates108: I don't really get your argument here. The factory method is a member of the class and keeps the internals  encapsulated. I'm changing the code, but the reason that a factory method must be incredibly simple isn't clear to me (and this really isn't complicated). If the factory function wasn't a member of the class, I would agree due to breaking encapsulation, but that isn't the case.

I usually think constructors need to be simple since you're usually trying to get everything defined in the block before the constructor body.

iagaponenko: The basic rule for what one can do in the factory method and what can not is - the factory method should do what can't be done in the constructors.

src/protojson/JobReadyMsg.h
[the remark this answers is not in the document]
iagaponenko: You may see an example of using `CONTEXT_` macro in: https://github.com/lsst/qserv/blob/main/src/http/AsyncReq.cc

Another good example in Qserv where you can see the benefits of the mactros is the query parser: 
https://github.com/lsst/qserv/blob/main/src/ccontrol/ParseListener.cc

Or the macro `EMPTY_STR_IF_NULL` in https://github.com/lsst/qserv/blob/main/src/sql/SqlResults.cc

I'm not insisting that macros should be used all around. However, there are use cases for those.

PR #922, opened 2025-05-04
DM-50679: Fixed race conditions in the query cancelation logic

iagaponenko:


src/ccontrol/UserQuerySelect.h
```diff
@@ -152,7 +152,10 @@ class UserQuerySelect : public UserQuery {
 
     void _expandSelectStarInMergeStatment(std::shared_ptr<query::SelectStmt> const& mergeStmt);
 
-    void _discardMerger();
+    /// Discard the merger if it exists.
+    /// @param lock a lock on the mutex _killMutex that prevents the race condition if the query
+    /// cancellation happens at the same time the query resources are being discarded.
+    void _discardMerger(std::lock_guard<std::mutex> const& lock);
```
jgates108, line 158: Calling `.reset()` for various pointers in `discard()` for the children of `UserQuery` is dangerous and probably unnecessary in most cases, as `discard()` is called right before the `uq` is reset in the `finalizer`  in `Czar::submitQuery`, so they are about to be deleted anyway. It makes the code more complicated and the benefit isn't clear.

src/ccontrol/UserQuerySelect.cc
```diff
@@ -156,27 +156,29 @@ std::string UserQuerySelect::getError() const {
...
-        }
-        // Since this is being aborted, collectedRows and collectedBytes are going to
-        // be off a bit as results were still coming in. A rough idea should be
-        // good enough.
-        _qMetaUpdateStatus(qmeta::QInfo::ABORTED, collectedRows, collectedBytes, 0);
+    if (_killed) return;
+
+    // If either pointer is nullptr then it's too late for killing the query. The query
+    // has already been finished or it's in a process of beging finished.
+    if (_executive == nullptr || _infileMerger == nullptr) {
```
jgates108, line 166: While, in theory, the mutex should make this ok, it's better to copy the pointer before testing. That way, even if there's some place in the code where the value of the pointer got changed without locking the mutex, it will still be ok. Also, it may be possible to remove the mutex in some cases.
```
auto exec = _executive;
auto infM = _infileMerger;
if (exec == nullptr || infM == nullptr) {
```
Also, everything after this has to use `exec` and `infM` as `_executive` and `_infileMerger` can be changed by other threads. Since they can change, `_executive` and `_infileMerger` should probably be weak pointers, but that gets weird as their lifetime is controlled this class `UserQuerySelect`.

In the branch, I copied and checked `_executive` before using it, as it's a flag for the query being alive. It looks like I may need to do the copy and check for `_infileMerger` as well. Or just don't reset that and the other shared pointers.

jgates108 (approved):
I think it should work, but I think we'd be better off not reseting most of the shared pointers. I have an incomplete solution here [https://github.com/lsst/qserv/blob/tickets/DM-43715/src/ccontrol/UserQuerySelect.cc](https://github.com/lsst/qserv/blob/tickets/DM-43715/src/ccontrol/UserQuerySelect.cc) where there is at least one place I need to check the value of `_infileMerger` before use. (I was looking at a different problem when I made those changes).

Reconciling the different approaches during rebasing will not be fun.

src/ccontrol/UserQuerySelect.cc
iagaponenko: I must admit that I really hate this pattern of relying on a state of the shared pointers that I keep seeing in Qserv, at both sides (czar and worker). A more reliable and easy to comprehend design pattern is the FSM (Finite State Machine) approach where the object state is explicitly defined via some state variable. The state transitions must be well-defined and guarded by a mutex (atomics, etc.). Eventually, I would like to redesign all these classes based on the FSM paradigm. The redesign would require some effort to revisit the very design of these classes instead of making some minor cosmetic changes by pushing things around.

What I did in this PR was a fix that works in the current (fuzzy state approach based on the state of `_killed` and the shared pointers). My short-term intent is to make the production Qserv work now as we already have existing users.

src/ccontrol/UserQuerySelect.h
iagaponenko: The `.reset()` is already in the code. I don't want to change the logic flow of the existing code. My changes were just to prevent existing code from resetting the pointers while there is a chance that the query cancellation event may arrive. Hence I was very careful (not) to make any further mods to that code. Let's delay any dramatic changes to some significant redesign of this (and similar) classes.

src/ccontrol/UserQuerySelect.cc
jgates108: The finite state machine solution can easily become a nightmare in a multithread environment. That's what xrootd does instead of using shared pointers, it depends on keeping all states in sync. Keeping states in sync in with multiple threads is hard. When cancelling, keeping all states in sync gets really hard.

I've changed the code on the branch to get rid of all of the `.reset()` calls except for the executive and it's passed integration tests. I'm considering changing `_executive` to be a weak pointer to `_executiveLifetime`. The weak pointer would force a lock, making it safer.

jgates108: `_executive` may not need to be reset() either, but I'll need to take a closer look at that.

iagaponenko: The FSM pattern is not the nightmare, this is the best solution exactly for the multithreaded and non-deterministic environments. It was introduced as a technique to organize the code around well-defined states & transitions. The relevant resources (shared pointers, other data members) just follow the states. Once an object in a given state you always know what is the state of the relevant resources. A typical problem that programmers face when they begin implementing FSM is the amount of mental effort to be put into the design of a class. With FSM one has to start with designing a well-defined state machine before even trying to write a single line of code. Once the model is there the rest (the actual coding) is easy.

The class `UserQuerySelect` is a good example of what happens when the functional extension allowing to cancel a query was added as a "fix" rather than a redesign of the class. Hence the crashes.

It's okay if you want to revisit the timing when `_executive` is reset (or if it needs to be reset at all). This could be added in a separate PR if needed.

iagaponenko: Regarding some parallel changes that you made in your development branch then in the future, I would encourage you to evaluate these changes to see if they apply to the `main` branch as well. In this (later) case, the change would be made first in the `main` branch and propagated to your development branch via rebase. This would reduce differences between the production branch and your development branch of Qserv. This would be easier than bringing changes to `main` "from the future".

PR #926, opened 2025-05-16
Added worker executable.

jgates108:


src/wmain/WorkerMain.h
```diff
@@ -0,0 +1,89 @@
...
+class Foreman;
+}  // namespace lsst::qserv::wcontrol
+
+namespace lsst::qserv::wcomms {
+class HttpSvc;
+}  // namespace lsst::qserv::wcontrol
+
+namespace lsst::qserv::wmain {
+
+class WorkerMain {
```
iagaponenko, line 49: `WorkerHttpApp` would be a more meaningful name.

src/wmain/qserv-worker-http.cc
```diff
@@ -0,0 +1,81 @@
+/*
```
iagaponenko, line 1: Why make another module for one class? Why not to put this class and `qserv-worker-http` into an existing module `worker`?

admin/local/docker/compose/docker-compose.yml
```diff
@@ -98,6 +106,37 @@ services:
               source: volume_worker_0_mariadb_run
               target: /var/run/mysqld  # This is where the mariadb container puts the socket file
         network_mode: "service:worker-xrootd-0"
+    worker-wkr-0:
```
iagaponenko, line 109: Could be just `worker-0`

admin/local/docker/compose/docker-compose.yml
```diff
@@ -98,6 +106,37 @@ services:
...
+          --db-admin-uri mysql://root:CHANGEME@127.0.0.1:3306
+          --vnid-config "@/usr/local/lib64/libreplica.so {{db_uri}}/qservw_worker 0 0"
+          --repl-instance-id qserv_proj
+          --repl-auth-key replauthkey
+          --repl-admin-auth-key=repladminauthkey
+          --repl-registry-host repl-registry
+          --repl-registry-port 25082
+          --results-dirname /qserv/data/results
+          --mysql-monitor-password CHANGEME_MONITOR
+          --log-cfg-file=/config-etc/log/log-worker-xrootd.cnf
```
iagaponenko, line 123: If you remove `xrootd` then it makes sense to call the config file as simply as `log-worker.cnf`.

admin/local/docker/compose/docker-compose.yml
```diff
@@ -98,6 +106,37 @@ services:
               source: volume_worker_0_mariadb_run
               target: /var/run/mysqld  # This is where the mariadb container puts the socket file
         network_mode: "service:worker-xrootd-0"
+    worker-wkr-0:
+        << : *worker-wkr
+        command: >
+          entrypoint worker-wkr
+          --db-uri mysql://qsmaster:CHANGEME2@127.0.0.1:3306
+          --db-admin-uri mysql://root:CHANGEME@127.0.0.1:3306
+          --vnid-config "@/usr/local/lib64/libreplica.so {{db_uri}}/qservw_worker 0 0"
```
iagaponenko, line 115: `vnid` is the XROOTD-specific thing. Why do you want it here?

src/admin/python/lsst/qserv/admin/cli/entrypoint.py
```diff
@@ -224,6 +228,15 @@ class CommandInfo:
 )
 
 
+option_worker_wkr_cfg_path = partial(
+    click.option,
+    "--worker-wkr-cfg-path",
```
iagaponenko, line 233: Same comment as earlier on the redundancy in `worker-wrk`.

src/wcomms/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+add_library(wcomms SHARED)
```
iagaponenko, line 1: Why do you need this module? You may have these classes in the existing module `worker`

src/xrdsvc/SsiProvider.cc
```diff
@@ -43,7 +43,7 @@
 #include "global/ResourceUnit.h"
 #include "wconfig/WorkerConfig.h"
 #include "wpublish/ChunkInventory.h"
-#include "xrdsvc/XrdName.h"
+#include "wcomms/XrdName.h"
```
iagaponenko, line 46: I don't think you need `XrdName`. I'm not seeing any real use for this class and the environment `XRDNAME`. I think this mechanism is obsolete. I might be used in the very long past to allow many workers to coexist within the same MySQL database. The current integration test launches a dedicated MariaDB server for each worker. Hence the name of the worker's database is now fixed in all test or deployment environments:
```
qservw_worker
```

iagaponenko (commented):
I'm just posting comments on a few observations I made while looking through this PR.
And please - run `clang-format` before committing to GitHub.

admin/local/docker/compose/docker-compose.yml
jgates108: I went with `worker-wkr` as I can search on that and `worker_wkr` and find just about everything associated with it. Searching on `worker` returns far too many unrelated things. I'm not set on the name, but something that can be found with grep will make life easier in the future.

iagaponenko (approved):
Looks good. However, not all artifacts related to XROOTD may have been removed. For instance, class `wpublish::ChunkInventory` was meant to serve as a source of truth for which chunks are available to the worker. The inventory was mainly used by the SSI-based worker for publishing SSI resources and verifying which resources are valid.
I think `ChunkInventory` needs to be eliminated.
The only method of this class that still makes sense is `ChunkInventory::id()`. This method is needed by the worker to pull the unique identifier of the worker from the worker-side metadata table `qservw_worker.Id`. The identity is used in the Czar-worker protocol, in the worker monitoring, and elsewhere.
I think the whole class `ChunkInventory` can be replaced by a simpler class.

src/qmeta/QMetaMysql.cc
```diff
@@ -834,7 +834,9 @@ void QMetaMysql::addQueryMessages(QueryId queryId, shared_ptr<MessageStore> cons
 
     for (auto const& elem : msgCountMap) {
         if (elem.second.count > _maxMsgSourceStore) {
+            // QMessages source column is VARCHAR(63)
```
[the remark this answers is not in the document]
jgates108: The truncation happens here as the message may have distinct identiifiers that happen after the 63 character boundary. Those identifiers can be used to separate, sort, and count them (`msgCountMap`). It would be useful to put the extra characters in the database, but the information won't fit. I don't want to alter the database schema as it needs to be the same for both versions of qserv so we can easily switch between them.

Something seems to have gone sideways leading up to this point as the `source` became more of a message than the `desc` and this seems to be widespread and somewhat inconsistent. Making it consistent is beyond the scope of this ticket.

src/wmain/WorkerMain.cc
```diff
@@ -151,6 +151,14 @@ WorkerMain::Ptr WorkerMain::setup() {
     return ptr;
 }
 
+std::shared_ptr<WorkerMain> WorkerMain::get() {
+    auto ptr = _globalWorkerMain.lock();
```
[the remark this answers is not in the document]
jgates108: Primarily, for program termination and access to command line argument values. It would be nice to have a clean shutdown to find things like memory leaks.

I think it would be reasonable to have most or all values in wconfig setable via command line arguments and this may be useful for that.

src/wmain/qserv-worker-http.cc
jgates108: The things in `worker`  now have very little to do with anything in `wmain` or `wcomms`. I'm trying to look at this from the point of view of me trying to find something after 5 years have gone by without touching the code, or someone new. I'm still thinking over names and locations.

PR #940, opened 2025-06-27
DM-51585: The unconstrained COUNT(*) queries in Qserv are not optimized when LIMIT is present

iagaponenko:


fritzm (approved):
LGTM -- thanks!

PR #942, opened 2025-07-14
Changed FQDN calls to blocking or using stored value.

jgates108:
Changed FQDN calls to prevent crashes.

### DIFF
--- a/src/czar/CzarRegistry.cc
+++ b/src/czar/CzarRegistry.cc
@@ -80,13 +80,14 @@ void CzarRegistry::_registryUpdateLoop() {
     string const url = "http://" + _czarConfig->replicationRegistryHost() + ":" +
                        to_string(_czarConfig->replicationRegistryPort()) + "/czar";
     vector<string> const headers = {"Content-Type: application/json"};
+    string const fqdn = util::getCurrentHostFqdnBlocking();
     json const request = json::object({{"instance_id", _czarConfig->replicationInstanceId()},
                                        {"auth_key", _czarConfig->replicationAuthKey()},
                                        {"czar",
                                         {{"name", _czarConfig->name()},
                                          {"id", _czarConfig->id()},
                                          {"management-port", _czarConfig->replicationHttpPort()},
-                                         {"management-host-name", util::get_current_host_fqdn()}}}});
+                                         {"management-host-name", fqdn}}}});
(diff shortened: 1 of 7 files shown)


src/czar/CzarRegistry.cc
```diff
@@ -80,13 +80,15 @@ void CzarRegistry::_registryUpdateLoop() {
     string const url = "http://" + _czarConfig->replicationRegistryHost() + ":" +
                        to_string(_czarConfig->replicationRegistryPort()) + "/czar";
     vector<string> const headers = {"Content-Type: application/json"};
+    LOGS(_log, LOG_LVL_INFO, "CzarRegistry::_registryUpdateLoop() getting FQDN");
```
iagaponenko, line 83: Do you want to see this message in the elog stream? You could downgrade it to be at most `DEBUG` or maybe `TRACE`. A better option is not to pollute the code or the log stream with this.

src/czar/CzarRegistry.cc
```diff
@@ -80,13 +80,15 @@ void CzarRegistry::_registryUpdateLoop() {
...
+    string const fqdn = util::getCurrentHostFqdnBlocking();
```
iagaponenko, line 84: Why would you put this into a separate variable instead of feeding the value directly into:
```
     json const request = json::object(...,
        {"management-host-name", util::getCurrentHostFqdnBlocking()},
        ...
     );
```

iagaponenko (approved):
LGTM.

Please note that the integration test in this PR is failing. I don't see an immediate reason for the failure in any changes made in the context of this particular PR. Perhaps the base (feature?) branch has problems.

src/czar/CzarRegistry.cc
jgates108: That's in a loop that will get called a lot, and the function requires contacting a separate program which could block for a significant. By moving it, it's called when the program starts and never again.

src/czar/CzarRegistry.cc
iagaponenko: If the FQDN can't be obtained then an exception is thrown (by the called method). In this case, it doesn't matter from what context you're calling this. No one is going to investigate this. What would matter here is that there is a problem with the system/networking infrastructure.

PR #944, opened 2025-07-16
DM-51775: Special symbols in the database and/or table names need to be properly handled by Qserv

iagaponenko:


fritzm (approved):
LGTM -- thanks!

PR #947, opened 2025-07-22
DM-51700: Discarded result files at workers are not deleted automatically before query processing finishes

iagaponenko:


fritzm (approved):
LGTM -- thanks!

PR #950, opened 2025-07-25
DM-51963: Upgraded Apache Arrow and Parquet libs to version 21

iagaponenko:
Otherwise applications crash with:

  error while loading shared libraries: libarrow.so.2100: cannot open shared object file:
  No such file or directory


fritzm (approved):
LGTM -- thanks!

PR #951, opened 2025-07-30
DM-50683: Completed queries may not be removed from the in-progress metadata table of Qserv

iagaponenko:


fritzm (approved):
LGTM -- thanks!

PR #957, opened 2025-08-29
tickets/DM-52349

jgates108:
I decided to keep this as simple as possible and just move the functions to the Executive class as their are cases in new qserv where no attempt at merging is ever made.

Any significant changes to what is put into the database should probably wait until after we all meet to decided what we want to happen with qmeta and css.


iagaponenko:
Looks good to me.

PR #960, opened 2025-09-13
DM-48953: `ruff` Python linter fixes 

fritzm:
* Updates all qserv Python code to Python 3.12
* Introduce `ruff` linter and correct Python lint
* Fix for openssl MD5 deprecation warn during build

### DIFF
--- a/doc/misc/examples/obsolete/rebalance.py
+++ b/doc/misc/examples/obsolete/rebalance.py
@@ -86,7 +86,6 @@ class RebalanceProgram:
             parser.error("Must specify at least one path. (How about . ?)")
         self._pathList = args
         self._opts = opts
-        pass
 
     def _gather_chunk_files(self):
         chunk_files = chain(*imap(loader.find_chunk_files, self._pathList, repeat(self._opts.prefix)))
@@ -100,7 +99,6 @@ class RebalanceProgram:
                 if name in check_dupe:
                     raise RuntimeError(f"Found 2 identically named chunk files named {name}")
                 check_dupe.add(name)
-        pass
 
     def _move(self):
         chunk_count = len(self._chunkFiles)
@@ -122,7 +120,6 @@ class RebalanceProgram:
                 os.listdir(targetdir)
             except OSError:
                 os.makedirs(targetdir)
-                pass
 
             for j in range(start, stop):
                 for f in self._chunkFiles[j]:
@@ -130,7 +127,6 @@ class RebalanceProgram:
                     print("rename ", f, os.path.join(targetdir, name))
                     os.rename(f, os.path.join(targetdir, name))
             start = stop
-        pass
 
     def run(self):
         self._gather_chunk_files()
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,10 @@ select = [
     "B",    # flake8-bugbear
     "E",    # pycodestyle
     "F",    # pyflakes
+    "FURB", # refurb
     "N",    # pep8-naming
+    "PGH",  # pygrep-hooks
+    "PIE",  # flake8-pie
     "RUF",  # Ruff-specific
     "UP",   # pyupgrade
     "W",    # pycodestyle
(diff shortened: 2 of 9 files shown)


src/util/StringHash.cc
```diff
@@ -89,7 +97,7 @@ namespace lsst::qserv::util {
 /// @return a hexadecimal representation of the MD5 hash of the input buffer
 /// 128 bits -> 16 bytes -> 32 hex digits
 std::string StringHash::getMd5Hex(char const* buffer, int bufferSize) {
-    return wrapHashHex<MD5, MD5_DIGEST_LENGTH>(buffer, bufferSize);
+    return wrapHashHex<WMD5, MD5_DIGEST_LENGTH>(buffer, bufferSize);
```
iagaponenko, line 100: Though I 100% agree with the fix as a short-term solution, the very presence of the method `StringHash::getMd5Hex(` is a design mistake in Qserv. The only use of the method is for generating random names of the temporary tables in:
```
void UserQuerySelect::submit() {
    ...
    TmpTableName ttn(_queryId, _qSession->getOriginal());
```
Since `queryId` is unique in each Qserv deployment, it's easy to make unique temporary table names by simply naming those as:
```
"temporary_" + std::to_string(_queryId)
```
Instead of dragging the whole `openssl/MD5` machinery for that.

I'm not proposing to make this change now. Your change was the right thing to do in the scope of the current PR. We should probably do the right thing in a separate PR.

doc/misc/examples/obsolete/rebalance.py
```diff
@@ -50,15 +50,31 @@ class RebalanceProgram:
     """)
     parser = optparse.OptionParser(usage)
     parser.add_option(
-        "-d", "--dirs", dest="dirs", type="int", default=22, help=dedent("""\
-        Number of directories among which to balance (default=%default)."""))
+        "-d",
```
iagaponenko, line 53: I have inspected files in this subtree:
```
doc/misc
```
And I think it's nearly 100% obsolete. Besides, the subtree is not used for generating documentation on Qserv. See:
```
doc/index.rst
```
I vote for eliminating the subtree altogether.

doc/misc/examples/obsolete/loader.py
```diff
@@ -23,16 +23,17 @@
...
-from itertools import cycle, groupby
-import MySQLdb as sql
 import optparse
 import os
 import os.path
 import re
-from textwrap import dedent
 import time
 import warnings
+from glob import glob
```
iagaponenko, line 33: The whole `doc/misc` sub-tree should be eliminated. It's totally obsolete.

admin/local/cli/qserv
```diff
@@ -37,15 +37,15 @@ from lsst.qserv.admin.qservCli.qserv_log import log_level_from_args
 from lsst.qserv.admin.qservCli.subproc import QservSubprocessError
 
 if __name__ == "__main__":
-    success, levelMsg = log_level_from_args(sys.argv)
+    success, level_msg = log_level_from_args(sys.argv)
```
iagaponenko, line 40: Just for my own education. Did you migrate this name to stick to the "all-underscores" naming convention, or `ruff` did it?

iagaponenko (approved):
Looks good to me.

src/util/StringHash.cc
fritzm: Thanks, Igor -- good suggestion.  Would gladly tackle on a separate ticket.

doc/misc/examples/obsolete/loader.py
fritzm: I largely agree, but I left it here for now for our own reference while improving the newer, more up-to-date documentation.  In particular, some bits (like "scheduler-foreman-rationale", "guide-to-source", etc.) cover some historical mysteries not covered elsewhere.  The tcl script to plot chunks on a 3d-sphere might still be useful to make an illustration for the model docs, etc.

We should cut this material loose bit by bit as we continue to expand the newer docs.  I tried to sweep all the obsolete stuff scatter around into this one place at least.

admin/local/cli/qserv
fritzm: I probably let ruff make that change.  I used it's auto-editors as much as possible, since they are generally very principled about what they will auto-edit, and what the will leave to you as "possibly dangerous."

PR #961, opened 2025-09-13
DM-49098: source tree reorg

fritzm:
The main goal here was to collect all Python code under a single directory -- this makes Python developer tooling (linter, mypy) work a lot better, and removes the oft-forgotten `make install` from the Python edit/run/test developer loop.

Along the way also did some documentation link fixes, made the code `mypy` clean, and made the code compliant with a few additional `ruff` linters.


iagaponenko (approved):
Looks good to me!

PR #1016, opened 2026-04-07
DM-54589: Qserv Replication system records incorrect timestamps of chunk replicas in the system's metadata database

iagaponenko:
The bug was maxde when migrating Qserv code from `boost::filesystem` to `std::filesystem`. A problem with the migration path is that an implementation of the new library introduces a new type of clock that has an "Epoch" unrelated to the customary UNIX Epoch. Working with the new library requires writing complex code that would be difficult to comprehend.

Migrated the code to use the straightforward C language implementation instead of the complex API of the `std::filesystem` and std::chrono libraries. Added two simple methods for obtaining and setting values of the files' `mtime`.


fritzm (approved):
LGTM -- thanks!

PR #1017, opened 2026-04-09
DM-54599: validateDominantDbs() is called more frequently than seems necessary 

iagaponenko:
The PR addresses a design problem of the original implementation. The new implementation:
- Pushed dominant DB validation into the query analysis stage
- Made the validator a private method of the QuerySession class
- Eliminated all other calls to the validator
- Adjusted unit tests to skip the validations since this operation requires non-trivial CSS


jgates108 (approved):
That looks like the correct way to handle that.

PR #1019, opened 2026-04-12
DM-54632: Eliminated obsolete code

iagaponenko:
There are three commits in this PR:

- Remove `queryDistributionTestVer` from modules `czar` and `cconfig`

  - The code represents an experimental attempt to speed up query dispatch based on feedback from workers. The eliminated code was never finished, nor did it mature into any working prototype.
  - Another problem with the removed code was that it was making the package `qana` directly dependent on the Qserv frontend module `czar`. This was not right from the dependency management standpoint.

- Remove `QuerySession.h:  int _isFinal`

  This member has no actual effect on the query processing

- Remove unnecessary mutex locking from:

  ```
    // make new UserQuery
    // this is atomic
    ccontrol::UserQuery::Ptr uq;
    {
        lock_guard<mutex> lock(_mutex);
        uq = _uqFactory->newUserQuery(query, defaultDb, getQdispSharedResources(), userQueryId, msgTableName,
                                      resultDb);
    }
  ```

  The current query creation code is thread-safe after migrating the CSS database connector to the thread-safe implementation.

A few notes:
- The lock is known not to affect the performance of the `mysql-proxy` frontend due to another lock within the proxy that serializes query submission.
- The lock removal improves the performance of the `http` frontend since it doesn’t have any additional locks
- The ANTL4-based query parser is still the major factor that serializes query submission for the large query texts in both frontends.

### DIFF
--- a/src/cconfig/CzarConfig.h
+++ b/src/cconfig/CzarConfig.h
@@ -135,8 +135,6 @@ public:
      */
     int getXrootdCBThreadsInit() const { return _xrootdCBThreadsInit->getVal(); }
 
-    bool getQueryDistributionTestVer() const { return _queryDistributionTestVer->getVal(); }
-
     /*
      * @return A value of the "spread" parameter. This may improve a performance
      * of xrootd for catalogs with the large number of chunks. The default value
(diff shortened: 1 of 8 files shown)


jgates108 (approved):
LGTM

jgates108 (approved):
LGTM

PR #1020, opened 2026-04-16
Tickets/dm 54109

jgates108:
Fixed http-czar to check message severity to detect errors for failed queries.


iagaponenko (approved):
Looks good to me. One question - did you test the HTTP frontend to see the responses of the failed queries? Does it return what you expect based on changes made in this PR?

jgates108:
I tested it by adding junk to the end of the SQL query in QueryRunner::runQuery and the appropriate SQL errors showed when using the http czar. After removing the poison pill, queries worked normally.

PR #1021, opened 2026-04-28
DM-54798: The Controller should not terminate the process after failed self-registration

iagaponenko:
The Controller still posts warnings into the log stream upon self-registration failures and makes another registration attempt after the configurable timeout.


fritzm (approved):
LGTM -- thanks!

PR #1022, opened 2026-04-29
DM-54815: Small Qserv fixups

fritzm:
A few small fixups:

* Update `sphgeom` module to latest
* `ruff` formatting fixes to `itest.py`
* Add a `vscode` workspace file

### DIFF
--- a/.code-workspace
+++ b/.code-workspace
@@ -1,0 +1,18 @@
+{
+    "folders": [
+        {
+            "path": "."
+        }
+    ],
+    "settings": {
+        "files.associations": {
+            "**/bin/entrypoint": "python",
+            "**/bin/qserv": "python",
+            "**/bin/qserv-kraken": "python",
+            "**/bin/qserv-smig": "python"
+        },
+        "python.analysis.extraPaths": [
+            "./python"
+        ]
+    }
+}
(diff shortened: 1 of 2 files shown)


iagaponenko (approved):
Looks good. Though I still have a couple of questions, mostly out of curiosity:

- Do we need to run the Python formatter manually before submitting code to GitHub, as we do with `clang-format` (C++ code)?
- What are the benefits of keeping the vscode workspace file in the Qserv package?

fritzm:
Python formatting: yes, we should start doing this, but I need to do a little work first to support this in the makefile and the `qserv` CLI tool and add checks to CI before rolling this out as a requirement to you and John.  I'll make a ticket for that work, but I can continue to do it myself until we get to that -- we don't edit the Python codes all that frequently right now.

Workspace file: these particular settings are user-independent, project-specific, and necessary to have integrated python linting and code navigation work correctly in the vscode environment.  vscode  has a workspace file feature that lets you separate user-indpendent settings like this from user-specific customizations/preferences.  That way you can commit the project-specific settings to source control along with the project so they can track changes to the project structure, possibly vary per-branch, all vscode users get them automatically on pulls, etc.

PR #1025, opened 2026-05-20
DM-55010: Helm fixes for http-czar

fritzm:
* Correct bad ssl key file path arg for http-czar
* Add `publishNotReadyAddresses: true` for czar svc
* Reorder keys in svc yamls for consistency
* Clue vscode for syntax awareness in helm charts

### DIFF
--- a/deploy/helm/templates/czar-svc.yaml
+++ b/deploy/helm/templates/czar-svc.yaml
@@ -6,9 +6,11 @@ metadata:
     {{- include "qserv.labels" . | nindent 4 }}
 spec:
   clusterIP: "None"
+  publishNotReadyAddresses: true
   selector:
     {{- include "qserv.selectorLabels" . | nindent 4 }}
     app.kubernetes.io/component: czar
+  {{- if (include "qserv.enable" . | fromYaml).replication }}
   ports:
     - name: czar-http
       port: 4048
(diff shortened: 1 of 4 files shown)


deploy/helm/templates/czar-svc.yaml
```diff
@@ -6,13 +6,16 @@ metadata:
     {{- include "qserv.labels" . | nindent 4 }}
 spec:
   clusterIP: "None"
+  publishNotReadyAddresses: true
```
iagaponenko, line 9: Is this meant to register the DNS entry for a pod before it's ready?

iagaponenko (commented):
LGTM

iagaponenko (approved):

deploy/helm/templates/czar-svc.yaml
fritzm: Yes -- specifically, it means to publish the DNS address for a service as soon as the pod is created, and not to wait until whatever is running in the pod is answering at its declared readiness probes endpoints.  This is needed because the registry "calls back" for DNS resolution during service registration, before the http service is officially open for business.

Same issue (and same workaround) that we saw for the replication system registering itself in the registry at launch a few weeks back.

PR #1027, opened 2026-06-02
Returning async results no longer marks incorrect czarId as an error.

jgates108:
This would be very short, except clang-format made unrelated changes. The only changes that matter here are in ccontrol.

### DIFF
--- a/src/ccontrol/UserQueryAsyncResult.cc
+++ b/src/ccontrol/UserQueryAsyncResult.cc
@@ -77,7 +77,8 @@ void UserQueryAsyncResult::submit() {
     // if there are messages already it means the error was detected, stop right here
     if (_messageStore->messageCount() > 0) {
         LOGS(_log, LOG_LVL_WARN,
-             "UserQueryAsyncResult::submit giving up, messageCount=" << _messageStore->messageCount());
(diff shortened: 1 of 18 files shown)


src/ccontrol/UserQueryAsyncResult.cc
```diff
@@ -83,11 +83,9 @@ void UserQueryAsyncResult::submit() {
 
     // Presently we cannot return query results that originated from different czar
     if (_qInfo.czarId() != _czarId) {
-        // TODO: tell user which czar was it?
-        std::string message = "Query originated from different czar";
-        LOGS(_log, LOG_LVL_WARN, "UserQueryAsyncResult::submit giving up, message=" << message);
-        _messageStore->addErrorMessage("SYSTEM", message);
-        return;
+        LOGS(_log, LOG_LVL_WARN,
+             "UserQueryAsyncResult::submit Query originated from different czar=" << _qInfo.czarId()
```
iagaponenko, line 88: This is not critical. To make the messages immutable for possible changes in the names of the enclosing methods, I use the following pattern:
```
"UserQueryAsyncResult::" << __func__ << " Query originated from different czar="
```

iagaponenko (approved):
Looks good to me.

PR #1029, opened 2026-06-09
Increased the number of threads for various operations and added temp…

jgates108:
Draft PR mostly for simplifying loading onto the kubernetes cluster.
Increased thread available for certain operations. Primarily http services, the worker scheduler, and the czar thread pool. Many of these were much lower than they should have been for 64 core machines.
Several temporary log statements have been added to try and determine the load from various sources. These need to be removed before merging.

### DIFF
--- a/src/wpublish/QueriesAndChunks.cc
+++ b/src/wpublish/QueriesAndChunks.cc
@@ -249,10 +249,8 @@ void QueriesAndChunks::removeDead() {
             if (statPtr->isDead(_deadAfter, now)) {
                 LOGS(_log, LOG_LVL_TRACE, "QueriesAndChunks::removeDead added to list");
                 deadList.push_back(statPtr);
-                iter = _deadQueries.erase(iter);
-            } else {
-                ++iter;
             }
+            iter = _deadQueries.erase(iter);
         }
     }
 
@@ -264,10 +262,8 @@ void QueriesAndChunks::removeDead() {
     {
         lock_guard<mutex> g(_queryStatsMapMtx);
         lock_guard<mutex> gd(_deadMtx);
-        for (auto const& elem : _queryStatsMap) {
-            auto const& statPtr = elem.second;
+        for (auto const& [qId, statPtr] : _queryStatsMap) {
             if (statPtr->isMostlyDead()) {
-                QueryId qId = statPtr->getQueryId();
                 _deadQueries[qId] = statPtr;
             }
         }
@@ -458,9 +454,8 @@ nlohmann::json QueriesAndChunks::statusToJson(wbase::TaskSelector const& taskSel
     ///       a similar argument could be made for "tasks". DM-55247
     status["query_stats"] = nlohmann::json::object();
     lock_guard<mutex> g(_queryStatsMapMtx);
-    for (auto&& itr : _queryStatsMap) {
-        string const qId = to_string(itr.first);  // forcing string type for the json object key
-        QueryStatistics::Ptr const& qStats = itr.second;
+    for (auto const& [queryId, qStats] : _queryStatsMap) {
+        string const qId = to_string(queryId);  // forcing string type for the json object key
         if (false) {
             status["query_stats"][qId]["histograms"] = qStats->getJsonHist();
         } else {
--- a/src/wpublish/QueriesAndChunks.cc
+++ b/src/wpublish/QueriesAndChunks.cc
@@ -445,7 +445,6 @@ void QueriesAndChunks::buildCancelledAndDeletedLists(
 nlohmann::json QueriesAndChunks::statusToJson(wbase::TaskSelector const& taskSelector) const {
     nlohmann::json status = nlohmann::json::object();
     {
-        util::InstanceCount icA("QandC::statToJ-lck1 &&&");
         auto bSched = _blendSched.lock();
         if (bSched == nullptr) {
             LOGS(_log, LOG_LVL_WARN, "blendSched undefined, can't check user query");
@@ -455,17 +454,19 @@ nlohmann::json QueriesAndChunks::statusToJson(wbase::TaskSelector const& taskSel
         }
     }
     /// TODO: This just shouldn't be a part of this at all. It's accrues data over time
-    ///       so frequent updates don't help and it's big.
+    ///       so frequent updates don't help and it's big. "histograms" is the worst offender,
+    ///       a similar argument could be made for "tasks". DM-55247
     status["query_stats"] = nlohmann::json::object();
-    if (false) {
-        util::InstanceCount icB("QandC::statToJ-lck12 &&&");
-        lock_guard<mutex> g(_queryStatsMapMtx);
-        for (auto&& itr : _queryStatsMap) {
-            string const qId = to_string(itr.first);  // forcing string type for the json object key
-            QueryStatistics::Ptr const& qStats = itr.second;
+    lock_guard<mutex> g(_queryStatsMapMtx);
+    for (auto&& itr : _queryStatsMap) {
+        string const qId = to_string(itr.first);  // forcing string type for the json object key
+        QueryStatistics::Ptr const& qStats = itr.second;
+        if (false) {
             status["query_stats"][qId]["histograms"] = qStats->getJsonHist();
-            status["query_stats"][qId]["tasks"] = qStats->getJsonTasks(taskSelector);
+        } else {
+            status["query_stats"][qId]["histograms"] = nlohmann::json::object();
         }
+        status["query_stats"][qId]["tasks"] = qStats->getJsonTasks(taskSelector);
     }
     return status;
 }
--- a/src/wpublish/QueriesAndChunks.cc
+++ b/src/wpublish/QueriesAndChunks.cc
@@ -451,16 +451,14 @@ nlohmann::json QueriesAndChunks::statusToJson(wbase::TaskSelector const& taskSel
     }
     /// TODO: This just shouldn't be a part of this at all. It's accrues data over time
     ///       so frequent updates don't help and it's big. "histograms" is the worst offender,
-    ///       a similar argument could be made for "tasks". DM-55247
+    ///       a similar argument could be made for "tasks" as this is just too
+    ///       much information to send every couple of seconds. DM-55247
     status["query_stats"] = nlohmann::json::object();
+    if (true) return status;  // TODO: DM-55247
     lock_guard<mutex> g(_queryStatsMapMtx);
     for (auto const& [queryId, qStats] : _queryStatsMap) {
         string const qId = to_string(queryId);  // forcing string type for the json object key
-        if (false) {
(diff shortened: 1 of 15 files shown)


iagaponenko:
@fritzm , @jgates108 :
In general, I'm a bit puzzled by an attempt to tune the configuration parameters of Qserv in the source tree.
(see the commit: https://github.com/lsst/qserv/pull/1029/commits/56d1cae46a5e51f817aa0b251604bca2e30b73b0).
My understanding is that this kind of tuning is normally done later, at the targeted site-specific deployments. The tuning (especially when it comes to the number of threads or the memory buffers) should be done to match the amount of resources (core, etc.) available at the given site, or the desired performance profile (like the expected query payload). Using the source tree as the ultimate source of truth looks like a flashback of the older (problematic) configuration model, in which we had to build a new Qserv image each time we needed to reconfigure Qserv deployments. My understanding is that the configuration files are presently totally decoupled from any direct dependency on a specific Qserv image tag in all known deployment models of Qserv.

src/wpublish/QueriesAndChunks.cc
```diff
@@ -455,17 +454,19 @@ nlohmann::json QueriesAndChunks::statusToJson(wbase::TaskSelector const& taskSel
...
     status["query_stats"] = nlohmann::json::object();
-    if (false) {
-        util::InstanceCount icB("QandC::statToJ-lck12 &&&");
-        lock_guard<mutex> g(_queryStatsMapMtx);
-        for (auto&& itr : _queryStatsMap) {
-            string const qId = to_string(itr.first);  // forcing string type for the json object key
-            QueryStatistics::Ptr const& qStats = itr.second;
+    lock_guard<mutex> g(_queryStatsMapMtx);
+    for (auto&& itr : _queryStatsMap) {
+        string const qId = to_string(itr.first);  // forcing string type for the json object key
```
iagaponenko: I'm sorry, this way of iterating over associative collections is outdated. The right way (as of C++17) over maps in the JSON Nlohmann library is:
```
for (auto& [key, val] : j_object.items())
{
    std::cout << "key: " << key << ", value:" << val << '\n';
}
```
Please, read the documentation on iterators in: https://json.nlohmann.me/features/iterators/#range-based-for-loops

src/wpublish/QueriesAndChunks.cc
```diff
@@ -239,43 +234,49 @@ ChunkStatistics::Ptr QueriesAndChunks::getChunkStatistics(int chunkId) {
...
     }
-    if (LOG_CHECK_LVL(_log, LOG_LVL_DEBUG)) {
-        lock_guard<mutex> gdend(_deadMtx);
-        LOGS(_log, LOG_LVL_DEBUG,
-             "QueriesAndChunks::removeDead end deadQueries size=" << _deadQueries.size());
+
+    /// Find mostly dead queries and add them to the dead map to be checked next time.
+    {
+        lock_guard<mutex> g(_queryStatsMapMtx);
+        lock_guard<mutex> gd(_deadMtx);
```
iagaponenko, line 264: I see two locks acquired in a row. Normally, this triggers concerns about the possibility of deadlocks ... should these locks be acquired in an opposite order in some other method?

src/wpublish/QueriesAndChunks.cc
```diff
@@ -239,43 +234,49 @@ ChunkStatistics::Ptr QueriesAndChunks::getChunkStatistics(int chunkId) {
...
+            if (statPtr->isMostlyDead()) {
+                QueryId qId = statPtr->getQueryId();
+                _deadQueries[qId] = statPtr;
+            }
+        }
+        deadQueriesSizeEnd = _deadQueries.size();
+        queryStatsMapSize = _queryStatsMap.size();
     }
+    LOGS(_log, LOG_LVL_INFO,
+         "removeDead queryStatsMapSize=" << queryStatsMapSize << " deadQueriesSize start="
```
iagaponenko, line 274: Do you really want to see this in the `INFO` logging stream? In essence, these reports are meant for code debugging purposes. Hence, `DEBUG` better suits in this context.

src/wpublish/QueryStatistics.h
```diff
@@ -78,6 +78,12 @@ class QueryStatistics {
 
     void setQueryBooted(bool booted, TIMEPOINT now);
 
+    /// Return true if all tasks are completed.
+    bool isMostlyDead() const {
+        std::lock_guard<std::mutex> gs(_qStatsMtx);
```
iagaponenko, line 90: Why not call the method `allTasksAreCompleted()` instead? This name would explain exactly what this method is meant for, instead of some fuzzy concept of something being "mostly dead."?

[ For example, you already have another metgod called `addTaskCompleted`. And it's clear what that method does. The proposed `allTasksAreCompleted()` would be perfectly consistent in a context of tracking task statuses. ]

Note that when you name your methods and variables, you're introducing a vocabulary to explain what is actually happening in your code. It's much easier to read and comprehend the code in which the vocabulary matches the reality.

iagaponenko:
My other general comment concerns the concept of "dead" queries (tasks, etc.). In reality (based on what the code actually does), this word refers to the completed queries. Yes, the code intermittently mentions both "completed" and "dead". This creates unnecessary impedance in reading and comprehending the code. Why not come up with a consistent naming convention? 

I would argue that "completed" would be the right term in the current context.

iagaponenko:
My next comment (perhaps I should direct it to @fritzm ) is about:
```
qservImageName: ghcr.io/lsst/qserv:2026.5.1-rc1-60-g56d1cae46
```
This sounds like a recursive dependency. Should the Qserv image tags be tuned at a specific deployment _after_ building the right image? I would argue that changes to the Helm charts should be made in separate PRs. The proposed model/policywould be to have 2 PRs:
* Firstly, the PR for changes in the Qserv code per se
* And the dependent PR for changes to the deployment charts

This model is cleaner, and it would avoid possible traps (references to outdated Qserv images) should iterative changes (Git pushes) be made into a single (monolithic) PR combining both code and Helm chart changes.

src/wpublish/QueriesAndChunks.cc
jgates108: Wouldn't that print everything to the console? I just need a string version to use as a key in multiple assignments.

iagaponenko: Since I anticipated your question, I posted the link to the documentation on the range-based loop over maps in the JSON library: https://json.nlohmann.me/features/iterators/#range-based-for-loops
If you had a chance to look into the documentation, then you would see the following statement:
```
Note

When iterating over an array, key() will return the index of the element as string. For primitive types (e.g., numbers), key() returns an empty string.
```

src/wpublish/QueriesAndChunks.cc
jgates108: It shows up about every 5 minutes and will be the only indicator that workers are removing these objects appropriately. This and some other items may be future candidates for the dashboard.

src/wpublish/QueriesAndChunks.cc
jgates108: It's not iterating over a json object. It's iterating over a `std::map` and putting the contents into json maps. `[key, val]` would be an improvement, but the documentation you linked to has nothing to do with`std::map`.

iagaponenko: I missed that. In this case, `to_string()` is the right thing to do.
However, the general recommendation is the same - for `std::map` you may want to use the modern range-based loop pattern:
```
for (const auto& [queryId, stats] : _queryStatsMap) {
        ...
}
```

iagaponenko (approved):
LGTM. I have posted my comments and suggestions.

src/wpublish/QueriesAndChunks.cc
jgates108: I'm not that happy about doing this, but the `_deadQueires` object could be tied up for a while and I would really rather that did not also tie up  `_queryStatsMap` unnecessarily, as that is used very frequently. 
So, two separate mutex..
On the other hand, `_deadMtx` is only needed at 2 locations and its declaration include the stipulation that it be locked after `_queryStatsMap` if both need to be locked, so the risks of dead lock are low.  
Edit: Adding comments about life time of this object to the code.

src/wpublish/QueryStatistics.h
jgates108: `QueryStatistics` have a complicated life cycle. There are two main issues. First is the statistics are not complete, and often not very useful, until the query has completed. And then nobody is likely to come looking for it for a few minutes. The next problem is that more work can come in for a query where all tasks had already finished. So, it's completely possible to have something return true from `isMostlyDead()` and then return false a little later. I think that would be even more confusing is the function was called `allTasksAreCompleted()`

After something has been identified as "mostlyDead", it is put on list. If nothing changes it for a while, it is declared dead by being deleted. Otherwise it may come back to life.

iagaponenko: I see it. Based on your explanation, it sounds like the method is meant to say: "No active tasks have been detected" (at the time of the call). I still don't think that "mostly dead" is the right definition for this situation. The method should say exactly what its implementation does, like `noActiveTasksFound()` (or in the reverse logic `activeTasksFound()`).
Perhaps I'm just confused by the concept of the "dead queries"? In my view, the query processing may be in one of the following states:
* created
* in-progress
* finished
* failed
I just don't see where "dead" fits into this state machine.

Since you introduced this concept a long time ago, you have already gotten used to it, and it makes perfect sense to you.

In the end, it's up to you how you call these methods as long as you're the only one who maintains this code. It's just hard to review the code that introduces unusual vocabulary.

src/wpublish/QueriesAndChunks.cc
iagaponenko: Regarding 5 minutes. This translates into 300 seconds. What if someone submitted 3,000 trivial queries within 300 seconds? It's just 10 queries per second. In this case, there would be 18,000 histogram objects in the query stats. I don't think 10 qps (or even 100 qps, which will translate into 180,000 histograms at a time) is such an unimaginable number for Qserv. Right now, fast queries are throttled by the poor performance of ANTL4. However, in a few weeks from now, we might get this block eliminated.

As a design option, you may consider creating a metadata table in the worker's database, dumping query stats into the table after the query is over, and eliminating the in-memory entry right away. In this case, the query stats becomes permanent, and one can pull he stats for various performance and monitoring purposes for any period of time.

[ You may still do the garbage collection of the stats to catch missed notifications on the query completion sent by Czar to the worker. ]

This (retaining stats in the worker's persistent state) is just an idea to think about. I'm not arguing that it has to be implemented in the short term.

jgates108: There's a lot to consider here. All of this information gathering also relates to the decision to boot a user query, and I still don't have a good way to do that.

Edit: The cluster has shown that it can handle storing a lot of these objects. That doesn't mean it is wise to do so, just that there's some leway.

PR #1032, opened 2026-07-13
DM-55424: Support scisql_s2PtInCircle Spatial Joins and scisql_angSep Area Restrictors

malensek:
Currently, Qserv infers spatial join conditions only from appropriately formed `scisql_angSep` expressions.  These conditions could, however, also validly be expressed in terms of `scisql_s2PtInCircle`. Similarly, a query can express an area restriction (cone evaluated against a single director table) using `scisql_angSep` instead of `scisql_s2PtInCircle`, as long as one side of the coordinate pair is a constant. For example:
`WHERE scisql_angSep(dias.coord_ra, dias.coord_dec, 55.5, -30.2) < 0.05`
is equivalent to
`WHERE scisql_s2PtInCircle(dias.coord_ra, dias.coord_dec, 55.5, -30.2, 0.05) = 1`
which Qserv already recognizes as an area restrictor.

This PR adds the missing direction for both, so either scisql form can be used interchangeably for spatial joins and area restrictors. It also allows director-child and child-child spatial joins (where admissible), which were previously rejected outright.

## Summary of Changes
* RelationGraph.cc:
    - `_addSpEdges` now tries `scisql_angSep` first, then falls back to the equivalent `scisql_s2PtInCircle(...) = 1` form. Refactored the per-form extraction logic. Improved error reporting in the case of malformed queries.
    - `_addSpEdges` now allows director-child and child-child spatial joins, provided that they have spatial columns configured and appropriate partitioning. This required a small refactor of `TableInfo.h` and `TableInfoPool.cc` to query the necessary information from CSS.
* QservRestrictorPlugin.cc - `extractSingleAreaRestrictor` now also recognizes `scisql_angSep(lon, lat, constLon, constLat) <= radius` as equivalent to `scisql_s2PtInCircle(...) = 1`, producing the same AreaRestrictorCircle.
* testQueryAnaGeneral.cc - new test coverage for the features described above

### Behavior Changes
* `extractSingleScisqlAreaFunc` accepted `scisql_s2PtInCircle(...) = <anything>` as long as one side was a `scisql_s2Pt*` call. It did not check that the compared value was 1, so `scisql_s2PtInCircle(...) = 0` would have been accepted even though the opposite was intended. A check for == 1.0 was added.
* Previously if we had a `BoolFactor` with an associated `NOT`, it was getting ignored and treated as the positive case. This is now caught in RelationGraph / QservRestrictorPlugin.
* We now disallow `!=` and `<>` for `angSep` spatial joins. Such a join was accepted by the system previously, but would not have actually been evaluable because it references data outside chunk overlap.
* Director-child and child-child spatial joins were previously disallowed completely. We now allow them **IF** they are admissible.


src/qana/QservRestrictorPlugin.cc
```diff
@@ -210,6 +198,141 @@ std::shared_ptr<query::AreaRestrictor> makeAreaRestrictor(query::FuncExpr const&
...
+                radiusSide = compPredicate.left;
+            }
+            break;
+        default:
+            return nullptr;
+    }
+
+    // Validate the scisql_angSep function and radius
+    if (not funcSide->isFunction()) return nullptr;
+    auto&& funcExpr = funcSide->getFunction();
```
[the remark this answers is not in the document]
malensek: This should be essentially the same thing in this case. I like the readability improvement though.

src/qana/RelationGraph.cc
```diff
@@ -314,12 +294,109 @@ FuncExpr::Ptr getAngSepFunc(ValueExprPtr const& ve) {
...
+            // point numbers are being tested for equality), it is
+            // technically evaluable.
+            fe = getAngSepFunc(cp.left);
+            if (!fe) {
+                angSep = cp.left->getNumericConst();
+                fe = getAngSepFunc(cp.right);
+            } else {
+                angSep = cp.right->getNumericConst();
+            }
+            break;
```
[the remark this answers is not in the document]
malensek: This was how it worked before this PR (just relocated), so I guess this definitely warrants some discussion...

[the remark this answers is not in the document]
malensek: Our conclusion from discussion in yesterday's meeting and on Slack was that this should be indeed disallowed. I added a note to the PR description and will update RelationGraph shortly.

fritzm (approved):
This look great -- thanks!

src/qana/QservRestrictorPlugin.cc
```diff
@@ -210,6 +198,141 @@ std::shared_ptr<query::AreaRestrictor> makeAreaRestrictor(query::FuncExpr const&
...
+            // Area functions must be of the form scisql_s2Pt*(...) = 1 (or <=> 1):
+            if (compPredicate->op != query::CompPredicate::EQUALS_OP &&
+                compPredicate->op != query::CompPredicate::NULL_SAFE_EQUALS_OP) {
+                continue;
+            }
+
+            // Check both sides of the '='
+            if (isScisqlAreaFunc(*compPredicate->left) &&
+                areaParamsAreConst(*compPredicate->left->getFunction()) &&
+                compPredicate->right->getNumericConst() == 1.0) {
```
iagaponenko, line 316: As far as I know, the function returns `double`:
```
double getNumericConst(ValueExprPtr const& ve);
```
(Depending on the context) The floating-point comparisons may be tricky at times. My understanding is that the safest approach is to use an "epsilon". For numbers around 0 - 1, the safest approach would be to use the following pattern:
```
#include <algorithm>
#include <limits>

if (std::fabs(compPredicate->right->getNumericConst() - 1.0) <= std::numeric_limits<double>::epsilon()) {
   ...
}
```
For further details, read:
* https://en.cppreference.com/cpp/types/numeric_limits/epsilon
* https://www.reddit.com/r/cpp_questions/comments/1ap668e/float_comparison/

src/qana/QservRestrictorPlugin.cc
```diff
@@ -210,6 +198,141 @@ std::shared_ptr<query::AreaRestrictor> makeAreaRestrictor(query::FuncExpr const&
...
+                ++count;
+                scisqlFunc = compPredicate->left->getFunction();
+            }
+            if (isScisqlAreaFunc(*compPredicate->right) &&
+                areaParamsAreConst(*compPredicate->right->getFunction()) &&
+                compPredicate->left->getNumericConst() == 1.0) {
```
iagaponenko, line 322: See my previous comment on `epsilon`

src/qana/TableInfoPool.cc
```diff
@@ -152,12 +153,25 @@ TableInfo const* TableInfoPool::get(std::string const& db, std::string const& ta
...
     infoPtr->fk = partParam.dirColName;
     if (infoPtr->fk.empty()) {
         throw InvalidTableError("Child table " + db_ + "." + table +
                                 " metadata"
                                 " does not contain a director column name!");
     }
+    // Only reject a collision when lon/lat are both configured -- a single empty coordinate
+    // already leaves the child with no usable position (see hasSpatialCols()) regardless of fk.
+    if (!partParam.lonColName.empty() && !partParam.latColName.empty() &&
+        (partParam.lonColName == partParam.latColName || partParam.lonColName == infoPtr->fk ||
```
iagaponenko: I'm a bit confused about this comparison:
```
partParam.lonColName == infoPtr->fk
```
The LH represents the name of a spatial coordinate, while the RH is the name of the "director" key.

iagaponenko:
I have one general comment about comparing floating-point values using the `==` operator. The comparisons need to be done using an "epsilon".

The second comment concerns case-sensitive comparisons involving column names. In MySQL (and Qserv) these names are case-insensitive. Therefore (in some cases) the operator `==` may produce a false-positive result. Perhaps one should always use `boost::iequals()`?

src/qana/TableInfoPool.cc
malensek: Oh, I see how this looks weird here. This is following a similar pattern as done above for director tables, `latColName`, `lonColName`, and the FK all have to be distinct strings (i.e., distinct column names).

(The director check is:)
```
 if (v.size() != 3 || v[0].empty() || v[1].empty() || v[2].empty() || v[0] == v[1] || 
            v[1] == v[2] ||
            v[0] == v[2]) {
```

To make both of these more readable, it might actually be nice to have a helper function that calls this out explicitly, something like `checkDistinct` and then we could use it in both places.

EDIT: and we could take care of the case insensitivity thing here too.

src/qana/QservRestrictorPlugin.cc
```diff
@@ -210,6 +198,141 @@ std::shared_ptr<query::AreaRestrictor> makeAreaRestrictor(query::FuncExpr const&
...
+
+    // Validate the scisql_angSep function and radius
+    if (not funcSide->isFunction()) return nullptr;
+    auto funcExpr = funcSide->getFunction();
+    if (funcExpr->getName() != "scisql_angSep" || funcExpr->getParams().size() != 4) {
+        return nullptr;
+    }
+    auto const& params = funcExpr->getParams();
+    double const radius = radiusSide->getNumericConst();
+    if (std::isnan(radius)) return nullptr;
```
iagaponenko: A better (more comprehensive) check is to use: https://en.cppreference.com/cpp/numeric/math/isnormal
Which will ensure the radius is s neither zero, subnormal, infinite, nor NaN.

i.e.:
```
    if (!std::isnormal(radius)) return nullptr;
```

There are other similar uses of `std::isnan` in this commit.

src/qana/TableInfoPool.cc
iagaponenko: If you are going to look at case insensitivity, then you may also expand one of the unit tests for testing the case-sensitivity behavior of the query analyzer. Though, this could be done in a separate (dedicated ticket/PR).

iagaponenko (commented):
One more suggestion while you're still on the PR. We have two obsolete cases of using the old Boost library:
```
src/qana/RelationGraph.cc:    if (!fe || boost::math::isnan(angSep)) {
src/qana/RelationGraph.h:    bool isSpatial() const { return !(boost::math::isnan)(angSep); }
```
These can be replaced by either `std::isnan()` or (where applies) `!std::isnormal()`

iagaponenko (approved):
LGTM

src/qana/QservRestrictorPlugin.cc
malensek: For a bit of context, these previously didn't do the check at all, so even `0.0` would work in the source query here too (i.e. `scisql_s2PtInCircle( . . . ) = 0` was still processed as if it was `= 1`). This was the source of my question on Slack about whether this check is even necessary, I guess it mainly comes from ADQL upstream.

My understanding is that using an epsilon for comparison is necessary when we are evaluating the result of  arithmetic operations. `getNumericConst` does return `double`, but with IEEE 754 we can represent integers up to 2^53 exactly. So we're going `String` -> `strtod()` -> exactly-representable int in an IEEE 754 double -> equality comparison, so I think this is safe. (I don't foresee any arithmetic happening on either side of this getting added later)

For what it's worth, I originally considered just checking String equality, but then that'd only allow one variant and not `1`, `1.0`, `1.00`, etc

src/qana/QservRestrictorPlugin.cc
malensek: Oh yes, makes sense, thank you!!

malensek:
Will fix those obsolete boost-isms too! Thank you :-)

src/qana/QservRestrictorPlugin.cc
iagaponenko: Please note that `std::isnormal(radius)` treats `0.0` as not "normal". So, if `0.0` is allowed in this context, then the test should probably be:
```
    if (!(radius == 0.0 || std::isnormal(radius))) return nullptr;
```
On the other hand, I have an impression that `radius=0.0` should never be passed into these UDFs.

src/qana/TableInfoPool.cc
malensek: I'm going to add some tests for this part specifically, but looking through `qana` I think it would be worth filing a separate ticket to do some in-depth case sensitivity wrangling because there are a few more cases that would be worth looking at.

fritzm: Some care is required when getting around to that -- MySQL defaults are case-insenstive column and index names everywhere, but case _sensitive_ table names on Linux...

PR #1033, opened 2026-07-16
DM-55404: A low-latency synchronization mechanism on the data management events sent by the Replication/Ingest system to Czar in Qserv

iagaponenko:


src/http/MetaModule.cc
```diff
@@ -36,7 +36,7 @@ lsst::qserv::http::AuthContext const authContext;
 
 namespace lsst::qserv::http {
 
-unsigned int const MetaModule::version = 56;
+unsigned int const MetaModule::version = 57;
```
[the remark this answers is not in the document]
iagaponenko: This is not relevant in the context of the PR.

fritzm (approved):
LGTM -- thanks!

src/cconfig/DataManagementEvent.cc
```diff
@@ -0,0 +1,114 @@
...
+        case Type::CHUNK_MAP_REBUILT:
+            return "CHUNK_MAP_REBUILT";
+        case Type::DATABASE_PUBLISHED:
+            return "DATABASE_PUBLISHED";
+        case Type::DATABASE_DELETED:
+            return "DATABASE_DELETED";
+        case Type::TABLE_DELETED:
+            return "TABLE_DELETED";
+    }
+    throw invalid_argument("Unknown DataManagementEvent::Type");
```
jgates108: Adding `type` to error message may be useful if it ever gets triggered.

src/cconfig/EventService.cc
```diff
@@ -0,0 +1,120 @@
...
+    }
+    _subscriptions[subscriptionId] = callback;
+    LOGS(_log, LOG_LVL_INFO, _CONTEXT << "new subscription added with ID " << subscriptionId);
+}
+
+void EventService::unsubscribe(string const& subscriptionId) {
+    lock_guard<mutex> lock(_mtx);
+    auto it = _subscriptions.find(subscriptionId);
+    if (it == _subscriptions.end()) {
+        throw runtime_error(_CONTEXT + "subscription ID not found: " + subscriptionId);
```
jgates108: This is probably a discussion for a later time, and not really needed for this ticket:

I'd prefer to use a derived exception here. `runtime_error` is a pretty vast collection of exceptions thrown from many places and this is a case where I'd probably want to catch this exception specifically. I'd rather be be catching a fairly generic `QservRuntimeError` as I'd know the source was qserv and not something from the std library. I'm generally fine with things like `invalid_argument` as they are smaller sets, but this isn't really an invalid argument.

src/cconfig/testEventService.cc
```diff
@@ -0,0 +1,444 @@
...
+    service->subscribe(callback, "sub1");
+
+    DataManagementEvent event;
+    event.type = DataManagementEvent::Type::DATABASE_PUBLISHED;
+    event.database = "test_db";
+
+    service->postEvent(event);
+
+    // Give async callback time to execute
+    this_thread::sleep_for(chrono::milliseconds(100));
```
jgates108: I like using a loop to help things be consistent on heavily loaded machines

`for (int i=0; receivedCount < 1 && i < 100; ++i)  this_thread::sleep_for(chrono::milliseconds(100));`

It clears quickly in most cases, and doesn't fail unless things are incredibly slow. This applies to all `sleep` statements in unit tests.

jgates108 (approved):
LGTM - I focused mostly on the qserv parts.

src/cconfig/DataManagementEvent.cc
iagaponenko: Good suggestion

src/cconfig/EventService.cc
iagaponenko: Good suggestion

src/cconfig/testEventService.cc
iagaponenko: Good suggestion

PR #1034, opened 2026-07-16
Created Czar::FamilyMapRead function to be called on map update notif…

jgates108:
When the tables are updated, the notification just needs to call `Czar::FamilyMapRead`.

### DIFF
--- a/src/czar/Czar.cc
+++ b/src/czar/Czar.cc
@@ -98,6 +98,8 @@ Czar::Ptr Czar::createCzar(string const& configFilePath, string const& czarName)
 
 void Czar::_monitor() {
     string const funcN("Czar::_monitor");
+
+    bool familyMapUpdated = false;
     while (_monitorLoop) {
         this_thread::sleep_for(_monitorSleepTime);
         LOGS(_log, LOG_LVL_DEBUG, funcN << " start0");
@@ -106,11 +108,10 @@ void Czar::_monitor() {
         // check if workers have died by timeout.
         _czarRegistry->sendActiveWorkersMessages();
 
-        // Check if the family map should be updated
-        bool const jobsAssigned = familyMapRead(TIMEPOINT(chrono::seconds(0)));
-        if (!jobsAssigned) {
-            _assignJobsToUberJobs();
-        }
+        // After one successful read, the czar will depend on EventService for updates.
+        if (!familyMapUpdated) { familyMapUpdated = familyMapRead(TIMEPOINT(chrono::seconds(0))); }
+
+        _assignJobsToUberJobs();
 
         // To prevent anything from slipping through the cracks:
         // Workers will keep trying to transmit results until they think the czar is dead.
@@ -146,8 +147,8 @@ bool Czar::familyMapRead(TIMEPOINT const requestTime) {
     } else {
         LOGS(_log, LOG_LVL_DEBUG,
              funcN << " family map read already completed since request time "
-                   << util::TimeUtils::timePointToDateTimeString(requestTime) << " latestFamilyUpdate="
-                   << util::TimeUtils::timePointToDateTimeString(_latestFamilyUpdate));
+                   << util::TimeUtils::timePointToDateTimeString(requestTime, true) << " latestFamilyUpdate="
+                   << util::TimeUtils::timePointToDateTimeString(_latestFamilyUpdate, true));
     }
     // To prevent anything from slipping through the cracks:
     // Workers will keep trying to transmit results until they think the czar is dead.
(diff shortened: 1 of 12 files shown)


src/czar/Czar.h
```diff
@@ -282,6 +292,15 @@ class Czar {
...
     mutable std::mutex _commErrCountMtx;  ///< protects _commErrCountMap
+
+    /// Limits Czar::FamilyMapRead() to one copy running at a time. FamilyMapRead() is expensive
+    /// and running multiple copies concurrently is not useful.
+    std::mutex _latestFamilyUpdateMtx;
+    /// The last time a family read was started. If the request time is less than this, the change has already
+    /// been read and the request can be safely ignored.
+    TIMEPOINT _latestFamilyUpdate;
+    /// If more than this time has past, check the family map.
+    std::chrono::seconds _familyMapmaxUpdateWait{600};
```
iagaponenko: For the sake of consistency, when using the Camel-case naming convention:
```
_familyMapMaxUpdateWait
```

src/czar/Czar.h
```diff
@@ -139,6 +139,13 @@ class Czar {
 
     std::shared_ptr<CzarFamilyMap> getCzarFamilyMap() const { return _czarFamilyMap; }
 
+    /// Check if the family map has been updated since update time and if so, read the new family map.
+    /// @param requestTime The time of the request.
+    /// If the requestTime is > _latestFamilyUpdate (start of last read), the family map is out probably out
+    /// of date.
+    /// @return true if the family map was read.
+    bool FamilyMapRead(TIMEPOINT const requestTime);
```
iagaponenko: Two questions/comments:
* Is this method supposed to be `public`?
* The name of the method should begin with lowercase as per: https://developer.lsst.io/cpp/style.html

src/czar/Czar.cc
```diff
@@ -153,6 +118,69 @@ void Czar::_monitor() {
     }
 }
 
+bool Czar::FamilyMapRead(TIMEPOINT const requestTime) {
+    string const funcN("Czar::FamilyMapRead");
+    bool familyMapRead = false;
```
iagaponenko: Could you pick a different name for the variable? It's basically (module lower/upper case first letter) the same as the name of the enclosing method. Something like:
```
bool mapWasUpdated = false;
```

src/czar/Czar.cc
```diff
@@ -153,6 +118,69 @@ void Czar::_monitor() {
...
+    LOGS(_log, LOG_LVL_DEBUG, funcN << " start0");
+    // Only one thread at a time should do this.
+    lock_guard familyUpdateLock(_latestFamilyUpdateMtx);
+    if (requestTime >= _latestFamilyUpdate || CLOCK::now() - _latestFamilyUpdate > _familyMapmaxUpdateWait) {
+        /// Check database for changes in worker chunk assignments and aliveness
+        try {
+            auto familyUpdateStart = CLOCK::now();
```
iagaponenko, line 138: Why is this variable here? It doesn't seem to be used anywhere. Did you mean to measure the performance of the table read operation?

src/czar/Czar.cc
```diff
@@ -153,6 +118,69 @@ void Czar::_monitor() {
     }
 }
 
+bool Czar::FamilyMapRead(TIMEPOINT const requestTime) {
+    string const funcN("Czar::FamilyMapRead");
+    bool familyMapRead = false;
+    LOGS(_log, LOG_LVL_DEBUG, funcN << " start0");
+    // Only one thread at a time should do this.
+    lock_guard familyUpdateLock(_latestFamilyUpdateMtx);
+    if (requestTime >= _latestFamilyUpdate || CLOCK::now() - _latestFamilyUpdate > _familyMapmaxUpdateWait) {
```
iagaponenko: I think you have a bug here:
```
CLOCK::now() - _latestFamilyUpdate > _familyMapmaxUpdateWait
```
Perhaps you meant to say this?
```
CLOCK::now() - _familyMapmaxUpdateWait > _latestFamilyUpdate
```

iagaponenko (commented):
I would highly recommend using Copilot to prescreen the PRs.

src/czar/Czar.cc
iagaponenko: I might be wrong here.

src/czar/Czar.cc
jgates108: Since the read may take a while (multiple milliseconds), the time used for comparisons needs to be before the read starts. That guarantees that updates aren't missed.

src/czar/Czar.h
jgates108: It's meant to be public as it can be called from anywhere when there's a notification that the tables have been changed.

First character is now lower case.

src/czar/Czar.cc
jgates108: latest = 8, now = 20, maxWait = 10  
20 - 8 > 10    looks ok

iagaponenko (approved):
Looks good to me

fritzm (approved):
LGTM, too -- thanks!

+1 on recommendation to have copilot review.  Why not?

PR #1038, opened 2026-07-28
DM-55611: REST services of the Qserv Replication-Ingest system should return 404 on attempts to access non-existing resources

iagaponenko:


fritzm (approved):
LGTM -- thanks!

PR #1039, opened 2026-07-28
Tickets/dm 55539

jgates108:
The main concern is that there is some very obscure corner case that doesn't use the same templates, which would explain why the templates were being generated for every job. Testing hasn't found and problems.


iagaponenko (approved):
LGTM

PR #1040, opened 2026-07-30
DM-55665: Use `qserv-xrd` as container base name when building on/from the `xrd` branch.

fritzm:
Since we want to start doing coordinated releases from both `main` and `xrd`, we need distinct container names to avoid clashes/confusion.

### DIFF
--- a/deploy/docker/build-user/Dockerfile
+++ b/deploy/docker/build-user/Dockerfile
@@ -12,7 +12,7 @@
 # account in the image. Values are provided at build time via the --build-arg option to `docker build`.
 #-------------------------------------------------------------------------------------------------------------
 
-ARG QSERV_BUILD_BASE=ghcr.io/lsst/qserv-build-base:latest
+ARG QSERV_BUILD_BASE=ghcr.io/lsst/qserv-xrd-build-base:latest
 FROM ${QSERV_BUILD_BASE} AS qserv-build-user
 
 ARG USER
--- a/deploy/docker/run/Dockerfile
+++ b/deploy/docker/run/Dockerfile
@@ -11,7 +11,7 @@
 # can be overridden by passing '--build-arg QSERV_RUN_BASE=...' to 'docker build'.
 #-------------------------------------------------------------------------------------------------------------
 
-ARG QSERV_RUN_BASE=ghcr.io/lsst/qserv-run-base:latest
+ARG QSERV_RUN_BASE=ghcr.io/lsst/qserv-xrd-run-base:latest
 FROM ${QSERV_RUN_BASE} AS qserv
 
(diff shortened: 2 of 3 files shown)


iagaponenko:
I have tested on `sdfqserv001` (as user `rubinqsv`). The following worked just fine:
```
% git checkout xrd
% git fetch --all --prune
% git pull -a
% git checkout tickets/DM-55665
% qserv build-images --pull-image
```
And I could see new containers locally:
```
% docker image ls

ghcr.io/lsst/qserv-xrd-build-base:2026.7.1-rc1
ghcr.io/lsst/qserv-xrd-build-rubinqsv:2026.7.1-rc1-3-g607e6192a
ghcr.io/lsst/qserv-xrd-mariadb:2026.7.1-rc1 
ghcr.io/lsst/qserv-xrd-run-base:2026.7.1-rc1
```
However, the next stage failed:
```
% rm -rvf build
% qserv build -j24 --no-build-image --cmake
CMake Error: Unable to (re)create the private pkgRedirects directory:
  /home/rubinqsv/code/qserv/build/CMakeFiles/pkgRedirects
This may be caused by not having read/write access to the build directory.
Try specifying a location with read/write access like:
  cmake -B build
If using a CMake presets file, ensure that preset parameter
'binaryDir' expands to a writable directory.
```

To debug this problem, I started the build container:
```
% docker run -d --init --rm -it --name rubinqsv_build -u rubinqsv  --mount src=/zfspool/management/qserv,dst=/home/rubinqsv/code/qserv,type=bind --cap-add sys_admin --cap-add sys_ptrace --security-opt seccomp=unconfined -w /home/rubinqsv/code/qserv/build ghcr.io/lsst/qserv-xrd-build-rubinqsv:2026.7.1-rc1-3-g607e6192a /bin/bash
```
And logged into the container to see if the folder is there:
```
% docker exec -it rubinqsv_build bash
% whoami
rubinqsv

% ls -al /home/rubinqsv/code/qserv/build/
total 111
drwxr-xr-x+  2 root     root  2 Jul 30 06:33 .
drwxrwxr-x+ 30 rubinqsv 1126 71 Jul 30 06:33 ..
```
The folder is not writable by the user `rubinqsv`

And here is how the folder is seen from the host:
```
% ls -al build/
total 111
drwxr-xr-x+  2 root     root  2 Jul 29 23:33 .
drwxrwxr-x+ 30 rubinqsv gu   71 Jul 29 23:33 ..
```
Apparently, the folder was NOT created by the `qserv ... --cmake` command as it used to be. And instead, it's created automatically by Docker if it's missing.

I don't understand how changes in the current PR might trigger this behavior.

iagaponenko (approved):
LGTM

fritzm:
That is very strange about the directory... will investigate...

PR #1044, opened 2026-08-01
DM-55559: Verify columns used for spatial area restrictors

malensek:
Consider a spatial query such as:
```
SELECT ... FROM Object
WHERE  scisql_s2PtInCircle(coord_ra, coord_decErr, 53.01, -28.35, 1.1111) = 1;
```
Assuming the table's partitioning columns are `coord_ra` and `coord_dec`, this query should **NOT** produce an area restrictor because it is not referring to the partitioning columns (note the 2nd parameter, coord_dec**Err**).

However, it turns out that the ra / dec passed into `scisql_s2PtInCircle` were assumed to always be the partitioning columns. This is a correctness bug because qserv may still return the correct results, only a plausible-looking subset of the correct results, or no results at all because pruning is being applied incorrectly. This could likely happen when querying u_ra/dec , sersic_ra/dec, etc. since there is some logical overlap with those types of queries. See DM-55559 for examples where only partial correct results are returned.

This PR adds several test cases to exhibit this bug, and adds a verification step (see `isAdmissibleAreaRestrictor` in `src/qana/QservRestrictorPlugin.cc`) to ensure only partitioning columns are used to create area restrictors. An object was added to the integration tests `Object.tsv.gz` (objectID 55559) that demonstrates this:
```
SELECT objectId, ra_SG, decl_SG
FROM   Object
WHERE  scisql_s2PtInCircle(ra_SG, decl_SG, 185.0, 5.0, 1.0) = 1;
```
Without the patch, this produces an area restrictor and the chunk holding objectId 55559 gets incorrectly pruned.

To reproduce this, roll back to the first commit of this branch and run the unit and integration tests: `git checkout $(git log main..tickets/DM-55559 --format=format:%H | tail -1)`


data/case01/queries/8012_areaRestrictorNonPartitionCol.sql
```diff
@@ -0,0 +1,10 @@
+-- Area restrictor chunk pruning regression test: Object is partitioned on
+-- ra_PS/decl_PS, so a scisql area function over ra_SG/decl_SG should NOT be
+-- prune chunks. Object 55559 is placed away from the rest of the data with
```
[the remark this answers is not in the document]
malensek: Matthew grammar does not good be.

src/qana/QservRestrictorPlugin.cc
```diff
@@ -164,12 +164,71 @@ bool areaParamsAreConst(query::FuncExpr const& scisqlFunc) {
...
+    }
+
+    if (partParam.lonColName.empty() || partParam.latColName.empty() ||
+        !boost::iequals(lonColumnRef.getColumn(), partParam.lonColName) ||
+        !boost::iequals(latColumnRef.getColumn(), partParam.latColName)) {
+        // column args passed into the area restrictor are NOT the partitioning columns
+        return false;
+    }
+
+    return true;
```
iagaponenko: [ This is just a matter of taste ]

The last 4 statements could be rewritten as:
```
css::PartTableParams const partParam = context.css->getPartTableParams(db, table);
return partParam.isChunked() &&
       !partParam.lonColName.empty() &&
       !partParam.latColName.empty() &&
       boost::iequals(lonColumnRef.getColumn(), partParam.lonColName) &&
       boost::iequals(latColumnRef.getColumn(), partParam.latColName);
```
Comments are not needed. This code is self-explanatory since the intent is expressed in a flow of the boolean expression and well-chosen names of the variables & methods.

malensek: Thanks, that does look a lot cleaner.

iagaponenko (approved):
LGTM

PR #1045, opened 2026-08-01
DM-55691: Add instance ID to web dash nav bar and page titles

fritzm:
Here is a small feature we previously discussed, vibe coded with Claude this morning over a ☕...


iagaponenko (approved):
LGTM

PR #1047, opened 2026-08-03
DM-52998: Remove serialization in logging checks

jgates108:


fritzm:
After further discussion with and research by John, we've decided not to proceed with this one.

PR #1049, opened 2026-08-05
tickets/DM-55724

jgates108:
This wound up being bigger than expected, but there's some old junk from way back that could be removed.

### DIFF
--- a/src/ccontrol/MergingHandler.cc
+++ b/src/ccontrol/MergingHandler.cc
@@ -195,13 +195,13 @@ std::ostream& MergingHandler::print(std::ostream& os) const {
     return os << "MergingRequester(flushed=" << (_flushed ? "true)" : "false)");
 }
 
-qdisp::MergeEndStatus MergingHandler::_mergeHttp(qdisp::UberJob::Ptr const& uberJob, string const& fileUrl,
+bool MergingHandler::_mergeHttp(qdisp::UberJob::Ptr const& uberJob, string const& fileUrl,
                                                  uint64_t fileSize) {
     if (_flushed) {
         throw util::Bug(ERR_LOC, "already flushed");
     }
 
-    if (fileSize == 0) return qdisp::MergeEndStatus(true);
+    if (fileSize == 0) return true;
     auto csvMemDisk = mysql::CsvMemDisk::create(fileSize, uberJob->getQueryId(), uberJob->getUjId());
     _csvMemDisk = csvMemDisk;
 
@@ -209,7 +209,7 @@ qdisp::MergeEndStatus MergingHandler::_mergeHttp(qdisp::UberJob::Ptr const& uber
     // race issues, and it needs to be before the thread starts.
     auto exec = uberJob->getExecutive();
     if (exec == nullptr || exec->getCancelled() || exec->isRowLimitComplete()) {
-        return qdisp::MergeEndStatus(true);
+        return true;
     }
 
     string fileReadErrorMsg;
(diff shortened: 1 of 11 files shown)


src/ccontrol/MergingHandler.cc
```diff
@@ -241,40 +241,45 @@ qdisp::MergeEndStatus MergingHandler::_mergeHttp(qdisp::UberJob::Ptr const& uber
...
     _flushed = true;
 
-    qdisp::MergeEndStatus mergeEStatus(fileMergeSuccess && fileReadErrorMsg.empty());
-    if (!mergeEStatus.success) {
+    bool mergeEStatus = fileMergeSuccess && fileReadErrorMsg.empty();
+    if (!mergeEStatus) {
         // This error check needs to come after the csvThread.join() to ensure writing
         // is finished. If any bytes were written, the result table is ruined.
-        mergeEStatus.contaminated = csvMemDisk->getBytesFetched() > 0;
+        if (csvMemDisk->getBytesFetched() > 0) uberJob->setContaminated();
```
jgates108: I'm adding an error message to this

iagaponenko (commented):
I have looked at the implementation of the merger, and a change where you're closing a connection after a failure. I don't think closing a connection after the last failure will help. You need to find a proactive solution to ensure the connection is useful _before_ doing `LOAD DATA INFILE`

Here is the best option that is provided by the MySQL C library:
```
// mysql_ping returns 0 if the connection is active
if (mysql_ping(conn) == 0) {
    return 1; // Connection is open and active
 } else {
    return 0; // Connection is dead or server went away
 }
```
You may add this check into the method `connected()`. This method simply checks if the connection was sucessfuly open in the past:
```
// File: src/mysq/MySqlConnection.{h,cc}

class MySqlConnection {
public:
    bool connected() const { return nullptr != _mysql; }
};
```
What you want to do here is to extend this method by:
```
class MySqlConnection {
public:
    bool connected() const {
        return (nullptr != _mysql) && (mysql_ping(conn) == 0);
    }
};
```
This is the most essential change that you need to do in this PR.

iagaponenko:
And I'm sorry for not mentioning this in my previous comment. You also need to. modify the result merger to test the status of the connection by calling `MySqlConnection::connected()` _before_ initiating `LOAD DATA INFILE`. And if the connection is not open, then you need to explicitly close the connection and reconnect. This will guarantee that you wil be pushing the result into the healthy connection that has been validated (and properly reconnected) as needed.

jgates108:
It's closing the connection after a success, not a failure. Closing after a success will cause it to reconnect the next time this is called.
Closing after a failure is redundant as this entire user query is going away, but it is easy enough to always close the connection.
I went with close the connection and reopen because it is simple and people want this quickly. Testing for error will require time for testing that test.

iagaponenko:
What if the "MySQL server gone" happened on the very first use of the previously open connection?

Did you test that this solution works to address the "MySQL server gone" failures? You may probably do this by reconfiguring the MySQL connection timeout to be (say 10 seconds) in your development Qserv and submitting some very long-running queries in which "uberjobs" will be running for much longer than 10 seconds.

jgates108:
The connection starts closed and needs to be opened on the first call.

jgates108:
InfileMerger.cc line line 138 is were I removed opening the connection from the constructor. It's been tested as it wouldn't pass any integration tests if it didn't work.

iagaponenko (approved):
LGTM

PR #1050, opened 2026-08-06
tickets/DM-55741

jgates108:
This works locally to show the tasks page in the dashboard.  Some discussion needs to be had before implementing DM-55247 as there should probably be separate messages for each section and/or some form of caching.


iagaponenko (approved):
LGTM

PR #1051, opened 2026-08-06
DM-55420: Large result fail reporting

jgates108:


jgates108:
Copy of my comment from Jira ticket:
I have a suspected cause, for which I put in a fix. I pushed the fixed but testing had odd results.
The suspect was that workers were hitting the query byte size limit and returning errors to save time and these errors weren’t setting the Executive::_resultFileSizeExceeded flag. Initial testing almost looked okay except some expected code wasn’t being called. At some point in the past FileChannelShared::_bytesWritten stopped getting updated with the number of bytes written to file. Being always zero, there was no way _bytesWritten could exceed the maximum number of bytes and trigger the error on the worker.

iagaponenko (commented):
How did you test an effect of these changes?
This can't be done in the ingtegraton test alone.
I would expect using some large-scale Qserv deployment (Docker-based 6-worker Qserv or 70-node Kubernetes-based deployment) and submitting a spectrum of many (hundreds) queries that were known to hit the large result limit in the Qserv deployemnets `prod` and `int`  used by the Rubin users.
You can easily find those queries on the Web dasboard of Qserv by going to the following page:
```
Status -> Past Queries
```
And selecting query statuses `FAILED` and `FAILED_LR`.

I will approve this PR only after I see it working on the Web Dashboard.

Tnak you!

jgates108:
Locally, I changed the maximum number of bytes on the worker (in the code) to a small value so that it would be sure to trip the limit on the worker and not the czar. The limit got tripped on the worker (as verified by log statements) and the FAILED_LR was on the dashboard.

iagaponenko:
I have a general comment about (opinion on) using signed vs. unsigned types for storing various sizes. Copilot already reported some concerns on this subject in this PR. I think that:
* signed types should never be used to represent sizes (of collections, etc.). The C++ standard library has a special type for this purpose: `std::size_t` (which is the same as `std::uint64_t` on 64-bit platforms in Linux). In most cases, the Library relies on this unsigned type in the relevant contexts.
* Mixing signed and unsigned types is a recipe for trouble. It also complicates code analysis and verification.
* "Dual-use" of the variables/members that are meant to store sizes to also store negative values like `-1` to indicate "a lack of information" (AKA `NULL` in SQL) is a bad practice. Good code should use a different mechanism for that.

[ Yes, in some cases the library methods (including the C++ standard library or other libraries) may return things like `int64_t`. However, the Qserv code should limit this. ]

jgates108:
I'm already well aware of these issues. Since the value is being placed into a 64 integer, the max value here is 9 exabytes, which is a physically impossible file size on a single qserv worker for at least many years. It is also simple, if need to convert an int64_t to uint64_t if really needed on the czar. 

I'm trying to be generic and an int64_t covers a lot of territory fairly easily as small negative values are likely to be useful. Generally, I think util:Error is too closely connected with database entries.

The correct value of this item is already in the string, and the number of bytes being sent to the czar is actually 0. 

Lastly, @iagaponenko expressed that having this displayed as a non-zero entry was both urgent and important, and this (int64_t) is how that happens without major rewrites.

src/qdisp/Executive.h
```diff
@@ -368,6 +379,8 @@ class Executive : public std::enable_shared_from_this<Executive> {
 
     /// Set to true if the result file is too large.
     std::atomic<bool> _resultFileSizeExceeded{false};
+    /// The size of the result that that exceeded the limit.
+    std::atomic<int64_t> _resultFileSizeErr{0};
```
iagaponenko, line 383: I'm not sure the names of the variable reflects its role:
* the `Err` at the end is a customary convention for strings (or numeric error types) where some information on an error is stored
* in this case it's a counter (the number of rows) 
Thi sis confusing for anyone who's seeing this name for the first time and before analyzing the full context where this variable is used. Perhaps some better name could be chosen here?
```
_resultFileSizeIfExceeded
```
This name would be 100% consistent with a data member that is defined above and which is used in conjunction with the new member:
```
    /// Set to true if the result file is too large.
    std::atomic<bool> _resultFileSizeExceeded{false};
```

src/qdisp/Executive.h
```diff
@@ -231,14 +231,25 @@ class Executive : public std::enable_shared_from_this<Executive> {
...
     void collectFile(std::shared_ptr<UberJob> ujPtr, protojson::FileUrlInfo const& fileUrlInfo,
                      std::string const& idStr);
 
     /// Return true if the result size limit has been exceeded.
-    bool resultSizeLimitExceeded() const { return _resultFileSizeExceeded; }
+    bool isResultSizeLimitExceeded() const { return _resultFileSizeExceeded; }
+
+    int64_t getResultFileSizeErr() const { return _resultFileSizeErr; }
+
+    void setResultSizeErr(int64_t resultFileSizeErr) {
```
iagaponenko, line 249: Same comment as for the new member `_resultFileSizeErr`. Perhaps a better name should be given here?
```
    void setResultSizeExeededErr(int64_t resultFileSize) {
```

iagaponenko (approved):
LGTM

Please, see my comments on the names of the new data member and methods (in `Executive`).

src/qdisp/Executive.h
jgates108: It's setting the flag and named after the flag.

PR #1062, opened 2026-08-09
DM-55153: Try to create czar transfer dir and httpczar tmp dirs if missing

fritzm:
Very simple code to call `std::filesystem::create_directories` on these two config-parameterized locations at czar construction and in http czar application `main`.  Deemed helpful in k8s environments where volume mounts will have been arranged, but directories within may not be created on fresh storage deployments.

### DIFF
--- a/src/czar/Czar.cc
+++ b/src/czar/Czar.cc
@@ -24,6 +24,7 @@
 #include "czar/Czar.h"
 
 // System headers
+#include <filesystem>
 #include <stdexcept>
 #include <sys/time.h>
 #include <thread>
@@ -75,6 +76,7 @@
 #include "util/TimeUtils.h"
 
 using namespace std;
+namespace fs = std::filesystem;
 
 // This macro is used to convert empty strings into "0" in order to avoid
 // problems with calling std::atoi() when the string is empty.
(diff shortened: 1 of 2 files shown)


iagaponenko:
I think attempts to create the temporary folder are misplaced. They should be done  in the following application:
```
// files: src/czar/qserv-czar-http.cc

int main(...) {
    ...
    // parse command line parameters
    ...
    // Create the folder
    ...
    fs::create_directories(httpCzarConfig.tmpDir, ec);
    if (ec) {
        ...
    }
```
This will also prevent duplicate efforts to create the missing folder, which you do in the classes `czar::Czar` and `czar::HttpCzarSvc`.

iagaponenko:
And yes, the application must be terminated if the folder can't be created.

Also note that `std::filesystem::create_directories` returns a `bool` to indicate if the folder/folders were actually created. In theory, you may post this event in the log stream:
```
    bool const created = fs::create_directories(httpCzarConfig.tmpDir, ec);
    if (ec) {
        ...
        return EXIT_FAILURE;
    }
    if (created) {
         LOGS(_log, LOG_LVL_INFO, "The temporary folder for the input data of the user tables created at: " << httpCzarConfig.tmpDir);
    }
```

fritzm:
> This will also prevent duplicate efforts to create the missing folder, which you do in the classes `czar::Czar` and `czar::HttpCzarSvc`.

Well, these are two separate directories -- the one in `czar::Czar` is usually mapped to `/qserv/data/results`, and the one in `czar::HttpCzarSvc` is usually mapped to `/qserv/data/ingest/`

iagaponenko:
@fritzm you're right regarding:

> > This will also prevent duplicate efforts to create the missing folder, which you do in the classes czar::Czar and czar::HttpCzarSvc.
> 
> Well, these are two separate directories -- the one in czar::Czar is usually mapped to /qserv/data/results, and the one in czar::HttpCzarSvc is usually mapped to /qserv/data/ingest/

I overlooked that!

Though, I still think that `httpCzarConfig.tmpDir` needs to be created at `src/czar/qserv-czar-http.cc` right after getting a value of the corresponding command-line parameter and before starting Czar.

The second folder should be created by the class Czar as per your implementation.

fritzm:
Thanks Igor -- I will update per your suggestions!

iagaponenko:
@fritzm , I bring you my apology! I have a small correction. In my original suggestion for creating the folder in `czar/qserv-czar-http.cc`, the message should state:
```
        LOGS(_log, LOG_LVL_INFO, "The temporary folder for the input data of the user tables created at: " << httpCzarConfig.tmpDir);
```

fritzm:
Thanks Igor.  I went with more generic strings "transfer directory" and "temporary directory" to better match the corresponding config args and help strings.

PR #1065, opened 2026-08-11
DM-55786: ScanTablePlugin misclassifies `SELECT *` queries

malensek:
This is a low-impact bug in the current ScanTablePlugin implementation: when finding `SELECT` column references, a bare `*` does not get counted as a column ref, resulting in the query being misclassified as NOT a scan. The query then lands on the GROUP scheduler and consumes excess resources. This is not an issue in production because the upstream TAP service already fully expands incoming queries before they reach qserv. 

Resolution:
* If a query contains a `SELECT *`, properly classify it as having column refs. We can short circuit the rest of the checks.


fritzm (approved):
LGTM -- thanks!  (Particularly like seeing a few opportunistic cleanups/modernization on separate commit -- put litter in its place :-))

PR #1081, opened 2026-08-21
DM-55902: Fixed unnecessary locking during query cancellation in MySQL connector

iagaponenko:
The lock was acquired indircetly when checking a status of the connection during query cancellation.


fritzm (approved):
Nice find! LGTM.

PR #1084, opened 2026-08-24
DM-55795: Get correct error information to qmeta tables

jgates108:


iagaponenko:
I have a general comment on the commit structure of the PR. The second commit reverses a decision (on renaming `MULTIERROR` -> `QSERROR`) made in the first commit. You need to `squash` the second commit into the first one ... unless there is a good reason to keep the second one (`QSERROR` -> `MULTIERROR`) in the commit history of Qserv.

src/ccontrol/UserQueryAsyncResult.cc
```diff
@@ -97,33 +98,39 @@ void UserQueryAsyncResult::submit() {
...
+
     // Can only return results from mysql tables
     if (_qInfo.resultLocation().compare(0, 6, "table:") != 0) {
         std::string message = "Cannot return result as it is not stored in table.";
         LOGS(_log, LOG_LVL_DEBUG, message);
         _messageStore->addErrorMessage("SYSTEM", message);
-        return;
+        failed = true;
     }
     std::string const resultTableName = _qInfo.resultLocation().substr(6);
```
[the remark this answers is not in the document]
jgates108: Added string length check before string comparisson.

src/ccontrol/UserQueryAsyncResult.cc
```diff
@@ -113,13 +113,14 @@ void UserQueryAsyncResult::submit() {
                    qStatus == qmeta::QInfo::ABORTED);
 
     // Can only return results from mysql tables
-    if (_qInfo.resultLocation().compare(0, 6, "table:") != 0) {
+    auto const resultLoc = _qInfo.resultLocation();
+    if (resultLoc.size() < 6 || resultLoc.compare(0, 6, "table:") != 0) {
```
iagaponenko, line 116: Since this change fixed a bug introduced in the same commit as this PR, the fix needs to be squashed with the commit where the mistake was made.

src/ccontrol/UserQueryAsyncResult.cc
```diff
@@ -74,6 +74,7 @@ UserQueryAsyncResult::UserQueryAsyncResult(QueryId queryId, CzarId czarId,
 }
 
 void UserQueryAsyncResult::submit() {
+    LOGS(_log, LOG_LVL_INFO, "UserQueryAsyncResult::submit start QID=" << _queryId);
```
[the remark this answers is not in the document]
jgates108: This is short if there are no errors, and if there are errors then it is useful to seen what was presented to the user.

src/ccontrol/UserQueryAsyncResult.cc
jgates108: These have been squashed.

src/ccontrol/UserQueryAsyncResult.cc
```diff
@@ -97,33 +98,40 @@ void UserQueryAsyncResult::submit() {
 
     // If query has not finished yet return error
     // TODO: there may be more info available if status is FAILED or ABORTED
-    if (_qInfo.queryStatus() != qmeta::QInfo::COMPLETED) {
-        std::string message = "Query is still executing (or FAILED)";
+    auto const qStatus = _qInfo.queryStatus();
+    auto const qStatusStr = qmeta::QInfo::qstatusToStr(qStatus);
+    if (qStatus != qmeta::QInfo::COMPLETED) {
+        std::string message = "Query not COMPLETED status=" + qStatusStr;
```
iagaponenko: Since the message string is only posted on failed queries, the conversion of the query status code into a string has to be done within the block `if (qStatus != qmeta::QInfo::COMPLETED) {`
```
    auto const qStatus = _qInfo.queryStatus();
    if (qStatus != qmeta::QInfo::COMPLETED) {
        auto const qStatusStr = qmeta::QInfo::qstatusToStr(qStatus);
        std::string message = "Query not COMPLETED status=" + qStatusStr;
        ...
    }
```
Or, which is even better, you may not create this variable and just reduce the code to:
```
    auto const qStatus = _qInfo.queryStatus();
    if (qStatus != qmeta::QInfo::COMPLETED) {
        std::string const message = "Query not COMPLETED status=" + qmeta::QInfo::qstatusToStr(qStatus);
        ...
    }
```

src/czar/Czar.cc
```diff
@@ -395,9 +395,9 @@ SubmitResult Czar::submitQuery(string const& query, map<string, string> const& h
     SubmitResult result;
 
     // instantiate message table manager
-    MessageTable msgTable(lockName, _czarConfig->getMySqlResultConfig());
+    auto msgTable = make_shared<MessageTable>(lockName, _czarConfig->getMySqlResultConfig());
```
iagaponenko, line 398: Just curious - what's the point of switching to the shared pointer? What was wrong with allocating a small object on the stack? And how does this help to improve the error reporting?

The class has two small members:
```
class MessageTable {

private:
    std::string const _tableName;
    std::shared_ptr<sql::SqlConnection> _sqlConn;
};
```
The number of stack-allocated bytes for class `std::string` varies from  24 or 32 bytes. And the shared pointer class is much smaller.

src/czar/MessageTable.cc
```diff
@@ -148,6 +156,8 @@ void MessageTable::_saveQueryMessages(ccontrol::UserQuery::Ptr const& userQuery,
         LOGS(_log, LOG_LVL_ERROR, exc.message());
         throw exc;
     }
+    LOGS(_log, LOG_LVL_INFO,
+         "MessageTable::_saveQueryMessages QID=" << userQuery->getQueryId() << " errStr=" << multiErrStr);
```
iagaponenko, line 160: What is a reason for posting this at the `INFO` level instead of `DEBUG`?

src/czar/Czar.cc
jgates108: Just to make sure that any modifications are made to the same object, without having to dig into the specifics of what the objects contain. 

The difference in the amount of memory used is tiny as far as the qserv czar is concrened.

src/qmeta/QInfo.h
```diff
@@ -60,6 +60,22 @@ class QInfo {
...
+            case COMPLETED:
+                return "COMPLETED";
+            case FAILED:
+                return "FAILED";
+            case FAILED_LR:
+                return "FAILED_LR";
+            case ABORTED:
+                return "ABORTED";
+        }
+        return "unknown";
```
iagaponenko: The code may need to be reinforced to throw an exception instead of silently returning a string.
The second comment is that the exception should be thrown from `default:`
```
    static std::string qstatusToStr(QStatus qStatus) {
        switch (qStatus) {
            case EXECUTING:
                return "EXECUTING";
            case COMPLETED:
                return "COMPLETED";
            case FAILED:
                return "FAILED";
            case FAILED_LR:
                return "FAILED_LR";
            case ABORTED:
                return "ABORTED";
            default:
                throw std::runtime_error("QInfo::qstatusToStr invalid query status code: " + std::to_string(qStatus));
        }
    }
```

src/czar/MessageTable.cc
jgates108: There's one of these per user query, it's basically a small message saying how the query went. 
I would like INFO over DEBUG here in that this is generally useful and takes up very little space and the message table is likely to get deleted.

src/czar/Czar.cc
iagaponenko: I'm sorry, I still don't get what you mean to say here:

> Just to make sure that any modifications are made to the same object, without having to dig into the specifics of what the objects contain.

The `msgTable` is created at line `398` and consumed (locked in the calling thread and unlocked by the finalizer thread).

Another `msgTable` object is created later in a nested `if` block of the same method, `Czar::submitQuery` for the ASYNC queries.

These two objects didn't interfere. At least, they never interfered in the past. Did you run into any conflicts between these two objects in your query dispatch? And if you feel that having two variables with the same name is confusing, why not simply rename one (like changing the first one from `msgTable` to `submitMsgTable`) instead of changing the allocation method?

src/qmeta/QInfo.h
jgates108: Throwing could have some serious consequences (uncaught exception crashes the czar) and these are errors that may not occur frequently/easily, making them difficult to test. I'm reluctant to add that kind of consequence to this kind of translation function.

It looks like C++26 has a real fix for this. https://vittorioromeo.com/index/blog/refl_enum_to_string.html

jgates108: https://en.cppreference.com/cpp/header/meta

src/czar/Czar.cc
jgates108: It's about code readability. If I see a shared pointer being copied, I know all modifications are being made to the same object and I don't have to worry about the local object falling out of scope and being deleted.

The way it used to be, I had to go in and verify that copying the class was safe.

src/qmeta/QInfo.h
iagaponenko: Yes, exactly - that's the whole point of throwing an exception here. The application must report an error (along with the unexpected) status code and abort. Unknown query statuses should not be seen in this context. They can mean either of:
* a memory corruption within the application
* the table `qservMeta`.`QInfo` is corrupt
* the code is not compatible with the schema of the table `qservMeta`.`QInfo`

Returning the meaningless string `"unknown"` will be hiding these problems under the rug.

src/czar/Czar.cc
jgates108: I'm going to change the name of the second one of those. Risking shadowing variable names has no benefit.

src/qmeta/QInfo.h
jgates108: The most likely outcome would be "unknown" entries appearing in QInfo, which would lead back to the original problem fairly easily. If throwing is what you really want, okay.

iagaponenko (approved):
LGTM

PR #1088, opened 2026-08-26
DM-55957: Address worker pod container startup sequencing issue

fritzm:
When the `worker-svc` container starts before its associated `mariadb` container, an sql connection failure in the worker initialization code causes a throw, and exits the container.  The worker pod then sits in not-started state until k8s reaps and restarts it after 5 minutes.

Found the exception was coming from an incomplete refactor in `MySqlConnection`.  `populateErrorObject` was written expecting `getMySql` to return null on a failed connection (which it then checks for.)  After the refactor, that null check became dead code and the function threw instead.  This PR adds a `try`/`catch` to cover that path.

### DIFF
--- a/src/sql/MySqlConnection.cc
+++ b/src/sql/MySqlConnection.cc
@@ -25,11 +25,12 @@
 #include "sql/MySqlConnection.h"
 
 // System headers
-#include <cstdio>
 #include <cstddef>
+#include <cstdio>
 #include <fstream>
 #include <iostream>
 #include <sstream>
+#include <stdexcept>
 
 // LSST headers
 #include "lsst/log/Log.h"
@@ -41,13 +42,13 @@
 
 namespace {
 void populateErrorObject(lsst::qserv::mysql::MySqlConnection& m, lsst::qserv::sql::SqlErrorObject& o) {
-    MYSQL* mysql = m.getMySql();
-    if (mysql == nullptr) {
-        o.setErrNo(-999);
-        o.addErrMsg("Error connecting to mysql with config:" + m.getConfig().toString());
-    } else {
+    try {
+        MYSQL* mysql = m.getMySql();
         o.setErrNo(mysql_errno(mysql));
         o.addErrMsg(mysql_error(mysql));
+    } catch (std::logic_error const&) {
+        o.setErrNo(-999);
+        o.addErrMsg("Error connecting to mysql with config:" + m.getConfig().toString());
     }
 }
 


src/sql/MySqlConnection.cc
```diff
@@ -41,11 +41,11 @@
 
 namespace {
 void populateErrorObject(lsst::qserv::mysql::MySqlConnection& m, lsst::qserv::sql::SqlErrorObject& o) {
-    MYSQL* mysql = m.getMySql();
-    if (mysql == nullptr) {
+    if (!m.connected()) {
         o.setErrNo(-999);
         o.addErrMsg("Error connecting to mysql with config:" + m.getConfig().toString());
     } else {
+        MYSQL* mysql = m.getMySql();
```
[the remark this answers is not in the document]
fritzm: addressed

src/wpublish/ChunkInventory.cc
```diff
@@ -63,7 +62,6 @@ void fetchDbs(string const& instanceName, SqlConnection& sc, C& dbs) {
     while (true) {
         LOGS(_log, LOG_LVL_DEBUG, "Launching query: " << query);
         resultP = sc.getQueryIter(query);
-        assert(resultP.get());
         if (resultP->getErrorObject().isSet()) {
             SqlErrorObject& seo = resultP->getErrorObject();
```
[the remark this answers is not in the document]
fritzm: addressed

src/wpublish/ChunkInventory.cc
```diff
@@ -63,11 +62,14 @@ void fetchDbs(string const& instanceName, SqlConnection& sc, C& dbs) {
...
-            SqlErrorObject& seo = resultP->getErrorObject();
-            LOGS(_log, LOG_LVL_WARN,
-                 "Waiting to get list of publishable dbs: " << seo.errNo() << ": " << seo.errMsg());
+        if (!resultP || resultP->getErrorObject().isSet()) {
+            if (resultP) {
+                SqlErrorObject& seo = resultP->getErrorObject();
+                LOGS(_log, LOG_LVL_WARN,
+                     "Waiting to get list of publishable dbs: " << seo.errNo() << ": " << seo.errMsg());
+            } else {
+                LOGS(_log, LOG_LVL_WARN, "Waiting to get list of publishable dbs: null result");
```
iagaponenko: Will this block the calling thread potentially forever if the worker database doesn't have any databases read for Qserv? Or will it unblock if any database is found?

src/wpublish/ChunkInventory.cc
```diff
@@ -90,7 +92,10 @@ void fetchChunks(string const& instanceName, string const& db, SqlConnection& sc
     LOGS(_log, LOG_LVL_DEBUG, "Launching query: " << query);
 
     shared_ptr<SqlResultIter> resultP = sc.getQueryIter(query);
-    assert(resultP.get());
+    if (!resultP) {
+        LOGS(_log, LOG_LVL_ERROR, "ChunkInventory failed to get a list of published chunks for db: " << db);
+        return;
+    }
```
iagaponenko: While preventing the crash of the application, the graceful return from the method `fetchChunk` produces an ambiguous result:
- There was a failure to fetch the chunk list, or
- No chunks exist at the worker

The previous code was a bit more deterministic in this sense - the empty chunk list was returned only when no chunks existed.

Perhaps this is okay and the upstream code will be happy to have no chunks regardless of the reason(?)

src/wpublish/ChunkInventory.cc
```diff
@@ -116,7 +121,10 @@ void fetchId(string const& instanceName, SqlConnection& sc, string& id) {
     LOGS(_log, LOG_LVL_DEBUG, "Launching query: " << query);
 
     shared_ptr<SqlResultIter> resultP = sc.getQueryIter(query);
-    assert(resultP.get());
+    if (!resultP) {
+        LOGS(_log, LOG_LVL_ERROR, "ChunkInventory failed to get a unique identifier of the worker");
+        return;
+    }
```
iagaponenko: The same comment regarding changes in the behavior of the method. 

In principle, the method should either return some flag (like boolean `false`) to indicate the problem to fetch the identifier, or throw some exception recognisable by the caller

src/wpublish/ChunkInventory.cc
fritzm: This particular loop only iterates only on query failure (typically, the worker service reaching out to the worker db before the worker db has started up.)  A successful query with zero rows exits.

src/wpublish/ChunkInventory.cc
fritzm: I will drop this change from this PR -- this particular assert is not causing us trouble at the moment as far as I can tell, and it is probably better the leave things as they are than introduce subtle unintended behavioral changes.

Can be addressed in the future with systematic revisit of db/chunk update polling, attached to an appropriate testing campaign.

src/wpublish/ChunkInventory.cc
fritzm: [As above] I will drop this change from this PR -- this particular assert is not causing us trouble at the moment as far as I can tell, and it is probably better the leave things as they are than introduce subtle unintended behavioral changes.

Can be addressed in the future with systematic revisit of db/chunk update polling, attached to an appropriate testing campaign.

src/wpublish/ChunkInventory.cc
fritzm: On further reflection, I will drop this change from this PR as well -- this particular assert is not causing us trouble at the moment as far as I can tell, and it is probably better the leave things as they are than introduce subtle unintended behavioral changes.

Can be addressed in the future with systematic revisit of db/chunk update polling, attached to an appropriate testing campaign.

PR #1095, opened 2026-09-03
DM-55948: Cancel Queries that have taken too long on snail scan.

jgates108:


src/wpublish/QueriesAndChunks.h
```diff
@@ -401,6 +403,8 @@ class QueriesAndChunks {
     /// This should be set by the config with "scheduler.maxconcurrentbootedtasks"
     int _maxDarkTasks = 25;
 
+    unsigned int _maxQueryAgeMinutes = 0;  ///< Maximum age of a query in minutes, 0 means no limit.
```
iagaponenko, line 406: Seconds are a better unit for timing various activities. It would be consistent with the corresponding attributes of tasks and MySQL queries (shown in secondfs), etc. The unification (seconds) reduces the chance of bugs caused by Accidental conversion/interpretation. The max limit of the 32-bit unsigned integer is more than enough for this parameter.

As an example: NASA's Mars Climate Orbiter was lost in 1999 due to a software bug in the conversion from imperial units to metric units.

src/wpublish/QueryStatistics.cc
```diff
@@ -70,6 +70,11 @@ QueryStatistics::QueryStatistics(QueryId qId_, CzarId czarId_)
             string("TimeFillingBuffersPerTask_") + qidStr, {0.1, 1, 10, 30, 60, 120, 300, 600, 1200, 10000}));
 }
 
+double QueryStatistics::getAgeSeconds() const {
```
iagaponenko, line 73: This is another example where you're mixing units to represent the age of queries. This time is a `double` number of seconds. You may consider returning `unsigned int` (seconds) and unifying the query `age` attribute (and the relevant limits) as such.

src/wpublish/QueryStatistics.h
```diff
@@ -117,10 +117,7 @@ class QueryStatistics {
         _tasksBooted += 1;
     }
 
-    QueryId getQueryId() {
-        std::lock_guard<std::mutex> lg(_qStatsMtx);
-        return queryId;
-    }
+    QueryId getQueryId() const { return queryId; }
```
iagaponenko: Why this method here? The member is already public:
```
class QueryStatistics {
public:
    QueryId const queryId;
};
```

src/wpublish/QueriesAndChunks.cc
```diff
@@ -335,6 +337,24 @@ void QueriesAndChunks::examineAll() {
...
+    double const maxAgeSeconds = _maxQueryAgeMinutes * 60.0;
+
+    // Check all query for taking too long, and cancel it if it is.
+    if (maxAgeSeconds > 0) {
+        for (auto const& uqElem : uqStatList) {
+            //&&&; // don't check if query is finished.
+            auto const ageSeconds = uqElem->getAgeSeconds();
+            if (ageSeconds > maxAgeSeconds) {
+                auto const uqInfo = uqElem->getUserQueryInfo();
+                string errMsg = string("Query is too old, age=") + to_string(ageSeconds) +
```
iagaponenko: You don't need to explicitly turn the first part of the string composition statement into `string()` since the next part of the statement (`to_string`) is already `string`. The following would be easier to read, and it would compile just fine:
```
string const errMsg = "Query is too old, age=" + to_string(ageSeconds) + ...;
```

src/wconfig/WorkerConfig.h
```diff
@@ -379,6 +381,9 @@ class WorkerConfig {
             util::ConfigValTUInt::create(_configValMap, "czar", "DeadTimeSec", notReq, 180);
     CVTUIntPtr _czarComNumHttpThreads =
             util::ConfigValTUInt::create(_configValMap, "czar", "ComNumHttpThreads", notReq, 40);
+
+    CVTUIntPtr _maxQueryAgeMinutes =
+            util::ConfigValTUInt::create(_configValMap, "query", "maxQueryAgeMinutes", notReq, 0);
```
iagaponenko: Shouldn't this parameter (logically) better belong to the `scheduler` category instead of `query`?

src/util/Error.h
```diff
@@ -86,6 +86,7 @@ class Error {
         JOB_CANCEL,
         // Worker errors:
         WORKER_RESULT_TOO_LARGE,
+        WORKER_QUERY_TOO_OLD,
```
iagaponenko: Maybe something more specific like?
```
WORKER_QUERY_EXPIRED
```

src/wconfig/WorkerConfig.h
```diff
@@ -94,6 +94,8 @@ class WorkerConfig {
     /// @return maximum number of tasks that can be booted from a single user query
     unsigned int getMaxConcurrentBootedTasks() const { return _maxConcurrentBootedTasks->getVal(); }
 
+    unsigned int getMaxQueryAgeMinutes() const { return _maxQueryAgeMinutes->getVal(); }
```
iagaponenko, line 97: Seconds are a better unit here. Reasons:
* unified timing in all contexts
* _potentially_, you may want to implement the sub-minute expiration mechanism for specific queries. Right now, you seem to have the global query expiration mechanism for all queries.

src/wbase/UserQueryInfo.h
```diff
@@ -76,6 +76,10 @@ class UserQueryInfo {
     /// be killed but there's no need to track UberJob id's anymore.
     void cancelFromCzar();
 
+    /// Return true if no more Tasks need to be run for the query,
+    /// there may still be a need for the result files.
+    bool isWorkDone() const { return _cancelledByCzar; }
```
iagaponenko, line 81: Everything here is confusing (to the first-time reader of this code):
- the comment
- the name of the method (which is not in line with the comment)
- the decision-making (the implementation, which is based on the internal flag that seems to be irrelevant to the name of the method)

src/wpublish/QueryStatistics.cc
jgates108: Putting the configuration value for this in seconds makes for something really hard to read in the config file. If you see 44000 seconds in the configuration file, how much time is that?

src/wpublish/QueryStatistics.h
jgates108: Good point.

src/wconfig/WorkerConfig.h
jgates108: This really has little to do with the scheduler, and the scheduler has a lot of parameters as is. I'm thinking about it.

src/wconfig/WorkerConfig.h
jgates108: The limit for all queries at his time is measured in hours, seconds difficult for humans to read at that scale.

src/wpublish/QueryStatistics.cc
iagaponenko: The main problem is not the specific value you put into the config, which is not a big deal at all:
```
8 * 3600 = ..
44000 / 3600 = ...
```
It's about mixing different units in the code and converting between the units. What I have seen so far includes:
* using `double` to represent seconds
* using `unsigned int` to represent minutes
* using `unsigned int` to represent seconds
* translating minutes into integer seconds
* translating `double` seconds into integer seconds

This may not seem to be a problem to you (since you wrote all of that). A problem is maintaining this code in the long run, which may require others to look at this code and track all these types and transformations.

BTW, that's exactly one of the reasons why `std::chrono` was added to the C++ library.

jgates108: I'm aware of the arguments. This is a really straight forward calculation converting integer minutes to double seconds (that is the only conversion in this ticket).  I'm not doing something insane, like trying to use imperial units with metric units. Testing this is time consuming.

The configuration file has to use reasonable units, or it's asking for errors when the systems are setup. Using the same time units for everything in the code is a bad idea as some things should be in milliseconds and other things should be in hours, and integer truncation is a real thing.

chronos in C++  23 looks to be significantly improved (you can print duration without using count(), yay).

src/wbase/UserQueryInfo.h
jgates108: I have not been able to come up with a better name. The function returns true if there is no more work to be done for the query, which is true in all cases when `_cancelledByCzar` is true. 

`_cancelledByCzar` is set when the czar tells the worker it has what it needs to complete the user query, except for posibly some result files.

src/wdb/QueryRunner.cc
```diff
@@ -174,6 +173,7 @@ bool QueryRunner::runQuery(std::string& errMsg) {
 
 MYSQL_RES* QueryRunner::_primeResult(string const& query) {
     util::HoldTrack::Mark mark(ERR_LOC, "QR _primeResult() QID=" + _task->getIdStr());
+    LOGS(_log, LOG_LVL_TRACE, "QueryRunner::_primeResult() QID=" << _task->getIdStr() << " query=" << query);
```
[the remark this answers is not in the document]
jgates108: Removed

src/czar/HttpCzarWorkerModule.cc
```diff
@@ -181,9 +181,15 @@ json HttpCzarWorkerModule::_handleWorkerCzarComIssue(string const& func) {
...
+                LOGS(_log, LOG_LVL_DEBUG,
+                     "_handleWorkerCzarComIssue READY worker=" << wId << " QID=" << rdyMsg->queryId
+                                                               << " ujId=" << rdyMsg->uberJobId);
                 auto exRespMsg = czar::Czar::getCzar()->handleUberJobReadyMsgNoThrow(rdyMsg, fName);
                 execRespMsgs.push_back(exRespMsg);
             } else {
+                LOGS(_log, LOG_LVL_DEBUG,
+                     "_handleWorkerCzarComIssue ERROR worker=" << wId << " QID=" << rdyMsg->queryId
+                                                               << " ujId=" << rdyMsg->uberJobId);
                 auto errMsg = dynamic_pointer_cast<protojson::UberJobErrorMsg>(statusMsg);
```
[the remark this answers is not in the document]
jgates108: Removed

src/wpublish/QueriesAndChunks.cc
```diff
@@ -335,6 +338,25 @@ void QueriesAndChunks::examineAll() {
...
+            if (!uqInfo->isWorkDone()) {
+                auto const ageSeconds = uqElem->getAgeSeconds();
+                if (ageSeconds > maxAgeSeconds) {
+                    string errMsg = string("Query took too long, age=") + to_string(ageSeconds) +
+                                    " maxAge=" + to_string(maxAgeSeconds) + " seconds";
+                    util::Error error(util::Error::WORKER_QUERY_TOOK_TOO_LONG, util::Error::NONE, errMsg);
+                    uqInfo->fatalErrorForAllUberJobs(error);
+                }
+            }
+        }
```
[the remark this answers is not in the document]
jgates108: If `_cancelledByCzar` is true, all associated UberJobs (and their Tasks) should be done/cancelled already.
Re-signalling dead UberJobs and/or Tasks is harmless.

If, for some reason, the czar still cares about this, and it has taken too long, and there are any UberJobs that are still alive, this should be called to have those UberJobs inform the czar that there's a problem and also cancel any remaining Tasks, even if it is `isMostlyDead()`.

src/wpublish/QueryStatistics.cc
jgates108: Also, the conversion has been tested. 
The jira ticket has a comment that shows that 1 min is converted to 60 seconds in testing. 
Values in the configuration file are in minutes as that's easier for me (and I think most people) to understand, while integer hours doesn't have fine enough resolution.
To help future programmers, all the variables are labeled with the appropriate units.

src/wbase/FileChannelShared.cc
```diff
@@ -344,18 +344,31 @@ bool FileChannelShared::isRowLimitComplete() const {
...
+    _buildAndTransmitError(tMtxLock, multiErr, idStr, chunkId, LOG_LVL_ERROR, notCancelled);
+}
+
+void FileChannelShared::_buildAndTransmitError(lock_guard<mutex> const& streamMutexLock,
+                                               util::MultiError& multiErr, string const& idStr, int chunkId,
+                                               int logLvl, bool cancelled) {
     if (_rowLimitComplete) {
-        LOGS(_log, LOG_LVL_WARN,
-             __func__ << " already enough rows, this call likely a side effect" << task->getIdStr());
+        LOGS(_log, LOG_LVL_WARN, __func__ << " already enough rows, this call likely a side effect" << idStr);
```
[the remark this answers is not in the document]
jgates108: _streamMutex is not need in the http version of qserv, as there's no longer a stream to protect.

src/util/Error.h
jgates108: Changed to WORKER_QUERY_TOOK_TOO_LONG. For me, expired translates to dead.

iagaponenko (approved):
LGTM

PR #1096, opened 2026-09-03
DM-55959: add CertManager support to Helm chart

fritzm:
Adds chart support for optional per-deployment use of CertManager to manage frontend SSL cert


fritzm:
Go home copilot you're drunk

PR #1102, opened 2026-09-14
tickets/DM-56003

jgates108:
This went a bit out of scope as I found some useful mutex related options that I would like tested on a large scale. Mutex changes are confined to the second commit, so they can be removed easily.


src/qdisp/Executive.cc
```diff
@@ -913,6 +909,227 @@ void Executive::collectFile(std::shared_ptr<UberJob> ujPtr, protojson::FileUrlIn
     checkLimitRowComplete();
 }
 
+bool avoidThisWorker(czar::CzarChunkMap::WorkerChunksData::Ptr const& targetWorker,
+                     protojson::WorkerContactInfo::WCMapPtr const& wContactMap,
+                     qdisp::JobQuery::Ptr const& jqPtr,
+                     std::shared_ptr<czar::CzarFamilyMap> const& czFamilyMap) {
+    auto iter = wContactMap->find(targetWorker->getWorkerId());
```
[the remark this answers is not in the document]
jgates108: That check is made before calling this function, but it's still a reasonable thing to do.

src/ccontrol/UserQueryFactory.cc
```diff
@@ -431,26 +431,25 @@ UserQuery::Ptr UserQueryFactory::newUserQuery(std::string const& aQuery, std::st
         auto messageStore = std::make_shared<qmeta::MessageStore>();
         std::shared_ptr<qdisp::Executive> executive;
         std::shared_ptr<rproc::InfileMergerConfig> infileMergerConfig;
+        auto czarConfig = cconfig::CzarConfig::instance();
+        int const uberJobMaxChunks = czarConfig->getUberJobMaxChunks();
```
iagaponenko: It looks like the only user of this parameter `uberJobMaxChunks` is `Executive`. If so, then why not push this line into the implementation ofentation of `qdisp::Executive::create()`? In this case tou won't need to add parameter to the method `:create()`

src/qdisp/Executive.cc
```diff
@@ -94,14 +94,16 @@ namespace lsst::qserv::qdisp {
...
+                     int uberJobMaxChunks)
         : _messageStore(ms),
           _qdispPool(qdispPool),
           _queryProgress(queryProgress),
           _queryProgressHistory(queryProgressHistory),
           _secondsBetweenQMetaUpdates(chrono::seconds(secondsBetweenUpdates)),
           _querySession(querySession),
-          _jobMaxAttempts(jobMaxAttempts) {
+          _jobMaxAttempts(jobMaxAttempts),
+          _uberJobMaxChunks(uberJobMaxChunks) {
```
iagaponenko, line 106: I'm not seeing any good reason to pass the limit as a parameter to the constructor. You may replace the parameter initialization with:
```

    _uberJobMaxChunks(cconfig::CzarConfig::instance()->getUberJobMaxChunks()) {
}

iagaponenko (approved):
LGTM

PR #1104, opened 2026-09-16
DM-56124: Fixed a bug in the file status checking logic

iagaponenko:
Also did a minor refactoring of the for loop by migrating it to the for_range loop.


fritzm (approved):
LGTM

